Issue #93, opened 2018-08-28
PROW_JOB_ID should not be required


scothis:
Running the update-codegen.sh script from eventing fails because `PROW_JOB_ID` is unbounded. This var should have a default automatically set when not running on prow.

```shell
$ ./hack/update-codegen.sh
./hack/../vendor/github.com/knative/test-infra/scripts/library.sh: line 32: PROW_JOB_ID: unbound variable
```

adrcunha:
https://github.com/knative/serving/pull/1975 fixes the serving repo.

adrcunha:
https://github.com/knative/eventing/pull/413 fixes the eventing repo.

dprotaso:
https://github.com/knative/build/pull/327 fixes it for build repo

adrcunha:
https://github.com/knative/pkg/pull/61 fixes the pkg repo

adrcunha:
https://github.com/knative/build-templates/pull/62 fixes the build-templates repo

adrcunha:
https://github.com/knative/docs/pull/371 fixes the docs repo.

PR #94, opened 2018-09-04
Provide default (empty value) for PROW_JOB_ID if it's unset

dprotaso:
This will make the library script resilient to outer scripts that source library.sh with the `set -o nounset`


dprotaso:
Addresses: https://github.com/knative/test-infra/issues/93

adrcunha:
/approve
/lgtm

Thanks!

PR #98, opened 2018-09-05
Disable review_acts_as_approve

adrcunha:
It seems that it's causing some confusion during approval/lgtm, so let's disable it until things are figured out.


grantr:
/lgtm

Specifically a review-state approval in https://github.com/knative/serving/pull/1882#pullrequestreview-147843065 caused the approved label to be removed. I hope to dig into that today.

PR #105, opened 2018-09-07
Install wrk with prow

srinivashegde86:
Part of https://github.com/knative/serving/issues/1981

This will be used to run the perf tests to log latency values - https://github.com/knative/serving/tree/master/test/performance/observed-concurrency

### DIFF
--- a/images/prow-tests/Dockerfile
+++ b/images/prow-tests/Dockerfile
@@ -27,6 +27,7 @@ RUN docker-credential-gcr configure-docker
 # Extra tools through apt-get
 RUN apt-get install -y uuid-runtime  # for uuidgen
 RUN apt-get install -y npm  # for markdownk-link-check
+RUN apt-get install -y build-essential libssl-dev git # for wrk(perf benchmarking tool) 
 
 # Extra tools through go get
 RUN go get -u github.com/google/go-containerregistry/cmd/ko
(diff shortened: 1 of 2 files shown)


scripts/library.sh
```diff
@@ -58,6 +58,16 @@ function warning() {
   make_banner "!" "$1"
 }
 
+# Install wrk. Used for performance benchmarking
+# libssl-dev needs to be installed before running this.
+function install_wrk() {
```
adrcunha: Why not build and install the binary in the image itself, instead of building it every time you need it?

images/prow-tests/Dockerfile
```diff
@@ -27,6 +27,7 @@ RUN docker-credential-gcr configure-docker
 # Extra tools through apt-get
 RUN apt-get install -y uuid-runtime  # for uuidgen
 RUN apt-get install -y npm  # for markdownk-link-check
+RUN apt-get install -y build-essential libssl-dev git # for wrk(perf benchmarking tool)
```
adrcunha: git is already available in the image, you shouldn't need anything new to `git clone` stuff.

scripts/library.sh
srinivashegde86: Updated

images/prow-tests/Dockerfile
srinivashegde86: removed

PR #109, opened 2018-09-10
Setup Prow jobs for knative/caching

adrcunha:
Part of https://github.com/knative/caching/issues/1

### DIFF
--- a/ci/prow/config.yaml
+++ b/ci/prow/config.yaml
@@ -55,6 +55,7 @@ tide:
     - knative/docs
     - knative/test-infra
     - knative/pkg
+    - knative/caching
     labels:
     - lgtm
     - approved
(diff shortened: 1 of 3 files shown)


ci/prow/config.yaml
```diff
@@ -1789,6 +1953,68 @@ periodics:
...
+    containers:
+    - image: gcr.io/knative-tests/test-infra/coverage:latest
+      imagePullPolicy: Always
+      command:
+      - "/coverage"
+      args:
+      - "--artifacts=$(ARTIFACTS)"
+      - "--profile-name=coverage_profile.txt"
+      - "--cov-target=./pkg/"
+      - "--cov-threshold-percentage=80"
```
jessiezcc: 80 might be too high for a new repo?

adrcunha: It's not a blocker, so I don't think it's a problem.

jessiezcc: @steuhs, I see threshold parameter in post, continuous, pull coverage jobs. Which job will block PR, is it pull? If yes, 
what's the threshold usage in other post, continuous jobs?

adrcunha: And actually ALL repos have this threshold.

adrcunha:
@jessiezcc Prow won't understand `/lgtm` if you add something after it :(

jessiezcc:
opened https://github.com/kubernetes/test-infra/issues/9402 to track prow issu

PR #112, opened 2018-09-11
add missing paren to scripts/library.sh

aaron-prindle:
I think there might be a missing parenthesis in the library.sh script.  I've added a paren where I believe the missing one is.


adrcunha:
Good catch, thanks!

/lgtm
/approve

PR #121, opened 2018-09-17
add an option --skip-prepare-cluster for a local-up kubernetes cluster

### LINKED ISSUE #122: Propose: add an option --skip-prepare-cluster in e2e-test
## Expected

i have a kubernetes cluster that is bootstrapped by `kubeadm`, but currently, there is no way to let knative components run e2e on it

zrss:
Fix #122

### DIFF
--- a/scripts/e2e-tests.sh
+++ b/scripts/e2e-tests.sh
@@ -281,6 +281,7 @@ function success() {
 
 RUN_TESTS=0
 EMIT_METRICS=0
+PREPARE_CLUSTER=1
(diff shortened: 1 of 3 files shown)


adrcunha:
As I mentioned in the issue, the behavior you need can be achieved with `--run-tests`. You do need to call `setup_test_cluster()` unless you just want to run the e2e tests. But if you just want to run the e2e tests, then you just need to issue a `go test` command. Please consider enhancing the documentation if you think it would have make things easier/clearer for you.

PR #128, opened 2018-09-17
Add wait for app to be routable function in library.sh

srinivashegde86:


scripts/library.sh
```diff
@@ -146,6 +146,24 @@ function wait_until_service_has_external_ip() {
   return 1
 }
 
+# Waits for the endpoint to be routable
+# Parameters: $1 - External ingress IP address `kubectl get svc knative-ingressgateway -n istio-system -o jsonpath="{.status.loadBalancer.ingress[*].ip`
```
adrcunha: The kubectl commands are unnecessary, it's up to the caller to figure them out. You can, however, make the IP optional, which then would default to the result of the `kubectl` command above.

scripts/library.sh
```diff
@@ -146,6 +146,24 @@ function wait_until_service_has_external_ip() {
...
+#             $2 - Hostname `kubectl get route <route_name> -o jsonpath="{.status.domain}"`
+function wait_until_routable() {
+    for i in {1..150}; do  # timeout after 5 minutes
+        local val=$(curl -H "Host: $2" "http://$1" >/dev/null 2>&1)
+        if [[ -z "$val" ]]; then
+            echo -n "."
+            sleep 2
+        else
+            echo "App is now routable"
```
adrcunha: nit: add a linefeed before the status message, so it won't appear at the right of the "....".

scripts/library.sh
```diff
@@ -146,6 +146,24 @@ function wait_until_service_has_external_ip() {
   return 1
 }
 
+# Waits for the endpoint to be routable
+# Parameters: $1 - External ingress IP address `kubectl get svc knative-ingressgateway -n istio-system -o jsonpath="{.status.loadBalancer.ingress[*].ip`
+#             $2 - Hostname `kubectl get route <route_name> -o jsonpath="{.status.domain}"`
+function wait_until_routable() {
+    for i in {1..150}; do  # timeout after 5 minutes
+        local val=$(curl -H "Host: $2" "http://$1" >/dev/null 2>&1)
```
adrcunha: You're redirecting stderr to stdout, `val` will be populated in case of error, and your function will return success (when it shouldn't).

scripts/library.sh
```diff
@@ -146,6 +146,24 @@ function wait_until_service_has_external_ip() {
...
+        if [[ -z "$val" ]]; then
+            echo -n "."
+            sleep 2
+        else
+            echo "App is now routable"
+            return 0
+        fi                
+    done
+    echo "Timed out waiting for app to be routable"
```
adrcunha: nit: keep the logging format, add a linefeed and "ERROR: " to the beginning of the message.

scripts/library.sh
srinivashegde86: removed the commands.

scripts/library.sh
srinivashegde86: done

scripts/library.sh
srinivashegde86: updated.

scripts/library.sh
srinivashegde86: done

adrcunha:
Took care of a few nits myself instead of bugging you about that. Feel free to make more changes. :)
 
/lgtm

srinivashegde86:
thanks :)

PR #129, opened 2018-09-18
add format and spelling check

islinwb:
There's no format and spelling check in CI of repo build, serving and eventing.
This is the first step of adding format and spelling check.

### DIFF
--- a/scripts/library.sh
+++ b/scripts/library.sh
@@ -315,6 +315,51 @@ function check_licenses() {
   run_dep_collector -check $@
 }
 
+# Run gofmt to check format of .go files
+function check_format() {
+  if [[ -z "${GOROOT}" ]]; then
+    local GOROOT=$(go env GOROOT)
+  fi
+  gofmt="${GOROOT}/bin/gofmt"
+
+  diff=$(  find . -not \( \
+      \( \
+        -wholename './output' \
+        -o -wholename './_output' \
+        -o -wholename './_gopath' \
+        -o -wholename './release' \
+        -o -wholename './target' \
+        -o -wholename '*/third_party/*' \
+        -o -wholename '*/vendor/*' \
+      \) -prune \
+    \) -name '*.go' \
+   | xargs ${gofmt} -d -s 2>&1) || true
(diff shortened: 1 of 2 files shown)


scripts/library.sh
```diff
@@ -315,6 +315,51 @@ function check_licenses() {
...
+    exit 1
+  fi
+  echo `date`,"No errors found."
+}
+
+# Run misspell to check the spelling
+function check_spelling() {
+  local misspell="$(which misspell)"
+  if [[ -z ${misspell} ]]; then
+    go get -u github.com/client9/misspell/cmd/misspell
```
srinivashegde86: this can be part of the prow(test-infra/images/prow-tests/) image, before running the tests

scripts/library.sh
```diff
@@ -315,6 +315,51 @@ function check_licenses() {
   run_dep_collector -check $@
 }
 
+# Run gofmt to check format of .go files
+function check_format() {
```
srinivashegde86: maybe we can create a prow plugin instead of hard-coding this for gofmt. Then we can /gofmt or /fmt on the PR and run it independently.

scripts/library.sh
srinivashegde86: k8s already has a script https://github.com/kubernetes/test-infra/blob/master/hack/verify-gofmt.sh that does gofmt. We can run that with Prow as well instead of creating a new plugin

scripts/library.sh
srinivashegde86: k8s already has a script https://github.com/kubernetes/test-infra/blob/master/hack/verify-gofmt.sh that does gofmt. We can run that with Prow as well instead of creating a new plugin

adrcunha:
I appreciate the contribution, but before going into the several code issues I saw (scope, logging format, redundant settings, severity, etc), is there an alternative to misspell? I'm a bit concerned that it has been getting dust for a while...

islinwb:
I'm really curious why so many repos use it if it's like what you said.
/close

PR #134, opened 2018-09-19
Add function for checking that code dirs have test

greghaynes:
There is a known issue where directories which contain code but no tests
are skipped in coverage reports. We work around this by adding a
stub_tests.go file in these directories.

### DIFF
--- a/scripts/library.sh
+++ b/scripts/library.sh
@@ -374,3 +374,20 @@ function lint_markdown() {
   # https://github.com/markdownlint/markdownlint
   run_lint_tool mdl "linting markdown files" "-r ~MD013" $@
 }
+
+# Check for subdirs which do not contain any test module but contain code.
+# This is to prevent an issue where these directories are skipped in coverage
+# reports resulting in inaccurately high coverage results. Fix by adding a
+# stub_tests.go with no tests to these directories.
+function check_subdirs_without_tests() {
+  ret=0
+  for dir in $(find pkg -type d); do
+    if ls -l $dir/*.go 1> /dev/null 2>&1; then
+      if ! ls -l $dir/*_test.go 1> /dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


srinivashegde86:
cc @steuhs 
do we need a separate check here or this is a missing feature in the coverage tool?

scripts/library.sh
```diff
@@ -374,3 +374,20 @@ function lint_markdown() {
   # https://github.com/markdownlint/markdownlint
   run_lint_tool mdl "linting markdown files" "-r ~MD013" $@
 }
+
+# Check for subdirs which do not contain any test module but contain code.
+# This is to prevent an issue where these directories are skipped in coverage
+# reports resulting in inaccurately high coverage results. Fix by adding a
+# stub_tests.go with no tests to these directories.
+function check_subdirs_without_tests() {
+  ret=0
```
adrcunha: `local` please.

scripts/library.sh
```diff
@@ -374,3 +374,20 @@ function lint_markdown() {
...
+  for dir in $(find pkg -type d); do
```
adrcunha: I assume you intended to replace `pkg` by `$1` here?

scripts/library.sh
adrcunha: `local` please.

jessiezcc:
@steuhs is solving the same problem with same solution AFAIK.

steuhs:
For the code coverage tool, I am planning on add _test.go to all .go file on the repo checked out locally. Once that is done, it will not only solve the issue mentioned here, but also cover all .go files. For now, this solution is a doable approach, but it will become redundant after the code coverage feature is updated accordingly.

adrcunha:
@steuhs can we close this PR now that your solution is implemented?

steuhs:
@adrcunha yes, we can close this
/close

PR #142, opened 2018-09-25
Remove SSH access to Knative repos for Prow jobs

adrcunha:
The repos are public, SSH access is not necessary anymore.

### DIFF
--- a/ci/prow/config.yaml
+++ b/ci/prow/config.yaml
@@ -1382,7 +1382,6 @@ periodics:
     - image: gcr.io/knative-tests/test-infra/prow-tests:latest
       imagePullPolicy: Always
       args:
-      - "--ssh=/etc/ssh-knative/ssh-knative"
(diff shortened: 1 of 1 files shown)


adrcunha:
#141 was merged, rebased.

/hold cancel

ci/prow/config.yaml
```diff
@@ -1651,8 +1354,6 @@ periodics:
   name: ci-knative-serving-go-coverage
   agent: kubernetes
   decorate: true
-  ssh_key_secrets:
-  - ssh-knative
   extra_refs:
   - org: knative
     repo: serving
```
srinivashegde86: below this there is the perf cron, which also needs to be updated. I cant comment on that line :)

adrcunha: That's weird, my local copy was updated and had no ssh entries anymore... Thanks for pointing it out, let me fix it again.

PR #167, opened 2018-10-02
Add trigger for integration tests

bobcatfish:
The integration tests entrypoint in `presubmit-tests.sh`
(https://github.com/knative/build-pipeline/blob/master/test/presubmit-tests.sh#L46)
currently will just echo a `TODO` but I'd like to add this to Prow
before adding the logic to `presubmit-tests.sh` (especially b/c I don't
think I'll have permission to manually run these tests against the
boskos clusters in
https://github.com/knative/test-infra/blob/master/ci/prow/boskos/resources.yaml).

This is part of https://github.com/knative/build-pipeline/issues/16

@adrcunha do we need to do anything special to make it so that integration tests run against `knative/build-pipeline` can use [the boskos clusters](https://github.com/knative/test-infra/blob/master/ci/prow/boskos/resources.yaml)? Or would you prefer that we setup a separate set of clusters? (In which case could we start off using the same cluster?)

### DIFF
--- a/ci/prow/config.yaml
+++ b/ci/prow/config.yaml
@@ -482,6 +482,39 @@ presubmits:
           requests:
             memory: "1Gi"
 
+  - name: pull-knative-build-pipeline-integration-tests
+    agent: kubernetes
+    context: pull-knative-build-pipeline-integration-tests
+    always_run: true
+    rerun_command: "/test pull-knative-build-pipeline-integration-tests"
+    trigger: "(?m)^/test (all|pull-knative-build-pipeline-integration-tests),?(\\s+|$)"
+    labels:
+      preset-service-account: "true"
+      preset-bazel-scratch-dir: "true"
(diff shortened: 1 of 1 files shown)


adrcunha:
1. boskos is magic, you shouldn't worry about it. ;)
1. I'll update the cluster with your config, then you should start seeing the jobs. Remember to mark them as required checks for merging in the repo config.

/approve
/lgtm

adrcunha:
> Where do the e2e test scripts call out to boskos?

Oh, I see that you don't have E2E tests yet. Specifically, `kubetest` requests a project from boskos when you create a test cluster. All this logic is already ready to use in `e2e-tests.sh` if you follow the recipe we use for all the repos.

adrcunha:
Ok, I found your PR for the E2E tests and added a gazillion comments, please don't kill me. XD

PR #169, opened 2018-10-03
Allow base name of resource to be longer

bobcatfish:
When trying to run this script for knative/build-pipeline, the resource
name was getting truncated too early: 1) the resource names were all the
same and 2) they were invalid because they happened to end with `-`.

This resulted in errors like:
```
ERROR: (gcloud.compute.networks.describe) Could not fetch resource:
 - Invalid value 'kbuild-pipeline-e2e-'. Values must match the following regular expression: '[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?'
```

30 chars is pretty arbitrary so open to better solutions also!

This is for https://github.com/knative/build-pipeline/issues/16

### DIFF
--- a/scripts/e2e-tests.sh
+++ b/scripts/e2e-tests.sh
@@ -47,11 +47,11 @@ source $(dirname ${BASH_SOURCE})/library.sh
 function build_resource_name() {
   local prefix=${E2E_BASE_NAME}-$1
   local suffix=${BUILD_NUMBER}
-  # Restrict suffix length to 20 chars
+  # Restrict suffix length to 30 chars
   if [[ -n "${suffix}" ]]; then
-    suffix=${suffix:${#suffix}<20?0:-20}
+    suffix=${suffix:${#suffix}<30?0:-30}
   fi
-  echo "${prefix:0:20}${suffix}"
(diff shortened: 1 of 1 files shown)


adrcunha:
This will break the E2E tests:

https://github.com/knative/test-infra/blob/master/scripts/e2e-tests.sh#L44

You mentioned that the resource names are the same and ending in `-`: looks like you're not providing the suffix parameter to the function. If you follow the way the other `e2e-tests.sh` scripts are written, you should have no problem with resource names.

bobcatfish:
Ah okay, you're totally right, the suffix is missing - and it looks like from the code this will always be `$BUILD_NUMBER`, which is provided by Prow?

So if I want to run this without Prow (e.g. locally) it looks like I'll need to provide `$BUILD_NUMBER`

adrcunha:
> Ah okay, you're totally right, the suffix is missing - and it looks like from the code this will always be $BUILD_NUMBER, which is provided by Prow?

Yes.

> So if I want to run this without Prow (e.g. locally) it looks like I'll need to provide $BUILD_NUMBER

No. `build_resource_name` already takes care of this case, see the documentation of the function.

PR #173, opened 2018-10-03
Add infrastructure for publishing named releases

adrcunha:
Bonuses:
* move documentation about `release.sh` to `README.md`
* add reference to the documentation in `e2e-tests.sh` and `release.sh`
* turn `run_dep_collector` into a more generic function
* add a few smoke tests for `release.sh`

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -24,10 +24,11 @@ of the cluster (tipically using `kubectl`).
 `DOCKER_REPO_OVERRIDE`, `K8S_CLUSTER_OVERRIDE` and `K8S_USER_OVERRIDE` will be set
 according to the test cluster. You can also use the following boolean (0 is false,
 1 is true) environment variables for the logic:
-  * `EMIT_METRICS`: true if `--emit-metrics` is passed.
-  * `USING_EXISTING_CLUSTER`: true if the test cluster is an already existing one,
+    * `EMIT_METRICS`: true if `--emit-metrics` was passed.
+    * `USING_EXISTING_CLUSTER`: true if the test cluster is an already existing one,
 and not a temporary cluster created by `kubetest`.
-All environment variables above are marked read-only.
+
+    All environment variables above are marked read-only.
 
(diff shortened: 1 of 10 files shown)


adrcunha:
/hold

(a new commit is coming to fix the README)

scripts/README.md
```diff
@@ -60,3 +61,61 @@ kubectl get pods || fail_test
...
+1. Source the script.
+
+1. Call the `parse_flags()` function passing `$@` (without quotes).
+
+1. Call the `run_validation_tests()` function passing the script or executable that
+runs the release validation tests. It will call the script to run the tests unless
+`--skip_tests` was passed.
+
+1. Write logic for the release process. Call `publish_yaml()` to publish the manifest(s),
+`tag_releases_in_yaml()` to tag the generated images, `branch_release()` to branch
```
jessiezcc: once the environment variable is set, why the caller still need to know which release function to call? 
Wouldn't it be better if there is a generic function say do_release() which will determine whether to call publish or tag or branch based on environment variable?

jessiezcc:
/hold 
to address one of my comment before merge

scripts/README.md
adrcunha: The idea is that this script is a helper, not a framework. I don't want to make release scripts rigid, that will just make the helper more complex and incorporate too specific bits. If you check the release scripts, you'll see that they have too different approaches, which is not really suitable for what you're proposing (I mean, not without the disadvantages I mentioned).

test/presubmit-tests.sh
```diff
@@ -37,7 +37,11 @@ function build_tests() {
 
 function unit_tests() {
   header "Running unit tests"
-  ./test/library-tests.sh
+  local failed=0
+  for test in library release; do
```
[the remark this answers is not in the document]
adrcunha: http://tldp.org/HOWTO/Bash-Prog-Intro-HOWTO-7.html

PR #176, opened 2018-10-05
Don't return non-zero when there isn't another tag.

mattmoor:


mattmoor:
@adrcunha needs your approval

PR #189, opened 2018-10-15
Don't remove bazel stuff after parsing results

adrcunha:
Otherwise Prow won't show the test results summary.


scripts/library.sh
```diff
@@ -235,13 +235,7 @@ function report_go_test() {
   echo "Finished run, return code is ${failed}"
   # Tests didn't run.
   [[ ! -s ${report} ]] && return 1
-  # Create WORKSPACE file, required to use bazel, if necessary
```
srinivashegde86: keep this comment :)

adrcunha: Done, thanks. :)

PR #194, opened 2018-10-16
Install netbase with prow

srinivashegde86:
wrk uses ntp and fails when it does not find the ntp socket addresses with error " Servname not supported for ai_socktype"

Install netbase in our prow image so that we add the network services in /etc/services/


images/prow-tests/Dockerfile
```diff
@@ -30,6 +30,7 @@ RUN apt-get install -y uuid-runtime  # for uuidgen
 RUN apt-get install -y npm  # for markdown-link-check
 RUN apt-get install -y rubygems  # for mdl
 RUN apt-get install -y build-essential libssl-dev # for wrk
+RUN apt-get install -y netbase # sets up /etc/services needed for network calls
```
adrcunha: I would just comment this as "for wrk". There's nothing else network related that needs netbase.

srinivashegde86: done

PR #196, opened 2018-10-17
Separate gcs and testgrid into its own packages

srinivashegde86:
This will enable us to use gcs and testgrid packages to run different tools other than apicoverage like the latency metrics and the perf test results


tools/apicoverage/apicoverage.go
```diff
@@ -176,18 +181,26 @@ func main() {
...
 
 	// Calculate coverage
 	coverage := initCoverage()
-	calculateCoverage(parseLog(ctx, fmt.Sprintf("%s/%d", sourceDir, latestBuild), false, coverage), coverage)
+	calculateCoverage(
+		gcs.ParseLog(
+			ctx,
+			fmt.Sprintf("%s/%d/%s", logDir, num, buildFile),
+			func(fields []string) *string {
+				var covLogs []string
```
adrcunha: Unnecessary?

tools/apicoverage/apicoverage.go
```diff
@@ -176,18 +181,26 @@ func main() {
...
 	}
 
 	// Calculate coverage
 	coverage := initCoverage()
-	calculateCoverage(parseLog(ctx, fmt.Sprintf("%s/%d", sourceDir, latestBuild), false, coverage), coverage)
+	calculateCoverage(
+		gcs.ParseLog(
+			ctx,
+			fmt.Sprintf("%s/%d/%s", logDir, num, buildFile),
+			func(fields []string) *string {
```
adrcunha: nit: for readability sake, I would prefer a separate function.

tools/apicoverage/apicoverage.go
```diff
@@ -176,18 +181,26 @@ func main() {
...
+				var covLogs []string
+				// I0727 16:23:30.055] info	TestRouteCreation	test/configuration.go:34	resource {<resource_name>: <val>}"}
+				if len(fields) == 7 && fields[2] == "info" && fields[5] == "resource" {
```
adrcunha: https://github.com/knative/pkg/pull/122 will break this, I suggest you already deal with it now. See cl/217152312.

tools/apicoverage/testgrid.go
```diff
@@ -31,63 +31,21 @@ const (
 	overallService = "OverallService"
 )
```
adrcunha: I wonder if moving the remaining two functions to `apicoverage.go` wouldn't be better.

tools/gcs/gcs.go
```diff
@@ -69,33 +84,30 @@ func readGcsFile(ctx context.Context, filename string, sa string) ([]byte, error
 	return contents, nil
 }
 
-func parseLog(ctx context.Context, dir string, isLocal bool, coverage *OverallAPICoverage) []string {
-	var covLogs []string
+// ParseLog parses the log and returns the lines where the checkLog func does not return an empty slice.
+// checkLog function should take in the log statement and return a part from that statement that should be in the log output.ParseLog
```
adrcunha: s/output.ParseLog/output ?

tools/apicoverage/apicoverage.go
adrcunha: Please keep in mind that if you rebuild the apicoverage image before knative/pkg is updated in knative/serving, your metrics aggregation will be broken.

adrcunha:
/hold
/lgtm

Holding in case you want to address my comment about the knative/pkg update.

srinivashegde86:
I did address that and updated the relevant fields.

srinivashegde86:
oh did not see the update. I can use the older image to get the calculations.

PR #198, opened 2018-10-18
Allow defining custom flags for `e2e-tests.sh`

adrcunha:


scripts/README.md
```diff
@@ -50,11 +58,23 @@ function teardown() {
   echo "TODO: tear down test resources"
 }
 
+function parse_flags() {
+  if [[ "$1" == "--dont-wait-for-knative" ]]; then
```
srinivashegde86: we should call this --no-knative-wait instead? Shorter and follows the no-* convention

adrcunha: It's just an example, but anyway... Done.

PR #202, opened 2018-10-18
Update prow and testgrid configs to show perf latency numbers

srinivashegde86:
Prow config: Add a periodic job to parse the build-log from the daily perf run and get the latency numbers in the testgrid xml. This job is created in https://github.com/knative/test-infra/pull/201

Testgrid config: Add a new tab in knative/serving for showing performance latency from the xml generated from the tool and run in the above prow job.

Fixes knative/serving#1981

### DIFF
--- a/ci/testgrid/config.yaml
+++ b/ci/testgrid/config.yaml
@@ -68,6 +68,9 @@ test_groups:
 - name: ci-knative-serving-api-coverage
   gcs_prefix: knative-prow/logs/ci-knative-serving-api-coverage
   short_text_metric: api_coverage
+- name: ci-knative-serving-performance
+  gcs_prefix: knative-prow/logs/ci-knative-serving-performance-tool
+  short_text_metric: perf_latency
 - name: ci-knative-build-continuous
   gcs_prefix: knative-prow/logs/ci-knative-build-continuous
 - name: ci-knative-build-release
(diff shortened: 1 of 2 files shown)


ci/testgrid/config.yaml
```diff
@@ -125,6 +128,10 @@ dashboards:
     test_group_name: ci-knative-serving-api-coverage
     description: 'Conformance tests API coverage.'
     base_options: 'exclude-filter-by-regex=Overall$&group-by-directory=&expand-groups=&sort-by-name='
+  - name: performance
```
jessiezcc: now we have both Latency and Performance, might cause confusion. Should we rename Latency to something else?

ci/testgrid/config.yaml
```diff
@@ -125,6 +128,10 @@ dashboards:
...
+    test_group_name: ci-knative-serving-performance-tool
+    description: 'Performance tests numbers'
```
jessiezcc: can we replace numbers with something more meaningful such as duration or latency or else?

ci/testgrid/config.yaml
srinivashegde86: maybe we can rename latency to E2E latency or something like that to indicate that the numbers are from e2e tests?

srinivashegde86:
This is a temp solution to surface numbers from the one test we have. We will be updating it once we have the common framework to write tests and get the numbers.

ci/testgrid/config.yaml
jessiezcc: how about "E2E Metric" vs. "Perf Benchmark"?

srinivashegde86:
We dont need to update prow to create another job. 

This should now be handled by knative/serving#2320 by using the same job. Closing this PR.

PR #206, opened 2018-10-23
Bump hook to v20181023-ca14137 to enable review_acts_as_approve

adrcunha:
/cc @grantr


adrcunha:
I doesn't, I see my self-approve and yours. Yay! 🎉

PR #208, opened 2018-10-24
Expose docker daemon in knative/build e2e tests

dprotaso:
We're building a new build-base image in our e2e tests

Unblocks -  https://github.com/knative/build/pull/437

### DIFF
--- a/ci/prow/config.yaml
+++ b/ci/prow/config.yaml
@@ -359,6 +359,7 @@ presubmits:
       preset-service-account: "true"
       preset-bazel-scratch-dir: "true"
       preset-bazel-remote-cache-enabled: "true"
+      preset-dind-enabled: "true"
     spec:
       containers:
       - image: gcr.io/knative-tests/test-infra/prow-tests:latest


ci/prow/config.yaml
```diff
@@ -359,6 +359,7 @@ presubmits:
       preset-service-account: "true"
       preset-bazel-scratch-dir: "true"
       preset-bazel-remote-cache-enabled: "true"
+      preset-dind-enabled: "true"
```
jessiezcc: does it have any implication to other repos who don't need this tag?

dprotaso: I don't believe so as this is only for `pull-knative-build-integration-tests` job

dprotaso:
@jessiezcc thanks - when would this change take effect?

ci/prow/config.yaml
adrcunha: That's correct. It's a setting of a single job.

adrcunha:
I don't know if anyone else updated the Prow cluster with the new config while I was out, but I just did it.

PR #213, opened 2018-10-25
This moves the build of dep-collector into the Prow Dockerfile.

mattmoor:


mattmoor:
nudge (make gubernator clear this)

PR #217, opened 2018-10-27
Experiment with increasing the max nodes to 4

tcnghia:
I want to experiment with an increase in max nodes to see if that reduces flakes.

### DIFF
--- a/scripts/e2e-tests.sh
+++ b/scripts/e2e-tests.sh
@@ -41,7 +41,7 @@ readonly E2E_CLUSTER_NAME=$(build_resource_name e2e-cls)
 readonly E2E_NETWORK_NAME=$(build_resource_name e2e-net)
 readonly E2E_CLUSTER_REGION=us-central1
 readonly E2E_CLUSTER_ZONE=${E2E_CLUSTER_REGION}-a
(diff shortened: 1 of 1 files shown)


adrcunha:
I don't think this is the best place for such test, since the test-infra e2e test doesn't exercise knative itself. I would open a PR in knative/serving and hack `vendor/github.com/knative/test-infra/scripts/e2e-tests.sh` instead. If you want to run the tests several times, I would do it on my workstation, in a loop.

tcnghia:
I was hoping to check this in and update the `knative-serving` to use this change so that everyone's presubmit will get 4 pods.

On another thought, would it be good to allow this to be set in downstream script (in `knative-serving`, `knative-eventing`, etc ...?) since the needs may vary?

adrcunha:
> I was hoping to check this in and update the knative-serving to use this change so
> that everyone's presubmit will get 4 nodes.

Yes, you can do that, it will be just slower and hard to roll back if you have to...

> On another thought, would it be good to allow this to be set in downstream script (in knative-serving,
> knative-eventing, etc ...?) since the needs may vary?

Sounds like the best solution to me. :)

adrcunha:
This is already implemented, closing.

PR #228, opened 2018-10-31
When branching a release, accept YAMLs in any directory

adrcunha:
Bonus: normalize test banners.


adrcunha:
/hold

This requires a fix for `run_go_tool`.

adrcunha:
The fix is https://github.com/knative/test-infra/pull/229

PR #236, opened 2018-11-02
handle USING_EXISTING_CLUSTER teardown consistently

### LINKED ISSUE #235: EXISTING_CLUSTER still tears down the environment
Setting the K8S_CLUSTER_OVERRIDE is not being honored.

vaikas@penguin:~/projects/go/src/github.com/knative/serving$ ./test/e2e-tests.sh --run-tests
- Cluster is gke_quantum-reducer-434_us-west1-knative-eventing-e2e
- User is vaikas@google.com
- Docker is us.gcr.io/quantum-reducer-434
Deleting any previous SUT instance
>> Removing test resources (test/config/)
>> Bringing down Serving
namespace "knative-serving" deleted
clusterrole.rbac.authorization.k8s.io "knative-serving-admin" deleted
clusterrole.rbac.authorization.k8s.io "knative-serving-write" deleted
serviceaccount "controller" deleted
serviceaccount "autoscaler" deleted
clusterrolebinding.rbac.authorization.k8s.io "knative-serving-controller-admin" deleted
clusterrolebinding.rbac.authorization.k8s.io "knative-serving-autoscaler-write" deleted
gateway.networking.istio.io "knative-shared-gateway" deleted
service "knative-ingressgateway" deleted
deployment.extensions "knative-ingressgateway" deleted

vaikas-google:
Fixes #235 

Note that on Line 52 it's being handled differently and seems to work there.

### DIFF
--- a/scripts/e2e-tests.sh
+++ b/scripts/e2e-tests.sh
@@ -249,7 +249,7 @@ function setup_test_cluster() {
 
   trap teardown_test_resources EXIT
 
-  if (( USING_EXISTING_CLUSTER )) && function_exists teardown; then
+  if (( ! USING_EXISTING_CLUSTER )) && function_exists teardown; then
     echo "Deleting any previous SUT instance"
     teardown
   fi


adrcunha:
So, what this line is doing is deleting a previous SUT instance before a new one is installed. Line 52 is executed after the test finishes, and it deletes the SUT we created for the test (but only if we're using an exclusive, temporary test cluster).

Currently use cases are:
* I just want to run tests, let the script create a temp cluster for me and run them. In this case, the cluster is new so there's no cleanup required before running the tests, but there's a cleanup required after the tests.
* I want to run the tests against my existing k8s cluster. In this case, the cluster might be dirty, so the test script cleans it up first (but not at the end of the test, leaving everything untouched).

Looks like your use case is:
* I want to run the tests against my existing Knative installation. In this case, I don't want the test script to touch anything (no cleanup before or after, and no Knative installation of any kind).

If so, you should be able to do, say

`go test -v -tags=e2e -count=1 ./test/e2e -run ^TestKubernetesEvents$`

as explained in https://github.com/knative/eventing/tree/master/test#running-end-to-end-tests, assuming that your environment is setup (`K8S_CLUSTER_OVERRIDE` & co).

Looks like the source of the confusion is the documentation on eventing being unclear; knative/serving has a more detailed and up-to-date content.

vaikas-google:
Awesome! Thanks much!!! Yeah, there's some conflicting documentation. I guess I'm still confused if that particular line is wrong or not, seems like the condition is wrong?

adrcunha:
Those lines are OK. You should be able to run your tests against your cluster with the `go test` command above.

adrcunha:
Closing this in favor of https://github.com/knative/eventing/issues/581

PR #238, opened 2018-11-05
Move dummy.go to scripts/*

srinivashegde86:


scripts/dummy.go
```diff
@@ -0,0 +1,26 @@
...
+*/
+
+package scripts
+
+import (
+	"fmt"
+)
+
+func main() {
+	fmt.Println("This is a dummy go file so `go dep` can be used with knative/test-infra repo")
```
adrcunha:
```suggestion
	fmt.Println("This is a dummy go file so `go dep` can be used with knative/test-infra/scripts")
```

scripts/dummy.go
```diff
@@ -0,0 +1,26 @@
...
+	fmt.Println("This file can be removed once the repo contains real, useful go code in the root dir")
```
adrcunha:
```suggestion
	fmt.Println("This file can be safely removed if one day this directory contains real, useful go code")
```

adrcunha:
As synced offline, let's not remove the `dummy.go` file from the root dir to avoid breaking the other repos.

PR #240, opened 2018-11-05
Clearly display test failures on releases

adrcunha:
AKA "display a banner when tests failed for a release, instead of just silently aborting the release"

Bonus: factor out common code in tests.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -136,7 +136,10 @@ function run_validation_tests() {
   if (( ! SKIP_TESTS )); then
     banner "Running release validation tests"
     # Run tests.
-    $1
+    if ! $1; then
+      banner "Release validation tests failed, aborting"
+      exit 1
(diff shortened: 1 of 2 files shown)


scripts/release.sh
```diff
@@ -136,7 +136,10 @@ function run_validation_tests() {
   if (( ! SKIP_TESTS )); then
     banner "Running release validation tests"
     # Run tests.
-    $1
+    if ! $1; then
+      banner "Release validation tests failed, aborting"
+      exit 1
```
grantr: Is this additional exit necessary? Does it clobber the exit status of the test command?

adrcunha: It's not additional, and it ensures that the flow stops here. Technically it does clobber the status of the test command, but right now we're not using the status code to take any actions (except signaling a failure).

PR #243, opened 2018-11-06
Fix removal of -v flag in parameters

mgencur:
Remove `-v` only if it's a standalone flag. IOW, do not remove when it's part of a different string parameter (e.g. `registry.svc.ci.namespace.org/ci-op-v5x1m4tr/stable:queue`). Previously, it would remove -v from the ci-op-v5x1m4tr string as well, resulting in ci-op5x1m4tr.

### DIFF
--- a/scripts/library.sh
+++ b/scripts/library.sh
@@ -235,8 +235,8 @@ function acquire_cluster_admin_role() {
 function report_go_test() {
   # Run tests in verbose mode to capture details.
   # go doesn't like repeating -v, so remove if passed.
-  local args=("${@/-v}")
-  local go_test="go test -race -v ${args[@]}"
+  local args=" $@ "
+  local go_test="go test -race -v ${args/ -v / }"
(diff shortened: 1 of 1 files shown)


scripts/library.sh
```diff
@@ -235,7 +235,7 @@ function acquire_cluster_admin_role() {
 function report_go_test() {
   # Run tests in verbose mode to capture details.
   # go doesn't like repeating -v, so remove if passed.
-  local args=("${@/-v}")
+  local args=("${@/#-v}")
```
adrcunha: Thanks for catching this. However, this will only remove `-v` if it's at the front of the string. I suggest doing:

```
local args=" $@ "
local go_test="go test -race -v ${args/ -v / }"
```

This will correctly remove the first `-v` flag in the arguments.

mgencur:
Changed as suggested.

adrcunha:
Thanks!

/approve
/lgtm

Issue #244, opened 2018-11-06
API coverage flow is stale


adrcunha:
There are no results for the past 14 days:

![image](https://user-images.githubusercontent.com/35786489/48083655-9c6ed800-e1aa-11e8-911d-8b8a8026ea38.png)

/assign @srinivashegde86

adrcunha:
@srinivashegde86 I was about to close this as obsolete, but testgrid is missing runs from 12/04 an 12/05.

PR #250, opened 2018-11-08
Add more generic methods to testgrid

### LINKED ISSUE #241: Testgrid pkg should handle local storage with the artifacts dir
https://github.com/knative/test-infra/blob/master/tools/testgrid/testgrid.go#L53 currently accepts an artifacts directory where it creates the xml output. This should have one way of specifying local storage and not depend on the caller to specify this location.

For eg - https://github.com/knative/serving/pull/2320/files#r230900752

srinivashegde86:
Also, passes in the correct path for getting logs for apicoverage.

Fixes #241, #244

### DIFF
--- a/tools/gcs/gcs.go
+++ b/tools/gcs/gcs.go
@@ -50,8 +50,9 @@ func createStorageObject(filename string) *storage.ObjectHandle {
 
 // GetLatestBuildNumber gets the latest build number for the specified log directory
 func GetLatestBuildNumber(ctx context.Context, logDir string, sa string) (int, error) {
-	log.Printf("%s %s", logDir, latest)
-	contents, err := ReadGcsFile(ctx, logDir+latest, sa)
+	logFilePath := logDir + latest
+	log.Printf("Using %s to get latest build number", logFilePath)
+	contents, err := ReadGcsFile(ctx, logFilePath, sa)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


tools/gcs/gcs.go
```diff
@@ -50,6 +50,7 @@ func createStorageObject(filename string) *storage.ObjectHandle {
 
 // GetLatestBuildNumber gets the latest build number for the specified log directory
 func GetLatestBuildNumber(ctx context.Context, logDir string, sa string) (int, error) {
+	log.Printf("%s %s", logDir, latest)
```
adrcunha: Better add context to this log entry.

tools/testgrid/testgrid.go
```diff
@@ -49,6 +49,22 @@ type TestSuite struct {
 	TestCases []TestCase `xml:"testcase"`
 }
 
+// GetArtifactsDir gets the aritfacts directory where we should put the artifacts.
+// By default, it will look at the env var ARTIFACTS.
+func GetArtifactsDir() string {
+	dir := os.Getenv("ARTIFACTS")
+	if dir == "" {
+		return "./artifacts"
```
adrcunha: For the sake of debugging, add a log here saying we're using the hardcoded `./artifacts` dir.

PR #251, opened 2018-11-09
Remove wrk from prow image

srinivashegde86:
we dont use wrk now for perf tests. Remove this from the image.


adrcunha:
Yay!!

/lgtm

PR #255, opened 2018-11-14
Allow specifying a GCP project for E2E tests using a flag

adrcunha:
`--gcp-project` works locally or on Prow, `$PROJECT_ID` still works locally.

### DIFF
--- a/scripts/e2e-tests.sh
+++ b/scripts/e2e-tests.sh
@@ -45,6 +45,9 @@ readonly E2E_CLUSTER_NODES=3
 readonly E2E_CLUSTER_MACHINE=n1-standard-4
 readonly TEST_RESULT_FILE=/tmp/${E2E_BASE_NAME}-e2e-result
 
+# Flag whether test is using a boskos GCP project
+IS_BOSKOS=0
+
 # Tear down the test resources.
 function teardown_test_resources() {
   header "Tearing down test environment"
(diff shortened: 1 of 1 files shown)


scripts/e2e-tests.sh
```diff
@@ -319,22 +331,41 @@ function initialize() {
...
     esac
     shift
   done
+
+  # Use PROJECT_ID if set, unless --gcp-project was used.
+  if [[ -n "${PROJECT_ID:-}" && -z "${GCP_PROJECT}" ]]; then
+    echo "\$PROJECT_ID is set to '${PROJECT_ID}', using it to run the tests"
+    GCP_PROJECT="${PROJECT_ID}"
+  fi
+  if (( ! IS_PROW )) && [[ -z "${GCP_PROJECT}" ]]; then
```
srinivashegde86: should this be IS_BOSKOS as well?

scripts/e2e-tests.sh
```diff
@@ -45,6 +45,9 @@ readonly E2E_CLUSTER_NODES=3
 readonly E2E_CLUSTER_MACHINE=n1-standard-4
 readonly TEST_RESULT_FILE=/tmp/${E2E_BASE_NAME}-e2e-result
 
+# Flag whether test is using a boskos GCP project
+IS_BOSKOS=0
```
srinivashegde86: when will this be not 0?

scripts/e2e-tests.sh
adrcunha: No. IS_BOSKOS is set later.

scripts/e2e-tests.sh
adrcunha: See line 362.

PR #267, opened 2018-11-20
Enable and configure milestone plugin

grantr:
The milestone plugin enables milestone maintainers (members of the Knative Milestone Maintainers team, TBD) to use the `/milestone v1.2.3` command to assign an issue to a milestone. Milestones must be created in advance by a user with write access.

This won't actually work until the maintainers team is created and its team id is filled in.

/cc @mattmoor


grantr:
I added the real team id, PTAL.

PR #273, opened 2018-11-21
Append to the file if existing when creating testgrid output

srinivashegde86:
Also, add a test for testgrid to verify that we can create and append the xml output file


tools/testgrid/testgrid.go
```diff
@@ -25,9 +25,9 @@ import (
 )
 
 const (
-	// Default filename to store output that acts as input to testgrid.
+	// Filename to store output that acts as input to testgrid.
 	// Should be of the form junit_*.xml
-	filename = "/junit_knative.xml"
+	Filename = "/junit_knative.xml"
```
adrcunha: Removing the / from here, and add it to the relevant places to avoid surprises when using this const.

tools/testgrid/testgrid_test.go
```diff
@@ -0,0 +1,47 @@
...
+		t.Fatalf("Default value is %s and not artifacts", v)
+	}
+}
+
+func TestXMLOutput(t *testing.T) {
+	// Create a test file
+	if err := testgrid.CreateXMLOutput(testgrid.TestSuite{}, "."); err != nil {
+		t.Fatalf("Error when creating xml output file: %v", err)
+	}
+
```
adrcunha: You should check the contents of the file here and on line 42 to ensure the operations are working as expected.

adrcunha:
One last request: please add the tests to the presubmit

https://github.com/knative/test-infra/blob/6a266d8824043c54dfbcaca688b826458b77cb04/test/presubmit-tests.sh#L45

I think that a ```go test ./... || failed=1``` before the return should do.

adrcunha:
/lgtm

Thanks!

tools/testgrid/testgrid.go
```diff
@@ -80,9 +80,9 @@ func CreateXMLOutput(ts TestSuite, artifactsDir string) error {
 		return err
 	}
 
-	outputFile := artifactsDir + filename
+	outputFile := artifactsDir + "/" + Filename
```
jessiezcc: what was broken before this fix?

jessiezcc: got it. Will the file gets too big if we keep appending?

adrcunha: They can be as big as one want. It's the caller responsibility to keep that in mind. But I don't see that as an issue in the infra.

Issue #276, opened 2018-11-21
Create a job to cleanup boskos instances


adrcunha:
Instead of deleting the images in GCR after the test finishes, have a job to perform it from time to time (and only old images).

chaodaiG:
update, test-infra updated in all other knative repos

PR #277, opened 2018-11-21
Don't delete images after an E2E test finishes

adrcunha:
Part of https://github.com/knative/test-infra/issues/276

### DIFF
--- a/scripts/e2e-tests.sh
+++ b/scripts/e2e-tests.sh
@@ -56,15 +56,8 @@ function teardown_test_resources() {
     teardown
   fi
 
-  # Delete Knative Serving images when using boskos.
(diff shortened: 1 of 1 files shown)


scripts/e2e-tests.sh
```diff
@@ -56,15 +56,8 @@ function teardown_test_resources() {
     teardown
   fi
 
-  # Delete Knative Serving images when using boskos.
-  if (( IS_BOSKOS )); then
-    echo "Images in ${DOCKER_REPO_OVERRIDE}:"
```
jessiezcc: as long as it's not knative releases folder, it's still ok to delete, right? we don't have to block delete for all cases

PR #279, opened 2018-11-27
Use nightly images from knative-nightly

adrcunha:
Part of https://github.com/knative/serving/issues/2555.

Bonus: block running E2E tests using knative-nightly GCR as well.


adrcunha:
The new GCR/GCS is now populated.

/retest

PR #286, opened 2018-12-01
complete source code for code coverage project for Knative repos

steuhs:
Fixes #


steuhs:
> /hold
> 
> Please put it inside tools/, not at the root level.

@adrcunha done

coverage/main_test.go
```diff
@@ -0,0 +1 @@
+package main
```
adrcunha: This seems to be a leftover.

adrcunha:
Note: as previously agreed, the code in this PR was not reviewed. That was a one-time exception so we could get the source out of Stephen's personal repository and into test-infra ASAP, as properly reviewing it would take some time.

PR #288, opened 2018-12-03
Allow setting GCR/GCS on command line for releases

adrcunha:
Using environment variables can be dangerous, flags ensure we're explicitly setting values.
The default values point to the nightly GCR/GCS.

This also implies:
* no need for scripts to set `DOCKER_REPO_OVERRIDE` or `KO_DOCKER_REPO` by themselves
* helper functions use the values set by the flags

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -163,7 +163,7 @@ also overrides the value of the `TAG` variable as `v<version>`.
 it's empty and `master` HEAD will be considered the release branch.
     * `RELEASE_NOTES`: contains the filename with the release notes if `--release-notes`
 was passed. The release notes is a simple markdown file.
-    * `RELEASE_GCS`: contains the GCS bucket name to store the manifests if
+    * `RELEASE_GCS_BUCKET`: contains the GCS bucket name to store the manifests if
 `--release-gcs` was passed, otherwise the default value `knative-nightly/<repo>` will be
 used. It is empty if `--publish` was not passed.
     * `KO_DOCKER_REPO`: contains the GCR to store the images if `--release-gcr` was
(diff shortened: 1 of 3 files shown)


scripts/release.sh
```diff
@@ -80,6 +88,18 @@ function parse_flags() {
...
+        shift
+        [[ $# -ge 1 ]] || abort "missing GCR after --release-gcr"
+        KO_DOCKER_REPO=$1
+        has_gcr_flag=1
+        ;;
+      --release-gcs)
+        shift
+        [[ $# -ge 1 ]] || abort "missing GCS bucket after --release-gcs"
+        RELEASE_GCS=$1
+        has_gcs_flag=2
```
srinivashegde86: val can be 1

scripts/README.md
```diff
@@ -163,8 +163,14 @@ also overrides the value of the `TAG` variable as `v<version>`.
 it's empty and `master` HEAD will be considered the release branch.
     * `RELEASE_NOTES`: contains the filename with the release notes if `--release-notes`
 was passed. The release notes is a simple markdown file.
+    * `RELEASE_GCS`: contains the GCS bucket name to store the manifests if
```
srinivashegde86: can we call this RELEASE_GCS_BUCKET instead? makes it more explicit that we are referring to the GCS bucket

scripts/release.sh
adrcunha: Copy pasta... :P Done.

scripts/README.md
adrcunha: Makes sense. Done.

PR #289, opened 2018-12-03
Update serving continous run to every 2 hours

srinivashegde86:
Also, update timeout to 100 min.


adrcunha:
Looks like tests for `report_go_test()` are flaky.

/lgtm
/retest

PR #291, opened 2018-12-04
Fix image build in release

adrcunha:
`$KO_DOCKER_REPO` must be exported, or `ko` won't see it.

Bonuses:
* make `publish_yaml` more verbose
* log the config before building, for easier debugging/investigation/auditing

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -25,7 +25,7 @@ function banner() {
     make_banner "@" "$1"
 }
 
-# Tag images in the yaml file.
+# Tag images in the yaml file if $TAG is not empty.
 # $KO_DOCKER_REPO is the registry containing the images to tag with $TAG.
 # Parameters: $1 - yaml file to parse for images.
 function tag_images_in_yaml() {
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -30,9 +30,8 @@ function banner() {
 # Parameters: $1 - yaml file to parse for images.
 function tag_images_in_yaml() {
   [[ -z ${TAG} ]] && return 0
-  local src_dir="${GOPATH}/src/"
-  local BASE_PATH="${REPO_ROOT_DIR/$src_dir}"
-  local DOCKER_BASE="${KO_DOCKER_REPO}/${BASE_PATH}"
+  local SRC_DIR="${GOPATH}/src/"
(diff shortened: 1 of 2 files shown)


scripts/release.sh
```diff
@@ -43,9 +42,14 @@ function tag_images_in_yaml() {
 # If $TAG is not empty, also copy it to $RELEASE_GCS_BUCKET bucket's "previous" directory.
```
srinivashegde86: Update documentation

scripts/release.sh
```diff
@@ -174,10 +178,26 @@ function run_validation_tests() {
...
+    echo "- Artifacts WILL NOT be tagged"
+  fi
   if (( PUBLISH_RELEASE )); then
-    echo "- Destination GCS: ${RELEASE_GCS_BUCKET}"
+    echo "- Release WILL BE published to '${RELEASE_GCS_BUCKET}'"
+  else
+    echo "- Release will not be published"
   fi
+  if (( BRANCH_RELEASE )); then
+    echo "- Release WILL BE branched from '${RELEASE_BRANCH}'"
```
srinivashegde86: can we add this to the next if condition for branch_release.

scripts/release.sh
adrcunha: The documentation is up-to-date. What am I missing?

adrcunha: Hmmm... Maybe you mean the documentation about `tag_images_in_yaml()`?

scripts/release.sh
adrcunha: I don't want to, since these are two distinct functional blocks: one for logging purposes only, the other for actions.

srinivashegde86: okies.sg.

scripts/release.sh
adrcunha: Updated the documentation about tag_images_in_yaml().

srinivashegde86: Sorry, I meant publish_yaml. It only has $1 param in doc and you are also using $2

adrcunha: Read the code again, you missed the internal helper function `verbose_gsutil_cp()`.

srinivashegde86: oops. did not realize that it was an internal function

PR #293, opened 2018-12-04
Add performance wrapper on fortio

srinivashegde86:
This PR moves the fortio wrapper used for load testing in the test-infra repo. All the knative repos starting with serving will start using this to run their load tests.

### DIFF
--- a/performance/performance.go
+++ b/performance/performance.go
@@ -1,9 +1,12 @@
 /*
 Copyright 2018 The Knative Authors
+
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
+
     http://www.apache.org/licenses/LICENSE-2.0
+
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
--- a/performance/performance.go
+++ b/performance/performance.go
@@ -26,6 +26,12 @@ import (
 	"fortio.org/fortio/periodic"
 )
 
+const (
+	p50 = 50.0
+	p90 = 90.0
+	p99 = 99.0
+)
+
 // PerfOptions provides knobs to run the perf test
 type PerfOptions struct {
 	Duration       time.Duration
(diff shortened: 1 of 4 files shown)


performance/performance.go
```diff
@@ -0,0 +1,70 @@
+/*
```
adrcunha: mega-nit: please fix the boilerplate spacing, it hurts my eyes. :) :) :)

performance/performance_test.go
```diff
@@ -0,0 +1,69 @@
+/*
```
adrcunha: Ditto.

performance/performance.go
```diff
@@ -0,0 +1,70 @@
...
+	// If the url does not contains a resolvable domain, we need to add the domain as a header
+	if !resolvableDomain {
+		o.AddAndValidateExtraHeader(fmt.Sprintf("Host: %s", p.Domain))
+	}
+
+	return &fhttp.HTTPRunnerOptions{
+		RunnerOptions: periodic.RunnerOptions{
+			Duration:    p.Duration,
+			NumThreads:  p.NumThreads,
+			Percentiles: []float64{50.0, 90.0, 99.0},
```
adrcunha: Since we're here, what about making these constants for better readability?

adrcunha:
/lgtm

I'll hold for Dushyanth to take a look, and also in case you want to address my comments, but if you're in a hurry, unhold it.

/hold

performance/performance_test.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	if opts.HTTPOptions.HTTPReqTimeOut != testTime {
+		t.Fatalf("Request timeout is %v. Expected %v", opts.HTTPOptions.HTTPReqTimeOut, testTime)
+	}
+
+	if opts.HTTPOptions.URL != testUrl {
+		t.Fatalf("Url is %s. Expected %s", opts.HTTPOptions.URL, testUrl)
+	}
+}
```
dushyanthsc: Change looks good, but I do think that test is an overkill, because you are just assigning the values and test for the most part is verifying the assignments.

srinivashegde86: I debated about that as well :) But, I feel like we should not test the RunLoadTest() method from fortio and should only test that we are sending the correct values to it. This may be an overkill, but we can expand on this, as we add more features from fortio.

adrcunha:
I just noticed that this is a top level dir. This should go into a `src/performance` dir.

alculquicondor:
Maybe in `tools/` instead? But also, I would use the name `load` instead of `performance`, because it shouldn't be restricted to performance tests.
For instance, `serving/e2e/autoscale_test.go` should start using this wrapper too.

srinivashegde86:
Moved under tools and renamed to loadgenerator to make it more generic. PTAL.

PR #310, opened 2018-12-11
Add support to automated "dot" releases in release.sh

adrcunha:
"Dot" releases, enabled by the `--dot-release` flag, are incremental builds on a branch (e.g., v0.2.1, v0.2.2, etc).
The version of the last release published on GitHub is incremented (release branch is automatically detected), unless the release branch has no new commits.
A "dot" release can be restricted to a branch by using the `--branch` flag.
Release notes are copied from the previous "dot" release, unless the `--release-notes` flag is used.

Bonus: Added the flag `--github-token` so the script can be used with service accounts.

Part of https://github.com/knative/serving/issues/2427

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -43,7 +43,7 @@ function tag_images_in_yaml() {
 # Parameters: $1 - yaml file to copy.
 function publish_yaml() {
   function verbose_gsutil_cp {
-    local DEST=gs://${RELEASE_GCS_BUCKET}/$2/
+    local DEST="gs://${RELEASE_GCS_BUCKET}/$2/"
     echo "Publishing $1 to ${DEST}"
     gsutil cp $1 ${DEST}
   }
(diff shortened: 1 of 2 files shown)


scripts/release.sh
```diff
@@ -122,6 +154,51 @@ function parse_flags() {
     shift
   done
 
+  # Setup dot releases
+  if (( is_dot_release )); then
+    echo "Dot release requested"
+    TAG_RELEASE=1
+    PUBLISH_RELEASE=1
+    # List latest release
+    local releases
```
srinivashegde86: [nit] can be combined local releases="$(hub tool release)"

scripts/release.sh
```diff
@@ -122,6 +154,51 @@ function parse_flags() {
     shift
   done
 
+  # Setup dot releases
+  if (( is_dot_release )); then
+    echo "Dot release requested"
+    TAG_RELEASE=1
+    PUBLISH_RELEASE=1
```
srinivashegde86: do we allow not publishing for dot releases? What happens if I set --dot-release and --nopublish?

scripts/release.sh
adrcunha: No, it can't. If you do that the successful attribution will override the subshell return code, and you'll never catch errors from hub.

scripts/release.sh
adrcunha: Like any named releases, publishing is mandatory. This is explained in the documentation in each repo, and I intend to do the same about `--dot-release`. For example: https://github.com/knative/eventing/blob/master/hack/release.md#creating-versioned-releases

PR #312, opened 2018-12-12
 Add --service-account flag to cleanup.sh

adrcunha:
This is required when running the script from Prow jobs, to establish the authentication.

Bonuses:
* used the term "gcr" instead of project to clearly differentiate from the GCP projects parsed from a YAML file
* separated the deletion functions for testing purposes
* assorted code refactoring, documentation, simplification, fixing and nitpicking


chaodaiG:
/lgtm

nice work, good to learn

PR #315, opened 2018-12-12
Add Chao to OWNERS

adrcunha:


OWNERS
```diff
@@ -4,6 +4,7 @@ approvers:
 - adrcunha
 - dushyanthsc
 - ericKlawitter
+- chaodaig
```
srinivashegde86: can you add nikhil as well :)

srinivashegde86:
/lgtm
/hold
cancel and merge if we want to create another CL to add nikhil.

adrcunha:
/hold cancel

Let's do one thing at a time.

OWNERS
adrcunha: I prefer doing that later.

PR #317, opened 2018-12-14
x509: CPU denial of service in chain validation[CVE-2018-16875]

zouyee:
From golang/go#29233

> Package crypto/x509 parses and validates X.509-encoded keys and certificates. It's supposed to handle certificate chains provided by an attacker with reasonable resource use.

> The crypto/x509 package does not limit the amount of work performed for each chain verification, which might allow attackers to craft pathological inputs leading to a CPU denial of service. Go TLS servers accepting client certificates and TLS clients verifying certificates are affected.

Go 1.11.3 and 1.10.6 have been released with this fixed.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -17,16 +17,16 @@
 
 http_archive(
     name = "io_bazel_rules_go",
-    sha256 = "1868ff68d6079e31b2f09b828b58d62e57ca8e9636edff699247c9108518570b",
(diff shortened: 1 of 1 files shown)


adrcunha:
/hold

Thanks, @zouyee. We don't have any running go server in test-infra so we're good. The only tools built are for testing, and #317 is a superset of this change.

zouyee:
@adrcunha PTAL

adrcunha:
I still prefer #319 over this one, because that also keeps the external repository definitions in sync with the kubernetes/test-infra repo (from where the config checkers are built).

PR #319, opened 2018-12-14
Update WORKSPACE definitions

adrcunha:
The previous definitions don't work with bazel 0.2.


adrcunha:
Latest kubekins image still uses bazel 0.18.1, thus the test failure.

adrcunha:
My bad, the failure was caused by the oudated Prow configchecker rule. bazel 0.16 is able to run the tests with no problem.

/hold cancel

adrcunha:
GCP infra flakiness. :(

/retest

PR #327, opened 2018-12-17
Make cleanup job passing by adding placeholder in artifacts

chaodaiG:
Gubernator consider job failure if no `junit_*.xml` file found under artifacts, add placeholder file under artifacts to make it pass


adrcunha:
Nice!

/lgtm
/approve

PR #340, opened 2018-12-19
Create artifacts dir if not present when writing testgrid xml

srinivashegde86:


shared/testgrid/testgrid.go
```diff
@@ -62,6 +62,11 @@ func GetArtifactsDir() string {
 	dir := os.Getenv("ARTIFACTS")
 	if dir == "" {
 		log.Printf("Env variable ARTIFACTS not set. Using './artifacts' instead.")
+		err := os.Mkdir("artifacts", 0777)
+		if err != nil {
```
adrcunha: It should not fail if the directory already exists.

shared/testgrid/testgrid.go
```diff
@@ -62,6 +62,11 @@ func GetArtifactsDir() string {
...
+			log.Printf("Cannot create 'artifacts' directory. Using current directory instead.")
+			return "."
```
adrcunha: If it doesn't exist and you can't create it, then you should fail instead of falling back to using the current dir.

shared/testgrid/testgrid.go
srinivashegde86: it does not. You can verify by running the test multiple times.

shared/testgrid/testgrid.go
srinivashegde86: any reason for failing in this case? I would prefer not to fail unless its a real failure, which this does not seem so.

adrcunha: Consistency. If you can't create the directory, something is very wrong in your setup (eg, read-only FS), so probably something is gonna fail later. The current directory is not guaranteed to be writeable. Finally, if you set the env var, you're not checking for permissions as well, or making decisions about that, and the user can simply set it to the current dir if that's the desired output dir.

PR #341, opened 2018-12-19
webhook based API-Coverage tool: Value analysis

dushyanthsc:
This changeset adds next iteration in the webhook based API-Coverage
tool by introducing value analysis that updates the resource tree. Each
node is expected to implement NodeInterface method updateCoverage(v
reflect.Value). Inside this method each node updates its
nodeData.coverage field value based on wether the passed reflect.Value
parameter is set or not.

### DIFF
--- a/tools/webhook-apicoverage/resourcetree/arraykindnode.go
+++ b/tools/webhook-apicoverage/resourcetree/arraykindnode.go
@@ -20,7 +20,9 @@ import (
 	"reflect"
 )
 
-const arrayNodeNameSuffix = "-arr"
+const (
+	arrayNodeNameSuffix = "-arr"
+)
 
 // ArrayKindNode represents resource tree node of types reflect.Kind.Array and reflect.Kind.Slice
 type ArrayKindNode struct {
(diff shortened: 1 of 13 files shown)


tools/webhook-apicoverage/resourcetree/arraykindnode.go
```diff
@@ -43,3 +45,12 @@ func (a *ArrayKindNode) buildChildNodes(t reflect.Type) {
 	a.children[childName] = childNode
 	childNode.buildChildNodes(t.Elem())
 }
+
+func (a *ArrayKindNode) updateCoverage(v reflect.Value) {
+	if !v.IsNil() {
+		a.covered = true
+		for i := 0; i < v.Len(); i++ {
+			a.children[a.field + arrayNodeNameSuffix].updateCoverage(v.Index(i))
```
jessiezcc: might be easier to read with prefix

tools/webhook-apicoverage/resourcetree/resourcetree.go
```diff
@@ -62,4 +62,9 @@ func (r *ResourceTree) createNode(field string, parent NodeInterface, t reflect.
 func (r *ResourceTree) BuildResourceTree(t reflect.Type) {
 	r.Root = r.createNode(r.ResourceName, nil, t)
```
jessiezcc: should we make buildChildNodes as part of createNode? looks like it will make rest of code much simpler.

dushyanthsc: createNode is the tree method that has the common code used by all node types, where as buildChildNodes as node specific implementation as each node type would have different implementation.

tools/webhook-apicoverage/resourcetree/arraykindnode.go
jessiezcc: yes.

PR #343, opened 2018-12-20
When installing latest Knative, show what it is

adrcunha:
Also factor out the common address of the YAML files, so they can all be updated with a single change.

### DIFF
--- a/scripts/library.sh
+++ b/scripts/library.sh
@@ -23,11 +23,12 @@ readonly SERVING_GKE_VERSION=latest
 readonly SERVING_GKE_IMAGE=cos
 
 # Public latest stable nightly images and yaml files.
-readonly KNATIVE_ISTIO_CRD_YAML=https://storage.googleapis.com/knative-nightly/serving/latest/istio-crds.yaml
-readonly KNATIVE_ISTIO_YAML=https://storage.googleapis.com/knative-nightly/serving/latest/istio.yaml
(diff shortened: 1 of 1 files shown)


scripts/library.sh
```diff
@@ -351,11 +352,14 @@ function report_go_test() {
...
   header "Starting Knative Serving"
   subheader "Installing Istio"
+  echo "Installing Istio CRD from ${KNATIVE_ISTIO_CRD_YAML}"
   kubectl apply -f ${KNATIVE_ISTIO_CRD_YAML} || return 1
+  echo "Installing Istio from ${KNATIVE_ISTIO_YAML}"
   kubectl apply -f ${KNATIVE_ISTIO_YAML} || return 1
   wait_until_pods_running istio-system || return 1
   kubectl label namespace default istio-injection=enabled || return 1
   subheader "Installing Knative Serving"
+  echo "Installing Serving from ${KNATIVE_SERVING_RELEASE}"
```
chaodaiG: nit-picking: just for consistency between subheader and this message:
```suggestion
   echo "Installing Knative Serving from ${KNATIVE_SERVING_RELEASE}"
```

chaodaiG:
/lgtm

just a nit-picking comment on message

scripts/library.sh
adrcunha: You LGTMed but did not hold it, so the PR was merged without your suggestion. :/

chaodaiG: yeah it's no big deal, I don't think it should be a PR blocker. We can add it in the future

PR #346, opened 2018-12-20
Allow cluster size parameters to be set by caller.

tcnghia:
This would allow some PR/repository to change cluster size without changing it for everyone else.

### DIFF
--- a/scripts/e2e-tests.sh
+++ b/scripts/e2e-tests.sh
@@ -153,6 +153,8 @@ function create_test_cluster() {
 
   header "Creating test cluster"
 
+  echo "Cluster will have a minimum of ${E2E_MIN_CLUSTER_NODES} and a maximum of ${E2E_MIN_CLUSTER_NODES} nodes."
+
   # Smallest cluster required to run the end-to-end-tests
   local CLUSTER_CREATION_ARGS=(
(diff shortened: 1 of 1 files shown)


scripts/e2e-tests.sh
```diff
@@ -147,11 +151,12 @@ function create_test_cluster() {
   set -o errexit
   set -o pipefail
 
-  header "Creating test cluster"
+  header "Creating test cluster with --min-nodes=${E2E_MIN_CLUSTER_NODES} --max-nodes=${E2E_MAX_CLUSTER_NODES}"
```
chaodaiG: move `--min-node` and `max-nodes` out of "header" might be better

chaodaiG:
overall looks good to me, just a formatting comment

scripts/e2e-tests.sh
adrcunha: Agreed. Holding for the suggestion.

/hold

tcnghia: Done

chaodaiG: yeah that's exactly what I meant, something like this:
`echo "args:\n\t--min-nodes=${E2E_MIN_CLUSTER_NODES}\n\t--max-nodes=${E2E_MAX_CLUSTER_NODES}"`

adrcunha: I suggest a human-readable logging like

```echo "Cluster will have a minimum of ${E2E_MIN_CLUSTER_NODES} and a maximum of ${E2E_MIN_CLUSTER_NODES} nodes"```

chaodaiG: sounds good to me

tcnghia: Done

PR #356, opened 2019-01-03
Also update E2E test return value on failures

adrcunha:
Don't rely on "failed" being the default/current state.

Bonus: factor out the return code workaround in a separate function.

### DIFF
--- a/scripts/e2e-tests.sh
+++ b/scripts/e2e-tests.sh
@@ -271,6 +271,8 @@ function setup_test_cluster() {
   set +o pipefail
 }
 
+# Set the return code that the test script will return.
+# Parameters: $1 - return code (0-255)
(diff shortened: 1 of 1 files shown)


scripts/e2e-tests.sh
```diff
@@ -279,19 +271,32 @@ function setup_test_cluster() {
   set +o pipefail
 }
 
-function success() {
+function set_test_return_code() {
```
chaodaiG: should the argument be documented?

chaodaiG:
overall looks good

scripts/e2e-tests.sh
adrcunha: Usually I don't document trivial internal functions, but here it goes...

PR #357, opened 2019-01-03
delete images older than 30 days instead of 90 days

chaodaiG:


chaodaiG:
/cc @adrcunha 
FYI @jessiezcc

PR #358, opened 2019-01-03
Add save as json option to load generator results

srinivashegde86:
This will provide an option to perf tests to save the results in the artifacts directory as JSON. We will be using this later for benchmarking.

### DIFF
--- a/shared/loadgenerator/loadgenerator.go
+++ b/shared/loadgenerator/loadgenerator.go
@@ -19,17 +19,22 @@ limitations under the License.
 package loadgenerator
 
 import (
+	"encoding/json"
 	"fmt"
+	"log"
+	"os"
 	"time"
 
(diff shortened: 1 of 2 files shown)


shared/loadgenerator/loadgenerator.go
```diff
@@ -77,3 +82,30 @@ func (g *GeneratorOptions) RunLoadTest(resolvableDomain bool) (*GeneratorResults
...
+	if err != nil {
+		return err
+	}
+
+	defer f.Close()
+	json, err := json.Marshal(gr)
+	if err != nil {
+		return err
+	}
+	_, err = f.Write(json)
```
adrcunha: Merge with the line below? (holding so you can act on this if you want)

/lgtm
/hold

srinivashegde86: done

PR #360, opened 2019-01-04
raise coverage threshold for knative/build to 80%

steuhs:
As agreed here
https://github.com/knative/build/issues/447

### DIFF
--- a/ci/prow/config.yaml
+++ b/ci/prow/config.yaml
@@ -436,7 +436,7 @@ presubmits:
         - "--artifacts=$(ARTIFACTS)"
         - "--profile-name=coverage_profile.txt"
         - "--cov-target=."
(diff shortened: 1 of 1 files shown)


adrcunha:
In issue 447 you raised a question about some files not being above the 80% threshold yet, but didn't get explicit confirmation about them. Is it safe to merge this PR? (even though it might not be deployed now, it certainly will be at some point, much probably without any warning).  

/hold

adrcunha:
Sorry, there's a confusion here: this PR only raises the threshold, but doesn't block submission. 447 is about blocking the submission, so this PR does not fixes that issue, but is part of the solution.

/hold cancel

adrcunha:
Updated the description, so this PR doesn't close 447 by mistake.

/lgtm

Issue #101, opened 2018-09-05
Create helpers for build/unit/integration tests


adrcunha:
This will simplify presubmit tests in the repos, unify features and increase coverage.

* Build tests helper should:
  * build all go code, plus any e2e tests;
  * check for 3rd party licenses and also up-to-date depencencies (if `vendor` present);
  * check go formatting;
  * check links and formatting of markdown files (meaning that build tests won't be totally skipped anymore for documentation-only PRs)

* Unit test helper should
  * run all unit tests (but NOT e2e tests) and report each test;
  * properly identify non-fatal errors that cause the test job to fail (but NO test to fail), like
  ```
  W0905 20:12:53.535] # github.com/knative/pkg/test
  W0905 20:12:53.535] test/e2e_flags.go:28:2: imported and not used: "github.com/knative/pkg/test/logging"
  ```

* Integration test helper should
  * run all E2E tests and report each test;
  * properly take care of the `--emit-metrics` flag, like
  ```
  local options=""
  (( EMIT_METRICS )) && options="--emit-metrics"
  ./test/e2e-tests.sh ${options}
  ```

adrcunha:
All repos are now using the default runners.

PR #362, opened 2019-01-04
Create default build and unit test runners.

adrcunha:
Part of #101.


test/unit/test-helper.sh
```diff
@@ -64,3 +64,10 @@ function mock_gcloud_function() {
   }
   "$@" 2>&1
 }
+
+# Convenience method to display a test failure and exit the script.
+# Parameters: $1 - message to display.
+function failed() {
```
chaodaiG: this function name is also used as variable name in `scripts/presubmit-tests.sh`, is it on purpose?

test/unit/presubmit-custom-runners-tests.sh
```diff
@@ -0,0 +1,49 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+source $(dirname $0)/presubmit-integration-tests-common.sh
+
+function build_tests() {
+  RAN_BUILD_TESTS=1
```
chaodaiG: should it be `RUN` instead of `RAN`?

chaodaiG: oh never mind, I got it now

chaodaiG: just an idea, how about `BUILD_TESTS_RAN` ?

adrcunha: I think this is an unnecessary renaming.

test/unit/test-helper.sh
adrcunha: No, and one won't affect the other. Anyway, I'll rename for clarity.

chaodaiG:
/lgtm
this is much easier to read

PR #364, opened 2019-01-05
Update default build/unit test runners.

### LINKED ISSUE #65: Make it clear when build tests fail
For example, in https://gubernator.knative.dev/build/knative-prow/pr-logs/pull/knative_eventing/276/pull-knative-eventing-build-tests/1027295713966428160/ there's a single error:

```
W0808 20:50:44.825] pkg/sources/github/receive_adapter/receive_adapter.go:20:2: imported and not used: "context"
```

but it's hidden in a long list of files, which makes it hard to spot.

adrcunha:
Part of #101.
Fixes #65.

Bonuses:
* lint and link check markdown files in the default build test runner
* explicitly state if tests passed on failed
* add support to custom pre/post build/unit tests
* simplify and expand unit tests for `presubmit-tests.sh`
* also build go code containing build tags in the default build test runner
* update documentation missed in #362


scripts/presubmit-tests.sh
```diff
@@ -20,111 +20,206 @@
 source $(dirname ${BASH_SOURCE})/library.sh
 
 # Extensions or file patterns that don't require presubmit tests.
-readonly NO_PRESUBMIT_FILES=(\.md \.png ^OWNERS ^OWNERS_ALIASES)
+readonly NO_PRESUBMIT_FILES=(\.md \.png \.gitignore \.gitattributes ^OWNERS ^OWNERS_ALIASES ^AUTHORS)
```
chaodaiG: should `\.md` be removed from this list?

scripts/presubmit-tests.sh
```diff
@@ -20,111 +20,206 @@
...
+    subheader "Linting the markdown files"
+    lint_markdown ${mdfiles} || failed=1
+    subheader "Checking links in the markdown files"
+    check_links_in_markdown ${mdfiles} || failed=1
+  fi
+  # For documentation PRs, just check the md files
+  (( IS_DOCUMENTATION_PR )) && return ${failed}
+  # Ensure all the code builds
+  subheader "Checking that go code builds"
   go build -v ./... || failed=1
```
chaodaiG: I understand this is not part of this PR, just wondering while testing locally, would `go build -v ./...` scan everything under `${GOPATH}/src/github.com` as current `pwd` is `${REPO_ROOT_DIR}`

scripts/presubmit-tests.sh
adrcunha: Argh, rebasing copy-pasta. Fixed, thanks.

scripts/presubmit-tests.sh
adrcunha: No, it will scan everything under `${GOPATH}/src/github.com/YOUR_REPO_NAME_HERE`, which is `${REPO_ROOT_DIR}`.

chaodaiG:
overall lgtm, left 2 comments

PR #370, opened 2019-01-09
Don't fetch all remotes when releasing

adrcunha:
Instead, only fetch the remote branch we're interested in.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -19,8 +19,8 @@
 
 source $(dirname ${BASH_SOURCE})/library.sh
 
-# Github upstream.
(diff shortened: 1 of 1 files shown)


chaodaiG:
the `setup_branch` function does `git fetch`, and it's invoked by both `dot_release` and `initialize` functions. To avoid `git fetch` occurring twice, should there be a logic to prevent them from being executed in the same run?

adrcunha:
> the `setup_branch` function does `git fetch`, and it's invoked by both `dot_release` and `initialize`
> functions. To avoid `git fetch` occurring twice, should there be a logic to prevent them from being
> executed in the same run?

The double fetch won't hurt, and right now in light of 365 and 368 it's simpler keeping it this way.

PR #376, opened 2019-01-10
A helper funcion for removing broken symlinks

erain:
Two types of symlinks are considered broken:
1. Broken symlinks.
2. Symlinks pointing to a path outside of the source tree.

This function was impleneted in
  https://github.com/knative/serving/pull/2842
and should be useful for all knative modules.


adrcunha:
I kinda don't see this function being widely used. I would prefer having it in a helper `update-deps.sh` script (together with more boilerplate stuff), as I mentioned in your original PR.

erain:
Sure I misunderstood what you meant. Yes I think it makes a lot of sense to share a `update-deps.sh` script for different modules. Working on that.

adrcunha:
Thanks! Please use `release.sh`, `e2e-tests.sh` or `presubmit-tests.sh`  in `/scripts` (and the accompanying documentation in `README.md`) as a guidance.

erain:
/lifecycle frozen 
Will work on https://github.com/golang/dep/issues/1564 instead and we may not need this PR anymore if we can fix `dep`

PR #384, opened 2019-01-12
Allow customizing the markdown-link-checker tool

adrcunha:
* Ignore localhost links by default
* Allow per-repo custom config in /test

Bonus: make library tests less verbose, and its output clearer.

Addresses https://github.com/knative/docs/issues/750

CC @samodell


scripts/library.sh
```diff
@@ -356,16 +358,30 @@ function lint_markdown() {
...
 }
 
+# Returns the canonical path of a filesystem object.
+# Parameters: $1 - path to return in canonical form
+#             $2 - base dir for relative links; optional, defaults to current
+function get_canonical_path() {
+  # We don't use readlink because it's not available on every platform.
+  local path=$1
+  local pwd=${2:-.}
+  [[ ${path} == /* ]] || path="${pwd}/${path}"
```
chaodaiG: would this left statement ever be false?

chaodaiG:
overall lgtm, just a minor question

scripts/library.sh
adrcunha: It's false for all relative paths.

chaodaiG: oh right, misread it, thought it was `.*` initially

PR #392, opened 2019-01-14
Encapsulate the build process in a function for release.sh

### LINKED ISSUE #371: Latest release GCS might contain spurious files
`publish_yaml` simply copies the new YAML files to the GCS bucket. Since `latest` is always updated (never created), any YAML file not overwritten will remain in the bucket.

For example, after PR 363 Knative Serving doesn't include build's YAML file, but the `latest` dir in GCS will contain that file because it's not touched or overwritten anymore.

https://github.com/knative/test-infra/blob/fef8c6c10c686341b6d56beda01eac2b67fef99a/scripts/release.sh#L58

adrcunha:
Less boilerplate code, more control.

Bonus: fixes #371.


adrcunha:
CC @nbarthwal

PR #399, opened 2019-01-16
Automate the creation of major releases

### LINKED ISSUE #368: Automate the creation of major releases
Summary: a bi-hourly job (to ensure enough time for building a release) checks for new release branches (using the releases published on GitHub for comparison), then builds a new major release.

nbarthwal:
Addresses #368


scripts/release.sh
```diff
@@ -342,3 +341,14 @@ function branch_release() {
       ${commitish} \
       ${TAG}
 }
+
+function branch_release() {
+    local tags="$(git tag | cut -d 'v' -f2 | cut -d '.' -f1-2 | sort | uniq )"
```
adrcunha: You should not release on tags, as they can be different on a local repo. Furthermore, this will fail on Prow job as it doesn't fetch all remotes.

scripts/release.sh
```diff
@@ -342,3 +341,14 @@ function branch_release() {
...
+    local branches="$( { (git branch -r | grep origin/release-) ; (git branch  | grep release-); } | cut -d '-' -f2 | sort | uniq)"
```
adrcunha: You have to call `setup_upstream()` to fetch the remotes first, as they might not be available.

scripts/release.sh
```diff
@@ -312,8 +312,7 @@ function initialize() {
 # Create a new release on GitHub, also git tagging it (unless this is not a versioned release).
 # Parameters: $1 - Module name (e.g., "Knative Serving").
 #             $2 - YAML files to add to the release, space separated.
-function branch_release() {
-  # TODO(adrcunha): rename this function to publish_to_github()
+function publish_to_github() {
```
adrcunha: This broke the tests, as you probably noticed. Please update the tests.

scripts/release.sh
```diff
@@ -342,3 +341,14 @@ function branch_release() {
       ${commitish} \
       ${TAG}
 }
+
+function branch_release() {
+    local tags="$(git tag | cut -d 'v' -f2 | cut -d '.' -f1-2 | sort | uniq )"
+    local branches="$( { (git branch -r | grep origin/release-) ; (git branch  | grep release-); } | cut -d '-' -f2 | sort | uniq)"
+    if [ "$tags" == "$branches" ]
```
adrcunha: This is not protected by a flag, as we discussed in person.

scripts/release.sh
```diff
@@ -342,3 +341,14 @@ function branch_release() {
...
+    then
+        echo "Everything is released. Nothing to do!"
```
adrcunha: The no-op should come before the build itself. This way you're going through all the build process, but just not creating the release on GitHub.

adrcunha:
a) This doesn't fix 368, but is part of it. Once the script is ready, there's still the job to create.
a) Given the sensitive nature of this script, please add some unit testing if possible. Also as discussed, ensure through manual testing that it didn't break nightlies and that it does what's supposed to do (in a personal repo).

adrcunha:
One last comment: please be (way) more verbose in commit messages; just stating "changes" doesn't help figuring out what the commit is about in a change log.

adrcunha:
Also there's a misunderstanding about the goal of this PR: your code is just skipping publishing a release to GitHub if all release branches have releases associated. Which will only happen if you re-release a major version.

scripts/release.sh
nbarthwal: If I call setup_upstream(), then I do I have to grep remote branches? Shoudn't setup_upstream bring all the local branches & tags?

adrcunha: Yes. And that's exactly why you have to grep them.

scripts/release.sh
nbarthwal: So I can replace it with: git ls-remote --tags origin? What should work!

nbarthwal:
> One last comment: please be (way) more verbose in commit messages; just stating "changes" doesn't help figuring out what the commit is about in a change log.

nbarthwal:
I usually squash all commits and put a long descriptive message. The intermediate commits might have some messages like "changes"

PR #403, opened 2019-01-17
Save cluster metadata once it is created

### LINKED ISSUE #377: Log all the uber details like k8s version, go version, etc in metadata.json when running tests
Our testing tools look for metadata.json inside artifacts/ and incorporate it into the results. We can also use the names from this json to display custom headers on testgrid. So, we can display the k8s version, and #nodes used for the test run.

This will be particularly helpful to understand in testgrid, what params a test was run against.

We log all this information in the build log. Would be great if we can leverage this functionality as well.

srinivashegde86:
Fixes #377 

We can use this metadata mainly for testgrid to create custom column headers on each tab. this can also be used in prow/spyglass, if we enable that.

### DIFF
--- a/scripts/e2e-tests.sh
+++ b/scripts/e2e-tests.sh
@@ -100,7 +100,7 @@ function download_k8s() {
     echo "Using command-line supplied version ${E2E_CLUSTER_VERSION}"
   fi
   # Download k8s to staging dir
-  version=v${E2E_CLUSTER_VERSION}
+  E2E_CLUSTER_VERSION=v${E2E_CLUSTER_VERSION}
   local staging_dir=${GOPATH}/src/k8s.io/kubernetes/_output/gcs-stage
   rm -fr ${staging_dir}
   staging_dir=${staging_dir}/${E2E_CLUSTER_VERSION}
--- a/scripts/e2e-tests.sh
+++ b/scripts/e2e-tests.sh
@@ -143,6 +143,21 @@ function dump_cluster_state() {
   echo "***************************************"
 }
 
+# On a Prow job, save some metadata about the test for Testgrid.
+function save_metadata() {
+  (( ! IS_PROW )) && return
+  cat << EOF > ${ARTIFACTS}/metadata.json
+{
+  "Region": "${E2E_CLUSTER_REGION}",
+  "Zone": "${E2E_CLUSTER_ZONE}",
+  "Machine": "${E2E_CLUSTER_MACHINE}",
+  "Version": "${E2E_CLUSTER_VERSION}",
+  "MinNodes": "${E2E_MIN_CLUSTER_NODES}",
+  "MaxNodes": "${E2E_MAX_CLUSTER_NODES}"
+}
+EOF
+}
+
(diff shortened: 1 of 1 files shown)


scripts/e2e-tests.sh
```diff
@@ -143,6 +143,21 @@ function dump_cluster_state() {
...
 
+# On a Prow job, save some metadata about the test for Testgrid.
+function save_metadata() {
+  (( ! IS_PROW )) && return
+  cat << EOF > ${ARTIFACTS}/metadata.json
+{
+  "Region": "${E2E_CLUSTER_REGION}",
+  "Zone": "${E2E_CLUSTER_ZONE}",
+  "Machine": "${E2E_CLUSTER_MACHINE}",
+  "Version": "${E2E_CLUSTER_VERSION}",
```
adrcunha: This will be "latest" most of the time. The real version number is computed by `download_k8s()`. We should repurpose this variable to contain the real value, and create `E2E_CLUSTER_DEFAULT_VERSION` to hold the default test cluster version (it's defined in library.sh).

scripts/e2e-tests.sh
```diff
@@ -228,6 +243,10 @@ function create_test_cluster() {
   fi
   local result="$(cat ${TEST_RESULT_FILE})"
   echo "Test result code is $result"
+
+  # Save some metadata about cluster creation for using in prow and testgrid
+  save_metadata
```
adrcunha: I think this should be done during setup, so if for any reason the teardown doesn't execute, the metadata is still there.

scripts/e2e-tests.sh
```diff
@@ -82,7 +82,7 @@ function go_test_e2e() {
 # Download the k8s binaries required by kubetest.
 # Parameters: $1 - GCP project that will host the test cluster.
 function download_k8s() {
-  local version=${E2E_CLUSTER_VERSION}
+  local version=${E2E_DEFAULT_CLUSTER_VERSION}
```
adrcunha: I think we can remove E2E_DEFAULT_CLUSTER_VERSION, as it's not adding value here. Then, in line 113, readonly E2E_CLUSTER_VERSION, as it contains now the real version number.

scripts/e2e-tests.sh
```diff
@@ -82,33 +82,33 @@ function go_test_e2e() {
...
+  elif [[ "${E2E_CLUSTER_VERSION}" == "default" ]]; then
     echo "ERROR: `default` GKE version is not supported yet"
     return 1
   else
-    echo "Using command-line supplied version ${version}"
+    echo "Using command-line supplied version ${E2E_CLUSTER_VERSION}"
   fi
   # Download k8s to staging dir
-  version=v${version}
+  version=v${E2E_CLUSTER_VERSION}
```
adrcunha: Unused, remove.

adrcunha: Actually, this line must be:

```suggestion
  E2E_CLUSTER_VERSION=v${E2E_CLUSTER_VERSION}
```

That's why the integration tests failed.

PR #411, opened 2019-01-22
Add a presubmit job to run perf tests

srinivashegde86:
This just adds it to the config and allows us to test it on any PR's. Not adding this to list of auto running jobs per PR.

### DIFF
--- a/ci/prow/config.yaml
+++ b/ci/prow/config.yaml
@@ -314,6 +314,7 @@ presubmits:
       - name: github-token
         secret:
           secretName: covbot-token
+
   - name: pull-knative-serving-perf-tests
     agent: kubernetes
     context: pull-knative-serving-perf-tests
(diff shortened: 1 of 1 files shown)


ci/prow/config.yaml
```diff
@@ -314,6 +314,32 @@ presubmits:
       - name: github-token
         secret:
           secretName: covbot-token
+  - name: pull-knative-serving-perf-tests
```
adrcunha: One blank line before it, please.

ci/prow/config.yaml
```diff
@@ -314,6 +314,32 @@ presubmits:
...
+    agent: kubernetes
+    context: pull-knative-serving-perf-tests
+    always_run: false
+    rerun_command: "/test pull-knative-serving-perf-tests"
+    trigger: "(?m)^/test (all|pull-knative-serving-perf-tests),?(\\s+|$)"
+    labels:
+      preset-test-account: "true"
+    skip_branches:  # Skip these branches, as test isn't available.
```
adrcunha: You probably don't need this if they're not running on presubmit. If they do, then you'll need to skip branch release-0.3 until this is merged into the branch (or skip forever, if it never gets merged into the branch).

srinivashegde86: Added 0.3. Will keep this in from now, instead of remebering to add it later :D

PR #412, opened 2019-01-22
Do not lint or check links for deleted Markdown docs

dgerd:
Deleting a Markdown document results in a failed pre-submit test as it
attempts to lint and check the links of the document that no longer
exists.

See https://github.com/knative/serving/pull/2962 for example failure

### DIFF
--- a/scripts/presubmit-tests.sh
+++ b/scripts/presubmit-tests.sh
@@ -112,6 +112,14 @@ function markdown_build_tests() {
   (( DISABLE_MD_LINTING && DISABLE_MD_LINK_CHECK )) && return 0
   # Get changed markdown files (ignore /vendor)
   local mdfiles="$(echo "${CHANGED_FILES}" | grep \.md$ | grep -v ^vendor/)"
+  # Filter out deleted files
+  for file in $mdfiles; do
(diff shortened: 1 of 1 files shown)


scripts/presubmit-tests.sh
```diff
@@ -112,6 +112,14 @@ function markdown_build_tests() {
   (( DISABLE_MD_LINTING && DISABLE_MD_LINK_CHECK )) && return 0
   # Get changed markdown files (ignore /vendor)
   local mdfiles="$(echo "${CHANGED_FILES}" | grep \.md$ | grep -v ^vendor/)"
+  # Filter out deleted files
+  for file in $mdfiles; do
```
adrcunha: I suggest replacing lines 114-122 with something like:

```
local mdfiles=""
for file in $(echo "${CHANGED_FILES}" | grep \.md$ | grep -v ^vendor/); do
  # Filter out deleted files
  [[ -f "${file}" ]] && mdfiles="${mdfiles} ${file}"
done
```

dgerd: I like the suggestion. Updated.

PR #416, opened 2019-01-25
Adding MD024 rule for exclude

matzew:
When ignoring `MD024`, we are not getting bogus warnings about **Multiple headers with the same content** anymore.

This is sometimes needed. See this file:
https://github.com/knative/eventing/blob/master/docs/spec/interfaces.md#control-plane

IMO it's reasonable to have multiple Data Plane / Control Plane headers in one document, especially when that is in _different_ subsections

cc @evankanderson @grantr

### DIFF
--- a/scripts/library.sh
+++ b/scripts/library.sh
@@ -366,7 +366,7 @@ function check_links_in_markdown() {
 # Parameters: $1..$n - files to inspect
 function lint_markdown() {
   # https://github.com/markdownlint/markdownlint
-  run_lint_tool mdl "linting markdown files" "-r ~MD013" $@
+  run_lint_tool mdl "linting markdown files" "-r ~MD013,~MD024" $@
 }
 
 # Return whether the given parameter is an integer.


matzew:
I think we also should evaluate some `style` 

https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#md029---ordered-list-item-prefix 

`ordered` or `one`

adrcunha:
I prefer having the rule exclusion configurable per repo instead of globally. Like how it's done for the link checker.

/hold

adrcunha:
Also, the doc says that the right way of dealing with the problem stated in this PR is by setting `allow_different_nesting` to `true`.

matzew:
I think I am generally fine w/ a configuration per repo.

the thing that really was odd, that most of the _existing_ MD files have issues, and now the checker yells if they are touched. Would have been better to address that _before_.

That said, for eventing I am willing to tackle it: https://github.com/knative/eventing/issues/753

evankanderson:
Can we turn off the markdown linting in tests until the existing repos are clean?

/lgtm

adrcunha:
Yes. See item 3 in https://github.com/knative/test-infra/tree/master/scripts#helper-scripts
It's a one line change like in https://github.com/knative/docs/pull/769/files#diff-8f99251cc2fba45d21cb371ad880c91c

adrcunha:
Closing this as #417 took care of the issue.

PR #427, opened 2019-01-28
Assorted fixes to e2e test infra

adrcunha:
* stop using deprecate cluster creation flag in `kubetest`
* fix passing extra flags for cluster creation
* allow using gcloud beta features for cluster creation
* pass-through custom flags defined by the e2e script
* use 10GB disks in test cluster to save resources

Bonus:
* don't force zone f for e2e tests


srinivashegde86:
limiting the disk size still fails during error creation

adrcunha:
/hold

Until I squash the commits.

PR #429, opened 2019-01-29
Run all e2e tests on us-west1

adrcunha:
us-central1 is too busy and lately have been experiencing stockouts too frequently.

### DIFF
--- a/ci/prow/config.yaml
+++ b/ci/prow/config.yaml
@@ -77,6 +77,8 @@ presets:
   env:
   - name: GOOGLE_APPLICATION_CREDENTIALS
     value: /etc/test-account/service-account.json
+  - name: E2E_CLUSTER_REGION
+    value: us-west1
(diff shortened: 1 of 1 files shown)


ci/prow/config.yaml
```diff
@@ -105,6 +109,8 @@ presets:
   env:
   - name: GOOGLE_APPLICATION_CREDENTIALS
     value: /etc/release-account/service-account.json
+  - name: E2E_CLUSTER_REGION
```
srinivashegde86: do you also want to change the `# backups service account`?

srinivashegde86:
/lgtm
/hold

added a comment. PTAL.

ci/prow/config.yaml
adrcunha: No, it doesn't run E2E tests.

/hold cancel

PR #436, opened 2019-02-01
Check existence of GCS path should also work for directory path

chaodaiG:
Currently, existence of GCS path is implemented by looking at object handle, but GCS directories are virtual objects, so there is no object handle associated with it, thus it would return false if the path is a directory. This PR checks for both files and directories.

### DIFF
--- a/shared/gcs/gcs.go
+++ b/shared/gcs/gcs.go
@@ -41,18 +41,17 @@ func Authenticate(ctx context.Context, serviceAccount string) error {
 	return err
 }
 
-// Exist checks if path exist under gcs bucket,
+// Exists checks if path exist under gcs bucket,
 // this path can either be a directory or a file.
-func Exist(ctx context.Context, bucketName, storagePath string) bool {
-	// Check if this is a directory,
-	// gcs directory paths are virtual paths, they automatically got deleted if there is no child file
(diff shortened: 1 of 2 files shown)


shared/gcs/gcs.go
```diff
@@ -41,13 +41,25 @@ func Authenticate(ctx context.Context, serviceAccount string) error {
...
 
-// Exist checks if path exist under gcs bucket
-func Exist(ctx context.Context, bucketName, filePath string) bool {
-	handle := createStorageObject(bucketName, filePath)
+// Exist checks if path exist under gcs bucket,
+// this path can either be a directory or a file.
+func Exist(ctx context.Context, bucketName, storagePath string) bool {
+	// Check if this is a directory,
+	// gcs directory paths are virtual paths, they automatically got deleted if there is no child file
+	if len(ListChildrenFiles(ctx, bucketName, storagePath)) > 0 {
```
adrcunha: My only concern is that `Exist` will be a costly operation, when usually it's perceived as cheap. Comments?

chaodaiG: Good point, and you'are absolutely right about this. I will explore options to make this operation cheaper

chaodaiG: updated, please review again

PR #440, opened 2019-02-01
dep-update: knative/serving

dushyanthsc:
This change does a dep update to update vendor directory.

Bringing in latest knative/serving also resulted in changing the dependency in webhook apicoverage/webhook/webhook.go because of the change to serving/pkg/system package


vendor/github.com/knative/serving/pkg/reconciler/v1alpha1/clusteringress/config/testdata/config-istio.yaml
```diff
@@ -0,0 +1 @@
+../../../../../../config/config-istio.yaml
```
nbarthwal: Not sure what this is!

dushyanthsc: This is pulled from dep and it is specified in https://github.com/knative/serving/blob/master/pkg/reconciler/v1alpha1/clusteringress/config/testdata/config-istio.yaml

Where they are using the relative path to pull in the yaml file.

PR #447, opened 2019-02-06
Ignore links starting with '/' or './' for lint

### LINKED ISSUE #445: Markdown linter fails on working file links
I have a change that touches existing Markdown files in the repository. The change is https://github.com/knative/serving/pull/3029. I do not adjust any of the links in the files, but have build test failures for "dead links found!". See [output](https://gubernator.knative.dev/build/knative-prow/pr-logs/pull/knative_serving/3029/pull-knative-serving-build-tests/1090326122178023424/).

Going to a failed file from the PR such as the [atuoscale README](https://github.com/knative/serving/blob/8aa148b5c0525749adc2153c8af087c4a1205363/test/test_images/autoscale/README.md) shows a working link. 

The links in these documents use syntax like below where they are absolute paths that do not include a full URL.

`[section about test images](/test/README.md#test-images)`

ericKlawitter:
Links containing a file path relative to a Github repo were failing lint checks. Let's exclude those links so that lint doesn't give false errors.

Not sure how restrictive we want to be, but the regex matches for any link starting with a './' or '/', followed by at least one additional character. 

Fixes #445

### DIFF
--- a/scripts/markdown-link-check-config.json
+++ b/scripts/markdown-link-check-config.json
@@ -2,6 +2,9 @@
     "ignorePatterns": [
         {
             "pattern": "^https?://localhost($|[:/].*)"
+        },
+        {
+            "pattern": "^.?[/][\\w].*"
         }
     ]
 }


scripts/markdown-link-check-config.json
```diff
@@ -2,6 +2,9 @@
     "ignorePatterns": [
         {
             "pattern": "^https?://localhost($|[:/].*)"
+        },
+        {
+            "pattern": "^.?[/][\\w].*"
```
adrcunha: `[/]` can be simply `/` since it's just one character.

dgerd: What regex type is this? Do you know if the word character supports the character set for filenames, or should we make this more broad?

Also I don't think you want to escape `\w`.

ericKlawitter:
Still waiting to hear back from the [issue I opened](https://github.com/tcort/markdown-link-check/issues/68) with the markdown-link-check to see if we can support checking these link types instead of just ignoring them. 

In the meantime, to respond to @adrcunha and @dgerd points, I tested via a script with these inputs to determine the matches:

`/foo/bar/ true
/foo/bar true
./foo/bar/README.MD true
/123/4356 true
/123/434/readme.md true
.//// false
//foo/bar false
////foo/bar false
.../foo/bar false
./Aoo/bar true
./_foo/bar true
/_foo/bar true
/__foo/bar true
./!foo/bar false
z/asdf false
./ false
/ false`

and ran `grep -rls --include \*.md "]([^h./].*" .` on the serving repo to verify our link checker would pass for any other types of links we have. If there's anything else I should test, lmk.

scripts/markdown-link-check-config.json
ericKlawitter: \w supports alphanumeric characters and '_'. Maybe we should include '.' as well? Open to suggestions for how restrictive we want this to be. 

I think we do need the `\\w`, tough - according to jq, just a single `\w` results in invalid json.

dgerd: Thinking more about the character set that should be good for now. Since the long term plan is adding support in the linter library we can add additional cases here if needed at a later time

Issue #451, opened 2019-02-08
MD linting failing for serving due to pre-existing dead links


greghaynes:
We only run markdown linting when markdown files are editing. We also added dead link checking without verifying it would pass on the existing markdown files (which it does not). As a result pr's (such as https://github.com/knative/pkg/pull/143 and https://github.com/knative/serving/issues/2885) fail when they edit markdown files with pre-existing dead links.

adrcunha:
The proposed solution is unnecessarily complex. It's way simpler to fix dead links (they add no value to the documentation if they're dead) than to keep track of dead links and not check them.

greghaynes:
Sorry but I think theres a misunderstanding - my proposal is to remove this check for now as we cannot currently make edits to any markdowns. Then we fix all the dead links and update the link check test to run on *all* changes (the cost should be minimal and will avoid this situation).

adrcunha:
I don't understand why you can't make edits to any markdowns. I just did that in https://github.com/knative/pkg/pull/143. Anyway, to disable the link checker clone https://github.com/knative/docs/pull/773

Filed https://github.com/knative/test-infra/issues/453 to get a better link checker in place.

PR #452, opened 2019-02-08
Disable md link checking

greghaynes:
This is failing knative/serving due to preexisting dead links.

Fixes: #451


adrcunha:
You don't want to do this here, but in serving.
DO NOT submit.
/hold

PR #458, opened 2019-02-10
Auto-release jobs for build, eventing and serving

nbarthwal:
Setting up auto-release jobs for build, eventing and serving


srinivashegde86:
Updated to use WIP in title, as prow adds a hold on it. Remove it, and update the description once it is ready. Prow should remove the hold automatically.

PR #460, opened 2019-02-11
Create xml output from testcases for testgrid

srinivashegde86:
This prevents every test to create a testsuite from the testcases and pass it in. 

Also, allow perf test to set AllowForInitialErrors. Do not default to false as the test will still indicate PASS even if it fails.

Part of https://github.com/knative/serving/issues/3154

### DIFF
--- a/shared/loadgenerator/loadgenerator.go
+++ b/shared/loadgenerator/loadgenerator.go
@@ -46,6 +46,7 @@ type GeneratorOptions struct {
 	Domain         string
 	RequestTimeout time.Duration
 	QPS            float64
+	AllowInitialErrors bool
 }
 
 // GeneratorResults contains the results of running the per test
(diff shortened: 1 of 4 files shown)


adrcunha:
```
I0211 18:55:22.147] # github.com/knative/test-infra/tools/apicoverage
I0211 18:55:22.148] tools/apicoverage/apicoverage.go:223:37: cannot use &testSuites (type *junit.TestSuites) as type []junit.TestCase in argument to testgrid.CreateXMLOutput
```

chaodaiG:
/hold
/lgtm

This PR changes the default of `AllowForInitialErrors` from `true` to `false`, it's different from what's described in the PR body. Feel free cancel hold if this is intended

srinivashegde86:
This allows the flexibility of setting this in the config per test. The test can set it to true if needed.

/hold cancel

PR #476, opened 2019-02-13
let 0 statement go file pass coverage threshold

steuhs:
1. Go file with no statement should not be marked as low coverage file.
2. Added logs when a file is marked as low coverage file. This will help debugging to see if a file is marked wrongly

Fixes #

### DIFF
--- a/tools/coverage/calc/base.go
+++ b/tools/coverage/calc/base.go
@@ -147,6 +147,7 @@ func (c *Coverage) IsCoverageLow(covThresholdInt int) bool {
 	if err == nil {
 		return ratio < covThreshold
 	}
+	// go file with no statement should not be marked as low coverage file
 	return false
 }
 
--- a/tools/coverage/calc/base.go
+++ b/tools/coverage/calc/base.go
@@ -147,7 +147,7 @@ func (c *Coverage) IsCoverageLow(covThresholdInt int) bool {
 	if err == nil {
 		return ratio < covThreshold
 	}
-	return true
+	return false
 }
 
 func SortCoverages(cs []Coverage) {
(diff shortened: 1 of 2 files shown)


tools/coverage/calc/base.go
```diff
@@ -147,7 +147,7 @@ func (c *Coverage) IsCoverageLow(covThresholdInt int) bool {
 	if err == nil {
 		return ratio < covThreshold
 	}
-	return true
+	return false
```
chaodaiG: suggest add a comment explaining the logic

tools/coverage/calc/delta.go
```diff
@@ -112,16 +112,18 @@ func (changes *GroupChanges) processChangedFiles(
 		pathFromProfile := githubUtil.FilePathProfileToGithub(inc.base.Name())
 		fmt.Printf("checking if this file is in github change list: %s", pathFromProfile)
 		if (*githubFilePaths)[pathFromProfile] == true {
-			fmt.Printf("\tYes!\n")
+			fmt.Printf("\tYes!")
```
adrcunha: nit: for these cases, use `fmt.Println`.

steuhs: I intentionally don't want to print a newline until later

tools/coverage/calc/base.go
steuhs: 1. Go file with no statement should not be marked as low coverage file.

chaodaiG: sorry for not making it clear, I was suggesting adding a comment above this function explaining when this returns true and when this returns false

PR #477, opened 2019-02-13
Make --run-test skip unnecessary testing on presubmit

chaodaiG:
`presubmit-tests.sh` is used by both presubmit and continuous jobs, they don't always execute the same set of tests. Also add logic to skip running tests for documentation only PRs if it's presubmit test

### DIFF
--- a/scripts/presubmit-tests.sh
+++ b/scripts/presubmit-tests.sh
@@ -238,6 +238,7 @@ function default_integration_test_runner() {
 RUN_BUILD_TESTS=0
 RUN_UNIT_TESTS=0
 RUN_INTEGRATION_TESTS=0
+RUN_PRESUBMIT_TESTS=0
 EMIT_METRICS=0
 
 # Process flags and run tests accordingly.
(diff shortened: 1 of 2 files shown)


srinivashegde86:
The flag --all-tests is used in continous jobs and --run-test is used in integration tests. There is already different flags used in both these jobs.

Also, we have a IS_DOCUMENTATION_PR that checks if all the files in the PR as .md files and is not supposed to run the integration tests. Is that check missing something?

chaodaiG:
> The flag --all-tests is used in continous jobs and --run-test is used in integration tests. There is already different flags used in both these jobs.
> 
> Also, we have a IS_DOCUMENTATION_PR that checks if all the files in the PR as .md files and is not supposed to run the integration tests. Is that check missing something?

the `IS_DOCUMENTATION_PR` logic currently only works for `--build-tests`, `unit-tests`, and `integrations-tests`, but not for `--run-test`, and I just noticed there is a `IS_PRESUBMIT` variable, so this flag isn't really needed for this case

chaodaiG:
closed, need to rethink about the logic

scripts/presubmit-tests.sh
```diff
@@ -311,6 +311,7 @@ function main() {
     if (( RUN_BUILD_TESTS || RUN_UNIT_TESTS || RUN_INTEGRATION_TESTS )); then
       abort "--run-test must be used alone"
     fi
+    (( IS_DOCUMENTATION_PR )) && exit ${failed}
```
adrcunha: As discussed offline, for clarity, let's add a check for IS_PRESUBMIT (even though IS_DOCUMENTATION_PR is only set if IS_PRESUBMIT is set).

```
    # If this is a presubmit run, but a documentation-only PR, don't run the test
    (( IS_PRESUBMIT && IS_DOCUMENTATION_PR )) && exit 0
```

chaodaiG: done

adrcunha:
/lgtm

Thanks!

PR #482, opened 2019-02-14
Initial version of the Prow config generator

### LINKED ISSUE #382: Create a tool to generate the Prow config
From a simpler yaml file, generate the final config.yaml file.

Use templates for each kind of job (CI, presubmit, postsubmit, periodic, etc).

Allow generating one job or all.

Allow overriding parameters from command-line: service account, repository, log upload GCS, command to run before the real task.

adrcunha:
`make_config.go` generates `config.yaml` based on a much smaller and simpler config `config_knative.yaml`.

This initial version makes some concessions regarding the output generation in order to produce a predictable, trackable config file, which can be easily compared to the current version.

Fixes #382.


ci/prow/make_config.go
```diff
@@ -0,0 +1,1010 @@
...
+    value: "true"
+  volumes:
+  - name: docker-graph
+    emptyDir: {}
+  volumeMounts:
+  - name: docker-graph
+    mountPath: /docker-graph
+`
+
+	// presubmitJob is the template for presubmit jobs.
```
chaodaiG: just an idea, how bout storing each job template in separate files. For example, copy paste line 285 to line 310 into a file located at `ci/prow/job_templates/presubmit_job_template.yaml`

chaodaiG:
This is awesome. Would it make sense to add a unit test comparing generated yaml vs checked in `ci/prow/config.yaml`?

ci/prow/make_config.go
adrcunha: I thought about that, but was concerned that keeping them in separate files would make it harder to visualize common/shared settings/patterns, so I decided to keep them in a single place, at least for the initial version. There's a TODO in the header about factoring about some custom jobs from regular jobs; once that's done and the templates are simpler, I think that it would be a good thing keeping the templates in separate files.

chaodaiG: that totally make sense to me, since this is an initial version. This is much better than our 2224 lines config file

adrcunha:
> This is awesome. Would it make sense to add a unit test comparing generated yaml vs checked in
> `ci/prow/config.yaml`?

What a nice idea. :) Done.

ci/prow/config.yaml
```diff
@@ -115,10 +116,16 @@ presets:
   - name: account
     secret:
       secretName: release-account
+  - name: hub-token
```
chaodaiG: is this github token?

adrcunha: Yes, but it's specifically to be used with hub because of the ACL.

ci/prow/config.yaml
```diff
@@ -1804,9 +1763,8 @@ periodics:
       - "--artifacts=$(ARTIFACTS)"
       - "--profile-name=coverage_profile.txt"
       - "--cov-target=."
-      - "--cov-threshold-percentage=80"
-
-- cron: "30 * * * *" # Run every hour and 30 minutes
+      - "--cov-threshold-percentage=50"
```
chaodaiG: coverage threshold of 50 instead of 80, is this intended?

ci/prow/config.yaml
```diff
@@ -1738,9 +1672,8 @@ periodics:
...
```
chaodaiG: coverage threshold of 50 instead of 80, is this intended?

ci/prow/config.yaml
adrcunha: Yes, to keep it in sync with the threshold in the presubmit and postsubmit jobs.

ci/prow/config.yaml
```diff
@@ -1695,9 +1629,8 @@ periodics:
       - "--artifacts=$(ARTIFACTS)"
       - "--profile-name=coverage_profile.txt"
       - "--cov-target=."
-      - "--cov-threshold-percentage=80"
-
-- cron: "50 * * * *" # Run every hour and 50 minutes
+      - "--cov-threshold-percentage=50"
```
chaodaiG: coverage threshold of 50 instead of 80, is this intended?

ci/prow/config.yaml
adrcunha: Ditto.

ci/prow/config.yaml
adrcunha: DItto.

ci/prow/make_config.go
adrcunha: Yep. Now adding a CI flow is just a 2-line PR. :D

ci/prow/make_config.go
```diff
@@ -0,0 +1,1010 @@
+// The make_config tool generates a full Prow config for the Knative project,
```
chaodaiG: does this file also need the `Copyright` stuff?

ci/prow/make_config.go
```diff
@@ -0,0 +1,1010 @@
...
+// getInt casts the given interface (expected int) as int.
+func getInt(s interface{}) int {
+	if value, ok := s.(int); ok {
+		return value
+	}
+	log.Fatalf("Entry %v is not an integer", s)
+	return 0
+}
+
+// getInt casts the given interface (expected int) as int.
```
chaodaiG: comment doesn't match function

adrcunha: Fixed.

ci/prow/make_config.go
adrcunha: Absolutely. Added.

ci/prow/make_config.go
```diff
@@ -0,0 +1,1010 @@
...
+	flag.StringVar(&releaseAccount, "release-account", "/etc/release-account/service-account.json", "Path to the service account JSON for release jobs")
+	flag.StringVar(&coverageDockerImage, "coverage-docker", "gcr.io/knative-tests/test-infra/coverage:latest", "Docker image for coverage tool")
+	flag.StringVar(&prowTestsDockerImage, "prow-tests-docker", "gcr.io/knative-tests/test-infra/prow-tests:stable", "prow-tests docker image")
+	flag.StringVar(&presubmitScript, "presubmit-script", "./test/presubmit-tests.sh", "Executable for running presubmit tests")
+	flag.StringVar(&releaseScript, "release-script", "./hack/release.sh", "Executable for creating releases")
+	flag.StringVar(&performanceScript, "performance-script", "./test/performance-tests.sh", "Executable for running performance tests")
+	flag.StringVar(&repositoryOverride, "repo-override", "", "Repository path (github.com/foo/bar[=branch]) to use instead for a job")
+	flag.Parse()
+	if len(flag.Args()) != 1 {
+		log.Fatal("Pass the config file as parameter")
```
chaodaiG: this doesn't quite look like an error message, should it error out here?

chaodaiG: oh never mind, I got it now

PR #485, opened 2019-02-15
Update go version to 1.12rc1 in prow-tests docker image

chaodaiG:
A new feature in serving websockets need to use go1.12rc1 version. Instead of manually installing this version in serving repo, updating it in prow-tests image, so that all jobs use consistent go version

Related: https://github.com/knative/serving/pull/3240

Part of: #486

### DIFF
--- a/images/prow-tests/Dockerfile
+++ b/images/prow-tests/Dockerfile
@@ -29,6 +29,11 @@ RUN docker-credential-gcr configure-docker
 RUN apt-get install -y uuid-runtime  # for uuidgen
 RUN apt-get install -y rubygems  # for mdl
 
+# Update go version to 1.12rc1
+RUN go get golang.org/dl/go1.12rc1
+RUN go1.12rc1 download
+RUN cp $(which go1.12rc1) $(which go)
+
 # Extra tools through go get
 RUN go get -u github.com/google/go-containerregistry/cmd/ko
 RUN go get -u github.com/golang/dep/cmd/dep


srinivashegde86:
This change affects will affect all repos at the same time whenever we update the image and we should verify that nothing will be broken by this.

Also, looks like we are hard-coding to this version. What is the plan for removing this version dependency?

srinivashegde86:
Also, we are using a month old version of kubekins image `FROM gcr.io/k8s-testimages/kubekins-e2e:v20190103-edc7696cc-master`

Does the latest image have go1.12 already?

mattmoor:
/lgtm
/hold

For comments

chaodaiG:
> Also, we are using a month old version of kubekins image `FROM gcr.io/k8s-testimages/kubekins-e2e:v20190103-edc7696cc-master`
> 
> Does the latest image have go1.12 already?

Thanks for bringing this up, just noticed that my local k8s repo is a bit old, they have updated to newer version of go, but still is not go1.12 yet, they are at 1.11.5 now
for reference: https://github.com/kubernetes/test-infra/blob/master/images/kubekins-e2e/Makefile

chaodaiG:
> This change affects will affect all repos at the same time whenever we update the image and we should verify that nothing will be broken by this.
> 
> Also, looks like we are hard-coding to this version. What is the plan for removing this version dependency?

Matt had communicated with and got green lights from team leads, also we are already using this version on our workstation, so the chance of breaking anything should be low

srinivashegde86:
1.11.5 is the last stable go version. golang.org mentions that 1.12rc1 is still unstable https://golang.org/dl/#unstable

I am ok with change, but this should be communicated widely before we udpate the image as this can potentially break a few things.

/cc @jessiezcc

adrcunha:
/lgtm
/hold

Unhold once you confirm things as fine.

chaodaiG:
> 1.11.5 is the last stable go version. golang.org mentions that 1.12rc1 is still unstable https://golang.org/dl/#unstable
> 
> I am ok with change, butI think this should be communicated widely before we udpate the image as this can potentially break a few things. Also, create an issue to clean this up and use the one provided by kubekins, once it starts using 1.12
> 
> /cc @jessiezcc

Agreed, will do

chaodaiG:
> > 1.11.5 is the last stable go version. golang.org mentions that 1.12rc1 is still unstable https://golang.org/dl/#unstable
> > I am ok with change, butI think this should be communicated widely before we udpate the image as this can potentially break a few things. Also, create an issue to clean this up and use the one provided by kubekins, once it starts using 1.12
> > /cc @jessiezcc
> 
> Agreed, will do

Issue created tracking this issue, and updated in description

adrcunha:
> I am ok with change, butI think this should be communicated widely before we udpate the image as this
> can potentially break a few things. Also, create an issue to clean this up and use the one provided by
> kubekins, once it starts using 1.12

Yes. Also, the update can be tested isolated against already finished jobs by rerunning them using the new image (the new image can be pushed with just the date/SHA tag).

PR #487, opened 2019-02-15
gather directories for testgrid without relying on go test output

### LINKED ISSUE #442: coverage job is not reporting metrics to testgrid
https://testgrid.knative.dev/knative-serving#coverage shows blank rows for the last 3 days.

steuhs:
Reason: Existing implementation relies on `go test` to pass and print out
source file directories in fixed format. It will break when some flaky unit test
fail

Changes
1. Remove dependency on `go test`'s success. Keep the program running regardless of unit test failure
2. Gather directory names from file names, instead of from the `go test -coverprofile` output

Fixes #442


tools/coverage/calc/group.go
```diff
@@ -131,3 +132,16 @@ func (g *CoverageList) Report(itemized bool) {
...
+
+// GetDirs gets a list a sub-directories that contains source code. The list will be shown on Testgrid
+func (g *CoverageList) GetDirs() []string {
+	dirSet := map[string]bool{}
+	for _, cov := range g.group {
+		dirSet[path.Dir(cov.name)] = true
+	}
+	var result []string
+	for key := range dirSet {
+		result = append(result, key)
```
adrcunha: You can inline this in the previous loop.

PR #497, opened 2019-02-20
Allow setting custom environment variables when generating Prow config

adrcunha:


ci/prow/make_config.go
```diff
@@ -590,6 +594,20 @@ func configureServiceAccountForJob(data *baseProwJobTemplateData) {
 	addVolumeToJob(data, "/etc/"+name, name, true)
 }
 
+// addExtraEnvVarsToJob adds any extra environment variables (defined on command-line) to a job.
+func addExtraEnvVarsToJob(data *baseProwJobTemplateData) {
+	if len(extraEnvVars) == 0 {
```
chaodaiG: would it be better check for nil? as it would be no-op if len(extraEnvvars) == 0 below, but nil would fail the job

ci/prow/make_config.go
```diff
@@ -975,6 +999,17 @@ func executeTemplate(name, templ string, data interface{}) {
 	}
 }
 
+// Multi-value flag parser.
+
+func (a *stringArrayFlag) String() string {
+	return strings.Join(*a, ", ")
+}
+
+func (a *stringArrayFlag) Set(value string) error {
```
chaodaiG: this function always return nil, is it necessary?

adrcunha: It doesn't fail because there's no sanity check of the values. I can't break the contract otherwise flag parsing will break.

ci/prow/make_config.go
adrcunha: extraEnvVars won't be nil, because it's initialized as en empty slice. Removed the unnecessary check though.

Issue #132, opened 2018-09-18
automatically file git issues to track flaky tests


jessiezcc:
This way, working group triage will get the visibility to flaky tests and assign proper owners. 
The bug should provide data points about flakiness per test. Prefer one issue per flaky test.

adrcunha:
We can have a daily job that inspects the CI test jobs and files issues accordingly. Most of this work is already done in the latency and API coverage jobs.

adrcunha:
By "inspects the CI test jobs" I mean parsing the junit XML files containing the test results.

chaodaiG:
thanks for clarification!

chaodaiG:
I consider this work is done, can we close this issue? @jessiezcc

PR #507, opened 2019-02-21
Allow listing prow builds in gcs without precompute start/finish time

chaodaiG:
The current implementation of listing prow builds in gcs always precomputes start/finish time, by parsing `Started.json` and `Finished.json` on gcs, this is very expensive for listing latest N runs, especially when there are large number of builds, i.e. >5000 builds for serving continuous job. This PR adds a function for listing builds without precomputing start/finish time

Part of: #132

### DIFF
--- a/tools/flaky-test-reporter/result.go
+++ b/tools/flaky-test-reporter/result.go
@@ -26,6 +26,7 @@ import (
 	"io/ioutil"
 	"path"
 	"path/filepath"
+	"sort"
 
 	"github.com/knative/test-infra/shared/prow"
 	"github.com/knative/test-infra/shared/junit"
@@ -144,7 +145,7 @@ func collectTestResultsForRepo(jc *JobConfig) (*RepoData, error) {
 	if !job.PathExists() {
 		return rd, fmt.Errorf("job path not exist '%s'", jc.Name)
 	}
-	builds := job.GetLatestBuilds(buildsCount)
+	builds := getLatestFinishedBuilds(job, buildsCount)
 	
 	log.Printf("latest builds: ")
 	for i, build := range builds {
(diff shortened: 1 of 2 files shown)


tools/flaky-test-reporter/result.go
```diff
@@ -165,3 +166,23 @@ func collectTestResultsForRepo(jc *JobConfig) (*RepoData, error) {
 	}
 	return rd, nil
 }
+
+// getLatestFinishedBuilds is an inexpensive way of listing latest finished builds, in comparing to
+// the GetLatestBuilds function from prow package, as it doesn't precompute start/finish time before sorting.
+// This function takes the assumption that build IDs are always incremental integers
```
adrcunha: This is true if you see the implementation of the ID used by Prow for the jobs.

tools/flaky-test-reporter/result.go
```diff
@@ -165,3 +166,23 @@ func collectTestResultsForRepo(jc *JobConfig) (*RepoData, error) {
...
+func getLatestFinishedBuilds(job *prow.Job, count int) []prow.Build {
+	var builds []prow.Build
+	buildIDs := job.GetBuildIDs()
+	fmt.Println(len(buildIDs))
```
adrcunha: Debug, remove.

tools/flaky-test-reporter/result.go
```diff
@@ -165,3 +166,23 @@ func collectTestResultsForRepo(jc *JobConfig) (*RepoData, error) {
...
+	for _, buildID := range buildIDs {
+		if len(builds) >= count {
+			break
+		}
+		build := job.NewBuild(buildID)
+		if nil != build.FinishTime {
+			builds = append(builds, *build)
+		}
+	}
+	return builds
```
adrcunha: You could be paranoid here (since you're relying on the assumption that IDs are increasing integers) and ensure that the start time of the builds is indeed decreasing. Since you already computed the start/end time, and you're gonna iterate over `count` items, this won't make your function slower. It would be a clear signal if someday things change radically in the GCS bucket.

tools/flaky-test-reporter/result.go
chaodaiG: that's right
https://github.com/kubernetes/test-infra/blob/master/prow/pjutil/tot.go#L80

tools/flaky-test-reporter/result.go
chaodaiG: done

tools/flaky-test-reporter/result.go
chaodaiG: Good idea, implemented guard against it

PR #523, opened 2019-02-25
Add function for displaying pod logs.

ericKlawitter:
Added a more function for viewing kubectl pod logs. A similar function was added in the [build repo](https://github.com/knative/build/pull/466/commits/2a509650b7b99b907a70db28a44b2243e2ecaed2), but this provides a more general function for all test-infra clients. 

I will merge this change into Build and utilize this function there in a follow up commit.

### DIFF
--- a/scripts/library.sh
+++ b/scripts/library.sh
@@ -38,24 +38,11 @@ if [[ -z "${GOPATH:-}" ]]; then
   fi
 fi
 
-# Capitalize the first letter of each word.
-# Parameters: $1..$n - words to capitalize.
-function capitalize() {
-  local words=($1)
-  local capitalized=()
-  for word in $@; do
-    local initial="$(echo ${word:0:1}| tr 'a-z' 'A-Z')"
-    capitalized+=("${initial}${word:1}")
-  done
-  echo "${capitalized[@]}"
-}
-
 # Useful environment variables
 [[ -n "${PROW_JOB_ID:-}" ]] && IS_PROW=1 || IS_PROW=0
 readonly IS_PROW
 readonly REPO_ROOT_DIR="$(git rev-parse --show-toplevel)"
 readonly REPO_NAME="$(basename ${REPO_ROOT_DIR})"
-readonly REPO_NAME_FORMATTED="Knative $(capitalize ${REPO_NAME//-/})"
 
 # On a Prow job, redirect stderr to stdout so it's synchronously added to log
 (( IS_PROW )) && exec 2>&1
@@ -211,6 +198,18 @@ function get_app_pods() {
   local namespace=""
   [[ -n $2 ]] && namespace="-n $2"
   kubectl get pods ${namespace} --selector=app=$1 --output=jsonpath="{.items[*].metadata.name}"
+}
+
+
+# Capitalize the first letter of each word.
+# Parameters: $1..$n - words to capitalize.
+function capitalize() {
+  local capitalized=()
+  for word in $@; do
+    local initial="$(echo ${word:0:1}| tr 'a-z' 'A-Z')"
+    capitalized+=("${initial}${word:1}")
+  done
+  echo "${capitalized[@]}"
 }
 
 # Dumps pod logs for the given app.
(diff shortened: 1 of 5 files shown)


scripts/library.sh
```diff
@@ -84,6 +84,18 @@ function function_exists() {
   [[ "$(type -t $1)" == "function" ]]
 }
 
+# Capitalize the first letter of each word.
+# Parameters: $1..$n - words to capitalize.
+function capitalize() {
+  local words=("$1")
```
srinivashegde86: do you need this var?

adrcunha: You do, it's transforming a string into a list.

test/unit/test-helper.sh
```diff
@@ -89,6 +89,16 @@ function mock_gcloud_function() {
   "$@" 2>&1
 }
 
+# Mocks the kubectl functionality (does nothing and outputs nothing).
+# Parameters: $1..$n - parameters passed to the function.
+function mock_kubectl_function() {
+  set -e
+  function gcloud() {
```
adrcunha: You're mocking gcloud here, not kubectl.

scripts/library.sh
```diff
@@ -200,6 +212,18 @@ function get_app_pods() {
   kubectl get pods ${namespace} --selector=app=$1 --output=jsonpath="{.items[*].metadata.name}"
 }
 
+# Dumps pod logs for the given app.
+# Parameters: $1 - app name.
+#             $2 - namespace.
+function dump_app_logs() {
+  echo ">>> Knative $(capitalize ${REPO_NAME//-/}) $1 logs:"
```
adrcunha: There are now 2 places using this approach (here and release.sh). Create a readonly env var called something like "REPO_NAME_FORMATTED" in library.sh and use it.

adrcunha:
This PR addresses 466, but doesn't fix it (yet). Please update both build and serving, as both are using their own version if this function.

test/unit/test-helper.sh
ericKlawitter: Wow, fixed. Why do the tests still pass with this?

scripts/library.sh
ericKlawitter: Done.

ericKlawitter:
> This PR addresses 466, but doesn't fix it (yet). Please update both build and serving, as both are using their own version if this function.

Will do. Doesn't this need to be in master first?

scripts/library.sh
```diff
@@ -38,11 +38,24 @@ if [[ -z "${GOPATH:-}" ]]; then
   fi
 fi
 
+# Capitalize the first letter of each word.
+# Parameters: $1..$n - words to capitalize.
+function capitalize() {
+  local words=($1)
```
adrcunha: You added back the unused variable.

scripts/library.sh
```diff
@@ -38,11 +38,24 @@ if [[ -z "${GOPATH:-}" ]]; then
   fi
 fi
 
+# Capitalize the first letter of each word.
```
adrcunha: Move this to before dump_app_logs, to keep all functions together.

scripts/library.sh
```diff
@@ -38,11 +38,24 @@ if [[ -z "${GOPATH:-}" ]]; then
...
+  done
+  echo "${capitalized[@]}"
+}
+
 # Useful environment variables
 [[ -n "${PROW_JOB_ID:-}" ]] && IS_PROW=1 || IS_PROW=0
 readonly IS_PROW
 readonly REPO_ROOT_DIR="$(git rev-parse --show-toplevel)"
 readonly REPO_NAME="$(basename ${REPO_ROOT_DIR})"
+readonly REPO_NAME_FORMATTED="Knative $(capitalize ${REPO_NAME//-/})"
```
adrcunha: Move this to the end of the file, together with other env vars that depend on function.

test/unit/test-helper.sh
adrcunha: Because kubectl does return something in the test, and the function does what's supposed to do. You're just checking for the header, and that's written no matter what pods are in the cluster.

scripts/library.sh
```diff
@@ -200,6 +200,30 @@ function get_app_pods() {
   kubectl get pods ${namespace} --selector=app=$1 --output=jsonpath="{.items[*].metadata.name}"
 }
 
+
```
```suggestion
```

PR #526, opened 2019-02-26
Use stable GO version 1.12 instead of 1.12rc1

### LINKED ISSUE #486: Update go version to 1.12rc1
A new feature in serving websockets need to use go1.12rc1 version,
link: https://github.com/knative/serving/pull/3240
Instead of manually installing this version in serving repo, updating it in prow-tests image, so that all jobs use consistent go version. 
This proposal was communicated among team leads and green lighted. Also, our local workstations are already using this version, so the risk of it breaking tests badly is pretty low.

TODO: once 1.12 version becomes stable version, switch back to use standard kubekins-e2e images

chaodaiG:
As stated in issue #486 , GO1.12rc1 was installed in prow test image as needed for building serving. Now that official GO1.12 is released, update to use stable version.
The installation steps comes from https://github.com/kubernetes/test-infra/blob/master/images/kubekins-e2e/Dockerfile#L64

Fixes #486

### DIFF
--- a/images/prow-tests/Dockerfile
+++ b/images/prow-tests/Dockerfile
@@ -29,10 +29,12 @@ RUN docker-credential-gcr configure-docker
 RUN apt-get install -y uuid-runtime  # for uuidgen
 RUN apt-get install -y rubygems  # for mdl
 
-# Update go version to 1.12rc1
-RUN go get golang.org/dl/go1.12rc1
-RUN go1.12rc1 download
-RUN cp $(which go1.12rc1) $(which go)
+# Install go at 1.12
+ENV GO_TARBALL "go1.12.linux-amd64.tar.gz"
+RUN rm -rf /usr/local/go && \
+    wget -q "https://storage.googleapis.com/golang/${GO_TARBALL}" && \
+    tar xzf "${GO_TARBALL}" -C /usr/local && \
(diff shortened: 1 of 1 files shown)


images/prow-tests/Dockerfile
```diff
@@ -29,10 +29,12 @@ RUN docker-credential-gcr configure-docker
 RUN apt-get install -y uuid-runtime  # for uuidgen
 RUN apt-get install -y rubygems  # for mdl
 
-# Update go version to 1.12rc1
-RUN go get golang.org/dl/go1.12rc1
-RUN go1.12rc1 download
-RUN cp $(which go1.12rc1) $(which go)
+# Install go at 1.12
```
adrcunha: Thanks for the update, Chao. To avoid future questions, concerns or "smater updates" here, I suggest you add a comment here that this is the way it's done for kubekins, and a permalink to the original code:

https://github.com/kubernetes/test-infra/blob/1e9b5dc3de4b268aab57c9c48120aa3dcf096bc6/images/kubekins-e2e/Dockerfile#L64

chaodaiG: Sounds good, added

adrcunha:
/lgtm
/approve

@chaodaiG please generate a new docker image.

Issue #527, opened 2019-02-26
e2e/presubmit tests logs should be easily available locally


adrcunha:
Currently the test log is just dumped to stdout. Having it (and a summary) available for easy review would improve the local dev experience.

PR #528, opened 2019-02-26
Always store the logs in `go_test_e2e`

adrcunha:
Addresses #527.

So it's easy to check them when running E2E tests locally.

`e2e-tests.sh` now displays where the artifacts were written once it finishes. For local runs, this is a new temporary directory and will contain both the JUnit XML file and the full go test log.


adrcunha:
Sample local output:

```
$ ./test/e2e-tests.sh
...
Test subprocess exited with code 1
Artifacts were written to /tmp/tmp.lzRF52YwR8
Test result code is 0
$ ls /tmp/tmp.lzRF52YwR8
junit_Q5xbYWiB.xml  go_test_Q5xbYWiB.log
```

PR #506, opened 2019-02-21
Flaky test reporting(part #4) Github issues tracking for flaky tests

chaodaiG:
Creating/updating/closing flaky tests related issues, based on test results collected from most recent continuous flows and data structure defined in [PR465](https://github.com/knative/test-infra/pull/465).

Each flaky test is tracked by one issue, and if there are too many tests found flaky there will be a bulk issue created instead of flooding issues pool. Issue identity is embedded in a reserved part of issue body instead of title, so developers have enough freedom of making adjustment to title and other parts of issue body. History of test results are tracked by prepend latest result to the first comment, and all old histories are hidden by a collapsible Markdown template. Other than text description, test results history is also represented by a unicode graph similar to TestGrid.

Part of: #132

PR #536, opened 2019-02-28
Make all github calls as receivable by github client

srinivashegde86:
All the methods are direct function calls that can be made directly without initializing the github client. Updating this so that:

1. We can get a valid github client to perform github actions
1. Add the methods to the client so that we can call client.<Method> instead of relying on a global var
1. We can get a valid github user to perform actions on it directly
1. Update github_issue to hold these values and make calls to the issue struct directly


srinivashegde86:
will wait for https://github.com/knative/test-infra/pull/506 to be merged as there are some conflicts.

/hold

srinivashegde86:
/hold cancel
the PR has now merged.

tools/flaky-test-reporter/ghutil/ghutil.go
```diff
@@ -40,56 +40,65 @@ const (
...
+	*github.Client
+}
+
+// GithubUser provides methods to perform operations as a github user
+type GithubUser struct {
+	*github.User
+}
+
+// GetGithubClient explicitly authenticates to github with giving token and returns a handle
+func GetGithubClient(tokenFilePath string) (*GithubClient, error) {
```
adrcunha: nit: shouldn't this be called NewGithubClient then?

PR #465, opened 2019-02-11
Flaky test reporting(part #2): define data structure and data collection

chaodaiG:
Since flaky test reporting is designed to do both Github issues tracking and Slack notification, defines a common data structure to be used by data collector, Github issues creator, and Slack notifier is needed. This PR:
- defines data structure for storing configuration and test results for a repo
- collecting data from continuous runs and feed them into the data structure defined above

Part of: #132

PR #541, opened 2019-03-01
Flaky test reporting(part #5) Automatic Slack notification for flaky tests

chaodaiG:
Send Slack notification to channels, based on test results collected from most recent continuous flows and data structure defined in [PR465](https://github.com/knative/test-infra/pull/465).

Part of: #132


chaodaiG:
see example message here:
https://knative.slack.com/archives/CGJTUSL1K/p1551466230000800

tools/flaky-test-reporter/error.go
```diff
@@ -0,0 +1,36 @@
...
+package main
+
+import (
+	"fmt"
+	"strings"
+)
+
+// combineErrors combines slice of errors and return a single error
+func combineErrors(errs []error) error {
+	if nil == errs || 0 == len(errs) {
```
srinivashegde86: when can this be nil?

srinivashegde86: not needed. The loop below will handle empty slices just fine. Only things night be we will be adding a "\n" to the list if its empty, which should be fine.

shared/testgrid/testgrid.go
```diff
@@ -31,6 +31,8 @@ import (
 const (
 	filePrefix = "junit_"
 	extension  = ".xml"
+
```
srinivashegde86: remove this extra line

tools/flaky-test-reporter/main.go
```diff
@@ -30,14 +30,20 @@ import (
 
 func main() {
 	serviceAccount := flag.String("service-account", os.Getenv("GOOGLE_APPLICATION_CREDENTIALS"), "JSON key file for service account to use")
```
srinivashegde86: update the desc to indicate that this if for GCS auth

shared/testgrid/testgrid.go
```diff
@@ -43,6 +45,11 @@ func createDir(dirPath string) error {
 	return nil
 }
 
+// GetBaseURL returns baseURL of testgrid for knative
+func GetBaseURL() string {
```
srinivashegde86: can this be an exported var instead?

tools/flaky-test-reporter/result.go
```diff
@@ -68,18 +75,36 @@ func (ts *TestStat) hasEnoughRuns() bool {
...
-			flakyCount++
+			flakyTests = append(flakyTests, testName)
 		}
 	}
-	return float32(flakyCount)/float32(totalCount), nil
+	return flakyTests
+}
+
+
+func getFlakyRate(rd *RepoData) (float32, error) {
```
srinivashegde86: you are only returning nil as err. Do we need this as a return var?

tools/flaky-test-reporter/result.go
```diff
@@ -167,6 +192,34 @@ func collectTestResultsForRepo(jc *JobConfig) (*RepoData, error) {
...
+				res[i] = junit.Passed
+			default:
+				res[i] = junit.Skipped
+		}
+	}
+	return res
+}
+
+func intSliceContains(its []int, target int) bool {
+	if nil == its {
```
srinivashegde86: ditto. can this be nil?

tools/flaky-test-reporter/run.go
```diff
@@ -0,0 +1,32 @@
...
+
+// run.go controls how to run functions that needs dryrun support
+
+package main
+
+import (
+	"log"
+)
+
+func run(message string, call func() error, dryrun *bool) error {
```
srinivashegde86: any reason dryrun is a pointer and not a bool?

tools/flaky-test-reporter/slack_notification.go
```diff
@@ -0,0 +1,153 @@
...
+)
+
+var (
+	slackUsername = "Knative bot"
+	slackTokenStr string
+	slackIconEmoji *string
+)
+
+// jobNameTestgridURLMap contains harded coded mapping of job name: Testgrid tab URL relative to base URL
+var jobNameTestgridURLMap = map[string]string {
```
srinivashegde86: this can be in the var block above

tools/flaky-test-reporter/slack_notification.go
```diff
@@ -0,0 +1,153 @@
...
+	"ci-knative-serving-continuous": "knative-serving#continuous",
+}
+
+// slackChannel contains channel logical name and Slack identity
+type slackChannel struct {
+	name, identity string
+}
+
+// slackChannelsMap defines mapping of repo: slack channels
+var slackChannelsMap = map[string][]slackChannel{
```
srinivashegde86: part of the var block?

tools/flaky-test-reporter/slack_notification.go
```diff
@@ -0,0 +1,153 @@
...
+	b, err := ioutil.ReadFile(slackTokenPath)
+	if err != nil {
+		return err
+	}
+	slackTokenStr = strings.TrimSpace(string(b))
+	return nil
+}
+
+// postMessage does http post
+func postMessage(url string, uv *url.Values) error {
```
srinivashegde86: these messages should be on the slack client and not independent functions

tools/flaky-test-reporter/slack_notification.go
```diff
@@ -0,0 +1,153 @@
...
+	resp, err := http.PostForm(url, *uv)
```
srinivashegde86: do we need to sanitize the url before making the call here?

tools/flaky-test-reporter/slack_notification.go
```diff
@@ -0,0 +1,153 @@
...
+
+const (
+	slackChatPostMessageURL = "https://slack.com/api/chat.postMessage"
+	// default filter for testgrid link
+	testgridFilter          = "exclude-non-failed-tests=20"
+)
+
+var (
+	slackUsername = "Knative bot"
+	slackTokenStr string
```
srinivashegde86: these two can be part of the slack client

tools/flaky-test-reporter/slack_notification.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	return nil
+}
+
+// writeSlackMessage posts text to channel
+func writeSlackMessage(text, channel string) error {
+	uv := &url.Values{}
+	uv.Add("username", slackUsername)
+	uv.Add("token", slackTokenStr)
+	if nil != slackIconEmoji {
```
srinivashegde86: what happens if I send this as nil?

tools/flaky-test-reporter/slack_notification.go
```diff
@@ -0,0 +1,153 @@
...
+
+func sendSlackNotifications(repoDataAll []*RepoData, dryrun *bool) error {
+	var allErrs []error
+	for _, rd := range repoDataAll {
+		channels, ok := slackChannelsMap[rd.Config.Repo]
+		if !ok {
+			log.Printf("cannot find Slack channel for repo '%s', skip Slack notifiction", rd.Config.Repo)
+			continue
+		}
+		for _, channel := range channels {
```
srinivashegde86: create a go channel to send this parallely to all channels

tools/flaky-test-reporter/error.go
chaodaiG: removed

chaodaiG: removed

shared/testgrid/testgrid.go
chaodaiG: gone

tools/flaky-test-reporter/main.go
chaodaiG: done

shared/testgrid/testgrid.go
chaodaiG: done

tools/flaky-test-reporter/result.go
chaodaiG: removed

tools/flaky-test-reporter/result.go
chaodaiG: removed

tools/flaky-test-reporter/run.go
chaodaiG: it's for consistency, as it's a pointer in the beginning

tools/flaky-test-reporter/slack_notification.go
chaodaiG: moved

tools/flaky-test-reporter/slack_notification.go
chaodaiG: ditto

tools/flaky-test-reporter/slack_notification.go
chaodaiG: done

tools/flaky-test-reporter/slack_notification.go
chaodaiG: since this is not really exported function, I don't think there is a need sanitizing

tools/flaky-test-reporter/slack_notification.go
chaodaiG: moved

tools/flaky-test-reporter/slack_notification.go
chaodaiG: there is no emoji :)

tools/flaky-test-reporter/slack_notification.go
chaodaiG: done

tools/flaky-test-reporter/run.go
srinivashegde86: since its a flag, it will return the pointer. But, this function can take in a bool instead as we can catch this as compile time instead.

tools/flaky-test-reporter/slack_notification.go
```diff
@@ -0,0 +1,170 @@
...
+	"time"
+
+	"net/http"
+	"net/url"
+
+	"github.com/knative/test-infra/shared/testgrid"
+)
+
+const (
+	knativeBotName          = "Knative bot"
```
adrcunha: Please don't use EngProd in the bot's name. We call the Prow robot "Knative Prow Robot" and it has a robot icon. I suggest you use the icon for consistency, and rename the bot to something like "Knative Testgrid Robot" maybe? And if possible, fill its profile clearly indicating what it is and what it does (refer to Knative Prow Robot for example).

tools/flaky-test-reporter/slack_notification.go
```diff
@@ -0,0 +1,170 @@
+/*
+Copyright 2018 The Knative Authors
```
adrcunha: nit: 2019 here and everywhere.

tools/flaky-test-reporter/slack_notification.go
```diff
@@ -0,0 +1,170 @@
...
+	if !ok {
+		return "", fmt.Errorf("cannot find Testgrid tab for job '%s'", jobName)
+	}
+	return fmt.Sprintf("%s/%s&%s", testgrid.BaseURL, url, testgridFilter), nil
+}
+
+// createSlackMessageForRepo creates slack message layout from RepoData
+func createSlackMessageForRepo(rd *RepoData) string {
+	flakyTests := getFlakyTests(rd)
+	message := fmt.Sprintf("(As of %s)\nActive flaky tests in repo '%s': %d",
```
adrcunha: Rewording suggestion:

```suggestion
	message := fmt.Sprintf("As of %s, there are %d flaky tests in '%s'",
```

tools/flaky-test-reporter/slack_notification.go
```diff
@@ -0,0 +1,170 @@
...
+		time.Unix(*rd.LastBuildStartTime, 0).String(), rd.Config.Repo, len(flakyTests))
+	for _, testName := range flakyTests {
+		message += fmt.Sprintf("\n>- %s", testName)
+		// TODO(chaodaiG): try adding Github issues when PR #506 merged
+	}
+	if testgirdTabURL, err := getTestgridTabURL(rd.Config.Name); nil != err {
+		log.Println(err) // don't fail as this could be optional
+	} else {
+		message += fmt.Sprintf("\nSee Testgrid tab for most recent flaky tests: %s", testgirdTabURL)
```
adrcunha:
```suggestion
		message += fmt.Sprintf("\nSee Testgrid for up-to-date flaky tests information: %s", testgirdTabURL)
```

tools/flaky-test-reporter/slack_notification.go
```diff
@@ -0,0 +1,170 @@
...
+// createSlackMessageForRepo creates slack message layout from RepoData
+func createSlackMessageForRepo(rd *RepoData) string {
+	flakyTests := getFlakyTests(rd)
+	message := fmt.Sprintf("(As of %s)\nActive flaky tests in repo '%s': %d",
+		time.Unix(*rd.LastBuildStartTime, 0).String(), rd.Config.Repo, len(flakyTests))
+	for _, testName := range flakyTests {
+		message += fmt.Sprintf("\n>- %s", testName)
+		// TODO(chaodaiG): try adding Github issues when PR #506 merged
+	}
+	if testgirdTabURL, err := getTestgridTabURL(rd.Config.Name); nil != err {
```
adrcunha: typo: testgridTabURL

tools/flaky-test-reporter/slack_notification.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	return message
+}
+
+func sendSlackNotifications(repoDataAll []*RepoData, c *SlackClient, dryrun *bool) error {
+	var allErrs []error
+	for _, rd := range repoDataAll {
+		channels, ok := slackChannelsMap[rd.Config.Repo]
+		if !ok {
+			log.Printf("cannot find Slack channel for repo '%s', skip Slack notifiction", rd.Config.Repo)
```
adrcunha:
```suggestion
			log.Printf("cannot find Slack channel for repo '%s', skipping Slack notification", rd.Config.Repo)
```

tools/flaky-test-reporter/slack_notification.go
```diff
@@ -0,0 +1,170 @@
...
+		}
+
+		ch := make(chan bool, len(channels))
+		wg := sync.WaitGroup{}
+		for _, channel := range channels {
+			wg.Add(1)
+			go func(wg *sync.WaitGroup) {
+				message := createSlackMessageForRepo(rd)
+				if err := run(
+					fmt.Sprintf("post Slack message for repo '%s' in channel '%s'", rd.Config.Repo, channel.name),
```
adrcunha: Debug?

tools/flaky-test-reporter/run.go
chaodaiG: done

tools/flaky-test-reporter/slack_notification.go
chaodaiG: done

tools/flaky-test-reporter/slack_notification.go
chaodaiG: nice catch. fixed

tools/flaky-test-reporter/slack_notification.go
chaodaiG: not really debug, the `run` help function is a wrapper for dryrun, which prints this message with `[dryrun]` prefix if in dryrun mode, otherwise print this message straight

tools/flaky-test-reporter/slack_notification.go
chaodaiG: It is now recommended by Slack Web API to use application associated bot tokens. So Mark created a Slack app and named it `Knative Engprod`. The only thing we need from the app is a bot token, which allows us modify the bot display name, as `Knative bot` in this case, so changing this line to `knativeBotName = "Knative Testgrid Robot"` would do the job. I will try to figure out the icon part(which is also configurable with `iconEmoji`)

chaodaiG: Currently out Slack plan only supports up to 10 apps, so we might not want to rename the app to be `Testgrid` only

chaodaiG: As discussed in person, the current app is replaced with a new app named "Knative Testgrid Robot", and the default icon of this new app is the red green Knative robot

tools/flaky-test-reporter/slack_notification.go
```diff
@@ -121,26 +121,26 @@ func getTestgridTabURL(jobName string) (string, error) {
 // createSlackMessageForRepo creates slack message layout from RepoData
 func createSlackMessageForRepo(rd *RepoData) string {
 	flakyTests := getFlakyTests(rd)
-	message := fmt.Sprintf("(As of %s)\nActive flaky tests in repo '%s': %d",
-		time.Unix(*rd.LastBuildStartTime, 0).String(), rd.Config.Repo, len(flakyTests))
+	message := fmt.Sprintf("As of %s, there are %d flaky tests in '%s'",
+		time.Unix(*rd.LastBuildStartTime, 0).String(), len(flakyTests), rd.Config.Repo)
 	for _, testName := range flakyTests {
 		message += fmt.Sprintf("\n>- %s", testName)
 		// TODO(chaodaiG): try adding Github issues when PR #506 merged
```
adrcunha: Is it possible doing this TODO now?

chaodaiG: yep, and here it is

tools/flaky-test-reporter/main.go
```diff
@@ -64,9 +75,15 @@ func main() {
 		repoDataAll = append(repoDataAll, rd)
 	}
 
-	ghi, err := Setup(*githubToken)
-	if err != nil {
-		log.Fatalf("Cannot setup github: %v", err)
+	githubErr := ghi.processGithubIssues(repoDataAll, *dryrun)
```
srinivashegde86: we should fail fast here and not create both clients if there is an error.
if githubErr {
log.Printf("gituhb errors: %v, err)
os.Exit(1)
}

tools/flaky-test-reporter/main.go
```diff
@@ -64,9 +75,15 @@ func main() {
...
+	slackErr := sendSlackNotifications(repoDataAll, slackClient, ghi, *dryrun)
+	if nil != githubErr {
+		log.Printf("Github step failures:\n%v", githubErr)
+	}
+	if nil != slackErr {
+		log.Printf("Slack step failures:\n%v", slackErr)
+	}
+	if nil != githubErr || nil != slackErr { // Fail this job if there is any error
```
srinivashegde86: can we combine this in the if condition above?

tools/flaky-test-reporter/result.go
```diff
@@ -88,18 +88,23 @@ func (ts *TestStat) getTestStatus() string {
...
-			flakyCount++
+			flakyTests = append(flakyTests, testName)
 		}
 	}
-	return float32(flakyCount)/float32(totalCount), nil
+	return flakyTests
+}
+
+
+func getFlakyRate(rd *RepoData) float32 {
```
srinivashegde86: any reason this is float32 and not float64?

tools/flaky-test-reporter/slack_notification.go
```diff
@@ -0,0 +1,180 @@
...
+const (
+	knativeBotName          = "Knative Testgrid Robot"
+	slackChatPostMessageURL = "https://slack.com/api/chat.postMessage"
+	// default filter for testgrid link
+	testgridFilter = "exclude-non-failed-tests=20"
+)
+
+var (
+	// jobNameTestgridURLMap contains harded coded mapping of job name: Testgrid tab URL relative to base URL
+	jobNameTestgridURLMap = map[string]string{
```
srinivashegde86: maybe this var should be in testgrid.go and not here?

tools/flaky-test-reporter/slack_notification.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+	return nil
+}
+
+// writeSlackMessage posts text to channel
+func (c *SlackClient) writeSlackMessage(text, channel string) error {
+	uv := &url.Values{}
+	uv.Add("username", c.userName)
+	uv.Add("token", c.tokenStr)
+	if nil != c.iconEmoji {
```
srinivashegde86: we can remove this condition and just add this directly.
uv := url.Value {
username: c.userName,
token: c.tokenStr,
icon_emoji: c.iconEmoji,
...
}

return c.postMessage(slackUrl, &uv)

tools/flaky-test-reporter/slack_notification.go
```diff
@@ -0,0 +1,180 @@
...
+		uv.Add("icon_emoji", *c.iconEmoji)
+	}
+	uv.Add("channel", channel)
+	uv.Add("text", text)
+
+	return c.postMessage(slackChatPostMessageURL, uv)
+}
+
+// getTestgridTabURL gets Testgrid URL for giving job
+func getTestgridTabURL(jobName string) (string, error) {
```
srinivashegde86: should be part of testgrid?

tools/flaky-test-reporter/main.go
chaodaiG: This was by design. There are 2 different kinds of errors might occur here, first one is failure that could cause inaccuracy, in this case it should fast fail, and this is handled by `fatalf` in github notification. The other kind is github operations failure, which doesn't affect Slack notification accuracy, so Slack notification should still be sent here.

tools/flaky-test-reporter/main.go
chaodaiG: do you mean `defer`?

tools/flaky-test-reporter/result.go
chaodaiG: This rate is used to be compared with a threshold like 1%, so the precision of float32 is sufficient

tools/flaky-test-reporter/slack_notification.go
chaodaiG: I thought about it too. The problem of putting it there, is that it's under `shared`, so just put this one line mapping there would feel strange, and hard coding every mapping there would make it really hard to maintain. So put it here temporarily until we figure out a better way computing it in `testgrid.go`

tools/flaky-test-reporter/slack_notification.go
chaodaiG: ditto

tools/flaky-test-reporter/slack_notification.go
chaodaiG: the structure of `uv` is `map[string][]string`, so it would be more like `username: append(uv["username"], c.userName)`. So `.Add` seems like a better option

tools/flaky-test-reporter/main.go
srinivashegde86: ok. thanks for explaining, Probably add a comment here then we need to verify both.

tools/flaky-test-reporter/main.go
srinivashegde86: I mean, we dont need another if condition here. Maybe we can have a exit func and call that in the if conditions above?

func exit(msg string, args...) {
log.Printf(msg, args...)
os.Exit(1)
}

tools/flaky-test-reporter/slack_notification.go
srinivashegde86: I dont think its strange. We dont have to put all the mappings. Only the ones we need. So, for now we can have only one entry in the map.

tools/flaky-test-reporter/main.go
chaodaiG: there might be cases where there are errors caught from both github and slack, and they should both be printed at the end for easier debugging, it's not practical to exit in the first if condition

tools/flaky-test-reporter/main.go
chaodaiG: added

PR #554, opened 2019-03-07
Webhook-APICoverage: Prow job env variable fix

dushyanthsc:
This change adds an environment variable into the Webhook API Coverage job that is required by the tool.

### DIFF
--- a/ci/prow/make_config.go
+++ b/ci/prow/make_config.go
@@ -714,6 +714,7 @@ func generatePeriodic(title string, repoName string, periodicConfig yaml.MapSlic
 			}
 			jobNameSuffix = "webhook-apicoverage"
 			data.Base.Command = webhookAPICoverageScript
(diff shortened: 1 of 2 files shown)


dushyanthsc:
cc: @adrcunha

ci/prow/make_config.go
```diff
@@ -714,6 +714,7 @@ func generatePeriodic(title string, repoName string, periodicConfig yaml.MapSlic
 			}
 			jobNameSuffix = "webhook-apicoverage"
 			data.Base.Command = webhookAPICoverageScript
+			addEnvToJob(&data.Base, "SYSTEM_NAMESPACE", "knative-serving")
```
adrcunha: This is hardcoded for serving, I believe this is per repo, right? So it would be:

```suggestion
			addEnvToJob(&data.Base, "SYSTEM_NAMESPACE", data.Base.RepoNameForJob)
```

dushyanthsc: Fixed.

PR #559, opened 2019-03-07
Fail fast in release.sh when testing (#331)

### LINKED ISSUE #331: Fail fast in release.sh when testing
We don't want to wait until all unit/integration tests finish running if build tests failed, for example.

yt3liu:
Do not run any further tests if previous test already failed. 

Fixes #331

### DIFF
--- a/scripts/presubmit-tests.sh
+++ b/scripts/presubmit-tests.sh
@@ -22,6 +22,7 @@ source $(dirname ${BASH_SOURCE})/library.sh
 # Custom configuration of presubmit tests
 readonly DISABLE_MD_LINTING=${DISABLE_MD_LINTING:-0}
 readonly DISABLE_MD_LINK_CHECK=${DISABLE_MD_LINK_CHECK:-0}
+readonly PRESUBMIT_TEST_FAIL_FAST=${PRESUBMIT_TEST_FAIL_FAST:-0}
 
 # Extensions or file patterns that don't require presubmit tests.
 readonly NO_PRESUBMIT_FILES=(\.png \.gitignore \.gitattributes ^OWNERS ^OWNERS_ALIASES ^AUTHORS)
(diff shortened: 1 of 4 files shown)


scripts/presubmit-tests.sh
```diff
@@ -317,8 +317,14 @@ function main() {
   fi
 
   run_build_tests || failed=1
-  run_unit_tests || failed=1
-  run_integration_tests || failed=1
+  # Don't run unit tests if build tests failed
+  if (( ! failed )); then
+    run_unit_tests || failed=1
```
adrcunha: You're changing the behavior or the presubmit script, not the release script. The issue was about having this behavior only for releases. I suggest using a global environment variable to make the behavior configurable.

yt3liu: Updated to use the PRESUBMIT_TEST_FAIL_FAST global environment variable to configure the presubmit-test.

scripts/presubmit-tests.sh
```diff
@@ -317,8 +317,14 @@ function main() {
   fi
```
adrcunha: Also please update the README.md with the new feature.

test/unit/presubmit-tests.sh
```diff
@@ -117,6 +128,51 @@ function test_custom_runners_basic() {
   }
 }
 
+function test_custom_runners_fail_slow() {
```
adrcunha: Great work on the unit tests, thanks!

scripts/presubmit-tests.sh
yt3liu: Updated to initialize the readonly env var at the beginning of the script.

scripts/presubmit-tests.sh
```diff
@@ -317,8 +317,14 @@ function main() {
...
 
   run_build_tests || failed=1
-  run_unit_tests || failed=1
-  run_integration_tests || failed=1
+  # Don't run unit tests if build tests failed
```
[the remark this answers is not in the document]
yt3liu: Updated.

scripts/presubmit-tests.sh
yt3liu: Updated the README to include the new configuration.

adrcunha:
Nice PR, thanks Joyce!

/lgtm

PR #562, opened 2019-03-08
Enable dot releases for eventing

adrcunha:
Fixes knative/eventing#873

/hold
Requires knative/eventing#874 to be merged first.


adrcunha:
874 is merged
/hold cancel

PR #566, opened 2019-03-11
updating plank to v20190311

chaodaiG:
plank was recently updated to v20190109 but not it's utilities images, which makes us missing some features, for example `decoration_config` for overriding default timeout, which is required by currently failing images cleanup job due to timeout.

### DIFF
--- a/ci/prow/cluster.yaml
+++ b/ci/prow/cluster.yaml
@@ -148,7 +148,7 @@ spec:
       serviceAccountName: "plank"
       containers:
       - name: plank
-        image: gcr.io/k8s-prow/plank:v20190109-716fe27
+        image: gcr.io/k8s-prow/plank:v20190311-a967141
         args:
         - --dry-run=false
         volumeMounts:
--- a/ci/prow/cluster.yaml
+++ b/ci/prow/cluster.yaml
@@ -148,6 +148,7 @@ spec:
       serviceAccountName: "plank"
       containers:
       - name: plank
+        # Update plank utility_images versions in config.yaml when updating this version
         image: gcr.io/k8s-prow/plank:v20190311-a967141
         args:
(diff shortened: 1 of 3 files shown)


adrcunha:
Thanks for catching this!

/lgtm

ci/prow/cluster.yaml
```diff
@@ -148,7 +148,7 @@ spec:
       serviceAccountName: "plank"
       containers:
       - name: plank
-        image: gcr.io/k8s-prow/plank:v20190109-716fe27
+        image: gcr.io/k8s-prow/plank:v20190311-a967141
```
adrcunha: Please add a comment here and in the generator about keeping the tags in sync, so this doesn't happen again. Also thanks for switching to tags instead of SHA for the images, so we can have such parity.

chaodaiG: good idea, done

chaodaiG:
> Thanks for catching this!
> 
> /lgtm

credit to @cjwagner , he helped me identify this problem

ci/prow/config.yaml
```diff
@@ -25,6 +25,7 @@ plank:
     timeout: 7200000000000 # 2h
     grace_period: 15000000000 # 15s
     utility_images:
+      # Updatse these versions when updating plank version in cluster.yaml
```
adrcunha: Typo: Updatse

ci/prow/cluster.yaml
adrcunha: Please add a comment here as well.

chaodaiG: done

ci/prow/config.yaml
chaodaiG: nice catch, fixed

Issue #390, opened 2019-01-14
Extend cleanup job to delete old clusters as well


adrcunha:
Rationale: if a GKE cluster in a boskos project is older than 1h, much probably the test using it failed do clean it up. This will impact quotas in the boskos pool, and can cause E2E tests to fail.

It's safe to assume that clusters older than 6h are stale. As of now, no E2E test takes more than 20 minutes to run.

PR #570, opened 2019-03-13
Cleanup job also deletes stale Boskos clusters

chaodaiG:
there needs to be a mechanism scanning and cleaning up afterwards, in case Boskos clusters are not cleaned up by itself. Add this function to cleanup job for scanning all Boskos projects and deleting any clusters considered old, such as 24 hours.

Fixes: #390

### DIFF
--- a/scripts/library.sh
+++ b/scripts/library.sh
@@ -431,7 +431,7 @@ function remove_broken_symlinks() {
 # Return whether the given parameter is knative-tests.
 # Parameters: $1 - project name
 function is_protected_project() {
-  [[ -n $1 && "$1" =~ "^knative-tests$" ]]
+  [[ -n $1 && "$1" -eq "knative-tests" ]]
 }
 
 # Returns the canonical path of a filesystem object.
--- a/scripts/library.sh
+++ b/scripts/library.sh
@@ -427,6 +427,7 @@ function remove_broken_symlinks() {
       continue
     fi
   done
+}
 
 # Return whether the given parameter is knative-tests.
 # Parameters: $1 - project name
--- a/scripts/library.sh
+++ b/scripts/library.sh
@@ -427,6 +427,11 @@ function remove_broken_symlinks() {
       continue
     fi
   done
+
+# Return whether the given parameter is knative-tests.
+# Parameters: $1 - project name
+function is_protected_project() {
+  [[ -n $1 && "$1" =~ "^knative-tests$" ]]
 }
 
 # Returns the canonical path of a filesystem object.
--- a/scripts/library.sh
+++ b/scripts/library.sh
@@ -431,7 +431,7 @@ function remove_broken_symlinks() {
 # Return whether the given parameter is knative-tests.
 # Parameters: $1 - project name
 function is_protected_project() {
-  [[ -n $1 && "$1" -eq "knative-tests" ]]
+  [[ -n "$1" && "$1" == "knative-tests" ]]
 }
 
 # Returns the canonical path of a filesystem object.
--- a/tools/cleanup/README.md
+++ b/tools/cleanup/README.md
@@ -1,34 +1,26 @@
 # Resources Clean Up Tool
 
-This tool is designed to clean up stale resources from gcr, for now it only deletes old images created during testing.
+This tool is designed to clean up stale resources from gcr, for now it deletes old images and clusters created during testing.
 
 ## Basic Usage
 
-Directly invoke [cleanup.sh](cleanup.sh) script with certain flags. There is no-op if invoking or sourcing this script without arguments.
+Directly invoke [cleanup.sh](cleanup.sh) script with certain flags, but don't source this script.
 
 By default the current gcloud credentials are used to delete the images. If necessary, use the flag `--service-account _key-file.json_` to specify a service account that will be performing the access to the gcr.
 
-### Clean up old images from multiple gcrs
+Projects to be cleaned up are expected to be defined in a `resources.yaml` file. To remove old images and clusters from them, call [cleanup.sh](cleanup.sh) with following flags:
 
-Projects to be cleaned up are expected to be defined in a `resources.yaml` file. To remove old images from them, call [cleanup.sh](cleanup.sh) with action "delete-old-gcr-images" and following flags:
 - "--project-resource-yaml" as path of `resources.yaml` file - Mandatory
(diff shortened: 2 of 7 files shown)


tools/cleanup/README.md
```diff
@@ -1,34 +1,26 @@
...
-### Clean up old images from a specific gcr
-
-Cleaning up from a specific gcr is supported, except for some special ones (_knative-release_ and _knative-nightly_). Call [cleanup.sh](cleanup.sh) with action "delete-old-images-from-gcr" and following flags:
-- "--gcr-to-cleanup" as name of gcr, e.g. "gcr.io/foo" - Mandatory
-- "--days-to-keep" - Optional, default `365`
-
-Example:
-
-```./cleanup.sh "delete-old-images-from-gcr" --gcr-to-cleanup "gcr.io/foo" --days-to-keep 90```
+```./cleanup.sh "delete-old-gcr-images" --project-resource-yaml "ci/prow/boskos/resources.yaml" --days-to-keep-images 90 --days-to-keep-clusters 24```
```
adrcunha:
```suggestion
```./cleanup.sh --project-resource-yaml "ci/prow/boskos/resources.yaml" --days-to-keep-images 90 --days-to-keep-clusters 24```
```

tools/cleanup/cleanup-functions.sh
```diff
@@ -48,21 +48,51 @@ function delete_old_images_from_gcr() {
...
+
+  for project in $1; do
+    echo "Start deleting clusters from ${project}"
+
+    is_protected_project $project && \
+      abort "Target project set to $project, which is forbidden"
+
+    local current_time=$(date +%s)
+    local target_time=$(date -d "`date -d @${current_time}`-$2hours" +%s)
+    # Fail if the difference of current time and target time is not 3600 times hours to keep
```
adrcunha: Why this check? Did I miss something in the tests?

tools/cleanup/cleanup-functions.sh
```diff
@@ -48,21 +48,51 @@ function delete_old_images_from_gcr() {
...
+    if (( ! DRY_RUN )); then # This should only fail this job but not blocking PR
+      [[ "$((3600*$2))" -eq "$(($current_time-$target_time))" ]] || abort "date operation failed"
+    fi
+
+    gcloud --format='get(name,createTime,zone)' container clusters list --project=$project --limit=99999 | \
+    while read cluster_name cluster_createtime cluster_zone; do
+      (( name )) && (( ! cluster_zone )) && abort "list cluster output missing cluster zone"
```
adrcunha: Do you mean `cluster_name` instead of `name`? Also, prefer `-z` for checking for empty strings.

tools/cleanup/cleanup.sh
```diff
@@ -68,25 +68,21 @@ if [[ -z $1 ]]; then
   abort "missing parameters to the tool"
 fi
 
-FUNCTION_TO_RUN=$1
-shift
 parse_args $@
 
 (( DRY_RUN )) && echo "-- Running in dry-run mode, no image deletion --"
```
adrcunha:
```suggestion
(( DRY_RUN )) && echo "-- Running in dry-run mode, no resource deletion --"
```

tools/cleanup/cleanup.sh
```diff
@@ -68,25 +68,21 @@ if [[ -z $1 ]]; then
...
+echo "- from projects defined in '${PROJECT_RESOURCE_YAML}', matching '${RE_PROJECT_NAME}"
```
adrcunha:
```suggestion
echo "Iterating over projects defined in '${PROJECT_RESOURCE_YAML}', matching '${RE_PROJECT_NAME}"
```

tools/cleanup/cleanup-functions.sh
chaodaiG: This might not be necessary, just want to ensure that the second `date` command always does what it supposed to be: get a timestamp 2 hours before `${current_time}. 
Check it here instead of in unit testing as I don't think this should fail PR.

scripts/library.sh
```diff
@@ -408,6 +408,12 @@ function is_protected_gcr() {
   [[ -n $1 && "$1" =~ "^gcr.io/knative-(releases|nightly)/?$" ]]
 }
 
+# Return whether the given parameter is knative-tests.
+# Parameters: $1 - project name
+function is_protected_project() {
+  [[ -n $1 && "$1" =~ "^knative-tests$" ]]
```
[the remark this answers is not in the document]
chaodaiG: done

tools/cleanup/cleanup-functions.sh
chaodaiG: nice catch, this is certainly an error. Fixed

scripts/library.sh
```diff
@@ -408,6 +408,12 @@ function is_protected_gcr() {
   [[ -n $1 && "$1" =~ "^gcr.io/knative-(releases|nightly)/?$" ]]
 }
 
+# Return whether the given parameter is knative-tests.
+# Parameters: $1 - project name
+function is_protected_project() {
+  [[ -n $1 && "$1" -eq "knative-tests" ]]
```
```suggestion
  [[ -n "$1" && "$1" == "knative-tests" ]]
```

tools/cleanup/cleanup-functions.sh
adrcunha: Sorry, I'm failing to see why you check for dry-run here. Your tests only use --dry-run when checking for wrong/missing values, so they never reach this point.

adrcunha: @chaodaiG I don't think we clarified this point?

chaodaiG: sorry I thought we did. Now I understand the concern better, will get back to address this when I got a chance

chaodaiG: @adrcunha , I just double checked, the last test does reach this point, and it's the reason why `DRY_RUN` is checked here

adrcunha: Please clearly document the reason here. "but not blocking PR" is obscure. I suggest something like "Don't check on dry runs, as dry run is used for unit testing".

chaodaiG: done

PR #582, opened 2019-03-18
Increase serving test timeouts

ericKlawitter:
Per [testgrid](https://testgrid.knative.dev/knative-serving#continuous), the serving test suite is timing out. This changes back the timeout and run frequency configs of serving tests to address these timeouts.


adrcunha:
/lgtm

The root cause should be investigated, instead of simply addressing the timeout problem.

adrcunha:
/hold cancel

Let's get the flakiness addressed while the investigation is in progress.

PR #583, opened 2019-03-18
Also generate testgrid config with make_config.go

### LINKED ISSUE #483: Also generate testgrid config with make_config.go
With the Prow config being generated automatically (https://github.com/knative/test-infra/issues/382), we should extend this to also generate the Testgrid config file, thus keeping both configs in sync.

Fredy-Z:
Fixes #483

1. Wrote a `make_config.go` file to generate the testgrid configuration file from `config_knative.yaml`; added `make config` command
2. Refactored the original code, moved common helper functions to `shared/configgenerator/configgenerator.go`


adrcunha:
/hold

I don't think this is the best approach. 483 clearly states that the Prow config generator should be extended. This is desirable because reduces boilerplate and duplicated code and ensures that both configs are kept in sync. Obviously, for testing, debugging & etc the features should be enabled/disabled by a flag (i.e., by default both configs are generated, but you can generate only one).

Specifically, I would like to have something like:
`--generate-prow-config`: boolean, defaults to true, outputs to stdout unless destination is given
`--prow-config-output`: string, points to a file that will be created, defaults to stdout
`--generate-testgrid-config`: boolean, defaults to true, outputs to stdout unless destination is given
`--testgrid-config-output`: string, points to a file that will be created, defaults to stdout

I think the config generation rule can stay in `ci/prow`, so we avoid the generated configs being out of sync (unless someone intentionally does that for a legitimate reason).

Fredy-Z:
Hi Adriano, 
I was thinking about having the testgrid config generation code also in the original `make_config.go` file. But after some inspection, I decided to go for the current solution. 
There are two main reasons for this:
1. There is already lots of hard-code in the `make_config.go` file (the `switch...case...`s), and there are also lots of such configurations in the testgrid config file. Having them together will make the code harder to maintain.
2. The config generation logic is quite different from each other. For example, to generate the `test groups` and `dashboards` in the testgrid config file, we need to not only check the `periodics` job, but also the `presubmits` job to see if they have `test-coverage`. This makes it hard to leverage the original implementation for the new generator logic.

adrcunha:
This is a custom tool, so hardcoding stuff is fine and maintenance will be low (mostly small updates). Use multiple PRs, preprocess data and refactor code when necessary.

Fredy-Z:
Merged the two `make_config.go` file, please review again. Thanks.
/cat

ci/prow/Makefile
```diff
@@ -23,7 +23,7 @@ get-cluster-credentials:
 	gcloud container clusters get-credentials "$(CLUSTER)" --project="$(PROJECT)" --zone="$(ZONE)"
 
 config:
-	go run make_config.go config_knative.yaml > config.yaml
+	go run make_config.go --generate-testgrid-config=false --prow-config-output="config.yaml" config_knative.yaml
```
adrcunha: This should also generate the Testgrid config, so both configs are kept in sync.

ci/testgrid/Makefile
```diff
@@ -13,8 +13,17 @@
 # limitations under the License.
 
 TESTGRID_DIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
+GENERATOR_DIR := ../prow
+
+config:
+	go run ${GENERATOR_DIR}/make_config.go --generate-prow-config=false --testgrid-config-output="config.yaml" ${GENERATOR_DIR}/config_knative.yaml
```
adrcunha: Better call `make config` on ci/prow. It reduces duplicate code and ensure both configs are in sync.

ci/testgrid/config.yaml
```diff
@@ -50,27 +54,20 @@ default_dashboard_tab:
   code_search_url_template:      # The URL template to visit when searching for changelists
     url: https://github.com/knative/serving/compare/<start-custom-0>...<end-custom-0>
   alert_options:
-    alert_mail_to_addresses: 'knative-productivity-dev@googlegroups.com'
-
-#################################################################
-# Test groups
-#################################################################
-
+  alert_mail_to_addresses: 'knative-productivity-dev@googlegroups.com'
```
adrcunha: Bad indentation, see line 53.

ci/testgrid/config.yaml
```diff
@@ -50,27 +54,20 @@ default_dashboard_tab:
...
+######################################
```
adrcunha: I think we should either have the full comment, or no comment. A single separator won't help with readability in the generated file.

ci/testgrid/config.yaml
```diff
@@ -80,137 +77,104 @@ test_groups:
...
 - name: ci-knative-build-nightly-release
   gcs_prefix: knative-prow/logs/ci-knative-build-nightly-release
-- name: ci-knative-build-auto-release
-  gcs_prefix: knative-prow/logs/ci-knative-build-auto-release
 - name: ci-knative-build-dot-release
   gcs_prefix: knative-prow/logs/ci-knative-build-dot-release
-- name: pull-knative-build-test-coverage
-  gcs_prefix: knative-prow/logs/ci-knative-build-go-coverage
-  short_text_metric: coverage
+- name: ci-knative-build-auto-release
```
adrcunha: If the order of the test groups are different because of the go file, we should make them predictable. It will reduce the diff in future PRs and make reviews way easier.

ci/testgrid/config.yaml
```diff
@@ -80,137 +77,104 @@ test_groups:
...
-
-## Master branch (head)
-
-# knative/serving
 - name: knative-serving
   dashboard_tab:
   - name: continuous
     test_group_name: ci-knative-serving-continuous
     base_options: 'sort-by-name='
+  - name: conformance-tests
```
adrcunha: Can we keep the current order of tabs in Testgrid? If not 100%, at least most of it?

ci/testgrid/config.yaml
```diff
@@ -80,137 +77,104 @@ test_groups:
...
     test_group_name: ci-knative-serving-playground
-  - name: coverage
-    test_group_name: pull-knative-serving-test-coverage
-    base_options: 'exclude-filter-by-regex=Overall&group-by-directory=&expand-groups=&sort-by-name='
+    base_options: 'sort-by-name='
   - name: latency
     test_group_name: ci-knative-serving-latency
-    description: '95% latency in ms'
     base_options: 'exclude-filter-by-regex=Overall&group-by-directory=&expand-groups=&sort-by-name='
+    description: 95% latency in ms
```
adrcunha: Missing single quotes.

ci/testgrid/config.yaml
```diff
@@ -80,137 +77,104 @@ test_groups:
...
   - name: api-coverage
     test_group_name: ci-knative-serving-api-coverage
-    description: 'Conformance tests API coverage.'
-    base_options: 'exclude-filter-by-regex=Overall$&group-by-directory=&expand-groups=&sort-by-name='
-  - name: conformance-tests
-    test_group_name: ci-knative-serving-continuous
-    base_options: 'include-filter-by-regex=test/conformance\.&sort-by-name='
+    base_options: 'exclude-filter-by-regex=Overall&group-by-directory=&expand-groups=&sort-by-name='
```
adrcunha: Missing $ after Overall.

ci/testgrid/config.yaml
```diff
@@ -80,137 +77,104 @@ test_groups:
...
+    description: Conformance tests API coverage.
```
adrcunha: Missing single quotes.

ci/prow/Makefile
Fredy-Z: Add Testgrid config generation in prow's Makefile.

ci/testgrid/config.yaml
Fredy-Z: Removed the comment.

ci/testgrid/config.yaml
Fredy-Z: The current order is consistent with that in the template config file.

ci/testgrid/config.yaml
Fredy-Z: Same as above.

ci/testgrid/config.yaml
Fredy-Z: Added double quotes to keep the two config files consistent.

ci/prow/make_config.go
```diff
@@ -996,9 +1190,308 @@ func (a *stringArrayFlag) Set(value string) error {
...
+	data.Name = dashboardGroupName
+	data.RepoNames = dashboardRepoNames
+	executeTemplate("dashboard group", dashboardGroupTemplate, data)
+}
+
+// setOutput set the given file as the output target, then all the output will be written to this file
+func setOutput(fileName string) {
+	configFile, err := os.OpenFile(fileName, os.O_RDWR|os.O_CREATE, 0666)
+	if err != nil {
+		log.Fatal("Cannot open the configuration file")
```
adrcunha: Log the error.

ci/prow/make_config.go
```diff
@@ -996,9 +1190,308 @@ func (a *stringArrayFlag) Set(value string) error {
...
+	data.Extras = extras
+	executeTemplate("dashboard tab", dashboardTabTemplate, data)
+}
+
+// getTestGroupName get the testGroupName from the given repoName and jobName
+func getTestGroupName(repoName string, jobName string) string {
+	testGroupName := ""
+	switch jobName {
+	case "continuous", "dot-release", "auto-release", "performance", "latency", "api-coverage", "playground":
+		testGroupName = fmt.Sprintf("ci-%s-%s", repoName, jobName)
```
adrcunha: Just return the value, here and everywhere. Then the default case becomes the fallback.

ci/prow/make_config.go
```diff
@@ -996,9 +1190,308 @@ func (a *stringArrayFlag) Set(value string) error {
...
+			if repoName == "knative-serving" {
+				dashboardTabName := "conformance-tests"
+				baseOptions = "include-filter-by-regex=test/conformance\\\\.&sort-by-name="
+				executeDashboardTabTemplate(dashboardTabName, testGroupName, baseOptions, extras)
+			}
+		case "dot-release", "auto-release", "performance", "latency", "api-coverage", "playground":
+			dashboardTabName := jobName
+
+			if jobName == "latency" || jobName == "api-coverage" {
+				baseOptions = "exclude-filter-by-regex=Overall$&group-by-directory=&expand-groups=&sort-by-name="
```
adrcunha: This should be a const, as well in lines 1385 and 1400.

ci/prow/make_config.go
```diff
@@ -996,9 +1190,308 @@ func (a *stringArrayFlag) Set(value string) error {
...
+			executeDashboardTabTemplate(dashboardTabName, testGroupName, baseOptions, extras)
+		case "nightly":
+			dashboardTabName := "release"
+			executeDashboardTabTemplate(dashboardTabName, testGroupName, baseOptions, extras)
+		case "test-coverage":
+			dashboardTabName := "coverage"
+			baseOptions = "exclude-filter-by-regex=Overall$&group-by-directory=&expand-groups=&sort-by-name="
+			executeDashboardTabTemplate(dashboardTabName, testGroupName, baseOptions, extras)
+		default:
+			continue
```
adrcunha: Shouldn't you fail here due to an unknown job name?

ci/prow/make_config.go
```diff
@@ -996,9 +1190,308 @@ func (a *stringArrayFlag) Set(value string) error {
...
+	for _, jobName := range jobNames {
+		testGroupName := getTestGroupName(repoName, jobName)
+		baseOptions := "sort-by-name="
+		extras := make(map[string]string)
+		switch jobName {
+		case "continuous":
+			dashboardTabName := jobName
+			executeDashboardTabTemplate(dashboardTabName, testGroupName, baseOptions, extras)
+
+			if repoName == "knative-serving" {
```
adrcunha: Add a comment here saying this is a special case for knative/serving, as the conformance-tests tab is just a filtered view of the continuous tab.

ci/prow/make_config.go
```diff
@@ -996,9 +1190,308 @@ func (a *stringArrayFlag) Set(value string) error {
...
+func generateTestGroup(repoName string, jobNames []string) {
+	for _, jobName := range jobNames {
+		testGroupName := getTestGroupName(repoName, jobName)
+		gcsLogDir := ""
+		extras := make(map[string]string)
+		switch jobName {
+		case "continuous", "dot-release", "auto-release", "performance", "latency", "api-coverage", "playground":
+			gcsLogDir = fmt.Sprintf("%s/%s/ci-%s-%s", gcsBucket, logsDir, repoName, jobName)
+			if jobName == "playground" {
+				extras["alert_stale_results_hours"] = "168"
```
adrcunha: Please add a TODO here saying that this value should be derived from the cron string.

ci/prow/make_config.go
```diff
@@ -996,9 +1190,308 @@ func (a *stringArrayFlag) Set(value string) error {
...
+			}
+
+			// TODO: confirm if they are needed or not
```
```suggestion
			// TODO(Freddy-Z): confirm if they are needed or not
```

ci/prow/make_config.go
```diff
@@ -996,9 +1190,308 @@ func (a *stringArrayFlag) Set(value string) error {
...
+
+// generateTestGroup generates the test group configuration
+func generateTestGroup(repoName string, jobNames []string) {
+	for _, jobName := range jobNames {
+		testGroupName := getTestGroupName(repoName, jobName)
+		gcsLogDir := ""
+		extras := make(map[string]string)
+		switch jobName {
+		case "continuous", "dot-release", "auto-release", "performance", "latency", "api-coverage", "playground":
+			gcsLogDir = fmt.Sprintf("%s/%s/ci-%s-%s", gcsBucket, logsDir, repoName, jobName)
```
adrcunha: Why not use getTestGroupName, here and everywhere?

ci/prow/make_config.go
```diff
@@ -940,14 +1106,40 @@ func indentSection(indentation int, title string, array []string) string {
 	return indentSectionBase(indentation, title, "", array)
 }
 
+// indentMap returns the given map indented, with each key/value separated by ": "
+func indentMap(indentation int, mp map[string]string) string {
+	arr := make([]string, len(mp))
+	i := 0
+	for k, v := range mp {
+		_, err := strconv.ParseFloat(v, 64)
+		if err == nil {
```
adrcunha: Factor out the common part instead of duplicating the assignment.

ci/prow/make_config.go
```diff
@@ -906,24 +1063,33 @@ func indentBase(indentation int, prefix string, indentFirstLine bool, array []st
 		if i > 0 || indentFirstLine {
 			s += indent
 		}
-		s += prefix + quote(array[i]) + "\n"
+		if addQuote {
```
adrcunha: Factor out the common part instead of duplicating the assignment.

ci/prow/make_config.go
```diff
@@ -109,13 +114,42 @@ type sectionGenerator func(string, string, yaml.MapSlice)
...
-	// Array constants used throughout the jobs.
-	allPresubmitTests = []string{"--all-tests", "--emit-metrics"}
-	releaseNightly    = []string{"--publish", "--tag-release"}
-	releaseLocal      = []string{"--nopublish", "--notag-release"}
+// ####################################################################################################
+// ############## data definitions that are used for the testgrid config file generation ##############
+// ####################################################################################################
+// baseTestgridTemplateData contains basic data about the testgrid config file.
+type baseTestgridTemplateData struct {
+	TestGroupName string
```
adrcunha: You could use baseProwJobTemplateData instead of a separate structure, avoiding duplicate data and getting more info for generating the config. But it's not worth changing it now. Please add a TODO here about that for the future.

ci/prow/make_config.go
```diff
@@ -109,13 +114,42 @@ type sectionGenerator func(string, string, yaml.MapSlice)
...
+	Year          int
+}
+
+// testGroupTemplateData contains data about a test group
+type testGroupTemplateData struct {
+	Base      baseTestgridTemplateData
+	GcsLogDir string
```
adrcunha: This should be a base attribute. With Base being a baseProwJobTemplateData structure (and then renaming it to a more generic name would make more sense), it won' be necessary anymore. Please add a TODO.

ci/prow/make_config.go
```diff
@@ -142,15 +184,32 @@ var (
...
+	repoNames []string
+
+	// metaData saves the meta data needed to generate the final config file
+	// key is the main project version, value is another map containing job details
+	//     for the job detail map, key is the repo name, value is the list of job types, like continuous, latency, nightly, and etc.
+	metaData = make(map[string]map[string][]string)
 )
 
 // Templates for config generation.
 // TODO(adrcunha): eliminate redundant templates (e.g., latency job) by factoring them as standard jobs (e.g., periodic job).
```
adrcunha: This TODO is oudated, I forgot to remove it. Can you please do it?

ci/prow/make_config.go
```diff
@@ -996,9 +1190,308 @@ func (a *stringArrayFlag) Set(value string) error {
...
+				case "branch-ci":
+					enabled = getBool(item.Value)
+					jobName = "continuous"
+				case "release":
+					releaseVersion = getString(item.Value)
+				case "custom-job":
+					enabled = true
+					jobName = getString(item.Value)
+				default:
+					continue
```
adrcunha: Shouldn't you fail here, for safety?

ci/prow/make_config.go
```diff
@@ -996,9 +1190,308 @@ func (a *stringArrayFlag) Set(value string) error {
...
+				}
+			}
+			// add job types for the corresponding repos, if needed
+			if enabled {
+				// if it's an already released project
```
```suggestion
				// if it's a job for a release branch
```

ci/prow/Makefile
```diff
@@ -18,12 +18,14 @@ ZONE          ?= us-central1-f
...
 
 PROW_DIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
+TESTGRID_DIR := $(PROW_DIR)/../testgrid
 
 get-cluster-credentials:
 	gcloud container clusters get-credentials "$(CLUSTER)" --project="$(PROJECT)" --zone="$(ZONE)"
 
 config:
 	go run make_config.go --generate-testgrid-config=false --prow-config-output="config.yaml" config_knative.yaml
+	go run make_config.go --generate-prow-config=false --testgrid-config-output="$(TESTGRID_DIR)/config.yaml" config_knative.yaml
```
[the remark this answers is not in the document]
Fredy-Z: Joined the command. Also added code to reload the `config` object since it's modified when we generate the prow config file.

ci/testgrid/config.yaml
```diff
@@ -50,301 +54,233 @@ default_dashboard_tab:
...
-# knative/serving
 - name: knative-serving
   dashboard_tab:
   - name: continuous
     test_group_name: ci-knative-serving-continuous
-    base_options: 'sort-by-name='
+    base_options: "sort-by-name="
+  - name: conformance-tests
+    test_group_name: ci-knative-serving-continuous
+    base_options: "include-filter-by-regex=test/conformance\\.&sort-by-name="
```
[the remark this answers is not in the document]
Fredy-Z: Yes. In single quotes, `"` and `\` are freely used, but in double quotes they have to be escaped. More details can be found in the following links.
https://yaml.org/spec/current.html#id2534365
https://yaml.org/spec/current.html#id2532720

ci/prow/make_config.go
Fredy-Z: Done

ci/prow/make_config.go
Fredy-Z: Added fatal log.

ci/prow/make_config.go
Fredy-Z: Added fatal log.

ci/prow/make_config.go
Fredy-Z: Changed to testGroupName.
But for `test-coverage`, the two parts are not the same, so I left it as a special case.

ci/prow/make_config.go
Fredy-Z: I have removed the flag. 
Now all the arrays in the generated yaml file will be quoted.

ci/prow/make_config.go
Fredy-Z: Moved it to the `quote` function.

adrcunha:
@Fredy-Z please coordinate with the current oncall to update Testgrid and ensure it's working fine.

PR #599, opened 2019-03-20
Move tests to west2 region due to stockouts

ericKlawitter:
We are experiencing some stockout issues in west1, the current region. This changes to use us-west2, which seems like a healthier option.


ericKlawitter:
Ran a test job successfully in west2, gonna merge.

/hold cancel

PR #600, opened 2019-03-21
change prow config to us-central1 region

ericKlawitter:
After changing tests to run in us-west2, we hit a quota issue with the serving integration tests.  Central1 has sufficient quota allocated for the job, so this changes the prow config to use that region.


srinivashegde86:
I am adding a lgtm label as this seems ok. Cancel the hold once you test this.
/lgtm

ericKlawitter:
Successfully tested, gonna merge. 

/hold cancel

PR #607, opened 2019-03-21
Flaky test reporter: don't list individual test in Slack if flaky rate too high 

chaodaiG:
When the ratio of flaky tests is above certain threshold, there is only 1 issue created to avoid flooding github. Apply the same logic to Slack notification, so that we will not get a message with thousand lines

### DIFF
--- a/tools/flaky-test-reporter/github_issue.go
+++ b/tools/flaky-test-reporter/github_issue.go
@@ -102,6 +102,12 @@ func getIdentityForTest(testFullName, repoName string) string {
 	return fmt.Sprintf("'%s' in repo '%s'", testFullName, repoName)
 }
 
+// getBulkIssueIdentity creates a unique identify for bulk issue when flaky rate above threshold
+func getBulkIssueIdentity(rd *RepoData, flakyRate float32) string {
+	return fmt.Sprintf("%.2f%% tests failed in repo %s on %s",
+		flakyRate*100, rd.Config.Repo, time.Unix(*rd.LastBuildStartTime, 0).String())
+}
+
 // GithubIssue handles methods for github issues
 type GithubIssue struct {
 	user   *github.User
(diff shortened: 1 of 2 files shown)


tools/flaky-test-reporter/github_issue.go
```diff
@@ -102,6 +102,12 @@ func getIdentityForTest(testFullName, repoName string) string {
 	return fmt.Sprintf("'%s' in repo '%s'", testFullName, repoName)
 }
 
+// getBulkIssueIdentity creates a unique identify for bulk issue when flaky rate above threshold
```
adrcunha:
```suggestion
// getBulkIssueIdentity creates a unique identity for a bulk issue when the flaky rate is above a threshold
```

tools/flaky-test-reporter/slack_notification.go
```diff
@@ -109,13 +109,23 @@ func createSlackMessageForRepo(rd *RepoData, flakyIssuesMap map[string][]*flakyI
...
 	message := fmt.Sprintf("As of %s, there are %d flaky tests in '%s'",
 		time.Unix(*rd.LastBuildStartTime, 0).String(), len(flakyTests), rd.Config.Repo)
-	for _, testFullName := range flakyTests {
-		message += fmt.Sprintf("\n>- %s", testFullName)
-		if flakyIssues, ok := flakyIssuesMap[getIdentityForTest(testFullName, rd.Config.Repo)]; ok {
+	flakyRate := getFlakyRate(rd)
+	if flakyRate > threshold { // Don't list each test as this can be huge
+		message += fmt.Sprintf("\n>- skip displaying all tests as flaky rate above '%0.2f%%'", threshold)
+		if flakyIssues, ok := flakyIssuesMap[getBulkIssueIdentity(rd, flakyRate)]; ok {
 			for _, fi := range flakyIssues {
```
adrcunha: Move the loop to outside the if, then remove the same loop in line 124. Add a comment here expliciting that only one issue is created.

chaodaiG: Comment added. Moving the loop out might not be as clean as what it currently is, due to another for loop in else statement at line 123: `for _, testFullName := range flakyTests {`

adrcunha: Ah, yes.

PR #608, opened 2019-03-21
Update ko repo

jonjohnsonjr:
https://github.com/google/go-containerregistry/pull/415


adrcunha:
/approve
/lgtm

@jonjohnsonjr Do we need to generate a new prow-tests image now? If so, @ericKlawitter please do it.

jonjohnsonjr:
@adrcunha @ericKlawitter Yes, I believe that `go get` will fail until we regenerate the images.

ericKlawitter:
Pushed the updated image to GCR.

PR #613, opened 2019-03-22
Cleanup Testgrid a bit

adrcunha:
* rename "release" tab to "nightly" (this is a more appropriate and clear name)
* add back Testgrid config test that was dropped by mistake in #583
* remove the now unused api-coverage job (and tab)


ci/prow/make_config.go
```diff
@@ -1349,48 +1350,37 @@ func executeTestGroupTemplate(testGroupName string, gcsLogDir string, extras map
...
+			case "performance":
 				baseOptions = testgridTabGroupByTarget
-			}
-			if jobName == "latency" {
+			case "latency":
+				baseOptions = testgridTabGroupByDir
 				extras["description"] = "95% latency in ms"
-			}
-			if jobName == "api-coverage" {
+			case "api-coverage":
```
srinivashegde86: we recently removed api-coverage as a tab as it was fialing and we did not want to invest looking into it

adrcunha: Unrelated, but done. :P

adrcunha:
/hold cancel

We'll need another LGTM due to the changes.

PR #617, opened 2019-03-22
Enable dind for integration test in docs repo

chaodaiG:
Enable docker in docker for integration test in docs repo, as docker is needed for running e2e tests while testing sample apps deployment. For example: https://github.com/knative/docs/tree/master/docs/serving/samples/hello-world/helloworld-java


srinivashegde86:
should we be updating make_config here instead of directly updating the yaml?

srinivashegde86:
nm. looks like we are generating this from config_knative.yaml

PR #623, opened 2019-03-25
Use atleast 16 nodes for perf test cluster

srinivashegde86:
Fixes https://github.com/knative/serving/issues/3079

### DIFF
--- a/ci/prow/make_config.go
+++ b/ci/prow/make_config.go
@@ -634,6 +634,11 @@ func createCommand(data baseProwJobTemplateData) []string {
 
 // addEnvToJob adds the given key/pair environment variable to the job.
 func addEnvToJob(data *baseProwJobTemplateData, key, value string) {
+	// Value should always be string. Add quotes if we get a number
+	if isNum(value) {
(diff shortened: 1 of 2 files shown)


ci/prow/make_config.go
```diff
@@ -1039,9 +1047,17 @@ func gitHubRepo(data baseProwJobTemplateData) string {
 	return s
 }
 
+// isNum checks if the given string is a valid number
+func isNum(s string) bool {
+	if _, err := strconv.ParseFloat(s, 64); err == nil {
```
```
_, err := strconv.ParseFloat(s, 64)
return err == nil
```

ci/prow/make_config.go
```diff
@@ -842,6 +847,9 @@ func generatePeriodic(title string, repoName string, periodicConfig yaml.MapSlic
 			jobType = getString(item.Key)
 			jobNameSuffix = "performance"
 			data.Base.Command = performanceScript
+			// We need a larger cluster of atleast 16 nodes for perf tests
```
```suggestion
			// We need a larger cluster of at least 16 nodes for perf tests
```

PR #626, opened 2019-03-26
Also add coverage job/tab if no periodic job defined, but has go coverage in presubmit jobs

### LINKED ISSUE #615: make_config.go misses coverage job/tab if no periodic job defined
In https://github.com/knative/test-infra/pull/614 the knative/client repo has presubmit tests, but no periodic jobs defined yet (e.g., CI flow). The generator misses adding the coverage tab and the daily coverage job, since the presubmit section is empty.

/assign @Fredy-Z

Fredy-Z:
Fixes #615

### DIFF
--- a/ci/prow/make_config.go
+++ b/ci/prow/make_config.go
@@ -56,6 +56,7 @@ type repositoryData struct {
 	Name                string
 	EnableGoCoverage    bool
 	GoCoverageThreshold int
+	Processed           bool
 }
 
 // baseProwJobTemplateData contains basic data about a Prow job.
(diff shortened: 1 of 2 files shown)


ci/prow/make_config.go
```diff
@@ -916,6 +917,15 @@ func generatePeriodic(title string, repoName string, periodicConfig yaml.MapSlic
 	executeJobTemplate("periodic", jobTemplate, title, repoName, data.PeriodicJobName, false, data)
 }
 
+// generateRemainingGoCoveragePeriodicJobs generate the go coverage periodic jobs for repositories that are not processed in the periodic job configuration
+func generateRemainingGoCoveragePeriodicJobs() {
```
adrcunha: I think this approach is too specific. If we create a new kind of job that has implicit periodic jobs attached, this bug will manifest again. I would prefer a more generic approach that creates any required periodic jobs.

adrcunha: BTW, it would be nice to use that approach also for postsubmit jobs. Now that we're confident about the generated config, it doesn't need to strictly follow the order of the original config file.

Fredy-Z:
This PR involves further refactor combined with #615, close it for now.

PR #648, opened 2019-03-30
delete owner suggestion

steuhs:
GCP doesn't allow knative-productivity-admins@googlegroups.com to be added as owner because it is external

Fixes #


adrcunha:
/hold

The instructions says to add **one of the members of the group**, not the group.

PR #655, opened 2019-04-01
Let boskos handle test cluster teardown

adrcunha:
boskos-janitor automatically cleans up clusters and network resources left in the project, so we don't need to do it ourselves. Keep the cleanup for local runs though.

### DIFF
--- a/scripts/e2e-tests.sh
+++ b/scripts/e2e-tests.sh
@@ -120,6 +120,30 @@ function save_metadata() {
 EOF
 }
 
+# Delete target pools and health checks that might have leaked.
+# See https://github.com/knative/serving/issues/959 for details.
+# TODO(adrcunha): Remove once the leak issue is resolved.
+function delete_leaked_network_resources() {
+  (( IS_BOSKOS )) && return
+  # Ensure we're using the GCP project used by kubetest
+  local gcloud_project="$(gcloud config get-value project)"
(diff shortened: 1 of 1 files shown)


scripts/e2e-tests.sh
```diff
@@ -167,45 +191,28 @@ function create_test_cluster() {
   (( SKIP_KNATIVE_SETUP )) && test_cmd_args+=" --skip-knative-setup"
   [[ -n "${GCP_PROJECT}" ]] && test_cmd_args+=" --gcp-project ${GCP_PROJECT}"
   [[ -n "${E2E_SCRIPT_CUSTOM_FLAGS[@]}" ]] && test_cmd_args+=" ${E2E_SCRIPT_CUSTOM_FLAGS[@]}"
+  local extra_flags=()
+  # If using boskos, save time and let it tear down the cluster
+  (( ! IS_BOSKOS )) && extra_flags+=(--down)
```
chaodaiG: should this be `(( IS_BOSKOS ))` ?

adrcunha: No. We want to teardown the cluster only when running locally.

adrcunha:
/uncc @chaodaiG 

This is WIP.

adrcunha:
/assign @chaodaiG 

This is now ready for review. I confirmed the desired behavior in boskos-janitor.

PR #672, opened 2019-04-04
Update Istio installation

adrcunha:


tcnghia:
Thanks @adrcunha  for fixing this!!

Issue #656, opened 2019-04-01
Create an artifacts dir for local testing


srinivashegde86:
We use the artifacts dir to store artifacts when running tests. But, when I run tests locally(not on prow), we dont have the env var ARTIFACTS set to the artifacts dir handled by prow. We need a way to create a dir and use that to store the artifacts. These are useful for debugging and/or verifying the test results.

adrcunha:
I've seen the pattern "get artifacts dir; check if exists; create if not; fail if error" a few times in the codebase. This is a good candidate for refactoring/sharing, which will also simplify the resolution of this task.

jessiezcc:
not a priority, but good to have

Fredy-Z:
To add more background for this issue:
Currently when we run tests locally by using `go test` command, there is no standard and elegant way to export the test result as junit format `.xml` file like what `prow` does. 
Test cases like [this](https://github.com/knative/serving/blob/ce19a633478c7da01fec82826c11662364786be2/test/performance/observed_concurency_test.go#L209) explicitly create the file, but we want a more generic way to handle this.

adrcunha:
What you're describing is a different issue. This issue is about a simple refactoring.

PR #677, opened 2019-04-04
refactor mkdir into a common function

Fredy-Z:
#656 
Create `test-infra/shared/common/common.go` to hold the `CreateDir` function, and use it to replace the duplicate code.

We can continuously add common functions used by multiple components in this file.

### DIFF
--- a/shared/loadgenerator/loadgenerator.go
+++ b/shared/loadgenerator/loadgenerator.go
@@ -27,7 +27,7 @@ import (
 
 	"fortio.org/fortio/fhttp"
 	"fortio.org/fortio/periodic"
-	"github.com/knative/pkg/test/helpers"
+	"github.com/knative/test-infra/shared/common"
 	"github.com/knative/test-infra/shared/prow"
 )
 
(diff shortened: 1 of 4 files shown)


shared/loadgenerator/loadgenerator.go
```diff
@@ -109,10 +110,8 @@ func (g *GeneratorOptions) RunLoadTest(resolvableDomain bool) (*GeneratorResults
 // SaveJSON saves the results as Json in the artifacts directory
 func (gr *GeneratorResults) SaveJSON(testName string) error {
 	dir := prow.GetLocalArtifactsDir()
```
adrcunha: Is it worth making GetLocalArtifactsDir() creating the dir automatically, if it doesn't exist? Or maybe encapsulate this pattern in a "GetOrCreate" function?

Fredy-Z: Seems it's not worth changing this function, [here](https://github.com/knative/test-infra/blob/4ce16d390c55c71290f3b285bdfe37a1c5490304/tools/flaky-test-reporter/result.go#L118) it's only used for retrieving the path string. And several test cases are also using it for such purpose.

PR #685, opened 2019-04-11
Add a way to run multiple runs of load test

srinivashegde86:
Few more things in this PR

1. Create a new folder for e2e tests
1. Add an e2e test for testing localhost for fullLoad and step loads
1. Move the current smoke e2e to the new folder
1. Update `e2e-tests.sh` to run all the e2e tests


srinivashegde86:
/cc @hohaichi 
is adding numsteps enough to add the stepping load profile, or we need more knobs per step like QPS/duration we want to run for?

shared/loadgenerator/loadgenerator.go
```diff
@@ -37,12 +37,14 @@ const (
...
 	duration = 1 * time.Minute
 	qps      = 10
+	numSteps = 1
 	jsonExt  = ".json"
 )
 
 // GeneratorOptions provides knobs to run the perf test
 type GeneratorOptions struct {
 	// Duration is the total time to generate the load.
+	// If NumSteps is specified, this value will hold true for each step.
```
hohaichi: Do you mean if NumSteps is specified, this value is the duration of each step?  How about "Duration is the time to generate load in each step."?

shared/loadgenerator/loadgenerator.go
```diff
@@ -55,14 +57,18 @@ type GeneratorOptions struct {
 	// RequestTimeout is the maximum time waiting for a response.
 	RequestTimeout time.Duration
 	// QPS is the total QPS generated by all threads.
+	// If NumSteps is specified, for every step this value will be doubled.
```
hohaichi: "If NumSteps is specified" seems redundant.  How about "It will be doubled every step."?

shared/loadgenerator/loadgenerator.go
```diff
@@ -55,14 +57,18 @@ type GeneratorOptions struct {
...
 	QPS float64
 	// AllowInitialErrors specifies whether initial errors will abort the load.
 	AllowInitialErrors bool
+
+	// NumSteps controls the number of steps needed to reach full load.
+	NumSteps int
```
hohaichi: should we default NumSteps to 1 if not specified?  Right now I think a test will not run if its NumSteps is not specified.

shared/loadgenerator/loadgenerator.go
```diff
@@ -101,10 +111,32 @@ func (g *GeneratorOptions) CreateRunnerOptions(resolvableDomain bool) *fhttp.HTT
...
+By default, numsteps = 1 => test full load directly with no intermediate steps
+
+For numsteps=3, qps=q, duration=d
+	QPS
+	|		|---d---|
+	|		|   |	|
+	|       |---d---|   |4q	|
+	|---d---|   |2q	    |	|
+	|___|q______|_______|___|____duration(time)
+            1       2       3  <--- steps
```
hohaichi: nice diagram! :)

hohaichi:
> /cc @hohaichi
> is adding numsteps enough to add the stepping load profile, or we need more knobs per step like QPS/duration we want to run for?

Thank you for adding this capability.  It should be good for the ramp-up scenario.  Another scenario we are planning is bursting.  I think it could be a slight extension from your change:
- Instead of NumSteps, have an array of load factors, say [1, 2, 4] (for the sample you have in your diagram), or [1, 2, 1] for bursting.
- Add a BaseQPS to keep the base qps.  The user can set BaseQPS, but QPS is calculated each step based on the base and the factor.
- (optionally/later?) Change Duration to an array as well.  I'm thinking it may be useful, say, when we need more time for the system to stabilize in a larger increase (e.g., 2->4 may take more time than 1->2).  But this can be done by setting a common denominator value, so it may not be very valuable.

Thanks.

shared/loadgenerator/loadgenerator.go
srinivashegde86: it defaults to 1. This has been updated in addDefaults()

shared/loadgenerator/loadgenerator.go
srinivashegde86: added loadfactors. dont need this

shared/loadgenerator/loadgenerator.go
srinivashegde86: :D

srinivashegde86:
added loadfactors and baseqps.

>(optionally/later?) Change Duration to an array as well. I'm thinking it may be useful, say, when we need more time for the system to stabilize in a larger increase (e.g., 2->4 may take more time than 1->2). But this can be done by setting a common denominator value, so it may not be very valuable.

I was thinking about this as well. But, decided that this can be done later, when we want this variability.

PR #692, opened 2019-04-18
make E2E_CLUSTER_PROJECT globally available

Fredy-Z:
As an enhancement for the [PR in Eventing](https://github.com/knative/eventing/pull/1066#discussion_r276495895), add a flag `E2E_CLUSTER_PROJECT` that saves the actual project the `E2E_CLUSTER` resides in, and make it globally available.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -155,8 +155,9 @@ This is a helper script for Knative E2E test scripts. To use it:
 1. Write logic for the end-to-end tests. Run all go tests using `go_test_e2e()`
    (or `report_go_test()` if you need a more fine-grained control) and call
    `fail_test()` or `success()` if any of them failed. The environment variable
-   `KO_DOCKER_REPO` will be set according to the test cluster. You can also use
-   the following boolean (0 is false, 1 is true) environment variables for the logic:
+   `KO_DOCKER_REPO` and `E2E_PROJECT_ID` will be set according to the test cluster.
+   You can also use the following boolean (0 is false, 1 is true) environment
+   variables for the logic:
 
    - `EMIT_METRICS`: true if `--emit-metrics` was passed.
 
--- a/scripts/e2e-tests.sh
+++ b/scripts/e2e-tests.sh
@@ -273,6 +273,7 @@ function setup_test_cluster() {
 
   header "Setting up test cluster"
 
+  # Set the actual project the test cluster resides in
   E2E_CLUSTER_PROJECT="$(gcloud config get-value project)"
   readonly E2E_CLUSTER_PROJECT
 
--- a/scripts/e2e-tests.sh
+++ b/scripts/e2e-tests.sh
@@ -113,7 +113,7 @@ function save_metadata() {
     geo_key="Zone"
     geo_value="${E2E_CLUSTER_REGION}-${E2E_CLUSTER_ZONE}"
   fi
-  local cluster_version="$(gcloud container clusters list --project=${E2E_CLUSTER_PROJECT} --format='value(currentMasterVersion)')"
+  local cluster_version="$(gcloud container clusters list --project=${E2E_PROJECT_ID} --format='value(currentMasterVersion)')"
   cat << EOF > ${ARTIFACTS}/metadata.json
 {
   "E2E:${geo_key}": "${geo_value}",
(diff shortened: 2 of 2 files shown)


scripts/e2e-tests.sh
```diff
@@ -261,6 +260,11 @@ function create_test_cluster_with_retries() {
   done
 }
 
+# Flag which project holds the actual e2e test cluster.
+# It will be a project assigned by Boskos if test is running on Prow, 
+# otherwise will be ${GCP_PROJECT} set up by user.
+E2E_CLUSTER_PROJECT=""
```
adrcunha: Please put this declaration together with the other cars at the beginning of the file.

Fredy-Z: Sorry I'm not quite familiar with shell. But it seems we'll need to `export` the var for it to be used by other bash subprocesses? I've made some changes, could you check if it is the right way to go?

scripts/e2e-tests.sh
```diff
@@ -269,6 +268,11 @@ function setup_test_cluster() {
 
   header "Setting up test cluster"
 
+  # Set the actual project the test cluster resides in
+  # It will be a project assigned by Boskos if test is running on Prow, 
+  # otherwise will be ${GCP_PROJECT} set up by user.
+  export E2E_CLUSTER_PROJECT="$(gcloud config get-value project)"
```
srinivashegde86: make this readonly like the other vars

srinivashegde86: can we call this E2E_PROJECT_ID or something like that? cluster_project seems to inidcate it is a property of the cluster, which this is not as this can potentially be used for any non-cluster operation as well.

Fredy-Z: There are other vars like `E2E_CLUSTER_ZONE`, `E2E_CLUSTER_REGION`, `E2E_CLUSTER_NAME`, etc, that's why I pick up this name to follow the convention.
@adrcunha Do you have any suggestions for it?

adrcunha: I agree with E2E_PROJECT_ID, as it's in sync with PROJECT_ID.

scripts/e2e-tests.sh
adrcunha: Yes, `export` would be good.

scripts/e2e-tests.sh
Fredy-Z: Fixed.

Fredy-Z: Added `readonly`, and for the other two `export` vars as well.

scripts/README.md
```diff
@@ -155,8 +155,9 @@ This is a helper script for Knative E2E test scripts. To use it:
 1. Write logic for the end-to-end tests. Run all go tests using `go_test_e2e()`
    (or `report_go_test()` if you need a more fine-grained control) and call
    `fail_test()` or `success()` if any of them failed. The environment variable
-   `KO_DOCKER_REPO` will be set according to the test cluster. You can also use
-   the following boolean (0 is false, 1 is true) environment variables for the logic:
+   `KO_DOCKER_REPO` and `E2E_PROJECT_ID` will be set according to the test cluster.
```
adrcunha: Thanks for keeping the documentation up-to-date. 🎉

scripts/e2e-tests.sh
```diff
@@ -280,14 +284,15 @@ function setup_test_cluster() {
   if [[ -z "$(kubectl get clusterrolebinding cluster-admin-binding 2> /dev/null)" ]]; then
     acquire_cluster_admin_role ${k8s_user} ${E2E_CLUSTER_NAME} ${E2E_CLUSTER_REGION} ${E2E_CLUSTER_ZONE}
     kubectl config set-context ${k8s_cluster} --namespace=default
-    export KO_DOCKER_REPO=gcr.io/$(gcloud config get-value project)/${E2E_BASE_NAME}-e2e-img
+    readonly export KO_DOCKER_REPO=gcr.io/${E2E_PROJECT_ID}/${E2E_BASE_NAME}-e2e-img
```
adrcunha: Please don't readonly `KO_DOCKER_REPO`, as it might have undesired side-effects (although currently we don't change it later, it's not expected to be readonly).

scripts/e2e-tests.sh
```diff
@@ -280,14 +284,15 @@ function setup_test_cluster() {
...
   fi
 
+  echo "- Project is ${E2E_PROJECT_ID}"
   echo "- Cluster is ${k8s_cluster}"
   echo "- User is ${k8s_user}"
   echo "- Docker is ${KO_DOCKER_REPO}"
 
-  export KO_DATA_PATH="${REPO_ROOT_DIR}/.git"
+  readonly export KO_DATA_PATH="${REPO_ROOT_DIR}/.git"
```
adrcunha: Same here.

scripts/e2e-tests.sh
Fredy-Z: Removed.

scripts/e2e-tests.sh
Fredy-Z: Removed.

PR #695, opened 2019-04-19
E2E test allow arbitrary cluster name

chaodaiG:
In `setup_cluster` we dynamically set GCP cloud project, user, and cluster context, but not cluster name, as `E2E_CLUSTER_NAME` is a hard coded value. This PR enables arbitrary cluster name based on context.


adrcunha:
I don't get it. Please clarify and contextualize this PR, as the title alone isn't clear enough.

chaodaiG:
> I don't get it. Please clarify and contextualize this PR, as the title alone isn't clear enough.

Updated the context, but it was a wrong fix. I will try to make it work

Issue #520, opened 2019-02-25
Expose CPU amd memory metrics long term from perf tests


srinivashegde86:
Currently, for perf tests we bring up a prometheus instance when we run the tests. But, this instance dies when the cluster is deleted. We need a way to persist the **cpu and memory** metrics for the tests longer term so that we can evaluate it over time.

PR #701, opened 2019-04-22
Add ability to add cpu and memory profile with loadgenerator

srinivashegde86:
Fixes https://github.com/knative/test-infra/issues/520

the load generator library provides ability to create cpu and memory profiles. Add that capability to the load generator as well. Performance tests should be able to use them if possible.

Also, add a new var to generator options `FileNamePrefix` that will be used to store the profiles in artifacts

The profiles are generated using [pprof](https://github.com/google/pprof/blob/master/README.md) and can be read using the tool like `pprof -web TestProfiler.mem`

### DIFF
--- a/shared/loadgenerator/loadgenerator.go
+++ b/shared/loadgenerator/loadgenerator.go
@@ -118,7 +118,12 @@ func (g *GeneratorOptions) CreateRunnerOptions(resolvableDomain bool) *fhttp.HTT
 	}
 
 	if len(g.FileNamePrefix) != 0 {
-		ro.Profiler = g.FileNamePrefix
+		dir := prow.GetLocalArtifactsDir()
+		if err := common.CreateDir(dir); err != nil {
+			return nil
+		}
+
+		ro.Profiler = path.Join(dir, g.FileNamePrefix)
 	}
 
 	return &ro
--- a/shared/loadgenerator/loadgenerator.go
+++ b/shared/loadgenerator/loadgenerator.go
@@ -96,7 +96,7 @@ func (g *GeneratorOptions) addDefaults() {
 }
 
 // CreateRunnerOptions sets up the fortio client with the knobs needed to run the load test
-func (g *GeneratorOptions) CreateRunnerOptions(resolvableDomain bool) *fhttp.HTTPRunnerOptions {
+func (g *GeneratorOptions) CreateRunnerOptions(resolvableDomain bool) (*fhttp.HTTPRunnerOptions, error) {
 	o := fhttp.NewHTTPOptions(g.URL)
 	o.NumConnections = g.NumConnections
 	o.HTTPReqTimeOut = g.RequestTimeout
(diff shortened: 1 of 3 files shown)


shared/loadgenerator/loadgenerator.go
```diff
@@ -107,6 +116,17 @@ func (g *GeneratorOptions) CreateRunnerOptions(resolvableDomain bool) *fhttp.HTT
 		HTTPOptions:        *o,
 		AllowInitialErrors: g.AllowInitialErrors,
 	}
+
+	if len(g.FileNamePrefix) != 0 {
+		dir := prow.GetLocalArtifactsDir()
+		if err := common.CreateDir(dir); err != nil {
+			return nil
```
chaodaiG: shouldn't let the function return an error as well in this case?

test/e2e/load_generator_test.go
```diff
@@ -25,30 +25,41 @@ import (
...
 
 	res, err := opts.RunLoadTest(true)
 	if err != nil {
 		t.Fatalf("Error performing load test: %v", err)
 	}
 
-	if len(res.Result) != len(factors) {
-		t.Logf("got:%d, want: %d", len(res.Result), len(factors))
+	if factors && len(res.Result) != 3 {
+		t.Logf("got:%d, want: 3", len(res.Result))
```
chaodaiG: should this fail test? if not can you add a comment here explaining why this should not fail test?

shared/loadgenerator/loadgenerator.go
```diff
@@ -64,11 +65,19 @@ type GeneratorOptions struct {
 	// LoadFactors defines the multiplier for baseQPS.
 	// Len(loadfactors) defines the number of QPS changes.
 	LoadFactors []float64
+	// FileNamePrefix is the prefix used to identify the stored files in artifacts.
+	// If not empty, this can be used to store cpu/mem profile from loadgenerator.
+	// Typically, we can use t.Name() to differentiate between the tests.
+	FileNamePrefix string
```
chaodaiG: nit, how about renaming this with something including `cpu/mem/profiler`, so that it's easier to understand

shared/loadgenerator/loadgenerator.go
srinivashegde86: sg. Updated

test/e2e/load_generator_test.go
srinivashegde86: yup it should. Updated

shared/loadgenerator/loadgenerator.go
srinivashegde86: I would want to use this for various things. cpu/mem can be one of the few things. Plus, this is just a prefix. We will adding things like -Profile to indicate this is a profile and -Trace for trace. So, I Felt this cam eclose to explainig that. Added a comment to explain that as well.

PR #705, opened 2019-04-23
Add CI support for istio version and istio mesh

k4leung4:
This allows CI system to specify what version of istio to use and
whether to use istio mesh or not.

Fixes #

### DIFF
--- a/ci/prow/config_knative.yaml
+++ b/ci/prow/config_knative.yaml
@@ -108,6 +108,42 @@ periodics:
     - branch-ci: true
       release: "0.5"
       cron: "35 8 * * *" # Run at 01:35PST every day (08:35 UTC)
+    - custom-job: istio-1.0.7-mesh
+      cron: "7 */2 * * *" # Run every other hour and seven minute
+      command:
+      - "./test/e2e-tests.sh"
+      args:
+      - "--istio-version"
+      - "1.0.7"
+      - "--mesh"
+      timeout: 100
+    - custom-job: istio-1.0.7-no-mesh
+      cron: "22 */2 * * *" # Run every other hour and twenty-two minute
+      command:
+      - "./test/e2e-tests.sh"
+      args:
+      - "--istio-version"
+      - "1.0.7"
+      - "--no-mesh"
+      timeout: 100
+    - custom-job: istio-1.1.2-mesh
+      cron: "37 */2 * * *" # Run every other hour and thirty-seven minute
+      command:
+      - "./test/e2e-tests.sh"
+      args:
+      - "--istio-version"
+      - "1.1.2"
+      - "--mesh"
+      timeout: 100
+    - custom-job: istio-1.1.2-no-mesh
+      cron: "52 */2 * * *" # Run every other hour and fifty-two minute
+      command:
+      - "./test/e2e-tests.sh"
+      args:
+      - "--istio-version"
+      - "1.1.2"
+      - "--no-mesh"
+      timeout: 100
     - nightly: true
       cron: "1 8 * * *" # Run at 01:01PST every day (08:01 UTC)
     - dot-release: true
--- a/ci/prow/make_config.go
+++ b/ci/prow/make_config.go
@@ -1355,8 +1355,11 @@ func generateTestGroup(repoName string, jobNames []string) {
 			extras["short_text_metric"] = "coverage"
 			// Do not alert on coverage failures (i.e., coverage below threshold)
 			extras["num_failures_to_alert"] = "9999"
+		case "istio-1.0.7-mesh", "istio-1.0.7-no-mesh", "istio-1.1.2-mesh", "istio-1.1.2-no-mesh":
+			extras["alert_stale_results_hours"] = "3"
+			extras["num_failures_to_alert"] = "3"
 		default:
-			log.Fatalf("Unknown jobName: %s", jobName)
+			log.Fatalf("Unknown jobName for generateTestGroup: %s", jobName)
 		}
 		executeTestGroupTemplate(testGroupName, gcsLogDir, extras)
 	}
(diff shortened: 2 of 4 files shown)


scripts/presubmit-tests.sh
```diff
@@ -226,6 +226,14 @@ function default_integration_test_runner() {
   local options=""
   local failed=0
   (( EMIT_METRICS )) && options="--emit-metrics"
+  if [ $ISTIO_VERSION -ne 0 ]; then
```
adrcunha: Treat ISTIO_VERSION as a string, not as a value.

scripts/presubmit-tests.sh
```diff
@@ -240,6 +248,8 @@ RUN_BUILD_TESTS=0
 RUN_UNIT_TESTS=0
 RUN_INTEGRATION_TESTS=0
 EMIT_METRICS=0
+ISTIO_VERSION=0
```
adrcunha: ISTIO_VERSION=""

ci/prow/config_knative.yaml
```diff
@@ -100,6 +100,11 @@ presubmits:
 periodics:
   knative/serving:
     - continuous: true
+      istio: "1.0.7"
+      cron: "1 */2 * * *" # Run every other hour and one minute
```
adrcunha: Better distribute the jobs for better resource usage. Once day this should be automatic, but right now just use a different minute.

k4leung4:
What do you mean by a custom job?
Ideally build/unit tests would be separate setup.

Setups:
- build/unit tests
- master e2e test istio A mesh
- master e2e test istio A no-meah
- master e2e test istio B mesh
- master e2e test istio B no-mesh
- latest release branch e2e istio A mesh
- latest release branch e2e istio A no-mesh
- latest release branch e2e istio B mesh
- latest release branch e2e istio B no-mesh

adrcunha:
> What do you mean by a custom job?

https://github.com/knative/test-infra/blob/master/ci/prow/config_knative.yaml#L117

> Ideally build/unit tests would be separate setup.

Yes. See 459.

adrcunha:
> What do you mean by a custom job?

https://github.com/knative/test-infra/blob/master/ci/prow/config_knative.yaml#L117

> Ideally build/unit tests would be separate setup.

See 459 for the ideal setup. Meanwhile, we can get the setup above by just using different custom jobs.

k4leung4:
updated to use custom-job
ptal

adrcunha:
> updated to use custom-job
> ptal

LGTM, thanks. Please run `make config` to generate the Prow and TestGrid configs so I can review them.

k4leung4:
this seems to have failed on the bazel run step?
any idea how to debug the failure?

ci/testgrid/config.yaml
```diff
@@ -183,160 +191,3 @@ dashboards:
   - name: conformance-tests
     test_group_name: ci-knative-serving-continuous
     base_options: "include-filter-by-regex=test/conformance\\.&sort-by-name="
-  - name: nightly
```
adrcunha: This is wrong. These dashboards must not be affected.

ci/prow/make_config.go
```diff
@@ -1330,7 +1330,7 @@ func generateTestGroup(repoName string, jobNames []string) {
 		gcsLogDir := fmt.Sprintf("%s/%s/%s", gcsBucket, logsDir, testGroupName)
 		extras := make(map[string]string)
 		switch jobName {
-		case "continuous", "dot-release", "auto-release", "performance", "latency", "playground", "nightly":
+		case "continuous", "dot-release", "auto-release", "performance", "latency", "playground", "nightly", "istio-1.0.7-mesh", "istio-1.0.7-no-mesh", "istio-1.1.2-mesh", "istio-1.1.2-no-mesh":
```
adrcunha: Can we have a more generic approach here for custom jobs, instead of hardcoding too-specific job names?

adrcunha:
> this seems to have failed on the bazel run step?
> any idea how to debug the failure?

The bazel test should tell you why it failed. Logs, please.

k4leung4:
not sure if it is bazel, https://gubernator.knative.dev/build/knative-prow/pr-logs/pull/knative_test-infra/705/pull-knative-test-infra-build-tests/1121138581549617152/ doesnt include any thing useful about the make test target failed.

ci/prow/make_config.go
k4leung4: I doubt it. Custom jobs allows full control over what to run, we would effectively be defining what custom job is if we do that.

A custom type, say e2e_tests would be better suited as then we could guarantee that it is running the e2e_tests.sh script, which we know how to configure.

adrcunha:
> not sure if it is bazel, https://gubernator.knative.dev/build/knative-prow/pr-logs/pull/knative_test-infra/705/pull-knative-test-infra-build-tests/1121138581549617152/ doesnt include any thing useful about the make test target failed.

Oh, it does. It tells you that the generated config doesn't match the one in your PR. It's unrelated to bazel. Probably you forgot to commit the latest changes.

ci/prow/make_config.go
adrcunha: Generalize the config generation for custom jobs, don't mind the naming. Because that's effectively what you are doing here.

k4leung4: I'm confused now.

what is the change you want me to make?

adrcunha: Nevermind. The change needs to be deeper than that, as the generated tab names had bad readability. I'll file an issue for that, let's get this coverage in.

ci/prow/make_config.go
```diff
@@ -1355,8 +1355,10 @@ func generateTestGroup(repoName string, jobNames []string) {
 			extras["short_text_metric"] = "coverage"
 			// Do not alert on coverage failures (i.e., coverage below threshold)
 			extras["num_failures_to_alert"] = "9999"
+		case "istio-1.0.7-mesh", "istio-1.0.7-no-mesh", "istio-1.1.2-mesh", "istio-1.1.2-no-mesh":
+			extras["alert_stale_results_hours"] = "3"
```
adrcunha: Should we also alert after 3 failures, not on every failure, like the default continuous flow? If so, add `extras["num_failures_to_alert"] = "3"` here.

k4leung4: done.

adrcunha:
/lgtm
/approve

Please ask the [oncall](https://storage.googleapis.com/knative-infra-oncall/oncall.html) to update Prow and TestGrid.

PR #719, opened 2019-04-30
Also display username during e2e init

nbarthwal:
A Fix to display user name besides account in the e2e tests during cluster creation

Fixes #


scripts/e2e-tests.sh
```diff
@@ -187,8 +187,10 @@ function create_test_cluster() {
   # Assume test failed (see details in set_test_return_code()).
   set_test_return_code 1
   local gcloud_project="${GCP_PROJECT}"
+  local gcloud_user=$(gcloud config get-value core/account)
```
adrcunha: Inline this in the echo, you're not using this var anywhere.

PR #722, opened 2019-04-30
Add license header to all coverage files

srinivashegde86:
This PR does a few small things other than adding header:

1. Remove build.bazel files as we dont need them
1. Add a coverage-dev-image to makefile so that we can make prod image and a dev image. This can be simplified further as the basic instructions are the same, but can be worked on later.

This PR does not modify any existing functionality other than the doc updates. So, should not alter the behaviour in any way.


adrcunha:
Thanks for doing this, Srinivas.

/lgtm
/approve

Issue #679, opened 2019-04-07
Very stale coverage baselines


mattmoor:
In Serving a PR I just submitted got back:
![image](https://user-images.githubusercontent.com/2442466/55685124-83a85100-5907-11e9-95b8-992ceb79a411.png)

I'm wondering if whatever updates the coverage baselines is just broken because this file's existed for a while now...

/kind bug

jessiezcc:
@steuhs, can u pls look into this?

steuhs:
The baselines are stale because in recent runs, the unit test in github.com/knative/serving/pkg/activator/handler kept failing. We only use the most recent healthy run as baseline. I tried cloned a fresh copy of knative/serving and I see that test fail locally as well. I think the test need to be fixed @mattmoor

mattmoor:
We're talking like a month old, there's no way we've had a failing unit test for that long.

Here's a 100x run on master as of now:
```
$ go test -count=100 ./pkg/activator/handler/
ok      github.com/knative/serving/pkg/activator/handler        11.111s
```

With coverage (what I run within emacs):
```
go test -covermode=count -coverprofile=/tmp/count.out github.com/knative/serving/pkg/activator/handler/...
ok      github.com/knative/serving/pkg/activator/handler        0.341s  coverage: 97.1% of statements
```

The baselines are so old at this point, we may as well turn covbot off, so please get this fixed.

Also: why did it take me raising this as stale coverage for us to uncover this?  @jessiezcc can we add this to some sort of triage the oncall monitors and takes action on?

jessiezcc:
Matt is right. The coverage for this file shows between 96%-100% since April.8th on [testgrid](https://testgrid.knative.dev/knative-serving#coverage). 
This should be covered by monitoring spec @steuhs, pls make sure you add this scenario into functional spec. And before automation is in place, let's add a section to oncall playbook for manual triage. 
/cc @srinivashegde86 who is the project lead for monitoring/metrics.

jessiezcc:
@steuhs, can you pls update baseline asap for now?

steuhs:
added a temporary manual fix. the baseline should be up to date now.

This is a link showing the failure of github.com/knative/serving/pkg/activator/handler, mentioned before.
https://gubernator.knative.dev/build/knative-prow/logs/post-knative-serving-go-coverage/1123312530802872321/?log#log

steuhs:
fixed by #728

PR #728, opened 2019-05-01
add coverage completion marker regardless of unit test failure

steuhs:
Coverage completion marker is used in pre-submit workflow to find
most recent healthy post-submit coverage result as a basis for comparison.
We are adding the marker now regardless of unit test failure, to prevent
stale comparison baseline

#679

Fixes #


adrcunha:
If a unit test fails, would it corrupt the test coverage?

srinivashegde86:
similar question, is there any difference between coverage numbers for UT when it passes vs fails.

steuhs:
@adrcunha go test failure on a single file will not stop the process. Only data from failed test will be missing and the format of the coverage profile will not be corrupted.

@srinivashegde86 If a unit test fails, the data will not be there, as if the file does not exist. It will not be reported as 0 coverage. Considering we usually only see up to one failure and we only report on the files the PR contains, there shouldn't be surprising behavior

adrcunha:
Ok, so if a test fails we'll get a wrong base coverage for that file. That's better than having an outdated baseline. This should be addressed at some point though.

/lgtm

Issue #711, opened 2019-04-25
Integrate observability repo onto testgrid


jessiezcc:
The repo has been made public, it's necessary for this repo to be surfaced on knative testgrid.

adrcunha:
1. Update the vendored test-infra dir and the presubmit/e2e script on the repo.
2. Create the pre/post submit jobs, together with the Testgrid tabs.

adrcunha:
1 was fixed by https://github.com/knative/observability/pull/24

/close

PR #737, opened 2019-05-06
Update prow and testgrid for knative/observability

srinivashegde86:
Part of #711 

This updated the test-infra components.


adrcunha:
https://testgrid.knative.dev/knative-observability#Summary is stale, I believe TestGrid was updated, but not Prow?

srinivashegde86:
yup. just updated prow

PR #739, opened 2019-05-07
rename skip-istio flag to skip-istio-addon to make it more obvious

chaodaiG:


scripts/README.md
```diff
@@ -172,7 +172,7 @@ This is a helper script for Knative E2E test scripts. To use it:
    will immediately start the tests against the cluster currently configured for
    `kubectl`.
 
-1. By default Istio is installed on the cluster via Addon, using `--skip-istio` if
+1. By default Istio is installed on the cluster via Addon, using `--skip-istio-addon` if
```
srinivashegde86:
```suggestion
1. By default Istio is installed on the cluster via Addon, use `--skip-istio-addon` if
```

PR #743, opened 2019-05-07
Do not generate dashboard groups for versioned projects in testgrid

Fredy-Z:
Testgrid does not allow dashboard groups having the same name as other dashboards. So skip versioned projects when we create dashboard groups, since each of them has only one dashboard.


adrcunha:
This is just removing the offending groups, without actually fixing the problem. Please rollback the config and fix the dashboard grouping.

ci/prow/make_config.go
```diff
@@ -1485,15 +1485,16 @@ func generateDashboardsForReleases() {
...
-			repos := metaData[projName]
-			for _, repoName := range repoNames {
-				if _, exists := repos[repoName]; exists {
-					dashboardRepoNames = append(dashboardRepoNames, buildProjRepoStr(projName, repoName))
-				}
+			continue
+		}
+
+		dashboardRepoNames := make([]string, 0)
+		repos := metaData[projName]
```
srinivashegde86: additional var not needed.

Fredy-Z: Hmm, I think having this var can make the code a bit clearer.

PR #744, opened 2019-05-08
Wait until all containers in pods are running too.

markusthoemmes:
The two missing quotes cause the output from kubectl to being read in one go (rather than line by line). The container check logic therefore only checks the very first pod's containers.


adrcunha:
/approve
/lgtm

Please note that once this is merged you have to update the vendored test-infra in the affected repos.

PR #751, opened 2019-05-09
CI flows on branches use stable releases

### LINKED ISSUE #721: CI flows on branches should use stable releases
Currently CI flows on branches use nightly releases. For example,

https://gubernator.knative.dev/build/knative-prow/logs/ci-knative-eventing-0.5-continuous/1123145177083416576

fails when installing Istio from the latest Knative Serving nightly. The flow should be using artifacts from Knative Serving 0.5.

@mattmoor @tcnghia to validate the assumption.

Fredy-Z:
Fixes #721 

Notes:
1. This change can fix the CI flows for future branches(>=0.6). For fixing the tests failure of [Eventing-0.5](https://testgrid.knative.dev/knative-0.5#eventing), I'm not sure if it would be fine to use the Istio AddOn instead of Istio released by Serving-0.5.
2. We'll need to update the vendor for eventing-sources repo after this PR is merged.

### DIFF
--- a/scripts/library.sh
+++ b/scripts/library.sh
@@ -448,10 +448,9 @@ function get_knative_base_yaml_source() {
   local branch_name="$(git rev-parse --abbrev-ref HEAD)"
   # If it's a release branch, we should have a different knative_base_yaml_source.
   if [[ $branch_name =~ ^release-[0-9\.]+$ ]]; then
-    # Get the actual version number for the release branch.
-    # For example, branch_version will be "0.5" if the branch name is "release-0.5".
-    local branch_version=$(echo ${branch_name} | sed "s/release-//g")
-    knative_base_yaml_source=https://storage.googleapis.com/knative-releases/@/previous/v${branch_version}.0
+    # Get the latest tag name for the current branch, which is likely formatted as v0.5.0
+    local tag_name=$(git describe --tags)
+    knative_base_yaml_source=https://storage.googleapis.com/knative-releases/@/previous/${tag_name}
   fi
   echo ${knative_base_yaml_source}
 }
--- a/scripts/library.sh
+++ b/scripts/library.sh
@@ -442,17 +442,18 @@ function get_canonical_path() {
   echo "$(cd ${path%/*} && echo $PWD/${path##*/})"
 }
 
-# Return the version for the current branch.
-function get_branch_version() {
+# Return the base url we use to build the actual knative yaml sources.
+function get_knative_base_yaml_source() {
+  local knative_base_yaml_source=https://storage.googleapis.com/knative-nightly/@/latest
   local branch_name="$(git rev-parse --abbrev-ref HEAD)"
-  # Return the actual version number for the release branch.
-  # For example, we will return "0.5" if the branch name is "release-0.5".
-  if [[ $name =~ ^release-[0-9\.]+ ]]; then
-    echo ${branch_name} | sed 's/release-//g'
-  # Return empty string for any other branch.
-  else
-    echo ""
+  # If it's a release branch, we should have a different knative_base_yaml_source.
+  if [[ $branch_name =~ ^release-[0-9\.]+$ ]]; then
+    # Get the actual version number for the release branch.
+    # For example, branch_version will be "0.5" if the branch name is "release-0.5".
+    local branch_version=$(echo ${branch_name} | sed 's/release-//g')
+    knative_base_yaml_source=https://storage.googleapis.com/knative-releases/@/previous/v${branch_version}.0
   fi
+  echo ${knative_base_yaml_source}
 }
 
 # Initializations that depend on previous functions.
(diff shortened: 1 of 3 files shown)


adrcunha:
Adding an environment variable is unnecessary complexity. As I mentioned in the issue, just check the current git branch.

scripts/library.sh
```diff
@@ -23,7 +23,13 @@ readonly SERVING_GKE_VERSION=gke-latest
 readonly SERVING_GKE_IMAGE=cos
 
 # Public latest stable nightly images and yaml files.
-readonly KNATIVE_BASE_YAML_SOURCE=https://storage.googleapis.com/knative-nightly/@/latest
+KNATIVE_BASE_YAML_SOURCE=""
```
adrcunha: Move any code/condition dependent initialization to the end of the file, together with the other vars.

Fredy-Z: Done.

scripts/library.sh
```diff
@@ -448,8 +442,34 @@ function get_canonical_path() {
...
+}
+
 # Initializations that depend on previous functions.
 # These MUST come last.
 
 readonly _TEST_INFRA_SCRIPTS_DIR="$(dirname $(get_canonical_path ${BASH_SOURCE[0]}))"
 readonly REPO_NAME_FORMATTED="Knative $(capitalize ${REPO_NAME//-/})"
+
+# Public latest stable nightly images and yaml files.
+readonly BRANCH_VERSION="$(get_branch_version)"
```
adrcunha: Thanks, this is way nicer. Once last simplification: you don't need `BRANCH_VERSION` and top-level `if`. `get_branch_version()` can already return the right string to use as `KNATIVE_BASE_YAML_SOURCE`; it even makes sense to rename it to something like `get_knative_base_yaml_source()`.

scripts/library.sh
```diff
@@ -448,8 +442,34 @@ function get_canonical_path() {
   echo "$(cd ${path%/*} && echo $PWD/${path##*/})"
 }
 
+# Return the version for the current branch.
+function get_branch_version() {
+  local branch_name="$(git rev-parse --abbrev-ref HEAD)"
+  # Return the actual version number for the release branch.
+  # For example, we will return "0.5" if the branch name is "release-0.5".
+  if [[ $name =~ ^release-[0-9\.]+ ]]; then
```
adrcunha: nit: end the regex with $

scripts/library.sh
```diff
@@ -448,8 +442,34 @@ function get_canonical_path() {
...
+    echo ${branch_name} | sed 's/release-//g'
```
adrcunha: I think it's safer if, once in a branch, you find the latest tag and use it; e.g. in branch release-0.5 you would use the tag v0.5.2.

scripts/library.sh
Fredy-Z: Done.

scripts/library.sh
Fredy-Z: Sorry I didn't quite get your point. Shouldn't  the yaml files we use here correspond to the code we use to run test? Or do you mean when `Prow` runs the CI jobs for branch release-0.5, it actually uses the code tagged v0.5.2?

scripts/library.sh
Fredy-Z: Done.

scripts/library.sh
adrcunha: Yes, they should. Prow runs on the latest commit of a branch, but releases are tagged, so the best approximation is using the latest tag.

Fredy-Z: Thanks for the explanation. Done.

scripts/library.sh
```diff
@@ -448,8 +442,27 @@ function get_canonical_path() {
...
 }
 
+# Return the base url we use to build the actual knative yaml sources.
+function get_knative_base_yaml_source() {
+  local knative_base_yaml_source=https://storage.googleapis.com/knative-nightly/@/latest
+  local branch_name="$(git rev-parse --abbrev-ref HEAD)"
+  # If it's a release branch, we should have a different knative_base_yaml_source.
+  if [[ $branch_name =~ ^release-[0-9\.]+$ ]]; then
+    # Get the latest tag name for the current branch, which is likely formatted as v0.5.0
+    local tag_name=$(git describe --tags)
```
adrcunha: Better being safe here: sort tags and filter by the release number.

scripts/library.sh
```diff
@@ -448,8 +442,27 @@ function get_canonical_path() {
   echo "$(cd ${path%/*} && echo $PWD/${path##*/})"
 }
 
+# Return the base url we use to build the actual knative yaml sources.
+function get_knative_base_yaml_source() {
+  local knative_base_yaml_source=https://storage.googleapis.com/knative-nightly/@/latest
```
adrcunha: nit: quote the strings, please; here and everywhere.

PR #776, opened 2019-05-14
Add a config parser for knative monitoring

### LINKED ISSUE #761: monitoring config parser
/assign @steuhs

steuhs:
This commit complete the following steps in design doc:
- step 2.2 Parse monitoring config and retrieve all error patterns
- step 4
  - Parse monitoring config (we already parsed it in step 2, so we can simply use the result from there)
  - Retrieve the config for the error patterns we find in step 2
  - Retrieve the alert condition for the job name pattern that matches the current job

In addition,
- adds entry point to the main function;
- print the pattern collection on the served port

Fixes #761


tools/monitoring/main.go
```diff
@@ -18,11 +18,15 @@ package main
...
 import (
 	"fmt"
+	"github.com/knative/test-infra/tools/monitoring/yaml-handler"
 	"log"
 	"net/http"
 	"os"
 )
 
+// TODO: replace this with permanent url (or a function) that returns the yaml from test-infra master branch
+const yamlURL = "https://raw.githubusercontent.com/steuhs/test-infra-1/7282b8924d8e2f6f68fe3d34b33ec52a804790fb/tools/monitoring/config-getter/sample.yaml"
```
srinivashegde86: can we point to the sample yaml above instead of the one from your branch?

tools/monitoring/main.go
```diff
@@ -18,11 +18,15 @@ package main
 
 import (
 	"fmt"
+	"github.com/knative/test-infra/tools/monitoring/yaml-handler"
```
srinivashegde86: this should be after std imports

tools/monitoring/yaml-handler/parser.go
```diff
@@ -0,0 +1,116 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// The yamlHandler is responsible for fetching, parsing config yaml file. It also allows user to
+// retrieve a particular record from the yaml.
+
+package yamlHandler
```
srinivashegde86: can we call this config as this looks at the user defined config? We are defining it in yaml and using yaml parser, but that seems to be an implementation detail

tools/monitoring/yaml-handler/parser.go
```diff
@@ -0,0 +1,116 @@
...
+
+import (
+	"fmt"
+	"github.com/pkg/errors"
```
srinivashegde86: move to the end

tools/monitoring/yaml-handler/parser.go
```diff
@@ -0,0 +1,116 @@
...
+	"net/http"
+	"regexp"
+)
+
+type AlertConditions struct {
+	JobNameRegex string `yaml:"job-name-regex"`
+	Occurrences  int
+	JobsAffected string `yaml:"jobs-affected"`
+	PrsAffected  string `yaml:"prs-affected"`
+	Period       int
```
srinivashegde86: this should be time.Minutes

tools/monitoring/yaml-handler/parser.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+type PatternSpec struct {
+	ErrorPattern string `yaml:"error-pattern"`
+	Hint         string
+	Alerts       []AlertConditions
+}
+
+type YamlFile struct {
```
srinivashegde86: rename to ErrorPatternConfig?

tools/monitoring/yaml-handler/parser.go
```diff
@@ -0,0 +1,116 @@
...
+	"fmt"
+	"github.com/pkg/errors"
+	"gopkg.in/yaml.v2"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"regexp"
+)
+
+type AlertConditions struct {
```
srinivashegde86:
```suggestion
type AlertCondition struct {
```

tools/monitoring/yaml-handler/parser.go
```diff
@@ -0,0 +1,116 @@
...
+	ErrorPattern string `yaml:"error-pattern"`
+	Hint         string
+	Alerts       []AlertConditions
+}
+
+type YamlFile struct {
+	Spec []PatternSpec `yaml:"spec"`
+}
+
+type Output struct {
```
srinivashegde86: can we rename this to something that makes more sense about what output this is?

tools/monitoring/yaml-handler/parser.go
```diff
@@ -0,0 +1,116 @@
...
+	Hint         string
+	Occurrences  int
+	JobsAffected string
+	PrsAffected  string
+	Period       int
```
srinivashegde86: should be time.Minutes

tools/monitoring/yaml-handler/parser.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+type YamlFile struct {
+	Spec []PatternSpec `yaml:"spec"`
+}
+
+type Output struct {
+	Hint         string
+	Occurrences  int
+	JobsAffected string
```
srinivashegde86: int?

tools/monitoring/yaml-handler/parser.go
```diff
@@ -0,0 +1,116 @@
...
+	PrsAffected  string
```
srinivashegde86: int?

tools/monitoring/yaml-handler/parser.go
```diff
@@ -0,0 +1,116 @@
...
+	Period       int
+}
+
+//GetSpec gets the spec for a particular error pattern and a matching job name pattern
```
srinivashegde86:
```suggestion
// GetSpec gets the spec for a particular error pattern and a matching job name pattern
```

tools/monitoring/yaml-handler/parser.go
```diff
@@ -0,0 +1,116 @@
...
+func GetSpec(f YamlFile, pattern, jobName string) (output Output, noMatchError error) {
+	noMatchError = errors.New(fmt.Sprintf("No spec found for pattern[%s] and jobName[%s]", pattern, jobName))
```
srinivashegde86:
```suggestion
	noMatchError := errors.New(fmt.Sprintf("No spec found for pattern[%s] and jobName[%s]", pattern, jobName))
```

tools/monitoring/yaml-handler/parser.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	content, err := ioutil.ReadAll(res.Body)
+	res.Body.Close()
+	if err != nil {
+		log.Fatal(err)
+	}
+	return content
+}
+
+//ParseYaml reads the yaml text and convert it to the YamlFile struct defined
```
srinivashegde86:
```suggestion
// ParseYaml reads the yaml text and converts it to the config struct defined
```

tools/monitoring/yaml-handler/parser.go
```diff
@@ -0,0 +1,116 @@
...
+	return output, noMatchError
+}
+
+func getFileBytes(url string) []byte {
+	res, err := http.Get(url)
+	if err != nil {
+		log.Fatal(err)
+	}
+	content, err := ioutil.ReadAll(res.Body)
+	res.Body.Close()
```
srinivashegde86:
```suggestion
	defer res.Body.Close()
```

tools/monitoring/yaml-handler/parser.go
```diff
@@ -0,0 +1,116 @@
...
+					break
+				}
+			}
+			break
+		}
+	}
+	return output, noMatchError
+}
+
+func getFileBytes(url string) []byte {
```
srinivashegde86: return an error here instead of logging it

tools/monitoring/yaml-handler/parser.go
```diff
@@ -0,0 +1,116 @@
...
+	content, err := ioutil.ReadAll(res.Body)
+	res.Body.Close()
+	if err != nil {
+		log.Fatal(err)
+	}
+	return content
+}
+
+//ParseYaml reads the yaml text and convert it to the YamlFile struct defined
+func ParseYaml(url string) YamlFile {
```
srinivashegde86: this should return error instead of logging

tools/monitoring/yaml-handler/parser.go
```diff
@@ -0,0 +1,116 @@
...
+	content := getFileBytes(url)
+	file := YamlFile{}
+
+	if err := yaml.Unmarshal(content, &file); err != nil {
+		log.Fatalf("Cannot parse config %q: %v", url, err)
+	}
+	return file
+}
+
+//CollectErrorPatterns collects and returns all error patterns in the yaml file
```
srinivashegde86:
```suggestion
// CollectErrorPatterns collects and returns all error patterns in the yaml file
```

tools/monitoring/yaml-handler/parser.go
```diff
@@ -0,0 +1,116 @@
...
+func CollectErrorPatterns(f YamlFile) []string {
```
srinivashegde86: this can be a method on the struct instead of taking it as input

tools/monitoring/yaml-handler/parser.go
```diff
@@ -0,0 +1,116 @@
...
+type Output struct {
+	Hint         string
+	Occurrences  int
+	JobsAffected string
+	PrsAffected  string
+	Period       int
+}
+
+//GetSpec gets the spec for a particular error pattern and a matching job name pattern
+func GetSpec(f YamlFile, pattern, jobName string) (output Output, noMatchError error) {
```
srinivashegde86: can be a method on the struct

tools/monitoring/yaml-handler/parser.go
```diff
@@ -0,0 +1,116 @@
...
+	noMatchError = errors.New(fmt.Sprintf("No spec found for pattern[%s] and jobName[%s]", pattern, jobName))
+	for _, patternSpec := range f.Spec {
+		if pattern == patternSpec.ErrorPattern {
+			noMatchError = errors.New(fmt.Sprintf("Spec found for pattern[%s], but no match for job name[%s]", pattern, jobName))
```
srinivashegde86: can we use fmt.Errorf() instead of errors.New()?

tools/monitoring/yaml-handler/parser.go
```diff
@@ -0,0 +1,116 @@
...
+			output.Hint = patternSpec.Hint
+			for _, alertCondition := range patternSpec.Alerts {
+				matched, err := regexp.MatchString(alertCondition.JobNameRegex, jobName)
+				if err != nil {
+					continue
```
srinivashegde86: should we log here?

tools/monitoring/config-getter/sample.yaml
```diff
@@ -0,0 +1,28 @@
+spec:
```
srinivashegde86: dont think we need another folder here. we can prob have tools/monitoring/sampleConfig.yaml

tools/monitoring/main.go
steuhs: When the app is deployed on gke, it need to point to what is stored on github. We don't want to redeploy the app every time we update the yaml. That's why I don't point it to the sample yaml in the same directory

tools/monitoring/yaml-handler/parser.go
steuhs: time.Minutes is not a data type. It's a function
 func (t Time) Minute() int

tools/monitoring/yaml-handler/parser.go
steuhs: this struct is a collection of conditions instead of one single condition

tools/monitoring/yaml-handler/parser.go
steuhs: I want to emphasize that this struct is containing the entire yaml file, instead of part of the file. ErrorPatternConfig seems a bit vague and can apply to many structs

tools/monitoring/yaml-handler/parser.go
steuhs: time.Minutes is not a data type. It's a function
func (t Time) Minute() int

tools/monitoring/yaml-handler/parser.go
steuhs: by returning the error instead of stop the flow. We give the user of the function freedom to handle the error. It is possible that some jobs are not matched with any config intentionally.

tools/monitoring/main.go
srinivashegde86: yes. When the update the yaml, we should pick up the latest changes when we query github. Why would we need ot redeploy the app?

The URL here points to your personal branch, which it should not. It should point to the sample yaml being added in this PR.

tools/monitoring/config-getter/sample.yaml
srinivashegde86: also, missing header

tools/monitoring/config.go
```diff
@@ -0,0 +1,123 @@
...
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"regexp"
+
+	"gopkg.in/yaml.v2"
+)
+
+type alertConditions struct {
```
srinivashegde86: this should be alertCondition as this represent a single occurrence of an alert condition. The patternspec below has an array of these and can be called alerts.

tools/monitoring/config.go
```diff
@@ -0,0 +1,123 @@
...
+	return output, noMatchError
+}
+
+func getFileBytes(url string) ([]byte, error) {
+	res, err := http.Get(url)
+	if err != nil {
+		return nil, err
+	}
+	content, err := ioutil.ReadAll(res.Body)
+	defer res.Body.Close()
```
srinivashegde86:
```
defer res.Body.Close()
return ioutil.ReadAll(res.Body)

File: tools/monitoring/config.go [hunk 26]
```diff
@@ -0,0 +1,123 @@
...
+type SelectedConfig struct {
+	Hint         string
+	Occurrences  int
+	JobsAffected int
+	PrsAffected  int
+	Period       int
+}
+
+// Select gets the spec for a particular error pattern and a matching job name pattern
+func (f Config) Select(pattern, jobName string) (SelectedConfig, error) {
```
srinivashegde86: keep all methods on config together

PR #783, opened 2019-05-15
Fix the tag name in some situation

Fredy-Z:
Add `--abbrev=0` to use the closest tag.

According to https://linux.die.net/man/1/git-describe, if the current branch is based on one tag, e.g. v1.0.4, but it has a few commits on top of that, describe will add the number of additional commits ("14") and an abbreviated object name for the commit itself ("2414721") at the end. So when we run `git describe --tags` will get us `v1.0.4-14-g2414721`.

And with `--abbrev` set to 0, the command can be used to find the closest tagname without any suffix, so `git describe --tags --abbrev` will get us `v1.0.4`.


chaodaiG:
/lgtm

suggest update PR content reflecting why doing it

chaodaiG:
/approve
/hold

feel free unhold when you are done with the PR content

PR #790, opened 2019-05-16
Add directory containing resources for Devstats Dashboard site

### LINKED ISSUE #753: Merge devstats update into test-infra
/assign @ericKlawitter 

/milestone M5

ericKlawitter:
This adds the resources to run our [Devstats](https://github.com/cncf/devstats) instance, and any additional dashboards we want to run on top of it. This includes the Dockerfiles, Kubernetes manifests, and the scripts required to setup (see `INSTALL.md`). 

This commit includes a lot of scripts written by devstats.  There should be a follow up which removes this, so that test-infra contains only the necessary manifests. I added a [diff file,](https://github.com/knative/test-infra/files/3189324/diff.txt) which contains only the files I've modified, for review. 

I excluded the postgres-docker-entrypoint.sh file, which is a copy from [Postgres' GitHub repository](https://github.com/docker-library/postgres/blob/master/docker-entrypoint.sh) as a recommended entrypoint into docker containers, and the postgresql.conf which is a standard Postgres configuration file which I modified just to allow remote access for outside pods. 

[diff.txt](https://github.com/knative/test-infra/files/3189324/diff.txt)

Putting a hold on this large commit for now. 

Fixes #753 

/hold
/assign @srinivashegde86 @adrcunha


srinivashegde86:
Couple of generic comments:
* Add license header to all your files
* Move this under tools/ instead of the parent dir.

Other than that, agree with Adriano, that we dont have to review each file here.

srinivashegde86:
LGTM.

I will let adriano add the labels :)

PR #793, opened 2019-05-17
Determine the OS we're running on

adrcunha:
Set IS_LINUX, IS_OSX and IS_WINDOWS accordingly.

### DIFF
--- a/scripts/library.sh
+++ b/scripts/library.sh
@@ -36,6 +36,20 @@ readonly IS_PROW
 readonly REPO_ROOT_DIR="$(git rev-parse --show-toplevel)"
 readonly REPO_NAME="$(basename ${REPO_ROOT_DIR})"
 
+# Useful flags about the current OS
+IS_LINUX=0
+IS_OSX=0
+IS_WINDOWS=0
+case "${OSTYPE}" in
+  darwin*) IS_OSX=1 ;;
+  linux*) IS_LINUX=1 ;;
+  msys*) IS_WINDOWS=1 ;;
+  *) echo "** Internal error in library.sh, unknown OS" ; exit 1 ;;
+esac
+readonly IS_LINUX
+readonly IS_OSX
+readonly IS_WINDOWS
+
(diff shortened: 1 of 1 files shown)


scripts/library.sh
```diff
@@ -36,6 +36,20 @@ readonly IS_PROW
...
 
+# Useful flags about the current OS
+IS_LINUX=0
+IS_OSX=0
+IS_WINDOWS=0
+case "${OSTYPE}" in
+  darwin*) IS_OSX=1 ;;
+  linux*) IS_LINUX=1 ;;
+  msys*) IS_WINDOWS=1 ;;
+  *) echo "** Internal error in library.sh, unknown OS" ; exit 1 ;;
```
```suggestion
  *) echo "** Internal error in library.sh, unknown OS '${OSTYPE}'" ; exit 1 ;;
```

scripts/library.sh
```diff
@@ -36,6 +36,20 @@ readonly IS_PROW
 readonly REPO_ROOT_DIR="$(git rev-parse --show-toplevel)"
 readonly REPO_NAME="$(basename ${REPO_ROOT_DIR})"
 
+# Useful flags about the current OS
+IS_LINUX=0
+IS_OSX=0
+IS_WINDOWS=0
+case "${OSTYPE}" in
```
chaodaiG: What would the value of `${OSTYPE}` be for a linux container running on OSX?

adrcunha: I guess it would be Linux, because it's bash who defines OSTYPE.

chaodaiG: should we document it here or would it be the caller's job to figure this out?

adrcunha: This is a corner case and I can't imagine such situation in the present state of test-infra. If we ever get to this situation, I'm sure way more issues will arise when running these scripts (assuming they will be even being used then) than realizing that for some reason bash wrongly sets OSTYPE.

chaodaiG: make sense

Issue #736, opened 2019-05-06
Report flaky tests for tests of different istio/mesh combinations


chaodaiG:
We are running the following combinations of tests in serving:

- istio-1.0.7-mesh
- istio-1.0.7-no-mesh
- istio-1.1.2-mesh
- istio-1.1.2-no-mesh

As requested by #networking team, it would be nice to have automatic flaky tests reporting in #networking channel via Slack.

This work will require some code changes other than config changes, as flaky-test-reporter tool was designed to send notification to Slack channel based on repo. And since it is already sending notification to #serving-api from continuous run, sending to #networking channel requires some change.

Expected outcome:
- Flaky-test-reporter tool also scans these continuous runs: istio-1.0.7-mesh, istio-1.0.7-no-mesh, istio-1.1.2-mesh, and istio-1.1.2-no-mesh
- Report test failed in these runs in #networking channel via Slack
- [optional] Creating Github issues tracking these failures (This might not even be doable due to same set of tests run in all configurations)

dushyanthsc:
Current plan is to look at this issue in Q2(Apr-Jun)

PR #795, opened 2019-05-17
Remove pointer to JobConfig in RepoData struct to fix memory bug

TrevorFarrelly:
### Bug:
There seems to be a quirk with Go loops, where the local variable used is actually a copy of the current list item, with the same memory address across iterations. When adding multiple JobConfigs to run, the RepoData structs created referenced this memory address, thus making them all reference the last JobConfig when the loop exited, even though they ran the correct configuration on their iteration.
### Fix:
Remove pointers entirely. Seems similar to a quirk with [Goroutines](https://github.com/golang/go/wiki/CommonMistakes#using-goroutines-on-loop-iterator-variables), though I haven't found any documentation on this case specifically.

This was discovered while working on a fix for #736

### DIFF
--- a/tools/flaky-test-reporter/github_issue_test.go
+++ b/tools/flaky-test-reporter/github_issue_test.go
@@ -80,7 +80,7 @@ func createNewIssue(fgi *GithubIssue, title, body, testStat string) (*github.Iss
 }
 
 func createRepoData(passed, flaky, failed, notenoughdata int, startTime int64) *RepoData {
-	config := &JobConfig{
+	config := JobConfig{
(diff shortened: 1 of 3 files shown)


chaodaiG:
/approve

wait for @srinivashegde86 to lgtm

srinivashegde86:
Can you elaborate on the bug here? In the Pr description or create another issue that this will fix.

TrevorFarrelly:
> Can you elaborate on the bug here? In the Pr description or create another issue that this will fix.

Fixed.

PR #796, opened 2019-05-17
Adds unit tests for config yaml parser

steuhs:
also improved source code a bit to make unit testing easier

Fixes #

### DIFF
--- a/tools/monitoring/config_test.go
+++ b/tools/monitoring/config_test.go
@@ -120,7 +120,7 @@ func TestCollectErrorPatterns(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			if got := tt.args.f.CollectErrorPatterns(); !reflect.DeepEqual(got, tt.want) {
-				t.Errorf("CollectErrorPatterns() = %v, want %v", got, tt.want)
+				t.Errorf("Collected Error Patterns, got: %v, want: %v", got, tt.want)
 			}
 		})
 	}
@@ -184,11 +184,11 @@ func TestConfig_Select(t *testing.T) {
 			}
 			got, err := f.Select(tt.args.pattern, tt.args.jobName)
 			if (err != nil) != tt.wantErr {
-				t.Errorf("Config.Select() error = %v, wantErr %v", err, tt.wantErr)
+				t.Errorf("Config Selector returned error.")
 				return
 			}
 			if !reflect.DeepEqual(got, tt.want) {
-				t.Errorf("Config.Select() = %v, want %v", got, tt.want)
+				t.Errorf("Config selected, got : %v, want: %v", got, tt.want)
 			}
 		})
 	}
@@ -217,11 +217,11 @@ func Test_newConfig(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			got, err := newConfig(tt.args.text)
 			if (err != nil) != tt.wantErr {
-				t.Errorf("newConfig() error = %v, wantErr %v", err, tt.wantErr)
+				t.Errorf("config construction failed")
 				return
 			}
 			if !reflect.DeepEqual(got, tt.want) {
-				t.Errorf("newConfig() = %v, want %v", got, tt.want)
+				t.Errorf("config constructed,  got: %v, want: %v", got, tt.want)
 			}
 		})
 	}
--- a/tools/monitoring/config_test.go
+++ b/tools/monitoring/config_test.go
@@ -1,0 +1,228 @@
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"reflect"
+	"testing"
+)
+
+const (
+	sampleYaml = `spec:
+  - error-pattern: 'Something went wrong: starting e2e cluster: error creating cluster'
+    hint: 'Check gcp status'
+    alerts:
+      - job-name-regex: '^pull.*'
+        occurrences: 2
+        jobs-affected: 1
+        prs-affected: 2
+        period: 1440 # 1440 minutes = 24 hours
+      - job-name-regex: '.*'
+        occurrences: 5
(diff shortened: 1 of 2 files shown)


tools/monitoring/config_test.go
```diff
@@ -0,0 +1,228 @@
...
+	},
+}
+
+func TestCollectErrorPatterns(t *testing.T) {
+	conf, err := newConfig([]byte(sampleYaml))
+	if err != nil {
+		t.Errorf("cannot parse yaml: %v", err)
+	}
+
+	type args struct {
```
srinivashegde86: do we need this struct? can we directly have f *Config in the tests struct?

tools/monitoring/config_test.go
```diff
@@ -0,0 +1,228 @@
...
+			f := Config{
+				Spec: tt.fields.Spec,
+			}
+			got, err := f.Select(tt.args.pattern, tt.args.jobName)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("Config.Select() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("Config.Select() = %v, want %v", got, tt.want)
```
srinivashegde86: can you add more elaborate all the err messages? config.Select() does not mean much unless you know the code.

tools/monitoring/config_test.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := newConfig(tt.args.text)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("newConfig() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("newConfig() = %v, want %v", got, tt.want)
```
srinivashegde86: the pattern normally followed is "<msg> got:%v, want:%v"

tools/monitoring/config_test.go
```diff
@@ -0,0 +1,228 @@
...
+			},
+			want:    &sampleConfig,
+			wantErr: false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := newConfig(tt.args.text)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("newConfig() error = %v, wantErr %v", err, tt.wantErr)
```
srinivashegde86: all of t.Errorf() should be t.Fatalf()

tools/monitoring/config_test.go
```diff
@@ -0,0 +1,228 @@
...
+      - job-name-regex: '.*'
+        occurrences: 5
+        jobs-affected: 2
+        prs-affected: 1 # for non-pull jobs, we don't care about the number of prs affected, so we set the number to 1, which will basically make this particular condition always true
+        period: 1440
+
+  - error-pattern: 'sample*error2'
+    hint: 'hint_for_pattern_2'
+    alerts:
+      - job-name-regex: '^pull.*'
```
yt3liu: What happens if there's a missing field e.g. jobs-affected? What is the error propagated? Or is it ignoring that particular error pattern if that's the case?

yt3liu:
One general question: How do you think we should validate config yaml file? Or is it required to validate it? 

I think the current implementation only checks the config is a valid yaml, but doesn't check whether any of the fields that we want to read from are present until it is required. `Select(..)` is reading the those fields in yaml on the fly. 

My concern is if there's a new pattern that's added, how do we know the alert condition for that pattern in the config is in the right format?

tools/monitoring/config_test.go
srinivashegde86: We should have reasonable defaults for all the fields that are not madatory and check the ones that need a value

tools/monitoring/config_test.go
steuhs: Having this struct in all unit tests brings consistency. It's the default way intellij constructs unit tests

tools/monitoring/config_test.go
steuhs: Why? They are equivalent in all the cases because there is no statement after them

steuhs:
> One general question: How do you think we should validate config yaml file? Or is it required to validate it?
> 
> I think the current implementation only checks the config is a valid yaml, but doesn't check whether any of the fields that we want to read from are present until it is required. `Select(..)` is reading the those fields in yaml on the fly.
> 
> My concern is if there's a new pattern that's added, how do we know the alert condition for that pattern in the config is in the right format?

Validating the validity of actual input should not be part of the unit test. You can add a to-do in the source code and/or new github issue and we can discuss if we want to add that to the source code. @yt3liu . cc @srinivashegde86

tools/monitoring/config_test.go
steuhs: I agree we can add defaults for some fields. Yet we should do that in the source code, and we should add a github task for that to do later. This pr is for unit test and we shouldn't add the check here.

tools/monitoring/config_test.go
srinivashegde86: that is an IDE generalization and does not mean we should do that. It makes sense in most of the cases, but in this case, it adds an unnecessary level of indirection. We can keep it if you want to be consistent across the tests.

srinivashegde86:
@yt3liu please add the LGTM label once you have reviewed.

tools/monitoring/config_test.go
steuhs: @yt3liu some updates after our offline conversation. I check with both Unmarshal and UnmarshalStrict. None of them catches missing fields as errors. Instead, they use zero values for the type as defaults. Thus, if we want to have our own defaults and/or check missing fields, we should add a validator/processor on the returned structure of newConfig.

That said, UnmarshalStrict will report error on any field name that does not appear in the unmarshall target struct. So it will help us find some errors in the config such as field misspelling.

yt3liu: Thanks for looking into this. 

In this case, I think we do need to have an issue for setting the reasonable defaults so we don't accidentally set the state to always alarm (occurrence, prs-affected, jobs-affected=0) or never alarm (period=0). I think we can merge this in after we have an issue for to track it. 

I can't think of a use case for when the config.yaml would have fields not in the struct and still be a valid one except for the case where we'd like to update the names of the key i.e. job-name-regex to job-name-match-string, then config.yaml can be temporarily invalid if we use UnmarshalStrict. But I don't see we'll want to do that in the future, so do you think it make sense to change `Unmarshal` to `UnmarshalStrict`? 
(You can do that in a separate PR if you agree UnmarshalStrict handles better error catching than Unmarshal) 

Please create an issue to set the reasonable defaults referencing this issue.

steuhs: tracking issue: https://github.com/knative/test-infra/issues/810

Yes, I agree UnmarshalStrict is better

Issue #804, opened 2019-05-20
Prevent e2e test from running on Prow cluster


chaodaiG:
Prow cluster is in critical path of all our automated flows, running e2e on it could make it not working properly. Add a mechanism in e2e scripts to guard against this from happening

PR #808, opened 2019-05-20
Prevents e2e test from running on prow cluster

chaodaiG:
For debugging e2e tests on existing cluster, the cluster is determined by `kubectl config current-context`, and there is no guard against running on important cluster, such as prow. Add a guard here for safety concern

Fixes: #804

### DIFF
--- a/scripts/e2e-tests.sh
+++ b/scripts/e2e-tests.sh
@@ -283,6 +283,9 @@ function setup_test_cluster() {
   local k8s_user=$(gcloud config get-value core/account)
   local k8s_cluster=$(kubectl config current-context)
 
+  is_protected_cluster ${k8s_cluster} && \
+    abort "\$Kubeconfig context set to ${k8s_cluster}, which is forbidden"
+
   # If cluster admin role isn't set, this is a brand new cluster
   # Setup the admin role and also KO_DOCKER_REPO
   if [[ -z "$(kubectl get clusterrolebinding cluster-admin-binding 2> /dev/null)" ]]; then
--- a/scripts/e2e-tests.sh
+++ b/scripts/e2e-tests.sh
@@ -284,7 +284,7 @@ function setup_test_cluster() {
   local k8s_cluster=$(kubectl config current-context)
 
   is_protected_cluster ${k8s_cluster} && \
-    abort "\$Kubeconfig context set to ${k8s_cluster}, which is forbidden"
+    abort "kubeconfig context set to ${k8s_cluster}, which is forbidden"
 
   # If cluster admin role isn't set, this is a brand new cluster
   # Setup the admin role and also KO_DOCKER_REPO
--- a/scripts/library.sh
+++ b/scripts/library.sh
@@ -18,6 +18,9 @@
 # to be used in test scripts and the like. It doesn't do anything when
 # called from command line.
 
+# GCP project where all tests releate resources live
+readonly KNATIVE_TESTS_PROJECT=knative-tests
+
 # Default GKE version to be used with Knative Serving
 readonly SERVING_GKE_VERSION=gke-latest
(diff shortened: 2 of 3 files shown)


scripts/library.sh
```diff
@@ -418,6 +418,12 @@ function is_protected_gcr() {
   [[ -n $1 && "$1" =~ "^gcr.io/knative-(releases|nightly)/?$" ]]
 }
 
+# Return whether the given parameter is the knative-prow cluster.
+# Parameters: $1 - Kubernetes cluster context (output of kubectl config current-context)
+function is_protected_cluster() {
+  [[ -n $1 && "$1" = "gke_knative-tests_us-central1-f_prow" ]]
```
Fredy-Z: Will it be better to define "gke_knative-tests_us-central1-f_prow" as a var [here](https://github.com/knative/test-infra/blob/master/scripts/library.sh#L20)?

chaodaiG: prefer not to put it up there, as:
1. this variable is not used anywhere else, and it's not meant to be
1. consistent with is_protected_gcr above

scripts/e2e-tests.sh
```diff
@@ -283,6 +283,9 @@ function setup_test_cluster() {
   local k8s_user=$(gcloud config get-value core/account)
   local k8s_cluster=$(kubectl config current-context)
 
+  is_protected_cluster ${k8s_cluster} && \
+    abort "\$Kubeconfig context set to ${k8s_cluster}, which is forbidden"
```
```suggestion
    abort "kubeconfig context set to ${k8s_cluster}, which is forbidden"
```

scripts/library.sh
adrcunha: I suggest that:
* `knative_tests` becomes a constant, used here and in `is_protected_project()`
*  you use a regex to ignore the zone (I think you can even ignore the region)

scripts/library.sh
```diff
@@ -18,6 +18,9 @@
 # to be used in test scripts and the like. It doesn't do anything when
 # called from command line.
 
+# GCP project where all tests releate resources live
```
```suggestion
# GCP project where all tests related resources live
```

scripts/library.sh
```diff
@@ -415,7 +418,20 @@ function is_int() {
 # Return whether the given parameter is the knative release/nightly GCF.
 # Parameters: $1 - full GCR name, e.g. gcr.io/knative-foo-bar
 function is_protected_gcr() {
-  [[ -n $1 && "$1" =~ "^gcr.io/knative-(releases|nightly)/?$" ]]
+  [[ -n $1 && $1 =~ ^gcr.io/knative-(releases|nightly)/?$ ]]
```
adrcunha: Are the quotes giving trouble?

chaodaiG: not here, the one below given me trouble:
`[[ "gke_knative-tests_us-central1-f_prow" =~ "^gke_knative-tests_us\-[a-zA-Z0-9]+\-[a-z]+_prow$" ]] &
& echo "match" || echo "no match"` gave me `no match`, and removing `"` got `match`

change it here for consistency

adrcunha: I think it's the backslash. Can we get unit tests for the new/changed methods please? :)

chaodaiG: actually I just tested, `[[ "gcr.io/knative-releases" =~ "^gcr.io/knative-(releases|nightly)/?$" ]] && echo "match" || echo "no match"` gave me "no match` as well

chaodaiG: agree with unit test, will do

chaodaiG: unit tests added, and decided to extend protection on all clusters under `knative-tests`

adrcunha:
Thanks!
/lgtm

PR #809, opened 2019-05-20
Add flaky test reporting for various istio/mesh combinations

### LINKED ISSUE #736: Report flaky tests for tests of different istio/mesh combinations

TrevorFarrelly:
Refactored JobConfigs data structure to be more extensible in the future, and added a flag for GitHub issue creation if one does not want issues to be created for a specific job, as specified in #736.

Blocked on the Slack networking channel ID, as I do not have access to the team Slack. Once I have that, this change should be good to go.

If someone would like documentation on how to add test configurations for other jobs/repositories, I'd be happy to add it to the README.

/hold

Fixes #736

### DIFF
--- a/tools/flaky-test-reporter/config.go
+++ b/tools/flaky-test-reporter/config.go
@@ -15,6 +15,7 @@ limitations under the License.
 */
 
 // config.go contains configurations for flaky tests reporting
+// TODO: change testing customizations back to default before PR
 
 package main
 
@@ -30,14 +31,32 @@ const (
 	// Don't do anything if found more than 1% tests flaky, this is an arbitrary number
 	threshold = 0.01
 
-	org = "knative"
+	org = "TrevorFarrelly"
 )
 
 var (
-	jobConfigs = []JobConfig{
-		{"ci-knative-serving-continuous", "serving", prow.PostsubmitJob}, // CI flow for serving repo
+	// jobConfigs lists all repos and jobs to run within those repos
+	jobConfigs = map[string][]JobConfig{
+		// CI flows for serving repo
+		"serving": {{Name: "ci-knative-serving-continuous", Type: prow.PostsubmitJob},
+			{Name: "ci-knative-serving-istio-1.0.7-mesh", Type: prow.PostsubmitJob},
+			{Name: "ci-knative-serving-istio-1.0.7-no-mesh", Type: prow.PostsubmitJob},
+			{Name: "ci-knative-serving-istio-1.1.2-mesh", Type: prow.PostsubmitJob},
+			{Name: "ci-knative-serving-istio-1.1.2-no-mesh", Type: prow.PostsubmitJo},
+		},
 	}
-	repoIssueMap = map[string]string{
-		"serving": "serving",
+	// slackChannelsMap lists which slack channel to post results in for each job in repo
+	slackChannelsMap = map[string]map[string][]slackChannel{
+		// channel mapping for serving repo
+		"serving": {"default": {{"api", "CJEQ0MB50" /*"CA4DNJ9A4"*/}},
+			"ci-knative-serving-istio-1.0.7-mesh":    {{"networking", "CJS634MKP"}},
+			"ci-knative-serving-istio-1.0.7-no-mesh": {{"networking", "CJS634MKP"}},
+			"ci-knative-serving-istio-1.1.2-mesh":    {{"networking", "CJS634MKP"}},
+			"ci-knative-serving-istio-1.1.2-no-mesh": {{"networking", "CJS634MKP"}},
+		},
+	}
+
+	githubIssueMap = map[string]string{
+		"serving": "knative-testing",
 	}
 )
--- a/tools/flaky-test-reporter/config.go
+++ b/tools/flaky-test-reporter/config.go
@@ -38,11 +38,11 @@ var (
 	// jobConfigs lists all repos and jobs to run within those repos
 	jobConfigs = map[string][]JobConfig{
 		// CI flows for serving repo
-		"serving": {{Name: "ci-knative-serving-continuous", Type: prow.PostsubmitJob},
-			{Name: "ci-knative-serving-istio-1.0.7-mesh", Type: prow.PostsubmitJob},
-			{Name: "ci-knative-serving-istio-1.0.7-no-mesh", Type: prow.PostsubmitJob},
-			{Name: "ci-knative-serving-istio-1.1.2-mesh", Type: prow.PostsubmitJob},
-			{Name: "ci-knative-serving-istio-1.1.2-no-mesh", Type: prow.PostsubmitJo},
+		"serving": {{Name: "ci-knative-serving-continuous", Type: prow.PostsubmitJob, PostIssue: true},
+			{Name: "ci-knative-serving-istio-1.0.7-mesh", Type: prow.PostsubmitJob, PostIssue: true},
+			{Name: "ci-knative-serving-istio-1.0.7-no-mesh", Type: prow.PostsubmitJob, PostIssue: false},
+			{Name: "ci-knative-serving-istio-1.1.2-mesh", Type: prow.PostsubmitJob, PostIssue: false},
+			{Name: "ci-knative-serving-istio-1.1.2-no-mesh", Type: prow.PostsubmitJob, PostIssue: false},
 		},
 	}
 	// slackChannelsMap lists which slack channel to post results in for each job in repo
--- a/tools/flaky-test-reporter/config.go
+++ b/tools/flaky-test-reporter/config.go
@@ -15,7 +15,6 @@ limitations under the License.
 */
 
 // config.go contains configurations for flaky tests reporting
-// TODO: change testing customizations back to default before PR
 
 package main
 
@@ -31,7 +30,7 @@ const (
 	// Don't do anything if found more than 1% tests flaky, this is an arbitrary number
 	threshold = 0.01
 
-	org = "TrevorFarrelly"
+	org = "knative"
 )
 
 var (
@@ -48,15 +47,15 @@ var (
 	// slackChannelsMap lists which slack channel to post results in for each job in repo
 	slackChannelsMap = map[string]map[string][]slackChannel{
 		// channel mapping for serving repo
-		"serving": {"default": {{"api", "CJEQ0MB50" /*"CA4DNJ9A4"*/}},
-			"ci-knative-serving-istio-1.0.7-mesh":    {{"networking", "CJS634MKP"}},
-			"ci-knative-serving-istio-1.0.7-no-mesh": {{"networking", "CJS634MKP"}},
-			"ci-knative-serving-istio-1.1.2-mesh":    {{"networking", "CJS634MKP"}},
-			"ci-knative-serving-istio-1.1.2-no-mesh": {{"networking", "CJS634MKP"}},
+		"serving": {"default": {{"api", "CA4DNJ9A4"}},
+			"ci-knative-serving-istio-1.0.7-mesh":    {{"networking", ""}},
+			"ci-knative-serving-istio-1.0.7-no-mesh": {{"networking", ""}},
+			"ci-knative-serving-istio-1.1.2-mesh":    {{"networking", ""}},
+			"ci-knative-serving-istio-1.1.2-no-mesh": {{"networking", ""}},
 		},
 	}
 
 	githubIssueMap = map[string]string{
-		"serving": "knative-testing",
+		"serving": "serving",
 	}
 )
--- a/tools/flaky-test-reporter/config.go
+++ b/tools/flaky-test-reporter/config.go
@@ -44,10 +44,10 @@ var (
 			{Name: "ci-knative-serving-istio-1.1.2-no-mesh", Type: prow.PostsubmitJob, PostIssue: false},
 		},
 	}
-	// slackChannelsMap lists which slack channel to post results in for each job in repo
+	// slackChannelsMap lists which Slack channel to post results in for each job in repo
 	slackChannelsMap = map[string]map[string][]slackChannel{
 		// channel mapping for serving repo
-		"serving": {"default": {{"api", "CA4DNJ9A4"}},
+		"serving": {"ci-knative-serving-continuous": {{"api", "CA4DNJ9A4"}},
 			"ci-knative-serving-istio-1.0.7-mesh":    {{"networking", "CA9RHBGJX"}},
 			"ci-knative-serving-istio-1.0.7-no-mesh": {{"networking", "CA9RHBGJX"}},
 			"ci-knative-serving-istio-1.1.2-mesh":    {{"networking", "CA9RHBGJX"}},
--- a/tools/flaky-test-reporter/config.go
+++ b/tools/flaky-test-reporter/config.go
@@ -47,6 +47,8 @@ var (
 	// slackChannelsMap lists which Slack channel to post results in for each job in repo
 	slackChannelsMap = map[string]map[string][]slackChannel{
 		// channel mapping for serving repo
+		// "CA4DNJ9A4" => serving-api
+		// "CA9RHBGJX" => networking
 		"serving": {"ci-knative-serving-continuous": {{"api", "CA4DNJ9A4"}},
 			"ci-knative-serving-istio-1.0.7-mesh":    {{"networking", "CA9RHBGJX"}},
 			"ci-knative-serving-istio-1.0.7-no-mesh": {{"networking", "CA9RHBGJX"}},
--- a/tools/flaky-test-reporter/config.go
+++ b/tools/flaky-test-reporter/config.go
@@ -34,11 +34,11 @@ const (
 )
 
 var (
-	// jobConfigs lists all repos and jobs to run within those repos
+	// jobConfigs lists all repos and jobs to analyze within those repos
 	jobConfigs = map[string][]JobConfig{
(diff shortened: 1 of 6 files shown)


tools/flaky-test-reporter/config.go
```diff
@@ -34,10 +34,28 @@ const (
 )
 
 var (
-	jobConfigs = []JobConfig{
-		{"ci-knative-serving-continuous", "serving", prow.PostsubmitJob}, // CI flow for serving repo
+	// jobConfigs lists all repos and jobs to run within those repos
```
chaodaiG:
```suggestion
	// jobConfigs lists all repos and jobs to analyze within those repos
```

chaodaiG:
```suggestion
	// jobConfigs lists all repos and jobs to analyze within those repos
```

tools/flaky-test-reporter/config.go
```diff
@@ -34,10 +34,28 @@ const (
...
+	jobConfigs = map[string][]JobConfig{
+		// CI flows for serving repo
+		"serving": {{Name: "ci-knative-serving-continuous", Type: prow.PostsubmitJob, PostIssue: true},
+			{Name: "ci-knative-serving-istio-1.0.7-mesh", Type: prow.PostsubmitJob, PostIssue: true},
```
chaodaiG: this should be false

tools/flaky-test-reporter/config.go
```diff
@@ -34,10 +34,28 @@ const (
...
+			{Name: "ci-knative-serving-istio-1.0.7-no-mesh", Type: prow.PostsubmitJob, PostIssue: false},
+			{Name: "ci-knative-serving-istio-1.1.2-mesh", Type: prow.PostsubmitJob, PostIssue: false},
+			{Name: "ci-knative-serving-istio-1.1.2-no-mesh", Type: prow.PostsubmitJob, PostIssue: false},
+		},
 	}
-	repoIssueMap = map[string]string{
+	// slackChannelsMap lists which slack channel to post results in for each job in repo
```
chaodaiG: suggest mentioning fallback logic here

tools/flaky-test-reporter/config.go
```diff
@@ -34,10 +34,28 @@ const (
...
+	slackChannelsMap = map[string]map[string][]slackChannel{
```
chaodaiG: add a `getSlackChannel` helper function here maybe? so caller doesn't need to worry about fallback logic

chaodaiG: by the way, thank you for moving this logic to a more centralized place

tools/flaky-test-reporter/config.go
```diff
@@ -34,10 +34,28 @@ const (
...
+		// channel mapping for serving repo
+		"serving": {"default": {{"api", "CA4DNJ9A4"}},
```
srinivashegde86: I dont think we should have a default(catch-all) here. If we dont define a repo, we will not send the notification. We should not send it to a random one.

tools/flaky-test-reporter/config.go
```diff
@@ -34,10 +34,28 @@ const (
 )
 
 var (
-	jobConfigs = []JobConfig{
-		{"ci-knative-serving-continuous", "serving", prow.PostsubmitJob}, // CI flow for serving repo
+	// jobConfigs lists all repos and jobs to run within those repos
+	jobConfigs = map[string][]JobConfig{
+		// CI flows for serving repo
+		"serving": {{Name: "ci-knative-serving-continuous", Type: prow.PostsubmitJob, PostIssue: true},
```
srinivashegde86: Can we use the prowapi const here instead of defining our own? Looks like something existing. Feel free to no-op this.

tools/flaky-test-reporter/config.go
```diff
@@ -34,10 +34,28 @@ const (
...
 	}
-	repoIssueMap = map[string]string{
+	// slackChannelsMap lists which slack channel to post results in for each job in repo
+	slackChannelsMap = map[string]map[string][]slackChannel{
+		// channel mapping for serving repo
+		"serving": {"default": {{"api", "CA4DNJ9A4"}},
+			"ci-knative-serving-istio-1.0.7-mesh":    {{"networking", ""}},
+			"ci-knative-serving-istio-1.0.7-no-mesh": {{"networking", ""}},
+			"ci-knative-serving-istio-1.1.2-mesh":    {{"networking", ""}},
+			"ci-knative-serving-istio-1.1.2-no-mesh": {{"networking", ""}},
```
srinivashegde86: I am thinking we should move this to a user defined config than hard-coding it here. So, we can update the config and not know about the code. WDYT?

tools/flaky-test-reporter/result.go
```diff
@@ -49,9 +49,10 @@ type RepoData struct {
 
 // JobConfig is initial configuration for a given repo, defines which job to scan
 type JobConfig struct {
-	Name string
-	Repo string
-	Type string
+	Name      string
```
chaodaiG: we may need to rethink about the naming of `RepoData` and `JobConfig` since the first one doesn't make that much sense any more. We can discuss about this offline

tools/flaky-test-reporter/result.go
```diff
@@ -49,9 +49,10 @@ type RepoData struct {
...
+	Repo      string
+	Type      string
+	PostIssue bool
```
chaodaiG: A comment for this would be nice

chaodaiG:
@tcnghia  fyi

tools/flaky-test-reporter/config.go
TrevorFarrelly: Added helper function to slack_notification.go

tools/flaky-test-reporter/result.go
TrevorFarrelly: The most logical change will require more refactoring, will address in a future PR

tools/flaky-test-reporter/config.go
TrevorFarrelly: I am open to ideas. @chaodaiG implied he would like a fallback option, and I was assuming serving-api was the "standard" channel for the serving test jobs. I will modify this to whichever is desired.

tools/flaky-test-reporter/config.go
TrevorFarrelly: Good idea, will investigate.

tools/flaky-test-reporter/config.go
TrevorFarrelly: Not familiar with the prowapi. prow.PostsubmitJob references a const in shared/prow, is that what you mean?

tools/flaky-test-reporter/config.go
chaodaiG: @srinivashegde86 's suggestion makes sense to me, we can remove the fall back logic

chaodaiG: Just to be clear, the initial thought of this fall back logic is for `getSlackChannel` method being able to get a channel just by repo name, since this repo+jobname pattern doesn't seem to apply in other repos

tools/flaky-test-reporter/slack_notification.go
```diff
@@ -137,6 +136,22 @@ func createSlackMessageForRepo(rd *RepoData, flakyIssuesMap map[string][]*flakyI
 	return message
 }
 
+// getSlackChannel gets the channel(s) to send messages to for a given job configuration
+func getSlackChannel(rd *RepoData) []string {
+	jobMap, ok := slackChannelsMap[rd.Config.Repo]
+	if !ok {
+		log.Printf("cannot find Slack channel mapping for repo '%s', skipping Slack notification", rd.Config.Repo)
+		continue
```
chaodaiG: should be return

tools/flaky-test-reporter/slack_notification.go
```diff
@@ -137,6 +136,21 @@ func createSlackMessageForRepo(rd *RepoData, flakyIssuesMap map[string][]*flakyI
...
+	if !ok {
+		log.Printf("cannot find Slack channel mapping for repo '%s', skipping Slack notification", rd.Config.Repo)
+		return nil, ok
+	}
+	channels, ok := jobMap[rd.Config.Name]
+	if !ok {
+		log.Printf("cannot find Slack channel for job '%s' in repo '%s', skipping Slack notification", rd.Config.Name, rd.Config.Repo)
+		return nil, ok
+	}
+	return channels, ok
```
chaodaiG: seems like this function always return either `nil, false` or `channels, true`, but not the other combinations, is `ok` really needed to be returned?

tools/flaky-test-reporter/slack_notification.go
```diff
@@ -137,6 +136,21 @@ func createSlackMessageForRepo(rd *RepoData, flakyIssuesMap map[string][]*flakyI
 	return message
 }
 
+// getSlackChannel gets the channel(s) to send messages to for a given job configuration
+func getSlackChannel(rd *RepoData) ([]slackChannel, bool) {
```
chaodaiG: JobConfig is sufficient as input for this function, I personally would even prefer passing repo name and job name explicitly

tools/flaky-test-reporter/config.go
TrevorFarrelly: I think this is a big enough change to save for a future PR, if that's okay with you?

tools/flaky-test-reporter/config.go
srinivashegde86: shared/prow is like our version of [prow](https://github.com/kubernetes/test-infra/tree/master/prow/apis/prowjobs/v1). I would like to use the stuff in that lib directly instead of recreating this in our version. Maybe not part of this PR, but in the future.

TrevorFarrelly: Agreed. I will plan on looking at it down the road.

tools/flaky-test-reporter/slack_notification.go
```diff
@@ -137,6 +136,21 @@ func createSlackMessageForRepo(rd *RepoData, flakyIssuesMap map[string][]*flakyI
...
+func getSlackChannel(repo string, name string) []slackChannel {
+	jobMap, ok := slackChannelsMap[repo]
+	if !ok {
+		log.Printf("cannot find Slack channel mapping for repo '%s', skipping Slack notification", repo)
+		return nil
+	}
+	channels, ok := jobMap[name]
+	if !ok {
+		log.Printf("cannot find Slack channel for job '%s' in repo '%s', skipping Slack notification", name, repo)
+		return nil
```
srinivashegde86: not needed

tools/flaky-test-reporter/config.go
```diff
@@ -34,10 +34,28 @@ const (
...
+			{Name: "ci-knative-serving-istio-1.0.7-no-mesh", Type: prow.PostsubmitJob, PostIssue: false},
+			{Name: "ci-knative-serving-istio-1.1.2-mesh", Type: prow.PostsubmitJob, PostIssue: false},
+			{Name: "ci-knative-serving-istio-1.1.2-no-mesh", Type: prow.PostsubmitJob, PostIssue: false},
+		},
 	}
-	repoIssueMap = map[string]string{
+	// slackChannelsMap lists which Slack channel to post results in for each job in repo
+	slackChannelsMap = map[string]map[string][]slackChannel{
+		// channel mapping for serving repo
+		"serving": {"ci-knative-serving-continuous": {{"api", "CA4DNJ9A4"}},
```
srinivashegde86: can we add comments for ID -> slack channel. for eg. CA4DNJ9A4 => serving-api, CA9RHBGJX=>networking? The id's are difficult to read.

tools/flaky-test-reporter/slack_notification.go
```diff
@@ -107,25 +102,29 @@ func (c *SlackClient) writeSlackMessage(text, channel string) error {
...
 	} else {
 		for _, testFullName := range flakyTests {
 			message += fmt.Sprintf("\n>- %s", testFullName)
-			if flakyIssues, ok := flakyIssuesMap[getIdentityForTest(testFullName, rd.Config.Repo)]; ok {
+			if flakyIssues, ok := flakyIssuesMap[getIdentityForTest(testFullName, rd.Config.Repo)]; ok && rd.Config.PostIssue {
 				for _, fi := range flakyIssues {
 					message += fmt.Sprintf("\t%s", fi.issue.GetHTMLURL())
 				}
+			} else {
+				message += fmt.Sprintf("\t Job is marked to not create GitHub issues")
```
chaodaiG: should be my last comment, I would expect this line show up only once instead of multiple times for each message. You can also combine this line with line 117 if only show this line once

tools/flaky-test-reporter/slack_notification.go
srinivashegde86: [nit] return not needed

tools/flaky-test-reporter/config.go
TrevorFarrelly: I can add more than just the two we are currently using if you'd like.

tools/flaky-test-reporter/slack_notification.go
chaodaiG: To make it clear, the Slack message for with or w/o PostIssue flag should be almost identical. The only difference here is:
- (PostIssue = true) There is a github issue link for each failed test
- (PostIssue = false) Additional message indicates "Job is marked to not create GitHub issues"

An example message for PostIssue=false in my mind looks like this:
```
As of YYYY-MM-DD, there are 2 flaky tests in 'ci-knative-serving-istio-1.0.7-mesh' in repo 'serving' (Job is marked to not create GitHub issues):
>- TestNameA
>- TestNameB
See Testgrid for up-to-date flaky tests information: [TESTGRID_URL]
```

You don't have to follow exactly the same format, the idea is we should deliver same amount of information no matter PostIssue is true or false

tools/flaky-test-reporter/slack_notification.go
```diff
@@ -107,28 +102,33 @@ func (c *SlackClient) writeSlackMessage(text, channel string) error {
...
-			// so there is only one element in flakyIssues
-			for _, fi := range flakyIssues {
-				message += fmt.Sprintf("\t%s", fi.issue.GetHTMLURL())
-			}
-		}
-	} else {
-		for _, testFullName := range flakyTests {
-			message += fmt.Sprintf("\n>- %s", testFullName)
-			if flakyIssues, ok := flakyIssuesMap[getIdentityForTest(testFullName, rd.Config.Repo)]; ok {
+	if rd.Config.PostIssue {
```
chaodaiG: To make it clear, the Slack message for with or w/o PostIssue flag should be almost identical. The only difference here is:
- (PostIssue = true) There is a github issue link for each failed test
- (PostIssue = false) Additional message indicates "Job is marked to not create GitHub issues"

An example message for PostIssue=false in my mind looks like this:
```
As of YYYY-MM-DD, there are 2 flaky tests in 'ci-knative-serving-istio-1.0.7-mesh' in repo 'serving' (Job is marked to not create GitHub issues):
>- TestNameA
>- TestNameB
See Testgrid for up-to-date flaky tests information: [TESTGRID_URL]
```

You don't have to follow exactly the same format, the idea is we should deliver same amount of information no matter PostIssue is true or false

srinivashegde86:
LGTM, but I will let @chaodaiG add the label.

tools/flaky-test-reporter/slack_notification.go
TrevorFarrelly: I understand now, thank's for the clarification. Next commit will fix this.

tools/flaky-test-reporter/slack_notification.go
TrevorFarrelly: Messages now follow a similar format to what you specify above.

tools/flaky-test-reporter/slack_notification.go
```diff
@@ -137,6 +136,19 @@ func createSlackMessageForRepo(rd *RepoData, flakyIssuesMap map[string][]*flakyI
 	return message
 }
 
+// getSlackChannel gets the channel(s) to send messages to for a given job configuration
+func getSlackChannel(repo string, name string) []slackChannel {
+	jobMap, ok := slackChannelsMap[repo]
+	if !ok {
+		log.Printf("cannot find Slack channel mapping for repo '%s', skipping Slack notification", repo)
```
chaodaiG: still need to return here, otherwise it will be a runtime error on next line
I will lgtm if this is fixed

TrevorFarrelly: my bad, should be good to go now.

chaodaiG:
/lgtm
Good job!

PR #811, opened 2019-05-21
Build Log Parser

### LINKED ISSUE #762: Add log parser to parse build logs
/assign @steuhs

steuhs:
Fetch build log from given url and check it against given error patterns.
Return all found error patterns and error messages in pairs.

Fixes #762

### DIFF
--- a/tools/monitoring/logs.go
+++ b/tools/monitoring/logs.go
@@ -1,0 +1,74 @@
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"log"
+	"regexp"
+)
+
+// ErrorInLog stores the error pattern and the corresponding error message found in the log
+type ErrorInLog struct {
+	ErrorPattern string
+	ErrorMsg     string
+}
+
+func compilePatterns(patterns []string) ([]regexp.Regexp, []string) {
+	var regexps []regexp.Regexp
+	var badPatterns []string // patterns that cannot be compiled into regex
+
+	for _, pattern := range patterns {
+		r, err := regexp.Compile(pattern)
+		if err != nil {
+			log.Printf("Error compiling pattern [%s]: %v", pattern, err)
+			badPatterns = append(badPatterns, pattern)
+		} else {
+			regexps = append(regexps, *r)
+		}
+	}
+	return regexps, badPatterns
+}
+
+func findMatches(regexps []regexp.Regexp, text []byte) []ErrorInLog {
+	var errorLogs []ErrorInLog
+	for _, r := range regexps {
+		found := r.Find(text)
+		if found != nil {
+			errorLogs = append(errorLogs, ErrorInLog{
+				ErrorPattern: r.String(),
+				ErrorMsg:     string(found),
+			})
+		}
+	}
+	return errorLogs
+}
+
+// ParseLog fetches build log from given url and checks it against given error patterns. Return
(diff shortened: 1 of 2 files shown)


tools/monitoring/logs.go
```diff
@@ -0,0 +1,74 @@
...
+
+package main
+
+import (
+	"log"
+	"regexp"
+)
+
+// ErrorInLog stores the error pattern and the corresponding error message found in the log
+type ErrorInLog struct {
```
srinivashegde86:
```suggestion
type ErrorLog struct {
```

tools/monitoring/logs.go
```diff
@@ -0,0 +1,74 @@
...
+	ErrorPattern string
```
srinivashegde86: these vars can be pattern and message. No need to add error prefix

tools/monitoring/logs.go
```diff
@@ -0,0 +1,74 @@
...
+	ErrorMsg     string
+}
+
+func compilePatterns(patterns []string) ([]regexp.Regexp, []string) {
```
srinivashegde86: can you add func comments

tools/monitoring/logs.go
```diff
@@ -0,0 +1,74 @@
...
+// ParseLog fetches build log from given url and checks it against given error patterns. Return
+// all found error patterns and error messages in pairs.
+func ParseLog(url string, patterns []string) ([]ErrorInLog, error) {
+	content, err := getFileBytes(url)
+	if err != nil {
+		return nil, err
+	}
+	regexps, badPatterns := compilePatterns(patterns)
+	if len(badPatterns) != 0 {
+		log.Printf("The following pattern cannot be compiled: %v", badPatterns)
```
srinivashegde86:
```suggestion
		log.Printf("The following patterns cannot be compiled: %v", badPatterns)
```

tools/monitoring/logs_test.go
```diff
@@ -0,0 +1,120 @@
...
+				*regexp.MustCompile("Something went wrong:.*\n"),
+			},
+			wantedBadPatterns: []string{
+				"[0",
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, got1 := compilePatterns(tt.args.patterns)
```
srinivashegde86: we should make the vars more meaningful.

```suggestion
			patterns, badPatterns := compilePatterns(tt.args.patterns)
```

tools/monitoring/logs.go
```diff
@@ -0,0 +1,74 @@
...
+		// TODO: after email feature is done, send an email to oncall about the pattern issues
```
srinivashegde86: not sure if we should an email here as there is nothing the oncaller is expected to do. We should prob just log and ignore this pattern

tools/monitoring/logs.go
```diff
@@ -0,0 +1,74 @@
...
+		}
+	}
+	return regexps, badPatterns
+}
+
+func findMatches(regexps []regexp.Regexp, text []byte) []ErrorInLog {
+	var errorLogs []ErrorInLog
+	for _, r := range regexps {
+		found := r.Find(text)
+		if found != nil {
```
srinivashegde86: Find will give you an empty byte array and not nil

tools/monitoring/logs.go
```diff
@@ -0,0 +1,74 @@
...
+			regexps = append(regexps, *r)
+		}
+	}
+	return regexps, badPatterns
+}
+
+func findMatches(regexps []regexp.Regexp, text []byte) []ErrorInLog {
+	var errorLogs []ErrorInLog
+	for _, r := range regexps {
+		found := r.Find(text)
```
srinivashegde86: rename to matchTxt? found seems like a bool when this is a byte[]

tools/monitoring/logs.go
steuhs: The oncaller should fix the wrong pattern or at least add a issue to fix. We do want to be alerted if any supplied pattern is invalid. Otherwise the associated configurations would be useless, without being noticed

tools/monitoring/logs.go
steuhs: matchTxt sounds like a bool as well. How about collectMatches

tools/monitoring/logs.go
steuhs: see
https://golang.org/pkg/regexp/#Regexp.Find

PR #813, opened 2019-05-21
flaky-test-reporter: refactor logging to better support new job/repo mapping format

TrevorFarrelly:
Better support new job format introduced in #809 when logging.
Print messages that would be sent to Slack when running with -dry-run=true.

/lint

### DIFF
--- a/tools/flaky-test-reporter/github_issue.go
+++ b/tools/flaky-test-reporter/github_issue.go
@@ -428,7 +428,7 @@ func (gi *GithubIssue) processGithubIssueForRepo(rd *RepoData, flakyIssuesMap ma
 	var errs []error
 
 	// if flag is not set, do not create an issue at all
-	if !rd.Config.PostIssue {
+	if rd.Config.SkipGithubIssue {
 		messages = append(messages, "skip creating/updating issues, job is marked to not create GitHub issues\n")
 		return messages, combineErrors(errs)
 	}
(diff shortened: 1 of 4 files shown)


chaodaiG:
Nice!
/approve
Waiting on @srinivashegde86  to lgtm

tools/flaky-test-reporter/github_issue.go
```diff
@@ -427,6 +427,12 @@ func (gi *GithubIssue) processGithubIssueForRepo(rd *RepoData, flakyIssuesMap ma
 	var messages []string
 	var errs []error
 
+	// if flag is not set, do not create an issue at all
+	if !rd.Config.PostIssue {
```
chaodaiG: It might make more sense to default Post Github Issue, how about renaming `PostIssue` with `SkipGithubIssue`?

tools/flaky-test-reporter/github_issue.go
```diff
@@ -488,8 +494,8 @@ func (gi *GithubIssue) processGithubIssueForRepo(rd *RepoData, flakyIssuesMap ma
 
 // analyze all results, figure out flaky tests and processing existing auto:flaky issues
 func (gi *GithubIssue) processGithubIssues(repoDataAll []*RepoData, dryrun bool) error {
-	messagesMap := make(map[string][]string)
-	errMap := make(map[string][]error)
+	messagesMap := make(map[string]map[string][]string)
```
srinivashegde86: I just feel like this needs to be defined in a struct. Its difficult to keep track of which map we are using and what it relates to :thinking:

TrevorFarrelly: This would be good, but @chaodaiG has ideas for a larger change that would make this outdated anyways. I can add a comment to clarify for the time being.

tools/flaky-test-reporter/github_issue.go
TrevorFarrelly: fixed

tools/flaky-test-reporter/github_issue.go
chaodaiG: That's right, I totally agree with this idea. It would be even nicer to move this logic out of this file and make a shared components that can be used by both github and slack

PR #815, opened 2019-05-21
Knative monitoring cloud mail setup

### LINKED ISSUE #757: Add cloud mail to K8S app
/assign @yt3liu

yt3liu:
* added a setup script to setup the mail domain, sender, etc.
* integrate with the kubernetes app with a url to send the test mail

Fixes #757


images/monitoring/Dockerfile
```diff
@@ -21,10 +21,27 @@ ENV TOOL_NAME monitoring
 # Temporarily add test-infra to the image to build custom tools
 ADD ./tools $TEMP_TOOLS_REPO_DIR
 
+# These go dependencies need to be included before cloud mail files to avoid stale checkout due to
```
srinivashegde86: ` cloud mail files to avoid stale checkout `?

tools/monitoring/cloudmail/mail.go
```diff
@@ -0,0 +1,204 @@
...
+	addressSet := &mailpb.AddressSet{
+		AddressPatterns: []string{addressPattern},
+	}
+	req := &mailpb.CreateAddressSetRequest{
+		Parent:       fmt.Sprintf("regions/%s/domains/%s", region, domainID),
+		AddressSetId: addressSetID,
+		AddressSet:   addressSet,
+	}
+
+	_, err := client.CreateAddressSet(ctx, req)
```
srinivashegde86:
```suggestion
	if _, err := client.CreateAddressSet(ctx, req); err != nil {
	return err
	}
```

tools/monitoring/cloudmail/mail.go
```diff
@@ -0,0 +1,204 @@
...
+	if err != nil {
+		return err
+	}
+	fmt.Printf("Address set created.\n")
```
srinivashegde86: can be combined to one print statement

tools/monitoring/cloudmail/mail.go
```diff
@@ -0,0 +1,204 @@
...
+		DefaultHeaderFromAuthority:   addressSetPath,
+	}
+	req := &mailpb.CreateSenderRequest{
+		Parent:   fmt.Sprintf("projects/%s", projectID),
+		Region:   region,
+		SenderId: senderID,
+		Sender:   sender,
+	}
+
+	_, err := client.CreateSender(ctx, req)
```
srinivashegde86: ditto

tools/monitoring/cloudmail/mail.go
```diff
@@ -0,0 +1,204 @@
...
+	if err != nil {
+		return err
+	}
+	fmt.Printf("Sender created.\n")
```
srinivashegde86: ditto

tools/monitoring/cloudmail/mail.go
```diff
@@ -0,0 +1,204 @@
...
+		EnvelopeToPatterns: []*mailpb.ReceiptRule_Pattern{envelopeToPattern},
+		Action:             action,
+	}
+	createReceiptRuleReq := &mailpb.CreateReceiptRuleRequest{
+		Parent:      fmt.Sprintf("regions/%s/domains/%s", region, domainID),
+		RuleId:      receiptRuleID,
+		ReceiptRule: receiptRule,
+	}
+
+	_, err := client.CreateReceiptRule(ctx, createReceiptRuleReq)
```
srinivashegde86: combine

tools/monitoring/cloudmail/mail.go
```diff
@@ -0,0 +1,204 @@
...
+
+	mask := &field_mask.FieldMask{
+		Paths: []string{"receipt_ruleset"},
+	}
+	updateDomainReq := &mailpb.UpdateDomainRequest{
+		Domain:     domain,
+		UpdateMask: mask,
+	}
+
+	_, updateDomainErr := client.UpdateDomain(ctx, updateDomainReq)
```
srinivashegde86: ditto

tools/monitoring/cloudmail/mail.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+	fmt.Printf("New receipt rule %s applied to domain %s.\n", receiptRuleID, domainID)
+	return nil
+}
+
+// SendTestMessage sends a test email
+func SendTestMessage(ctx context.Context, client *mail.CloudMailClient, domainName string, toAddress string) error {
+	return SendEmailMessage(ctx, client, domainName, toAddress,
+		"Knative Monitoring Cloud Mail Test",
+		"This is a test message from Cloud Mail")
```
srinivashegde86:
```suggestion
		"This is a test message.")
```

tools/monitoring/cloudmail/mail.go
```diff
@@ -0,0 +1,204 @@
...
+		HeaderFromAuthority:   "",
+		EnvelopeFromAddress:   fromAddress,
+		Message:               message,
+	}
+
+	resp, err := client.SendMessage(ctx, req)
+	if err != nil {
+		return err
+	}
+	fmt.Printf("Messsage sent.\n")
```
srinivashegde86: can be combined

tools/monitoring/cloudmail/setup/README.md
```diff
@@ -0,0 +1,49 @@
+## Cloud Mail Setup
```
srinivashegde86: missing header

tools/monitoring/cloudmail/setup/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+1. `send-test-mail`: Send a test mail from the resources created in the previous setup steps
+   ```bash
+   go run setup.go -send-test-mail -domain-name "<Domain Name generated in (1)>" -to-address "<recipient email>"
+   ```
+
+### Cloud Mail Command Line Utility
+
+For more complex usage of cloud mail, use the command line interface directly. It is required
+to sign up for [cloud mail alpha](go/cloud-mail-alpha-form) before the command line utility
```
srinivashegde86: go links dont work. Just point to the cloud.google.com/mail links and not to actual forms.

tools/monitoring/cloudmail/setup/README.md
```diff
@@ -0,0 +1,49 @@
...
+can be used.
+
+##### Useful Commands
```
srinivashegde86: we should mention that we need to enable the cloud mail api as well

tools/monitoring/cloudmail/setup/setup.go
```diff
@@ -0,0 +1,69 @@
...
+	actionSendTestMail := flag.Bool("send-test-mail", false, "Send a test message")
+
+	domainID := flag.String("domain-id", "", "Cloud Mail domain ID")
+	domainName := flag.String("domain-name", "", "The domain name used to send the test email")
+	toAddr := flag.String("to-address", "", "The test email recipient address")
+
+	flag.Parse()
+
+	ctx := context.Background()
+
```
srinivashegde86: remove blank line

tools/monitoring/cloudmail/setup/setup.go
```diff
@@ -0,0 +1,69 @@
...
+		handleError("Failed to setup receipt rule", cloudmail.CreateAndApplyReceiptRuleDrop(ctx, client, *domainID))
+	}
+
+	if *actionSendTestMail {
+		fmt.Println("Sending a Test Email")
+		handleError("Failed to send test message", cloudmail.SendTestMessage(ctx, client, *domainName, *toAddr))
+	}
+}
+
+func handleError(errFmtMsg string, err error) {
```
srinivashegde86: should this be called something else? This does not always handle error.

tools/monitoring/gke_deployment/monitoring_service.yaml
```diff
@@ -41,11 +41,23 @@ spec:
       containers:
       - name: monitoring-app
         image: gcr.io/knative-tests/test-infra/monitoring:latest
+        command: ["/monitoring"]
+        env:
+        - name: GOOGLE_APPLICATION_CREDENTIALS
```
srinivashegde86: can we use the secret from the mount without using this env var?

tools/monitoring/main.go
```diff
@@ -57,3 +78,17 @@ func hello(w http.ResponseWriter, r *http.Request) {
 	errorPatterns := yamlFile.CollectErrorPatterns()
 	fmt.Fprintf(w, "error patterns collected from yaml:%s", errorPatterns)
 }
+
+func testSendMail(w http.ResponseWriter, r *http.Request) {
+	log.Printf("Serving request: %s", r.URL.Path)
+	log.Println("Testing sending an email.")
+
+	ctx := context.Background()
+	err := cloudmail.SendTestMessage(ctx, mailClient, mailDomain, alertAddr)
```
srinivashegde86:
```
 if err := <>; err != nil {
...
}

File: images/monitoring/Dockerfile (hunk 1 above)
Comment (yt3liu) [reply images/monitoring/Dockerfile]: The exact error is as follows for not including these packages before mail: 
```
package google.golang.org/genproto/googleapis/rpc/status: /go/src/google.golang.org/genproto exists but /go/src/google.golang.org/genproto/.git does not - stale checkout?
make: *** [all] Error 1
```

Basically, because .git files are not present in zip folder for mail client, when running `go get`, it cannot fetch the version information and get this error.

Another attempt to get the comment message more understandable, how about: 
```
These go dependencies need to be added before cloud mail dependencies. This is due to `go get` uses git to check and update the dependent packages. Since .git files are not present in the mail package, the subsequent `go get` on the cloud.google.com/go and google.golang.org/genproto/... would fail. 
https://golang.org/cmd/go/#hdr-Download_and_install_packages_and_dependencies
```

File: tools/monitoring/cloudmail/mail.go (hunk 2 above)
Comment (yt3liu) [reply tools/monitoring/cloudmail/mail.go]: Done

File: tools/monitoring/cloudmail/mail.go (hunk 3 above)
Comment (yt3liu) [reply tools/monitoring/cloudmail/mail.go]: Done

File: tools/monitoring/cloudmail/mail.go (hunk 4 above)
Comment (yt3liu) [reply tools/monitoring/cloudmail/mail.go]: Done

File: tools/monitoring/cloudmail/mail.go (hunk 5 above)
Comment (yt3liu) [reply tools/monitoring/cloudmail/mail.go]: Done

File: tools/monitoring/cloudmail/mail.go (hunk 6 above)
Comment (yt3liu) [reply tools/monitoring/cloudmail/mail.go]: Done

File: tools/monitoring/cloudmail/mail.go (hunk 7 above)
Comment (yt3liu) [reply tools/monitoring/cloudmail/mail.go]: Done

File: tools/monitoring/cloudmail/mail.go (hunk 8 above)
Comment (yt3liu) [reply tools/monitoring/cloudmail/mail.go]: Done

File: tools/monitoring/cloudmail/mail.go (hunk 9 above)
Comment (yt3liu) [reply tools/monitoring/cloudmail/mail.go]: Done

File: tools/monitoring/cloudmail/setup/README.md (hunk 10 above)
Comment (yt3liu) [reply tools/monitoring/cloudmail/setup/README.md]: Done

File: tools/monitoring/cloudmail/setup/README.md (hunk 11 above)
Comment (yt3liu) [reply tools/monitoring/cloudmail/setup/README.md]: The cloud mail page says do not share the link. Removed the go link.

File: tools/monitoring/cloudmail/setup/README.md (hunk 12 above)
Comment (yt3liu) [reply tools/monitoring/cloudmail/setup/README.md]: Done

File: tools/monitoring/cloudmail/setup/setup.go (hunk 13 above)
Comment (yt3liu) [reply tools/monitoring/cloudmail/setup/setup.go]: Done

File: tools/monitoring/cloudmail/setup/setup.go (hunk 14 above)
Comment (yt3liu) [reply tools/monitoring/cloudmail/setup/setup.go]: Updated

File: tools/monitoring/gke_deployment/monitoring_service.yaml (hunk 15 above)
Comment (yt3liu) [reply tools/monitoring/gke_deployment/monitoring_service.yaml]: This is for authenticating to the cloud api. From the code we write, we actually don't do anything with this secret / environment variable, but the cloud mail code fails authentication without this. 
https://cloud.google.com/docs/authentication/getting-started#setting_the_environment_variable

File: tools/monitoring/main.go (hunk 16 above)
Comment (yt3liu) [reply tools/monitoring/main.go]: done

File: images/monitoring/Dockerfile (hunk 1 above)
Comment (srinivashegde86) [reply images/monitoring/Dockerfile]: I just wasnt sure what we mean by stale checkout :)

I am fine with adding a simple comment like `Get cloud mail dependencies first as it is needed by the mail api` or something like that. Maybe even `Add cloud mail go dependencies first.`

File: images/monitoring/Dockerfile [hunk 17]
```diff
@@ -21,10 +21,27 @@ ENV TOOL_NAME monitoring
...
+# no .git file in cloud mail
+RUN go get -u cloud.google.com/go
+RUN go get -u google.golang.org/genproto/googleapis/api/annotations
+RUN go get -u google.golang.org/genproto/googleapis/iam/v1
+RUN go get -u google.golang.org/genproto/protobuf/field_mask
+
+#TODO(yt3liu): Use the alpha released after the go library builds correctly with the latest release
```
srinivashegde86: do we have any update on this?

tools/monitoring/cloudmail/setup/README.md
```diff
@@ -0,0 +1,61 @@
+# Cloud Mail Setup
```
srinivashegde86: missing header

tools/monitoring/cloudmail/setup/README.md
```diff
@@ -0,0 +1,61 @@
...
+go get -u google.golang.org/genproto/googleapis/iam/v1
+go get -u google.golang.org/genproto/protobuf/field_mask
+```
+
+2. Download the [zip file](https://storage.googleapis.com/cloud-mail-client-libraries/cloudmail-v1alpha3-go.zip).
+Extract its contents, copy and merge the content in the zip file with the src folder of your Go workspace.
+
+## Cloud Mail Command Line Utility
+
+For more complex usage of cloud mail, use the command line interface directly. It is required
```
srinivashegde86: I dont think we need to explain usage of gcloud here. Junt link to the actual cloud mail documentation

tools/monitoring/cloudmail/setup/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+1. Install the following packages with `go get`
+```bash
+go get -u cloud.google.com/go
+go get -u google.golang.org/genproto/googleapis/api/annotations
+go get -u google.golang.org/genproto/googleapis/iam/v1
+go get -u google.golang.org/genproto/protobuf/field_mask
+```
+
+2. Download the [zip file](https://storage.googleapis.com/cloud-mail-client-libraries/cloudmail-v1alpha3-go.zip).
```
srinivashegde86: Add a link to the actual documentation of this. Dont copy it here again

images/monitoring/Dockerfile
yt3liu: Updated.

images/monitoring/Dockerfile
yt3liu: Let's use the one in knative-monitoring for now until a more stable release is out.

tools/monitoring/cloudmail/setup/README.md
yt3liu: This is a README.md file. I don't think we include header in readme file.

tools/monitoring/cloudmail/setup/README.md
yt3liu: Done

tools/monitoring/cloudmail/setup/README.md
yt3liu: Done

images/monitoring/Dockerfile
```diff
@@ -21,10 +21,29 @@ ENV TOOL_NAME monitoring
...
+#RUN wget https://storage.googleapis.com/cloud-mail-client-libraries/cloudmail-v1alpha3-go.zip
+RUN wget https://storage.googleapis.com/knative-monitoring/cloudmail-v1alpha3-go.zip
+
+RUN apt-get install zip unzip
+RUN unzip cloudmail-v1alpha3-go.zip -d cloudmail-v1alpha3-go
+RUN ls
```
srinivashegde86: why do we need to run `ls`

tools/monitoring/cloudmail/mail.go
```diff
@@ -0,0 +1,197 @@
...
+	region    = "us-central1"
+
+	addressSetID   = "monitoring-address-set"
+	addressPattern = "monitoring-alert"
+	senderID       = "monitoring-alert-sender"
+	receiptRuleID  = "monitoring-receipt-drop-rule"
+)
+
+// CreateDomain creates a new project domain for cloud main
+func CreateDomain(ctx context.Context, client *mail.CloudMailClient) error {
```
srinivashegde86: seems like we should expand mail.CloudClient with our mailClient and add these methods to that
```
type struct MailClient{
*mail.CloudMailclient
}
```

tools/monitoring/cloudmail/mail.go
```diff
@@ -0,0 +1,197 @@
...
+		Domain: domain,
+	}
+	resp, err := client.CreateDomain(ctx, req)
+	if err != nil {
+		return err
+	}
+	fmt.Printf("Domain created.\n")
+	fmt.Printf("Domain name: %s", resp.GetDomainName())
+	fmt.Printf("Domain region: %s", region)
+	fmt.Printf("Domain ID: %s", strings.Replace(resp.GetName(), fmt.Sprintf("regions/%s/domains/", region), "", 1))
```
srinivashegde86: add the stuff we need from the reposnse to the struct. we should be holding onto these and not ask it from the caller?

tools/monitoring/cloudmail/mail.go
```diff
@@ -0,0 +1,197 @@
...
+	if _, err := client.CreateSender(ctx, req); err != nil {
+		return err
+	}
+	fmt.Printf("Sender created.\nSender ID: %s\n", senderID)
+	return nil
+}
+
+// CreateAndApplyReceiptRuleDrop configures the bounce message behaviour to do nothing when an email cannot be delivered
+func CreateAndApplyReceiptRuleDrop(ctx context.Context, client *mail.CloudMailClient, domainID string) error {
+	matchMode := "PREFIX"
```
srinivashegde86: const?

tools/monitoring/main.go
```diff
@@ -33,8 +36,14 @@ const (
 	yamlURL              = "https://raw.githubusercontent.com/knative/test-infra/master/tools/monitoring/sample.yaml"
 	dbUserSecretFile     = "/secrets/cloudsql/monitoringdb/username"
 	dbPasswordSecretFile = "/secrets/cloudsql/monitoringdb/password"
+
+	// TODO(yt3liu): Replace mail domain with the one created in mail setup
+	mailDomain = "REPLACE-WITH-DOMAIN-NAME-CREATED-IN-MAIL-SETUP"
+	alertAddr  = "knative-productivity-oncall@googlegroups.com"
 )
 
+var mailClient *mail.CloudMailClient
```
srinivashegde86: point to the our struct instead

images/monitoring/Dockerfile
yt3liu: oops. forgot to remove it after checking where the file is extracted. Thanks for pointing it out. Removed.

tools/monitoring/cloudmail/mail.go
yt3liu: Great point. Updated.

tools/monitoring/cloudmail/mail.go
yt3liu: This is only called for once to create the domain. For the rest of the resource creation, it requires the domain id only; for sending the test mail, it requires the domain name. 

Updated it to print the info and return the domain name. Added a method to do the complete setup. Note: there's still value in having standalone setup-sender (without setting up the domain) since we only have one mail domain per project, but can have multiple senders.

tools/monitoring/cloudmail/mail.go
yt3liu: updated.

tools/monitoring/main.go
yt3liu: done

tools/monitoring/cloudmail/mail.go
```diff
@@ -0,0 +1,221 @@
...
+
+	addressSetID   = "monitoring-address-set"
+	addressPattern = "monitoring-alert"
+	senderID       = "monitoring-alert-sender"
+	receiptRuleID  = "monitoring-receipt-drop-rule"
+)
+
+// MailClient holds instance of CloudMailClient for making mail-related requests
+type MailClient struct {
+	*mail.CloudMailClient
```
srinivashegde86: can we have domainID here as a var instead of returning it? The caller should not be responsible for handling that.

tools/monitoring/cloudmail/mail.go
```diff
@@ -0,0 +1,221 @@
...
+}
+
+// NewMailClient creates a new MailClient to handle all the mail interaction
+func NewMailClient(ctx context.Context) (*MailClient, error) {
+	var err error
+
+	client := &MailClient{}
+	if client.CloudMailClient, err = mail.NewCloudMailClient(ctx); err != nil {
```
srinivashegde86:
```
c, err := mail.NewCloudMailClient(ctx)
return &MailClient{CloudMailClient: c}, err
```

tools/monitoring/cloudmail/mail.go
```diff
@@ -0,0 +1,221 @@
...
+		return nil, err
+	}
+
+	return client, nil
+}
+
+// CreateDomain creates a new project domain for cloud main
+func (c *MailClient) CreateDomain(ctx context.Context) (string, error) {
+	var domainID string
```
srinivashegde86: add to struct

tools/monitoring/cloudmail/mail.go
```diff
@@ -0,0 +1,221 @@
...
+
+	if _, updateDomainErr := cloudMailClient.UpdateDomain(ctx, updateDomainReq); updateDomainErr != nil {
+		return updateDomainErr
+	}
+	fmt.Printf("New receipt rule %s applied to domain %s.\n", receiptRuleID, domainID)
+	return nil
+}
+
+// SendTestMessage sends a test email
+func (c *MailClient) SendTestMessage(ctx context.Context, domainName string, toAddress string) error {
```
srinivashegde86: add domainName to struct as well

tools/monitoring/cloudmail/mail.go
yt3liu: Done

tools/monitoring/cloudmail/mail.go
yt3liu: Done

tools/monitoring/cloudmail/mail.go
yt3liu: Done

tools/monitoring/cloudmail/mail.go
yt3liu: Done

hack/update-deps.sh
```diff
@@ -27,3 +27,15 @@ dep ensure
 
 rm -rf $(find vendor/ -name 'OWNERS')
 rm -rf $(find vendor/ -name '*_test.go')
+
```
adrcunha: This should come before the `rm`s above.

images/monitoring/Makefile
```diff
@@ -21,10 +21,14 @@ build:
 	docker build -t $(IMG):$(TAG) -f Dockerfile ../..
 	docker tag $(IMG):$(TAG) $(IMG):latest
 
+clean_build:
```
adrcunha: The usual practice is to have a "clean" rule that removes cached stuff, then you do `make clean && make build`. You can have a shortcut rule "build_clean" that depends on both, of course. But ideally not a duplicate build rule that can get out of sync easily.

tools/monitoring/cloudmail/setup/setup.go
```diff
@@ -0,0 +1,81 @@
...
+	actionSetupAll := flag.Bool("setup-all", false, "Setup up the domain, address set, sender, receipt rules with default settings")
+	actionSendTestMail := flag.Bool("send-test-mail", false, "Send a test message")
+
+	toAddr := flag.String("to-address", "", "The test email recipient address")
+
+	flag.Parse()
+
+	ctx := context.Background()
+	client, err := cloudmail.NewMailClient(ctx)
+	if err != nil {
```
adrcunha: You don't need the `if` if `failIfError` does nothing when `err` is `nil`.

adrcunha:
I reviewed the non-go stuff and added a few comments; I'll let Srinivas LGTM the go code.

tools/monitoring/cloudmail/setup/setup.go
yt3liu: Removed the if statement

images/monitoring/Makefile
yt3liu: Thanks for pointing this out. I removed the build_clean rule. 

I was using the build_clean rule for debugging the dependencies. Because I was using `ls` to check the mail dependencies are added to the correct directory. I want it to always print out the directory structure. I don't think the clean rule will be need in the future.

hack/update-deps.sh
yt3liu: Done

tools/monitoring/cloudmail/mail.go
```diff
@@ -0,0 +1,222 @@
...
+
+	addressSetID   = "monitoring-address-set"
+	addressPattern = "monitoring-alert"
+	senderID       = "monitoring-alert-sender"
+	receiptRuleID  = "monitoring-receipt-drop-rule"
+)
+
+// MailClient holds instance of CloudMailClient for making mail-related requests
+type MailClient struct {
+	CloudMailClient *mail.CloudMailClient
```
srinivashegde86: can we create an embedded type instead of a new field?

```
type MailClient struct {
	*mail.CloudMailClient
	DomainName      string
	DomainID        string
}
```

you can set as per https://stackoverflow.com/questions/27626765/mixture-of-fieldvalue-and-value-initializers

tools/monitoring/cloudmail/mail.go
```diff
@@ -0,0 +1,222 @@
...
+	DomainName      string
+	DomainID        string
+}
+
+// NewMailClient creates a new MailClient to handle all the mail interaction
+func NewMailClient(ctx context.Context) (*MailClient, error) {
+	c, err := mail.NewCloudMailClient(ctx)
+	return &MailClient{
```
srinivashegde86: Use https://stackoverflow.com/questions/27626765/mixture-of-fieldvalue-and-value-initializers to check how to init

tools/monitoring/cloudmail/mail.go
```diff
@@ -0,0 +1,222 @@
...
+	domain := &mailpb.Domain{
+		ProjectDomain: true,
+		DomainName:    "",
+	}
+	req := &mailpb.CreateDomainRequest{
+		Parent: fmt.Sprintf("projects/%s", projectID),
+		Region: region,
+		Domain: domain,
+	}
+	resp, err := c.CloudMailClient.CreateDomain(ctx, req)
```
srinivashegde86: We can call `c.CreateDomain()` directly with the change

tools/monitoring/cloudmail/mail.go
srinivashegde86: Also, do we need to export domainname and domainid or is it only for this wrapper?

tools/monitoring/cloudmail/mail.go
```diff
@@ -0,0 +1,222 @@
...
+	if _, err := c.CloudMailClient.CreateSender(ctx, req); err != nil {
+		return err
+	}
+	fmt.Printf("Sender created.\nSender ID: %s\n", senderID)
+	return nil
+}
+
+// CreateAndApplyReceiptRuleDrop configures the bounce message behaviour to do nothing when an email cannot be delivered
+func (c *MailClient) CreateAndApplyReceiptRuleDrop(ctx context.Context) error {
+	const matchMode = "PREFIX"
```
srinivashegde86: move it out of functions?

tools/monitoring/cloudmail/mail.go
```diff
@@ -0,0 +1,222 @@
...
+
+	cloudMailClient := c.CloudMailClient
```
srinivashegde86: var not needed

tools/monitoring/cloudmail/mail.go
```diff
@@ -0,0 +1,222 @@
...
+		EnvelopeToPatterns: []*mailpb.ReceiptRule_Pattern{envelopeToPattern},
+		Action:             action,
+	}
+	createReceiptRuleReq := &mailpb.CreateReceiptRuleRequest{
+		Parent:      fmt.Sprintf("regions/%s/domains/%s", region, c.DomainID),
+		RuleId:      receiptRuleID,
+		ReceiptRule: receiptRule,
+	}
+
+	if _, err := cloudMailClient.CreateReceiptRule(ctx, createReceiptRuleReq); err != nil {
```
srinivashegde86: should be able to use `c.CreateReceiptRule()`

tools/monitoring/cloudmail/mail.go
```diff
@@ -0,0 +1,222 @@
...
+
+	mask := &field_mask.FieldMask{
+		Paths: []string{"receipt_ruleset"},
+	}
+	updateDomainReq := &mailpb.UpdateDomainRequest{
+		Domain:     domain,
+		UpdateMask: mask,
+	}
+
+	if _, updateDomainErr := cloudMailClient.UpdateDomain(ctx, updateDomainReq); updateDomainErr != nil {
```
srinivashegde86:
```suggestion
	if _, err := cloudMailClient.UpdateDomain(ctx, updateDomainReq); err != nil {
```

tools/monitoring/cloudmail/mail.go
```diff
@@ -0,0 +1,222 @@
...
+		// sender. There must be an envelope from authority and a header from
+		// authority that Cloud Mail can use; otherwise the system will
+		// return an error.
+		EnvelopeFromAuthority: "",
+		HeaderFromAuthority:   "",
+		EnvelopeFromAddress:   fromAddress,
+		Message:               message,
+	}
+
+	resp, err := c.CloudMailClient.SendMessage(ctx, req)
```
srinivashegde86: `c.SendMessage()`

tools/monitoring/cloudmail/mail.go
yt3liu: It is only for this wrapper. I thought of using constants and set these fields at the top, but that would prevent us from having the one step setup that creates both the project domain as well as the sender. 

However, I'm not sure if the one step setup is valuable enough to add those fields in the struct though. The reason is that we can only set up one project domain per gcp project.So that function will only ever be run once. To test that flow works, I actually created a new GCP project to test it. Furthermore, I don't think we will use a custom domain with the current mail client in the future. Since if we have a custom email domain and can already authenticate to that, we can use [golang smtp](https://golang.org/pkg/net/smtp/) directly instead of going through cloud mail.

tools/monitoring/cloudmail/mail.go
yt3liu: Updated to use embedded type.

tools/monitoring/cloudmail/mail.go
yt3liu: Done

tools/monitoring/cloudmail/mail.go
yt3liu: Done

tools/monitoring/cloudmail/mail.go
yt3liu: Removed

tools/monitoring/cloudmail/mail.go
yt3liu: Done

tools/monitoring/cloudmail/mail.go
yt3liu: Done

tools/monitoring/cloudmail/mail.go
yt3liu: Done

yt3liu:
This PR is not longer need. https://github.com/knative/test-infra/pull/865 has the same functionality and is merged now.

PR #823, opened 2019-05-23
Add a more flexible flaky reporting threshold for jobs with less tests

TrevorFarrelly:
Previously, if 1% of tests were found flaky only 1 issue was created and Slack message didn't list the tests which failed. This makes sense for serving-continuous, but most other testing jobs have a much smaller number of tests, triggering this threshold after only one or two tests fail.

Now, the threshold is either 1% or 5 tests failing, whichever comes first.

/lint

### DIFF
--- a/tools/flaky-test-reporter/README.md
+++ b/tools/flaky-test-reporter/README.md
@@ -42,7 +42,7 @@ Efficient deduplication is crucial for the sustainability of this tool, and this
 
 #### Too many flaky tests identified
 
-When there are too many tests found to be flaky, most likely something abnormal is going on, and we don't want to create Github issues for all of them, or list all of them in Slack notifications. There is a threshold defined in [`config.go`](config.go), if the flaky rate went over the threshold there will be only 1 Github issue created, and Slack notification will not list all flaky tests.
+When there are too many tests found to be flaky, most likely something abnormal is going on, and we don't want to create Github issues for all of them, or list all of them in Slack notifications. There are thresholds defined in [`config.go`](config.go), if the flaky rate went over a threshold there will be only 1 Github issue created, and Slack notification will not list all flaky tests.
 
 #### Github issue updates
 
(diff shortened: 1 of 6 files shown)


tools/flaky-test-reporter/README.md
```diff
@@ -42,7 +42,7 @@ Efficient deduplication is crucial for the sustainability of this tool, and this
 
 #### Too many flaky tests identified
 
-When there are too many tests found to be flaky, most likely something abnormal is going on, and we don't want to create Github issues for all of them, or list all of them in Slack notifications. There is a threshold defined in [`config.go`](config.go), if the flaky rate went over the threshold there will be only 1 Github issue created, and Slack notification will not list all flaky tests.
+When there are too many tests found to be flaky, most likely something abnormal is going on, and we don't want to create Github issues for all of them, or list all of them in Slack notifications. There are thresholds defined in [`config.go`](config.go), if the flaky rate went over a threshold there will be only 1 Github issue created, and Slack notification will not list all flaky tests.
```
srinivashegde86: Ideally, this should be part of the config defined by the repo+job :) + @chaodaiG

tools/flaky-test-reporter/result.go
```diff
@@ -113,6 +113,19 @@ func getFlakyRate(rd *RepoData) float32 {
 	return float32(len(getFlakyTests(rd))) / float32(totalCount)
 }
 
+func flakyRateAboveThreshold(rd *RepoData) bool {
+	// if the percent determined by the test count threshold is higher than
+	// the percent threshold, use that instead of the percent threshold
+	var threshold float32
```
srinivashegde86: you dont need the else.
```
	threshold := float32(countThreshold) / float32(totalCount)
```

TrevorFarrelly: the main reason for doing this was to ensure totalCount wasn't zero, though that _shouldn't_ happen (no test jobs should have zero tests). I do agree this would be cleaner.

srinivashegde86: we can have a 
```
// Num of tests = 0 => no tests are flaky :)
if (totalCount == 0) {
return true
}

threshold := float32(countThreshold) / float32(totalCount)
if ...

Issue #829, opened 2019-05-27
Error happens occasionally when installing the nightly serving release


Fredy-Z:
Not sure if it's an infra issue. Add it here now for record.

Detailed log: https://gubernator.knative.dev/build/knative-prow/pr-logs/pull/knative_test-infra/828/pull-knative-test-infra-integration-tests/1133108957661892608/
```
I0527 20:38:13.477] Installing Serving from https://storage.googleapis.com/knative-nightly/serving/latest/serving.yaml
I0527 20:38:13.937] namespace/knative-serving created
I0527 20:38:13.958] clusterrole.rbac.authorization.k8s.io/knative-serving-certmanager created
I0527 20:38:13.977] clusterrole.rbac.authorization.k8s.io/knative-serving-istio created
I0527 20:38:14.001] clusterrole.rbac.authorization.k8s.io/knative-serving-admin created
I0527 20:38:14.025] clusterrole.rbac.authorization.k8s.io/knative-serving-core created
I0527 20:38:14.082] serviceaccount/controller created
I0527 20:38:14.103] clusterrolebinding.rbac.authorization.k8s.io/knative-serving-controller-admin created
I0527 20:38:14.347] customresourcedefinition.apiextensions.k8s.io/certificates.networking.internal.knative.dev created
I0527 20:38:14.377] customresourcedefinition.apiextensions.k8s.io/clusteringresses.networking.internal.knative.dev created
I0527 20:38:14.407] customresourcedefinition.apiextensions.k8s.io/configurations.serving.knative.dev created
I0527 20:38:14.442] customresourcedefinition.apiextensions.k8s.io/images.caching.internal.knative.dev created
I0527 20:38:14.480] customresourcedefinition.apiextensions.k8s.io/podautoscalers.autoscaling.internal.knative.dev created
I0527 20:38:14.516] customresourcedefinition.apiextensions.k8s.io/revisions.serving.knative.dev created
I0527 20:38:14.545] customresourcedefinition.apiextensions.k8s.io/routes.serving.knative.dev created
I0527 20:38:14.578] customresourcedefinition.apiextensions.k8s.io/services.serving.knative.dev created
I0527 20:38:14.615] customresourcedefinition.apiextensions.k8s.io/serverlessservices.networking.internal.knative.dev created
I0527 20:38:14.659] service/activator-service created
I0527 20:38:14.690] service/controller created
I0527 20:38:14.728] service/webhook created
I0527 20:38:14.771] deployment.apps/activator created
[...]

PR #834, opened 2019-05-28
Fixes flaky knative serving install

yt3liu:
Fixes flaky knative serving install
* wait on istio components to be installed before installing knative-serving
* retry the knative-serving install if it fails

Part of #829

### DIFF
--- a/scripts/e2e-tests.sh
+++ b/scripts/e2e-tests.sh
@@ -314,6 +314,7 @@ function setup_test_cluster() {
   set +o pipefail
 
   if (( ! SKIP_KNATIVE_SETUP )) && function_exists knative_setup; then
+    echo "Waiting for Istio installation to complete before starting knative_setup"
     (( ! SKIP_ISTIO_ADDON )) && (wait_until_batch_job_complete istio-system || return 1)
     knative_setup || fail_test "Knative setup failed"
   fi
--- a/scripts/e2e-tests.sh
+++ b/scripts/e2e-tests.sh
@@ -279,7 +279,7 @@ function setup_test_cluster() {
   header "Setting up test cluster"
 
   # Set the actual project the test cluster resides in
-  # It will be a project assigned by Boskos if test is running on Prow, 
+  # It will be a project assigned by Boskos if test is running on Prow,
   # otherwise will be ${GCP_PROJECT} set up by user.
   readonly export E2E_PROJECT_ID="$(gcloud config get-value project)"
 
@@ -314,6 +314,7 @@ function setup_test_cluster() {
   set +o pipefail
 
   if (( ! SKIP_KNATIVE_SETUP )) && function_exists knative_setup; then
+    (( ! SKIP_ISTIO_ADDON )) && (wait_until_batch_job_complete istio-system || return 1)
     knative_setup || fail_test "Knative setup failed"
   fi
   if function_exists test_setup; then
--- a/scripts/e2e-tests.sh
+++ b/scripts/e2e-tests.sh
@@ -314,7 +314,7 @@ function setup_test_cluster() {
   set +o pipefail
 
   if (( ! SKIP_KNATIVE_SETUP )) && function_exists knative_setup; then
-    echo "Waiting for Istio installation to complete before starting knative_setup"
+    # Wait for Istio installation to complete, if necessary, before calling knative_setup.
     (( ! SKIP_ISTIO_ADDON )) && (wait_until_batch_job_complete istio-system || return 1)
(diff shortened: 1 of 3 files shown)


scripts/library.sh
```diff
@@ -332,10 +332,39 @@ function start_latest_knative_serving() {
   header "Starting Knative Serving"
   subheader "Installing Knative Serving"
   echo "Installing Serving from ${KNATIVE_SERVING_RELEASE}"
-  kubectl apply -f ${KNATIVE_SERVING_RELEASE} || return 1
+  # Some CRDs defined in serving YAML are also referenced by other components in serving. As it takes
+  # time for CRDs to become effective, there is a race condition between when the CRDs are effective
+  # and when the resources that references those CRDs are created.
+  # The current workaround is to re-apply serving.yaml if it fails. Remove the retry logic after the
+  # race condition is fixed.
```
adrcunha: Please link to an open issue for the problem.

scripts/library.sh
```diff
@@ -332,10 +332,39 @@ function start_latest_knative_serving() {
...
+  execute_with_retries 30 3 kubectl apply -f ${KNATIVE_SERVING_RELEASE} || return 1
```
adrcunha: This is a workaround. I think it would be simpler and more readable if you inline the 3 retries instead of writing a function for that. Also, can we make it work with a single retry?

scripts/library.sh
yt3liu: Done. Open https://github.com/knative/serving/issues/4176

scripts/library.sh
yt3liu: Updated it to use single retry. The single retry should be sufficient as long as there's enough time between the two install. Change the wait to sleep 1 minute for now and will increase it if it's still flaky.

scripts/e2e-tests.sh
```diff
@@ -314,6 +314,7 @@ function setup_test_cluster() {
   set +o pipefail
 
   if (( ! SKIP_KNATIVE_SETUP )) && function_exists knative_setup; then
+    (( ! SKIP_ISTIO_ADDON )) && (wait_until_batch_job_complete istio-system || return 1)
```
adrcunha: A comment here would be nice.

scripts/library.sh
```diff
@@ -332,7 +332,15 @@ function start_latest_knative_serving() {
   header "Starting Knative Serving"
   subheader "Installing Knative Serving"
   echo "Installing Serving from ${KNATIVE_SERVING_RELEASE}"
-  kubectl apply -f ${KNATIVE_SERVING_RELEASE} || return 1
+  # Some CRDs defined in serving YAML are also referenced by other components in serving. As it takes
+  # time for CRDs to become effective, there is a race condition between when the CRDs are effective
+  # and when the resources that references those CRDs are created.
+  # The current workaround is to re-apply serving.yaml if it fails. Remove the retry logic after the
+  # race condition is fixed. (https://github.com/knative/serving/issues/4176)
+  if [[ `kubectl apply -f ${KNATIVE_SERVING_RELEASE}` ]]; then
```
```suggestion
  if ! kubectl apply -f ${KNATIVE_SERVING_RELEASE}; then
```

scripts/library.sh
```diff
@@ -332,7 +332,15 @@ function start_latest_knative_serving() {
...
+    sleep 60
```
adrcunha: I would also add a

```
echo "Install failed, waiting 60s and then retrying..."
```

here so it's clear in the logs what's going on.

scripts/e2e-tests.sh
yt3liu: Done.

scripts/library.sh
yt3liu: Done. this made the code much more readable :)

scripts/library.sh
yt3liu: Thanks. Missed the log statement after moving the retry out of the function.

scripts/e2e-tests.sh
```diff
@@ -314,6 +314,8 @@ function setup_test_cluster() {
   set +o pipefail
 
   if (( ! SKIP_KNATIVE_SETUP )) && function_exists knative_setup; then
+    echo "Waiting for Istio installation to complete before starting knative_setup"
```
```suggestion
    # Wait for Istio installation to complete, if necessary, before calling knative_setup.
```

Why?
1. Because we don't wait if Istio add-on is not installed.
1. wait_until_batch_job_complete() is verbose enough.

yt3liu: Thanks for the explanation :)

Issue #835, opened 2019-05-29
Create database tables specified in monitoring design doc


steuhs:
This need to be completed before #764 can be verified 

Step 1:
Create an sql file that contains the table creation statements

Step 2:
execute the file created in step 1 and load the table in monitoring database

PR #836, opened 2019-05-29
Create the tables specified in monitoring design doc

steuhs:
Step 1 of #835 
TODO: execute this file and load the table in monitoring database

### DIFF
--- a/tools/monitoring/mysql/tables.sql
+++ b/tools/monitoring/mysql/tables.sql
@@ -22,7 +22,7 @@ CREATE TABLE ErrorLogs
   ID           int           NOT NULL AUTO_INCREMENT,
   ErrorPattern varchar(4095) NOT NULL,
   ErrorMsg     varchar(4095) NOT NULL,
-  Jobname      varchar(1023) NOT NULL, /*e.g. pull-knative-serving-integration-tests*/
+  JobName      varchar(1023) NOT NULL, /*e.g. pull-knative-serving-integration-tests*/
   PRNumber     int, /*pull request number; null for non pull jobs*/
   BuildLogURL  varchar(1023) NOT NULL, /*link to build-log.txt file*/
   TimeStamp    timestamp, /* stamps the time the record is added*/
--- a/tools/monitoring/mysql/tables.sql
+++ b/tools/monitoring/mysql/tables.sql
@@ -31,6 +31,8 @@ CREATE TABLE ErrorLogs
 
 CREATE TABLE Alerts
 (
+  ID           int           NOT NULL AUTO_INCREMENT,
   ErrorPattern varchar(4095) NOT NULL UNIQUE,
-  Sent         timestamp
+  Sent         timestamp,
+  PRIMARY KEY (ID)
 )
--- a/tools/monitoring/mysql/tables.sql
+++ b/tools/monitoring/mysql/tables.sql
@@ -1,0 +1,36 @@
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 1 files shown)


tools/monitoring/mysql/tables.sql
```diff
@@ -0,0 +1,36 @@
...
+  ErrorPattern varchar(4095) NOT NULL,
+  ErrorMsg     varchar(4095) NOT NULL,
+  Jobname      varchar(1023) NOT NULL, /*e.g. pull-knative-serving-integration-tests*/
+  PRNumber     int, /*pull request number; null for non pull jobs*/
+  BuildLogURL  varchar(1023) NOT NULL, /*link to build-log.txt file*/
+  TimeStamp    timestamp, /* stamps the time the record is added*/
+  PRIMARY KEY (ID)
+);
+
+CREATE TABLE Alerts
```
yt3liu: Do we want a primary key for the alerts table? Maybe make the ErrorPattern the primary key so it's automatically indexed?

srinivashegde86: yes. there should be an id here as well.

tools/monitoring/mysql/tables.sql
```diff
@@ -0,0 +1,36 @@
...
+
+# These tables only need to be created once,
+# and they should be created before any run of the application
+
+CREATE TABLE ErrorLogs
+(
+  ID           int           NOT NULL AUTO_INCREMENT,
+  ErrorPattern varchar(4095) NOT NULL,
+  ErrorMsg     varchar(4095) NOT NULL,
+  Jobname      varchar(1023) NOT NULL, /*e.g. pull-knative-serving-integration-tests*/
```
srinivashegde86:
```suggestion
  JobName      varchar(1023) NOT NULL, /*e.g. pull-knative-serving-integration-tests*/
```

tools/monitoring/mysql/tables.sql
```diff
@@ -0,0 +1,36 @@
...
+  PRNumber     int, /*pull request number; null for non pull jobs*/
```
srinivashegde86:
```suggestion
  PRNumber     int, /*pull request number; 0 for non pull jobs*/
```

tools/monitoring/mysql/tables.sql
steuhs: We cannot make error pattern the primary key - it is not unique for each record

tools/monitoring/mysql/tables.sql
steuhs: I think null makes more sense, by default mysql allows this to be nullable. Null makes it a lot easier to query than examining whether the field has a value of zero

steuhs: checking null should be optimized in sql, and is supposed to be more efficient than checking a field of specific value

tools/monitoring/mysql/tables.sql
srinivashegde86: I was suggesting to have another column ID here and make it the primary key.

yt3liu: +1 on adding column id and make it the primary key. 

If the error pattern is not unique, remove `UNIQUE` from the column definition? And what makes the error pattern not unique in the alerts table?

steuhs: Sorry, I thought you are talking about the other table, because I see "PRIMARY KEY (ID)" quoted.

Yes, in the Alert table the pattern is unique and I think we should make it a primary key. In this way it is easier to index. @srinivashegde86 I am not sure if you made the same assumption as me. In this case I think what Joyce said made sense.

yt3liu: I think there's a restriction on the length of the primary key, I don't remember the exact limit though. varchar(4095) *may* be too large to make it the primary key

steuhs: Let's add an auto id here then. In practice this table shouldn't have too many records and we probably don't need indexing

PR #837, opened 2019-05-29
[monitoring] update db record after parsing config yaml and build log

### LINKED ISSUE #764: Store error patterns in the error logs table
/assign @steuhs

steuhs:
1. update db record after parsing config yaml and build log
2. moved config and build log parsing to their own packages so that the sql package can import them

Fixes #764


tools/monitoring/mysql/db_mysql.go
```diff
@@ -71,3 +75,42 @@ func (c DBConfig) dataStoreName(dbName string) string {
...
+	if err != nil {
+		return err
+	}
+
+	tx, err := db.Begin()
+	if err != nil {
+		return err
+	}
+	defer tx.Rollback()
+	stmt, err := tx.Prepare("INSERT INTO ErrorLogs VALUES()")
```
yt3liu: Could you specify the column names in the insert statement? This makes insert statement still valid in case we want to add new columns in the future.

tools/monitoring/mysql/db_mysql.go
```diff
@@ -71,3 +75,42 @@ func (c DBConfig) dataStoreName(dbName string) string {
...
+	logs, err := logs.ParseLog(buildLogURL, config.CollectErrorPatterns())
+	if err != nil {
+		return err
+	}
+
+	tx, err := db.Begin()
+	if err != nil {
+		return err
+	}
+	defer tx.Rollback()
```
yt3liu: Does this line always roll back the transaction?

steuhs: my understanding is that it rolls back transactions that are not committed. I can make it more explicit however

tools/monitoring/logs/logs.go
```diff
@@ -14,11 +14,13 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package main
+package logs
```
srinivashegde86: rename this to log_parser? logs seems like it refers to logging.

tools/monitoring/mysql/db_mysql.go
```diff
@@ -71,3 +75,42 @@ func (c DBConfig) dataStoreName(dbName string) string {
...
+
+// PubsubMsgHandler adds record(s) to ErrorLogs table in database,
+// after parsing build log and compares the result with config yaml
+func PubsubMsgHandler(db *sql.DB, configURL, buildLogURL, jobname string, prNumber int) error {
+	config, err := config.ParseYaml(configURL)
+	if err != nil {
+		return err
+	}
+
+	logs, err := logs.ParseLog(buildLogURL, config.CollectErrorPatterns())
```
srinivashegde86:
```suggestion
	pattern, err := logs.ParseLog(buildLogURL, config.CollectErrorPatterns())
```

tools/monitoring/mysql/db_mysql.go
```diff
@@ -71,3 +75,42 @@ func (c DBConfig) dataStoreName(dbName string) string {
...
+	stmt, err := tx.Prepare("INSERT INTO ErrorLogs VALUES()")
+
+	for _, errorLog := range logs {
+		_, err := stmt.Exec(errorLog.Pattern, errorLog.Msg, jobname, prNumber, buildLogURL, time.Now())
+		if err != nil {
+			return err
+		}
+	}
+
+	err = stmt.Close()
```
srinivashegde86: combine the if condition

if err = stmt.Close(); err != nil {
return err
}

tools/monitoring/mysql/db_mysql.go
```diff
@@ -71,3 +75,42 @@ func (c DBConfig) dataStoreName(dbName string) string {
...
+	if err != nil {
+		return err
+	}
+
+	err = tx.Commit()
```
srinivashegde86: return tx.Commit()?

tools/monitoring/mysql/db_mysql.go
steuhs: this is a list of ErrorLog structs, not just patterns.

tools/monitoring/mysql/db_mysql.go
steuhs: This line is too long to be combined.

tools/monitoring/mysql/db_mysql.go
steuhs: if we specify column names, then a subtle change in column name would fail the statement. I think there is drawbacks either ways

steuhs:
@srinivashegde86 @yt3liu addressed all comments. Please re-review

tools/monitoring/mysql/db_mysql.go
yt3liu: Could you elaborate what you mean by the line is too long to be combined? Did you get a golint error or something similar?

tools/monitoring/config/config_test.go
```diff
@@ -256,6 +257,49 @@ func getConfigYaml() []byte {
 	return text
 }
 
+func Test_compilePatterns(t *testing.T) {
```
srinivashegde86:
```suggestion
func TestCompilePatterns(t *testing.T) {
```

tools/monitoring/mysql/db_mysql.go
srinivashegde86: If we change the order of columns or add new ones, we will fail here, which is not desirable. If we are changing the column name, then the query should change IMO and you will be responsible to make sure that you handle cases for the old name.

tools/monitoring/mysql/db_mysql.go
srinivashegde86: should this be `defer stmt.Close()`

Issue #689, opened 2019-04-15
Automate Prow components versions bumping


chaodaiG:
Current process of updating Prow components versions is manual, which is time consuming and error prone. Automating the versions bumping can make this process more resilient and we always get latest features and fixes. 
We might be able to borrow from automating script in Kubernetes/test-infra repo:
https://github.com/kubernetes/test-infra/blob/0f376b4ea70b4d3d8f51c622158a623fee2e9744/prow/autobump.sh

adrcunha:
Synced with the k8s team about this. We should only automatically bump our components version up to the one used by prow.k8s.io, which is safe.

dushyanthsc:
We would be taking a look at this in Q3 (Jul - Sept)

adrcunha:
Note: autobump.sh was replaced by experiment/autobumper. See https://github.com/kubernetes/test-infra/pull/12471

chaodaiG:
Kubernetes is updating Prow version almost daily now: https://github.com/kubernetes/test-infra/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aclosed+label%3Aarea%2Fprow%2Fbump+

So if we choose to use the same version as Kubernetes, it's not too much different from using latest Prow version. An idea of getting a tested version is going through kubernetes/test-infra PR history, and pick a PR  with `area/prow/bump` label that is roughly one week old,  merged and was not reverted later

adrcunha:
Fejta mentioned something about test-infra updating frequently and being the "staging" version, while k8s would update less frequently, on stable versions. Maybe that's what he was referring to? We should get confirmation from the k8s team before proceeding.

chaodaiG:
Just checked with @cjwagner , the entire kubernetes org is managed by a single Prow instance.

adrcunha:
Thanks. +1 for the ideas of using a slightly older, not rolled-back version.

chaodaiG:
This is done after #1057 merged
/close

PR #838, opened 2019-05-29
Split shared/ghutil based on functionalities, add pullrequest support

chaodaiG:
Splitting shared/ghutil package into several files based on functionalities. Also add pullrequest related generic helper functions, which will be consumed by Prow version auto bump tool

Partof: #689


chaodaiG:
Note for code reviewers: `ghutil.go` was splitted into `client.go` and `issue.go`, with no code change. The only thing new in this PR is `pullrequest.go`

adrcunha:
/approve

I'll let Srinivas LGTM.

shared/ghutil/client.go
```diff
@@ -0,0 +1,148 @@
...
+	maxRetryCount = 5
+	tokenReserve  = 50
+)
+
+var (
+	ctx = context.Background()
+)
+
+// GithubClientInterface contains a set of functions for Github operations
+type GithubClientInterface interface {
```
srinivashegde86: rename this to GithubOperations? I dont think we need to add interface siffux

shared/ghutil/pullrequest.go
```diff
@@ -0,0 +1,148 @@
...
+	// PullRequestOpenState is the state of open PullRequest
+	PullRequestOpenState PullRequestStateEnum = "open"
+	// PullRequestCloseState is the state of closed PullRequest
+	PullRequestCloseState PullRequestStateEnum = "closed"
+	// PullRequestAllState is the state for all, useful when querying PullRequest
+	PullRequestAllState PullRequestStateEnum = "all"
+)
+
+// PullRequestStateEnum represents different states of PullRequest
+type PullRequestStateEnum string
```
srinivashegde86:
```suggestion
type PullRequestState string
```

shared/ghutil/pullrequest.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/google/go-github/github"
+)
+
+const (
+	// PullRequestOpenState is the state of open PullRequest
+	PullRequestOpenState PullRequestStateEnum = "open"
+	// PullRequestCloseState is the state of closed PullRequest
+	PullRequestCloseState PullRequestStateEnum = "closed"
+	// PullRequestAllState is the state for all, useful when querying PullRequest
+	PullRequestAllState PullRequestStateEnum = "all"
```
srinivashegde86: this seems weird as a pr state. What does it mean to have pr with all state? Maybe replace with any?

shared/ghutil/pullrequest.go
```diff
@@ -0,0 +1,148 @@
...
+)
+
+// PullRequestStateEnum represents different states of PullRequest
+type PullRequestStateEnum string
+
+// ListPullRequests lists pull requests within given repo, filters by head user and branch name if
+// provided as "user:ref-name", and by base name if provided, i.e. "master"
+func (gc *GithubClient) ListPullRequests(org, repo, head, base string) ([]*github.PullRequest, error) {
+	PRsListOptions := github.PullRequestListOptions{
+		State: "all",
```
srinivashegde86: use enum for all

shared/ghutil/pullrequest.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+	if "" != head {
```
srinivashegde86: what happens if I dont check these and do

options := PRLIstOptions { Head: head, Base: base, State: "all"}

These should be empty by default?

shared/ghutil/pullrequest.go
srinivashegde86: keep this all if GH expects all though.

chaodaiG: yep, GH only accepts `open`, `closed`, or `all`

shared/ghutil/client.go
chaodaiG: done

shared/ghutil/pullrequest.go
chaodaiG: done

shared/ghutil/pullrequest.go
chaodaiG: good catch, done

shared/ghutil/pullrequest.go
chaodaiG: yes that's right, done

shared/ghutil/pullrequest.go
```diff
@@ -0,0 +1,144 @@
...
+
+import (
+	"fmt"
+
+	"github.com/google/go-github/github"
+)
+
+const (
+	// PullRequestOpenState is the state of open PullRequest
+	PullRequestOpenState PullRequestState = "open"
```
srinivashegde86: [nit] can we rename this to open, close and all? the type indicates that it is a PRState.

I like concise var names :D

shared/ghutil/pullrequest.go
```diff
@@ -0,0 +1,144 @@
...
+			var interfaceList []interface{}
+			if nil == err {
+				for _, PR := range page {
+					interfaceList = append(interfaceList, PR)
+				}
+			}
+			return interfaceList, resp, err
+		},
+	)
+	for _, PR := range genericList {
```
srinivashegde86: will we get anything other than PR's here? If no, we should prob use:
res := make([]*PR, len(genericList))
for i,PR := range genericList {
  res[i] = PR
}

this is faster than append. Ditto for the other methods here.

shared/ghutil/pullrequest.go
chaodaiG: since this is cross package var, I would prefer it keep prefix and postfix to avoid potential naming clash

shared/ghutil/pullrequest.go
chaodaiG: done, cross all other methods as well

PR #839, opened 2019-05-29
Add a new performance-mesh job to run perf tests with mesh config

srinivashegde86:
Update prow and testgrid configs to add a perf tests with mesh config

### DIFF
--- a/ci/prow/make_config.go
+++ b/ci/prow/make_config.go
@@ -806,6 +806,7 @@ func generatePeriodic(title string, repoName string, periodicConfig yaml.MapSlic
 	jobNameSuffix := ""
 	jobTemplate := periodicTestJob
 	jobType := ""
+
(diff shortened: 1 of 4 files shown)


ci/prow/make_config.go
```diff
@@ -857,12 +858,12 @@ func generatePeriodic(title string, repoName string, periodicConfig yaml.MapSlic
 				"--github-token /etc/hub-token/token"}
 			addVolumeToJob(&data.Base, "/etc/hub-token", "hub-token", true)
 			data.Base.Timeout = 90
-		case "performance":
+		case "performance", "performance-mesh":
```
adrcunha: Reuse "performance" and just override the args. Because the name will conflict, create a generic key for jobs called "job-name-suffix" and use it to generate the job name.

Next time you create a job like this you won't need any code changes.

PR #842, opened 2019-05-30
Add mock functions for ghutil pullrequest related functions

chaodaiG:
Add mock for ghutil pullrequest related functions, so that they can be used for unit tests

Part of: #689

### DIFF
--- a/shared/ghutil/fakeghutil/fakeghutil.go
+++ b/shared/ghutil/fakeghutil/fakeghutil.go
@@ -20,6 +20,7 @@ package fakeghutil
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/google/go-github/github"
 	"github.com/knative/test-infra/shared/ghutil"
@@ -27,10 +28,13 @@ import (
 
 // FakeGithubClient is a faked client, implements all functions of ghutil.GithubOperations
 type FakeGithubClient struct {
-	User     *github.User
-	Repos    []string
-	Issues   map[string]map[int]*github.Issue       // map of repo: map of issueNumber: issues
-	Comments map[int]map[int64]*github.IssueComment // map of issueNumber: map of commentID: comments
+	User         *github.User
+	Repos        []string
+	Issues       map[string]map[int]*github.Issue       // map of repo: map of issueNumber: issues
+	Comments     map[int]map[int64]*github.IssueComment // map of issueNumber: map of commentID: comments
+	PullRequests map[string]map[int]*github.PullRequest // map of repo: map of PullRequest Number: pullrequests
+	PRCommits    map[int][]*github.RepositoryCommit     // map of PR number: slice of commits
+	CommitFiles  map[string][]*github.CommitFile        // map of commit SHA: slice of files
 
 	NextNumber int    // number to be assigned to next newly created issue/comment
 	BaseURL    string // base URL of Github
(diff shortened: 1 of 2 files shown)


shared/ghutil/fakeghutil/fakeghutil.go
```diff
@@ -185,6 +189,134 @@ func (fgc *FakeGithubClient) RemoveLabelForIssue(org, repo string, issueNumber i
...
+			("" == base || *PR.Base.Ref == base) {
+			res = append(res, PR)
+		}
+	}
+	return res, nil
+}
+
+// ListCommits lists commits from a pull request
+func (fgc *FakeGithubClient) ListCommits(org, repo string, ID int) ([]*github.RepositoryCommit, error) {
+	commits, ok := fgc.PRCommits[ID]
```
srinivashegde86: why no just `return fgc.PRCommits[ID]`

The caller will get an empty array, if not present

shared/ghutil/fakeghutil/fakeghutil.go
```diff
@@ -185,6 +189,134 @@ func (fgc *FakeGithubClient) RemoveLabelForIssue(org, repo string, issueNumber i
...
+		PR.Base = &github.PullRequestBranch{
+			Label: &l,
+			Ref:   &base,
+		}
+	}
+
+	if _, ok := fgc.PullRequests[repo]; !ok {
+		fgc.PullRequests[repo] = make(map[int]*github.PullRequest)
+	}
+	fgc.PullRequests[repo][PRNumber] = PR
```
srinivashegde86: should this be wiuthin if? If the if conditon fails, you will not be creating the map and go will panic

shared/ghutil/fakeghutil/fakeghutil.go
```diff
@@ -185,6 +189,134 @@ func (fgc *FakeGithubClient) RemoveLabelForIssue(org, repo string, issueNumber i
...
+
+// AddCommitToPullRequest adds commit to pullrequest
+// This is complementary of mocking CreatePullRequest, so that newly created pull request can have commits
+func (fgc *FakeGithubClient) AddCommitToPullRequest(org, repo string, ID int, SHA string) error {
+	PRs, ok := fgc.PullRequests[repo]
+	if !ok {
+		return fmt.Errorf("repo %s not exist", repo)
+	}
+	_, ok = PRs[ID]
+	if !ok {
```
srinivashegde86: combine with above

shared/ghutil/fakeghutil/fakeghutil.go
```diff
@@ -185,6 +189,134 @@ func (fgc *FakeGithubClient) RemoveLabelForIssue(org, repo string, issueNumber i
...
+		return fmt.Errorf("Pull Request %d not exist", ID)
+	}
+	c := &github.RepositoryCommit{
```
srinivashegde86: no need of another var. can be inlined

shared/ghutil/fakeghutil/fakeghutil.go
chaodaiG: ok is a bool, but the function needs error, so it would not work

shared/ghutil/fakeghutil/fakeghutil.go
chaodaiG: it would not be added if the repo already exist if moved within

shared/ghutil/fakeghutil/fakeghutil.go
chaodaiG: done

shared/ghutil/fakeghutil/fakeghutil.go
chaodaiG: done

PR #843, opened 2019-05-30
Fix depagination in ghutil package

chaodaiG:
Currently depagination keeps on querying the first page, due to ListOptions not passed in as reference while querying. Passing in correct ListOptions to make it work

### DIFF
--- a/shared/ghutil/issue.go
+++ b/shared/ghutil/issue.go
@@ -38,13 +38,13 @@ type IssueStateEnum string
 
 // ListRepos lists repos under org
 func (gc *GithubClient) ListRepos(org string) ([]string, error) {
-	options := &github.ListOptions{}
+	repoListOptions := &github.RepositoryListOptions{}
 	genericList, err := gc.depaginate(
 		"listing repos",
 		maxRetryCount,
-		options,
+		&repoListOptions.ListOptions,
 		func() ([]interface{}, *github.Response, error) {
-			page, resp, err := gc.Client.Repositories.List(ctx, org, nil)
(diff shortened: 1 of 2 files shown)


shared/ghutil/issue.go
```diff
@@ -63,20 +63,19 @@ func (gc *GithubClient) ListRepos(org string) ([]string, error) {
...
 		issueListOptions.Labels = labels
 	}
 
-	options := &github.ListOptions{}
 	genericList, err := gc.depaginate(
 		fmt.Sprintf("listing issues with label '%v'", labels),
 		maxRetryCount,
-		options,
+		&issueListOptions.ListOptions,
 		func() ([]interface{}, *github.Response, error) {
```
adrcunha: I can't help thinking that you could simplify depaginate even more, by encapsulating this repeated pattern (call gc.Client/for/append/return) into it and just passing a function that calls gc.Client.X. But I won't block on it.

chaodaiG: I think you're absolutely right. I will let this PR in first, and see what I can do refactoring this in another PR

chaodaiG: I have done some investigation, the sad news is this is almost not possible, due to the lack of generic support in Go. As in this case, `ListByRepo` function can not be passed in as a  parameter as the arguments vary among other `List` functions. And this line `page, resp, err := gc.Client.Issues.ListByRepo(ctx, org, repo, issueListOptions)` can not be used as an argument, as the return type(`page`) is different among other `List` functions. So seems like converting `page` to an interface slice locally is not avoidable. @adrcunha

adrcunha: :(

PR #851, opened 2019-06-02
Allow standard install of a custom Knative Serving

adrcunha:
Delegate the process in `start_latest_knative_serving()` to `start_knative_serving()` to the standard install can be used for custom testing as well.

Part of knative/serving#4202

### DIFF
--- a/scripts/library.sh
+++ b/scripts/library.sh
@@ -333,9 +333,9 @@ function start_knative_serving() {
   header "Starting Knative Serving"
   subheader "Installing Knative Serving"
   echo "Installing Serving CRDs from $1"
-  kubectl apply --selector knative.dev/crd-install=true -f $1
+  kubectl apply --selector knative.dev/crd-install=true -f "$1"
   echo "Installing the rest of serving components from $1"
-  kubectl apply -f $1
+  kubectl apply -f "$1"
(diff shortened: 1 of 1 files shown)


scripts/library.sh
```diff
@@ -327,17 +327,23 @@ function report_go_test() {
...
+# Parameters: $1 - Knative Serving manifest.
+function start_knative_serving() {
   header "Starting Knative Serving"
   subheader "Installing Knative Serving"
-  echo "Installing Serving CRDs from ${KNATIVE_SERVING_RELEASE}"
-  kubectl apply --selector knative.dev/crd-install=true -f ${KNATIVE_SERVING_RELEASE}
-  echo "Installing the rest of serving components from ${KNATIVE_SERVING_RELEASE}"
-  kubectl apply -f ${KNATIVE_SERVING_RELEASE}
+  echo "Installing Serving CRDs from $1"
+  kubectl apply --selector knative.dev/crd-install=true -f $1
```
chaodaiG: should this `$1` be double quoted?

chaodaiG: also, what would happen if `$1` is empty string?

chaodaiG:
/lgtm

/hold
Feel free unhold after resolving comment

PR #853, opened 2019-06-03
Move Prow/Testgrid templates into individual yaml files

chaodaiG:
All of our Prow versions live in `*.yaml` files, other than the template inlined in `ci/prow/make_config.go`. Which makes Prow auto version bumper tool need to parse Go files. Move all templates to individual yaml files to avoid parsing Go files.

Part of: #689


ci/prow/templates/general_prow_config.yaml
```diff
@@ -0,0 +1,86 @@
+# Copyright [[.Year]] The Knative Authors
```
adrcunha: I think "prow_config_header.yaml" would be better.

ci/prow/make_config.go
```diff
@@ -218,320 +219,57 @@ var (
...
-  [[indent_array 2 .RepoNames]]
-	`
+	dashboardGroupTemplate = "testgrid_dashboardgroup.yaml"
 )
 
 // Yaml parsing helpers.
 
+// read template yaml file content
+func readTemplate(fp string) string {
+	content, err := ioutil.ReadFile(path.Join(templateDir, fp))
```
adrcunha: I can't help saying that this is I/O inefficient. :) You're reading it from disk every time you need a template. You could store them in a map once they're read, so you only read them from disk once. But I'm OK with a TODO if you don't want to do it now, this is not a critical performance issue.

ci/prow/templates/testgrid_general_config.yaml
```diff
@@ -0,0 +1,60 @@
+# Copyright [[.Year]] The Knative Authors
```
adrcunha: I think "testgrid_config_header.yaml" would be better.

ci/prow/templates/general_prow_config.yaml
chaodaiG: done

ci/prow/make_config.go
chaodaiG: done

ci/prow/templates/testgrid_general_config.yaml
chaodaiG: done

ci/prow/make_config.go
```diff
@@ -213,325 +214,68 @@ var (
 	// key is the main project version, value is another map containing job details
 	//     for the job detail map, key is the repo name, value is the list of job types, like continuous, latency, nightly, and etc.
 	metaData = make(map[string]map[string][]string)
+
+	// tempaltesCache caches templates in memory to avoid I/O
```
```suggestion
	// templatesCache caches templates in memory to avoid I/O
```

PR #859, opened 2019-06-04
Adding CloudRunEvents to prow.

n3wscott:
CloudRunEvents are GCP Knative components.

### DIFF
--- a/ci/prow/config_knative.yaml
+++ b/ci/prow/config_knative.yaml
@@ -98,7 +98,7 @@ presubmits:
     - unit-tests: true
     - integration-tests: true
 
-  GoogleCloudPlatform/cloud-run-events:
+  googlecloudplatform/cloud-run-events:
     - build-tests: true
     - unit-tests: true
     - integration-tests: true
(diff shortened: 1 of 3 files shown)


ci/prow/config_knative.yaml
```diff
@@ -254,3 +260,13 @@ periodics:
   knative/observability:
     - continuous: true
       cron: "30 * * * *" # Run every hour and 30 minutes
+
+  GoogleCloudPlatform/cloud-run-events:
```
adrcunha:
```suggestion
  googlecloudplatform/cloud-run-events:
```

n3wscott: It is CamelCased.

n3wscott: https://github.com/GoogleCloudPlatform/cloud-run-events

adrcunha: Yes, I saw that. But it messes up with all the buckets and tabs and names generated from it. I don't think it will break fetching repos; but if it does I'll revert it.

n3wscott: Got it. Fixed.

PR #864, opened 2019-06-04
add knative-metrics-robot to coverage tool & config

### LINKED ISSUE #863: remove previous posts of knative-metrics-robot
remove previous posts of knative-metrics-robot

steuhs:
The rename from covbot to knative-metrics-robot prevents garbage collection of bot comments.

Fixes #863 

### DIFF
--- a/tools/coverage/main.go
+++ b/tools/coverage/main.go
@@ -50,14 +50,14 @@ func main() {
 	coverageProfileName := flag.String("profile-name", defaultCoverageProfileName, "file name for coverage profile")
 	githubTokenPath := flag.String("github-token", "", "path to token to access github repo")
 	covThresholdFlag := flag.Int("cov-threshold-percentage", defaultCovThreshold, "token to access github repo")
-	covbotUserName := flag.String("covbot-username", "covbot", "github user name for coverage robot")
+	postingBotUserName := flag.String("posting-robot", "knative-metrics-robot", "github user name for coverage robot")
 	flag.Parse()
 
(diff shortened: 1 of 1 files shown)


ci/prow/templates/prow_presubmit_gocoverate_job.yaml
```diff
@@ -18,6 +18,7 @@
         - "--artifacts=$(ARTIFACTS)"
         - "--cov-threshold-percentage=[[.Base.GoCoverageThreshold]]"
         - "--github-token=/etc/covbot-token/token"
+        - "--posting-robot=knative-metrics-robot"
```
srinivashegde86: do we need this, if that is the default?

steuhs: In the future if bot name changes, only the template needs to be updated accordingly.

ci/prow/config.yaml
```diff
@@ -246,6 +246,7 @@ presubmits:
         - "--artifacts=$(ARTIFACTS)"
         - "--cov-threshold-percentage=80"
         - "--github-token=/etc/covbot-token/token"
+        - "--posting-robot=knative-metrics-robot"
```
srinivashegde86: why are we adding this to the jobs, if its the default value? We should only be adding it if it deviates from default

ci/prow/templates/prow_presubmit_gocoverate_job.yaml
srinivashegde86: I understand the reqmt here. I am suggesting that we dont need to add this flag as that is already the default value in code

PR #872, opened 2019-06-06
update boskos images

steuhs:
the previous images are over 2 months old and may contribute to the janitor failures on June 5th

Fixes #

### DIFF
--- a/ci/prow/boskos/config.yaml
+++ b/ci/prow/boskos/config.yaml
@@ -53,7 +53,7 @@ spec:
       terminationGracePeriodSeconds: 30
       containers:
       - name: boskos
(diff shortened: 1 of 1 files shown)


ci/prow/boskos/config.yaml
```diff
@@ -53,7 +53,7 @@ spec:
       terminationGracePeriodSeconds: 30
       containers:
       - name: boskos
-        image: gcr.io/k8s-testimages/boskos:v20190326-21e7f896f
+        image: gcr.io/k8s-testimages/boskos:v20190605-77876e8
```
srinivashegde86: can we use an image a few days old instead of latest? We should use what k8s uses

adrcunha: Yes, use what k8s is using.

PR #878, opened 2019-06-06
Prow version auto bumper part 3: Update Prow versions and create PullRequest

chaodaiG:
Part 3 of Prow version auto bumper, updating versions and creating pullrequest, contains:
- Splitting main.go into `main.go` and `parser.go` without logic change
- Add `file.go` for updating Prow versions in local files
- Add `pullrequest.go` for creating/updating pullrequest
- Add missing ghutil functions

Part of: #689

### DIFF
--- a/tools/prow-auto-bumper/file.go
+++ b/tools/prow-auto-bumper/file.go
@@ -38,15 +38,8 @@ func cdToRootDir() error {
 	return os.Chdir(d)
 }
 
-func call(cmd string, args ...string) error {
-	c := exec.Command(cmd, args...)
-	c.Stdout = os.Stdout
-	c.Stderr = os.Stderr
-	return c.Run()
-}
-
 // update all tags in a byte slice
-func updateAllTags(pv *PRVersions, content []byte, imageFilter *regexp.Regexp) ([]byte, string, []string) {
+func (pv *PRVersions) updateAllTags(content []byte, imageFilter *regexp.Regexp) ([]byte, string, []string) {
 	var msg string
 	var errMsgs []string
 	indexes := imageFilter.FindAllSubmatchIndex(content, -1)
@@ -58,9 +51,13 @@ func updateAllTags(pv *PRVersions, content []byte, imageFilter *regexp.Regexp) (
 	var res string
 	lastIndex := 0
 	for _, m := range indexes {
+		// append from end of last match to end of image part, including ":"
 		res += string(content[lastIndex : m[imageImagePart*2+1]+1])
+		// image part of a version, i.e. the portion before ":"
 		image := string(content[m[imageImagePart*2]:m[imageImagePart*2+1]])
+		// tag part of a version, i.e. the portion after ":"
 		tag := string(content[m[imageTagPart*2]:m[imageTagPart*2+1]])
+		// m[1] is the end index of current match
 		lastIndex = m[1]
 
 		iv := pv.getIndex(image, tag)
@@ -80,26 +77,27 @@ func updateAllTags(pv *PRVersions, content []byte, imageFilter *regexp.Regexp) (
 }
 
 // updateFile updates a file in place.
-func updateFile(pv *PRVersions, filename string, imageFilter *regexp.Regexp, dryrun bool) ([]string, error) {
+func (pv *PRVersions) updateFile(filename string, imageFilter *regexp.Regexp, dryrun bool) ([]string, error) {
 	var errMsgs []string
 	content, err := ioutil.ReadFile(filename)
 	if err != nil {
 		return errMsgs, fmt.Errorf("failed to read %s: %v", filename, err)
 	}
 
-	newContent, msg, errMsgs := updateAllTags(pv, content, imageFilter)
+	newContent, msg, errMsgs := pv.updateAllTags(content, imageFilter)
 	if err := run(
 		fmt.Sprintf("Update file '%s':%s", filename, msg),
 		func() error {
 			return ioutil.WriteFile(filename, newContent, 0644)
 		},
-		dryrun); err != nil {
+		dryrun,
+	); err != nil {
 		return errMsgs, fmt.Errorf("failed to write %s: %v", filename, err)
 	}
 	return errMsgs, nil
 }
 
-func updateAllFiles(pv *PRVersions, fileFilters []*regexp.Regexp, imageFilter *regexp.Regexp,
+func (pv *PRVersions) updateAllFiles(fileFilters []*regexp.Regexp, imageFilter *regexp.Regexp,
 	dryrun bool) ([]string, error) {
 	var errMsgs []string
 	if err := cdToRootDir(); err != nil {
@@ -109,7 +107,7 @@ func updateAllFiles(pv *PRVersions, fileFilters []*regexp.Regexp, imageFilter *r
 	err := filepath.Walk(".", func(filename string, info os.FileInfo, err error) error {
 		for _, ff := range fileFilters {
 			if ff.Match([]byte(filename)) {
-				msgs, err := updateFile(pv, filename, imageFilter, dryrun)
+				msgs, err := pv.updateFile(filename, imageFilter, dryrun)
 				errMsgs = append(errMsgs, msgs...)
 				if err != nil {
 					return fmt.Errorf("Failed to update path %s '%v'", filename, err)
--- a/tools/prow-auto-bumper/file.go
+++ b/tools/prow-auto-bumper/file.go
@@ -1,0 +1,122 @@
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"regexp"
+	"strings"
+)
+
+func cdToRootDir() error {
+	cmd := exec.Command("git", "rev-parse", "--show-toplevel")
+	output, err := cmd.Output()
+	if err != nil {
+		return err
+	}
+	d := strings.TrimSpace(string(output))
+	log.Printf("Changing working directory to %s...", d)
+	return os.Chdir(d)
+}
+
+func call(cmd string, args ...string) error {
+	c := exec.Command(cmd, args...)
+	c.Stdout = os.Stdout
+	c.Stderr = os.Stderr
+	return c.Run()
+}
+
+// update all tags in a byte slice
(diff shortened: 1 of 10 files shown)


tools/prow-auto-bumper/file.go
```diff
@@ -0,0 +1,122 @@
...
+	"log"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"regexp"
+	"strings"
+)
+
+func cdToRootDir() error {
+	cmd := exec.Command("git", "rev-parse", "--show-toplevel")
```
srinivashegde86: use the call method below?

tools/prow-auto-bumper/file.go
```diff
@@ -0,0 +1,122 @@
...
+
+func call(cmd string, args ...string) error {
+	c := exec.Command(cmd, args...)
+	c.Stdout = os.Stdout
+	c.Stderr = os.Stderr
+	return c.Run()
+}
+
+// update all tags in a byte slice
+func updateAllTags(pv *PRVersions, content []byte, imageFilter *regexp.Regexp) ([]byte, string, []string) {
```
srinivashegde86: all the methods seem to be method on pv?

tools/prow-auto-bumper/file.go
```diff
@@ -0,0 +1,122 @@
...
+	indexes := imageFilter.FindAllSubmatchIndex(content, -1)
+	// Not finding any images is not an error.
+	if indexes == nil {
+		return content, msg, errMsgs
+	}
+
+	var res string
+	lastIndex := 0
+	for _, m := range indexes {
+		res += string(content[lastIndex : m[imageImagePart*2+1]+1])
```
srinivashegde86: Add a comment on what content will be. These numbers are confusing

tools/prow-auto-bumper/file.go
srinivashegde86: should this be ?
```suggestion
func updateAllTags(pv *PRVersions, content string, imageFilter *regexp.Regexp) (string, string, []string) {
```

tools/prow-auto-bumper/file.go
```diff
@@ -0,0 +1,122 @@
...
+		image := string(content[m[imageImagePart*2]:m[imageImagePart*2+1]])
+		tag := string(content[m[imageTagPart*2]:m[imageTagPart*2+1]])
+		lastIndex = m[1]
```
srinivashegde86: lastIndex will always be m[1] execpt for first one?

tools/prow-auto-bumper/parser.go
```diff
@@ -0,0 +1,197 @@
...
+	"log"
+	"math"
+	"strings"
+	"time"
+
+	"github.com/knative/test-infra/shared/ghutil"
+)
+
+// Helper method for adding a newly discovered tag into pv
+func (pv *PRVersions) getIndex(image, tag string) int {
```
srinivashegde86: move the methods in the same file with the definition of PRVersions

tools/prow-auto-bumper/pullrequest.go
```diff
@@ -0,0 +1,161 @@
...
+		func() error { return call("git", commitArgs...) },
+		dryrun); err != nil {
+		return fmt.Errorf("failed to git commit: %v", err)
+	}
+	pushArgs := []string{"push", "-f", fmt.Sprintf("git@github.com:%s/%s.git", gi.userID, gi.repo),
+		fmt.Sprintf("HEAD:%s", gi.head)}
+	if err := run(
+		fmt.Sprintf("Running 'git %s'", strings.Join(pushArgs, " ")),
+		func() error { return call("git", pushArgs...) },
+		dryrun); err != nil {
```
Fredy-Z: The format looks harder to read, I would suggest
```suggestion
		dryrun,
	); err != nil {
```
The same applies to a couple of other function calls as well.

tools/prow-auto-bumper/pullrequest.go
```diff
@@ -0,0 +1,161 @@
...
+
+func getMatchTitle() string {
+	return "Update prow from"
+}
+
+func generatePRBody(extraMsgs []string) string {
+	var body string
+	if len(extraMsgs) > 0 {
+		body += "Warnings:\n"
+		for _, msg := range extraMsgs {
```
Fredy-Z: Consider using [stringbuilder](https://www.calhoun.io/concatenating-and-building-strings-in-go/)?

tools/prow-auto-bumper/parser.go
```diff
@@ -0,0 +1,197 @@
...
+		}
+	}
+	return bestPv, overallErr
+}
+
+func retryGetBestVersion(gcw *GHClientWrapper, gi gitInfo) (*PRVersions, error) {
+	var bestPv *PRVersions
+	var overallErr error
+	// retry if there is github related error
+	for retryCount := 0; nil == overallErr && retryCount < maxRetry; retryCount++ {
```
Fredy-Z: There's a logic error here - if `nil == overallErr` you still retry until the `maxRetry` :-)

tools/prow-auto-bumper/parser.go
```diff
@@ -0,0 +1,197 @@
...
+			pv.images[plus[imageImagePart]][iv].newVersion = plus[imageTagPart]
+		}
+	}
+
+	return nil
+}
+
+// Query all PRs from "k8s-ci-robot:autobump", find PR roughly 7 days old and was not reverted later.
+// Only return error if it's github related
+func getBestVersion(gcw *GHClientWrapper, gi gitInfo) (*PRVersions, error) {
```
Fredy-Z: This function looks quite long... Can we possibly break it up?

tools/prow-auto-bumper/file.go
```diff
@@ -0,0 +1,122 @@
...
+		dryrun); err != nil {
+		return errMsgs, fmt.Errorf("failed to write %s: %v", filename, err)
+	}
+	return errMsgs, nil
+}
+
+func updateAllFiles(pv *PRVersions, fileFilters []*regexp.Regexp, imageFilter *regexp.Regexp,
+	dryrun bool) ([]string, error) {
+	var errMsgs []string
+	if err := cdToRootDir(); err != nil {
```
Fredy-Z: Do we need to change dir back after the function is done?

tools/prow-auto-bumper/pullrequest.go
```diff
@@ -0,0 +1,161 @@
...
+	body := generatePRBody(extraMsgs)
+	if err := makeGitCommit(gi, title, dryrun); nil != err {
+		return fmt.Errorf("failed git commit: '%v'", err)
+	}
+	existPR, err := getExistingPR(gcw, gi, matchTitle)
+	if nil != err {
+		return fmt.Errorf("failed querying existing pullrequests: '%v'", err)
+	}
+	if nil != existPR {
+		log.Printf("found open PR '%d'", *existPR.Number)
```
Fredy-Z: Nit - some of log messages are capitalized but some are not. Better to make them consistent.

tools/prow-auto-bumper/file.go
```diff
@@ -0,0 +1,122 @@
...
+		},
+		dryrun); err != nil {
+		return errMsgs, fmt.Errorf("failed to write %s: %v", filename, err)
+	}
+	return errMsgs, nil
+}
+
+func updateAllFiles(pv *PRVersions, fileFilters []*regexp.Regexp, imageFilter *regexp.Regexp,
+	dryrun bool) ([]string, error) {
+	var errMsgs []string
```
Fredy-Z: Do we indeed need to return this `errMsgs`? Seems it's empty if `err` is `nil`, and it's not being used if `err` is not `nil`.

tools/prow-auto-bumper/file.go
chaodaiG: They are slightly different, actually call method should belong to `pullrequest.go`, just moved it there

tools/prow-auto-bumper/file.go
chaodaiG: that's right,done

chaodaiG: []byte helps better with substring extraction

tools/prow-auto-bumper/file.go
chaodaiG: done

tools/prow-auto-bumper/file.go
chaodaiG: since it's looping over m, lastIndex is always m[1], I added comment there

tools/prow-auto-bumper/file.go
chaodaiG: It's not used here, the purpose of errMsgs is to be displayed in Version update PR, so approvers get first hand idea

tools/prow-auto-bumper/file.go
chaodaiG: I think it's a good idea, but when I tried to implement it, felt like introducing another possibility of error out

tools/prow-auto-bumper/parser.go
chaodaiG: done

tools/prow-auto-bumper/parser.go
chaodaiG: I have thought about it and tried it, it was not as clean as I thought, so kept it this way. Open to suggestion though

tools/prow-auto-bumper/parser.go
chaodaiG: you are right, not sure why this was not discovered in my test runs. Thanks!

tools/prow-auto-bumper/pullrequest.go
chaodaiG: This loop would rarely be executed, so this should be fine

tools/prow-auto-bumper/pullrequest.go
chaodaiG: done

tools/prow-auto-bumper/pullrequest.go
chaodaiG: Done.

tools/prow-auto-bumper/file.go
chaodaiG: To be clear, this one expects output, and the others are not

tools/prow-auto-bumper/parser.go
Fredy-Z: I did not find a way to break it down either, but I do think there should be more comments. `getDominantVersions` does not have comments on itself or in this function call, and it's not quite clear why you need the check `if updatePR, ok := visited[vs.newVersion]; ok `.

tools/prow-auto-bumper/parser.go
```diff
@@ -0,0 +1,178 @@
...
+	visited := make(map[string]PRVersions)
+	var bestPv *PRVersions
+	var overallErr error
+	var bestDelta float64 = maxDelta + 1
```
Fredy-Z: I was trying to find where `maxDelta` is, then I realized it was in `main.go`. I think it would clearer to put these constants in a separate file like `constants.go` or `config.go`.

tools/prow-auto-bumper/parser.go
chaodaiG: I was planning to add more in code README as well as a separate README file in another PR, but You're right, comments added here

tools/prow-auto-bumper/parser.go
chaodaiG: good idea, moved all these to config.go

PR #879, opened 2019-06-06
eventing-sources to eventing-contrib

Fredy-Z:
`eventing-sources` repo is renamed to `eventing-contrib`, we need to make corresponding changes in test-infra.


adrcunha:
/hold

This will break the GCP jobs, let me finish that first.

PR #881, opened 2019-06-07
Fix CamelCased org/repo names in Prow/Testgrid configs

adrcunha:
Also fix a few settings when the org is not Knative.

### DIFF
--- a/ci/prow/config.yaml
+++ b/ci/prow/config.yaml
@@ -1593,7 +1593,7 @@ presubmits:
       - name: test-account
         secret:
           secretName: test-account
-  googlecloudplatform/cloud-run-events:
+  GoogleCloudPlatform/cloud-run-events:
   - name: pull-googlecloudplatform-cloud-run-events-build-tests
     agent: kubernetes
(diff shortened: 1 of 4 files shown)


ci/prow/config.yaml
```diff
@@ -3654,7 +3654,7 @@ periodics:
       - "--" # end kubernetes_execute_bazel flags (consider following flags as text)
       - "./hack/release.sh"
       - "--dot-release"
-      - "--release-gcs googlecloudplatform/cloud-run-events"
+      - "--release-gcs knative-releases/cloud-run-events"
```
Fredy-Z: Will it be worth to create new gcs and gcr bucket for it? It's a bit weird to put it here.
I'm totally fine with this change as it's just one exception, but in the future if we have more such repos we might need to consider separating them.

adrcunha: I don't know what's the plan for the releases, so right now I prefer leaving them in knative-releases. Once that's figured out we can delete or move whatever was left in knative-releases.

PR #884, opened 2019-06-07
[monitoring] query db and checks if error condition is met

steuhs:
1. Query the ErrorLogs table and collect all matching entries within the time period specified in the monitoring config.
2. check against each alert condition

For more details refer to Step 5 in design doc

Fixes #


tools/monitoring/mysql/db_mysql.go
```diff
@@ -118,3 +118,45 @@ func PubsubMsgHandler(db *sql.DB, configURL, buildLogURL, jobname string, prNumb
 
 	return tx.Commit()
 }
+
+// CheckAlertCondition checks whether the given error pattern meets
+// the alert condition specified in config
+func CheckAlertCondition(errorPattern string, config *config.SelectedConfig, db *sql.DB) (bool, error) {
+	// the timestamp we want to start collecting logs
+	startTime := time.Now().Add(time.Minute * time.Duration(config.Period))
```
yt3liu: I think the `startTime` variable can be removed by using `now()` in `CREATE VIEW...` statement.

tools/monitoring/mysql/db_mysql.go
```diff
@@ -118,3 +118,45 @@ func PubsubMsgHandler(db *sql.DB, configURL, buildLogURL, jobname string, prNumb
...
+
+	_, err := db.Query(
+		"CREATE VIEW Matched AS\nSELECT Jobname, PrNumber FROM ErrorLogs\nWHERE\n    ErrorPattern=? and\n    TimeStamp > ?",
```
yt3liu: Remove `\n` in the query.

tools/monitoring/mysql/db_mysql.go
```diff
@@ -118,3 +118,45 @@ func PubsubMsgHandler(db *sql.DB, configURL, buildLogURL, jobname string, prNumb
...
+	if err != nil {
+		return false, err
+	}
+
+	var count int
+	if err = db.QueryRow("SELECT COUNT(*) FROM Matched;").Scan(&count); err != nil {
+		return false, err
+	}
+
+	if count < config.Occurrences {
```
yt3liu: In my opinion, I think updating `count` in after every `select` makes the code hard to read. I think you can combined the `select` statements and do `select count(*) as all_errors, count(distinct jobname) as failed_jobs, ...` and stores the count in three different vars and do a check at the end.

tools/monitoring/mysql/db_mysql.go
steuhs: Why do we want to do it there though? There is computation to be performed, as you can see in the assignment and I don't think it is a good idea to put the computation logic in the query.

yt3liu: Sorry, missed the period computation. Disregard my comment.

tools/monitoring/mysql/db_mysql.go
steuhs: I think \n is a better separator here than a space, though there is no semantic difference. Line separation is the norm in sql and makes the text more understandable

yt3liu: I agree line separator (with actual new line, not escape character) is better than spaces. With the current style, I would consider this as a single line. Having escape character directly in the query is difficult to read as we have to do the escape char conversion in our head. Maybe make the actual query multi-line for better readability?

steuhs: Agree

tools/monitoring/mysql/db_mysql.go
```diff
@@ -118,3 +118,35 @@ func PubsubMsgHandler(db *sql.DB, configURL, buildLogURL, jobname string, prNumb
...
+func CheckAlertCondition(errorPattern string, config *config.SelectedConfig, db *sql.DB) (bool, error) {
+	// the timestamp we want to start collecting logs
+	startTime := time.Now().Add(time.Minute * time.Duration(config.Period))
+
+	_, err := db.Query(
+		`CREATE VIEW Matched AS
+  SELECT Jobname, PrNumber FROM ErrorLogs
+  WHERE
+        ErrorPattern=? and TimeStamp > ?`,
+		errorPattern, startTime)
```
yt3liu: nit: indent or create a const for the sql statements and keep gocode in one line. 
```
_, err := db.Query(
		`CREATE VIEW Matched
		  AS
		     SELECT Jobname, PrNumber
		     FROM ErrorLogs
		     WHERE ErrorPattern = ? and TimeStamp > ?`,
		errorPattern, 
                 startTime)
```

tools/monitoring/mysql/db_mysql.go
```diff
@@ -118,3 +118,35 @@ func PubsubMsgHandler(db *sql.DB, configURL, buildLogURL, jobname string, prNumb
...
+
+	if err != nil {
+		return false, err
+	}
+
+	var nMatches int
+	var nJobs int
+	var nPRs int
```
yt3liu: combine this to one line `var nMatches, nJobs, nPRs int`

tools/monitoring/mysql/db_mysql.go
```diff
@@ -118,3 +118,35 @@ func PubsubMsgHandler(db *sql.DB, configURL, buildLogURL, jobname string, prNumb
...
+
+	if err = db.QueryRow(`SELECT 
+       COUNT(*),
+       COUNT (DISTINCT Jobname),
+       COUNT (DISTINCT PrNumber)
```
yt3liu: nit: indent. Or this may be short enough to do it in one line?

tools/monitoring/mysql/db_mysql.go
```diff
@@ -118,3 +118,35 @@ func PubsubMsgHandler(db *sql.DB, configURL, buildLogURL, jobname string, prNumb
...
+FROM Matched;`).Scan(&nMatches, nJobs, nPRs); err != nil {
+		return false, err
+	}
+
+	return nMatches > config.Occurrences && nJobs > config.JobsAffected && nPRs > config.PrsAffected, nil
```
yt3liu: Maybe `>` should be `>=`?

tools/monitoring/mysql/db_mysql.go
steuhs: my CREATE VIEW Matched AS use the same format as "https://www.w3schools.com/sql/sql_view.asp"

I can change the indent for where clause a bit

tools/monitoring/mysql/operations.go
```diff
@@ -67,3 +67,37 @@ func rollbackTx(tx *sql.Tx, err error) error {
...
+// the alert condition specified in config
+func CheckAlertCondition(errorPattern string, config *config.SelectedConfig, db *sql.DB) (bool, error) {
+	// the timestamp we want to start collecting logs
+	startTime := time.Now().Add(time.Minute * time.Duration(config.Period))
+
+	_, err := db.Query(`
+		CREATE VIEW Matched AS
+  	SELECT Jobname, PrNumber 
+  	FROM ErrorLogs
+  	WHERE ErrorPattern=? and TimeStamp > ?`,
```
yt3liu: nit: Is the indent of SELECT and CREATE not aligned due to the formatter? Or does github web rendering spaces differently?  Could you align it similar as follows? Maybe a mix of tabs and spaces in the preceding whitespaces? 
```
	_, err := db.Query(`
                 CREATE VIEW Matched 
                 AS
                        SELECT Jobname, PrNumber 
                        FROM ErrorLogs
                        WHERE ErrorPattern=? and TimeStamp > ?`,
```

tools/monitoring/mysql/operations.go
```diff
@@ -67,3 +67,37 @@ func rollbackTx(tx *sql.Tx, err error) error {
...
+	}
+
+	var nMatches, nJobs, nPRs int
+
+	row := db.QueryRow(`
+		SELECT 
+    	COUNT(*),
+    	COUNT (DISTINCT Jobname),
+    	COUNT (DISTINCT PrNumber)
+		FROM Matched;`)
```
yt3liu: ditto

tools/monitoring/mysql/operations.go
```diff
@@ -67,3 +67,37 @@ func rollbackTx(tx *sql.Tx, err error) error {
...
+
+	if err = row.Scan(&nMatches, nJobs, nPRs); err != nil {
```
yt3liu: My suspicion is that this wouldn't work. Maybe you meant: 

```suggestion
	if err = row.Scan(&nMatches, &nJobs, &nPRs); err != nil {
```

tools/monitoring/mysql/operations.go
srinivashegde86: do we need to create a view?

PR #892, opened 2019-06-10
Add common performance metrics methods

srinivashegde86:
Also, refactor some code:
* Add a few methods from serving/performance.go into common infra
* Move RollbackTx into SQL client
* Move errors percentage to loadgen from performance

### DIFF
--- a/shared/mysql/client.go
+++ b/shared/mysql/client.go
@@ -20,6 +20,7 @@ import (
 	"database/sql"
 	"database/sql/driver"
 	"fmt"
+	"io/ioutil"
 
 	_ "github.com/go-sql-driver/mysql"
 )
@@ -32,6 +33,27 @@ type DBConfig struct {
 	Password     string
 	Instance     string
 	DatabaseName string
+}
+
+func ConfigureDB(userSecret, passSecret, dbName, dbInstance string) (*DBConfig, error) {
+	user, err := ioutil.ReadFile(userSecret)
+	if err != nil {
+		return nil, err
+	}
+
+	pass, err := ioutil.ReadFile(passSecret)
+	if err != nil {
+		return nil, err
+	}
+
+	config := DBConfig{
+		Username:     string(user),
+		Password:     string(pass),
+		DatabaseName: dbName,
+		Instance:     dbInstance,
+	}
+
+	return &config, nil
 }
 
 func (c DBConfig) TestConn() error {
--- a/shared/mysql/client.go
+++ b/shared/mysql/client.go
@@ -35,7 +35,7 @@ type DBConfig struct {
 }
 
 func (c DBConfig) TestConn() error {
-	conn, err := c.getConn()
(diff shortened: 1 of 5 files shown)


shared/performance/performance.go
```diff
@@ -0,0 +1,119 @@
...
+	return tc
+}
+
+type DBConfig struct {
+	mysql.DBConfig
+}
+
+// Configure the db instance to store metrics information.
+// This will be later used to show the trending metrics on our grafana dashboard.
+func ConfigureDB() (*DBConfig, error) {
```
yt3liu: I think this is also defined in monitor/main.go. Move it to separate file (maybe `shared/mysql/config.go`?)

shared/performance/performance.go
```diff
@@ -0,0 +1,119 @@
...
+func CreatePerfTestCase(metricValue float32, metricName, testName string) junit.TestCase {
+	tp := []junit.TestProperty{{Name: perfLatency, Value: fmt.Sprintf("%f", metricValue)}}
+	tc := junit.TestCase{
+		ClassName:  testName,
+		Name:       fmt.Sprintf("%s/%s", testName, metricName),
+		Properties: junit.TestProperties{Properties: tp}}
+	return tc
+}
+
+type DBConfig struct {
```
yt3liu: Move this to a common file

shared/mysql/client.go
```diff
@@ -71,3 +71,12 @@ func (c DBConfig) dataStoreName(dbName string) string {
 
 	return fmt.Sprintf("%sunix(%s)/%s", cred, "/cloudsql/"+c.Instance, dbName)
 }
+
+// RollbackTx will try to rollback the transaction and return an error message accordingly
+func RollbackTx(tx *sql.Tx, err error) error {
+	rollbackErr := tx.Rollback()
+	if rollbackErr == nil {
```
yt3liu: Combine the statements
```
	if rollbackErr := tx.Rollback(); rollbackErr != nil {
             return fmt.Errorf("Statement execution failed: %v; rollback failed: %v", err, rollbackErr)
        }
        return fmt.Errorf("Statement execution failed: %v; rolled back", err)
```

shared/performance/performance.go
srinivashegde86: yes. but the dbname here is different and I feel every module should be able to define their own secrets and db names and does not make sense in common

shared/performance/performance.go
srinivashegde86: I dont think that makes sense. I want this here as I want methods on it that only perf scripts can call.

shared/performance/performance.go
yt3liu: I think we can take secret file and db name, db instance as params for this function. I think it still make sense to move it to common since the underlying logic is identical.

srinivashegde86: agreed. done.

PR #899, opened 2019-06-11
Add Prow auto bumper job

chaodaiG:
This PR adds Prow job for Prow auto bumper tool


ci/prow/make_config.go
```diff
@@ -1302,6 +1306,23 @@ func setOutput(fileName string) {
 	output = configFile
 }
 
+// Recursively finding all yaml files under "asis" dir, add them asis in prow jobs
+func generateAsisJobs() {
```
adrcunha: You should be able to generate this job as a custom job plus arguments override.

Like https://github.com/knative/test-infra/blob/master/ci/prow/config_knative.yaml#L121

chaodaiG: This is just an attempt of trying out ways of supporting very custom jobs. I will update this PR to use arguments override

adrcunha: I understand. But I would very much prefer having a framework around the jobs so all pieces are in sync, instead of going back to having several fully expanded job definitions (I know this is just one job, but that'll make it easy to start having copy&paste jobs added to Prow).

chaodaiG: Totally agreed, keeping the base in sync for future fully expanded jobs is the biggest con for this approach. I will explore other ways in future works

chaodaiG: done

chaodaiG:
/hold

Wait until all required permissions are granted on the bot user that runs this tool

chaodaiG:
Bot granted CLA
/unhold

PR #902, opened 2019-06-11
Add a configurable release number as a parameter to launch Knative serving

### LINKED ISSUE #897: Allow the version of the Knative installation be configurable
In https://github.com/knative/client/pull/170/files we are trying to establish E2E tests which tests against various version of Knative serving. Therefore we need to be able to parameterize the version of Knative to install in the E2E setup, but unfortunately the version in https://github.com/knative/client/blob/0d9494736616aa21e8a45da443cbc7a0a922ca7f/vendor/github.com/knative/test-infra/scripts/library.sh#L506 is currently hardcoded and not overwritable.

For a quick win it would be enough to allow this env variable to be overwritable, but for long term it would be much neither if the overall test could be parameterized with the Knative version to used for the tests)

houshengbo:
With the current script, we are only able to launch a Knative serving with the latest version from the source code. With this parameter added, we are able to launch Knative serving for a specific released version.

Fixes #897 

### DIFF
--- a/scripts/library.sh
+++ b/scripts/library.sh
@@ -339,6 +339,12 @@ function start_knative_serving() {
   wait_until_pods_running knative-serving || return 1
 }
 
+# Install the stable release Knative/serving in the current cluster.
+# Parameters: $1 - Knative Serving version number, e.g. 0.6.0.
+function start_release_knative_serving() {
+  start_knative_serving "https://storage.googleapis.com/knative-releases/serving/previous/v$1/serving.yaml"
+}
+
 # Install the latest stable Knative Serving in the current cluster.
 function start_latest_knative_serving() {
   start_knative_serving "${KNATIVE_SERVING_RELEASE}"


scripts/library.sh
```diff
@@ -327,21 +327,28 @@ function report_go_test() {
   return ${failed}
 }
 
-# Install Knative Serving in the current cluster.
-# Parameters: $1 - Knative Serving manifest.
-function start_knative_serving() {
```
adrcunha: This will break smoke serving tests, please don't remove this function.

scripts/library.sh
```diff
@@ -327,21 +327,28 @@ function report_go_test() {
...
-  header "Starting Knative Serving"
-  subheader "Installing Knative Serving"
-  echo "Installing Serving CRDs from $1"
-  kubectl apply --selector knative.dev/crd-install=true -f "$1"
-  echo "Installing the rest of serving components from $1"
-  kubectl apply -f "$1"
+# Install the stable release Knative/serving or the latest from source code in the current cluster.
+# Parameters: $1 - Knative Serving version number or latest.
+function start_release_knative_serving() {
```
adrcunha: I suggest you keep start_knative_serving() and create something like start_knative_serving_version() that will simply call start_knative_serving with the full URL for a versioned manifest. No need to touch start_latest_knative_serving.

scripts/library.sh
houshengbo: ok

houshengbo:
@adrcunha I committed the changes. Thx.

scripts/library.sh
adrcunha: No need to handle "latest". It's redundant since your doing exactly what start_latest_knative_serving() is doing.

houshengbo: I will remove it.

scripts/library.sh
```diff
@@ -339,6 +339,12 @@ function start_knative_serving() {
   wait_until_pods_running knative-serving || return 1
 }
 
+# Install the stable release Knative/serving in the current cluster.
+# Parameters: $1 - Knative Serving version number or latest.
+function start_release_knative_serving() {
+  start_knative_serving "https://github.com/knative/serving/releases/download/$1/serving.yaml"
```
adrcunha: For consistency, the URL should be

```suggestion
  start_knative_serving "https://storage.googleapis.com/knative-releases/serving/previous/v$1/serving.yaml"
```

scripts/library.sh
```diff
@@ -339,6 +339,12 @@ function start_knative_serving() {
   wait_until_pods_running knative-serving || return 1
 }
 
+# Install the stable release Knative/serving in the current cluster.
+# Parameters: $1 - Knative Serving version number or latest.
```
```suggestion
# Parameters: $1 - Knative Serving version number (e.g., 0.6.1).
```

houshengbo: done.

scripts/library.sh
houshengbo: done

PR #904, opened 2019-06-11
add endpoint test for alert condition check

steuhs:
verifies CheckAlertCondition handles no-match scenario correctly


tools/monitoring/main.go
```diff
@@ -106,7 +107,19 @@ func testCloudSQLConn(w http.ResponseWriter, r *http.Request) {
...
+
+	selectedConfig := config.SelectedConfig{}
+
+	db, _ := dbConfig.Connect()
+
+	fmt.Fprintf(w, "testing alert condition check...")
+	toAlert, err := mysql2.CheckAlertCondition("none", &selectedConfig, db)
+	if err != nil {
+		fmt.Fprintf(w, "error running alert condition check in no-match senario: %v", err)
+	} else if toAlert == true {
```
srinivashegde86:
```suggestion
	} else if toAlert {
```

tools/monitoring/main.go
```diff
@@ -106,7 +107,19 @@ func testCloudSQLConn(w http.ResponseWriter, r *http.Request) {
 		fmt.Fprintf(w, "Failed to ping the database %v", err)
 		return
 	}
-	fmt.Fprintf(w, "Success\n")
+	fmt.Fprintf(w, "DB Connection Succeeds\n")
+
+	selectedConfig := config.SelectedConfig{}
+
```
srinivashegde86: remove blankline

tools/monitoring/main.go
```diff
@@ -106,7 +107,19 @@ func testCloudSQLConn(w http.ResponseWriter, r *http.Request) {
...
+	db, _ := dbConfig.Connect()
```
srinivashegde86: any reason we are ignoring the err?

steuhs: Because that is already checked by TestConn() before this

yt3liu: I think TestConn can be removed from the `testCloudSQLConn` handler and check for `dbConfig.Connect()` error instead.

steuhs: Agree

tools/monitoring/main.go
```diff
@@ -113,14 +113,20 @@ func hello(w http.ResponseWriter, r *http.Request) {
...
 	log.Printf("Serving request: %s", r.URL.Path)
-	log.Println("Testing mysql database connection.")
+	fmt.Fprintf(w,"Testing mysql database connection...")
 
-	err := dbConfig.TestConn()
+	db, err := dbConfig.Connect()
 	if err != nil {
-		fmt.Fprintf(w, "Failed to ping the database %v", err)
-		return
+		log.Fatalln(err)
```
srinivashegde86: shouldnt we write the err message on the response?

steuhs: since we are doing the checking through the url in browser. It's easier to see it printed on browser instead of in console. If in the future, we make it into some automated test, then it's better to put it in console.

yt3liu: In this case, I think you meant to use `fmt.Fprintf(w, ...` as that makes the response viewable through the browser. [log.Fatalln](https://golang.org/pkg/log/#Fatalln) prints to the log and exits the app. I think for checking the alert condition, exiting the app after error may not be ideal.

steuhs: @yt3liu You are right. I read the change reversely ;)

Issue #871, opened 2019-06-05
Run multiple k8s/istio tests in parallel for Knative


chaodaiG:
Run multiple combinations of k8s/istio versions for regression detection, make sure the results are surfaced on Testgrid as well.

Start with latest 3 minor versions of k8s X Istio 1.0/1.1

chaodaiG:
/close
done in associated PRs

PR #905, opened 2019-06-11
Support cluster version as [major].[minor] or [major] in e2e tests

chaodaiG:
Currently e2e test would fail if pass in `cluster_version=1.12` or `cluster_version=v1.12`, as kubetest only accepts version in the format of `v[major].[minor].[patch]`, it's not easy to change this behavior as this version is consumed by both `--extract` and gke cluster creation, and `--extract` doesn't work with `v[major].[minor]` at all. So add logic finding latest gke version with same [major].[minor] version

part of: #871

### DIFF
--- a/scripts/e2e-tests.sh
+++ b/scripts/e2e-tests.sh
@@ -247,20 +247,19 @@ function create_test_cluster_with_retries() {
 
       local cluster_creation_zone="${E2E_CLUSTER_REGION}"
       [[ -n "${E2E_CLUSTER_ZONE}" ]] && cluster_creation_zone="${E2E_CLUSTER_REGION}-${E2E_CLUSTER_ZONE}"
-      local geoflag="--gcp-region=${cluster_creation_zone}"
-      # Kubetest --extract expects version string in the format of v[major].[minor].[patch], get latest and convert to this format
-      # if E2E_CLUSTER_VERSION is in the form of [major] or [major].[minor]
-      if [[ "${E2E_CLUSTER_VERSION}" =~ ^[0-9]+$ || "${E2E_CLUSTER_VERSION}" =~ ^[0-9]+\.[0-9]+$ ]]; then
-        E2E_CLUSTER_VERSION="v$(get_latest_gke_version ${E2E_CLUSTER_VERSION} ${cluster_creation_zone})"
-        [[ -z "${E2E_CLUSTER_VERSION}" ]] && fail_test "Unsupported cluster version provided"
+      # Convert cluster version to a full GKE version if we're not using latest
+      if [[ "${E2E_CLUSTER_VERSION}" =~ ^[0-9](\.[0-9]+)?$ ]]; then
+        E2E_CLUSTER_VERSION="v$(gcloud container get-server-config --zone=${cluster_creation_zone} --format="value(validMasterVersions)" 2> /dev/null \
+          | tr ';' '\n' | grep -E ^${E2E_CLUSTER_VERSION} | cut -f1 -d- | sort | tail -1)"
+        [[ "${E2E_CLUSTER_VERSION}" == "v" ]] && fail_test "Unsupported cluster version provided"
       fi
 
-      header "Creating test cluster in $E2E_CLUSTER_REGION $E2E_CLUSTER_ZONE with version $E2E_CLUSTER_VERSION"
+      header "Creating test cluster ${E2E_CLUSTER_VERSION} in ${cluster_creation_zone}"
       # Don't fail test for kubetest, as it might incorrectly report test failure
       # if teardown fails (for details, see success() below)
       set +o errexit
       { run_go_tool k8s.io/test-infra/kubetest \
-        kubetest --extract ${E2E_CLUSTER_VERSION} "$@" ${geoflag}; } 2>&1 | tee ${cluster_creation_log}
+        kubetest --extract ${E2E_CLUSTER_VERSION} "$@" --gcp-region=${cluster_creation_zone}; } 2>&1 | tee ${cluster_creation_log}
 
       # Exit if test succeeded
       [[ "$(get_test_return_code)" == "0" ]] && return 0
--- a/scripts/e2e-tests.sh
+++ b/scripts/e2e-tests.sh
@@ -204,7 +204,6 @@ function create_test_cluster() {
   (( ! IS_BOSKOS )) && extra_flags+=(--down)
   create_test_cluster_with_retries "${CLUSTER_CREATION_ARGS[@]}" \
     --up \
-    --extract "${E2E_CLUSTER_VERSION}" \
     --gcp-node-image "${SERVING_GKE_IMAGE}" \
     --test-cmd "${E2E_SCRIPT}" \
     --test-cmd-args "${test_cmd_args}" \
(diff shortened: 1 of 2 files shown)


scripts/library.sh
```diff
@@ -502,6 +502,22 @@ function get_latest_knative_yaml_source() {
   fi
 }
 
+# Returns the latest gke version in [major].[minor].[patch] with provided prefix
```
adrcunha: Move to e2e-tests.sh, since this is pretty specific.

scripts/e2e-tests.sh
```diff
@@ -246,15 +245,22 @@ function create_test_cluster_with_retries() {
...
+      fi
 
-      header "Creating test cluster in $E2E_CLUSTER_REGION $E2E_CLUSTER_ZONE"
+      header "Creating test cluster in $E2E_CLUSTER_REGION $E2E_CLUSTER_ZONE with version $E2E_CLUSTER_VERSION"
       # Don't fail test for kubetest, as it might incorrectly report test failure
       # if teardown fails (for details, see success() below)
       set +o errexit
       { run_go_tool k8s.io/test-infra/kubetest \
-        kubetest "$@" ${geoflag}; } 2>&1 | tee ${cluster_creation_log}
+        kubetest --extract ${E2E_CLUSTER_VERSION} "$@" ${geoflag}; } 2>&1 | tee ${cluster_creation_log}
```
adrcunha: You can inline geoflag here, there's no reason for it anymore with cluster_creation_zone defined.

scripts/e2e-tests.sh
```diff
@@ -246,15 +245,22 @@ function create_test_cluster_with_retries() {
...
       [[ "${E2E_CLUSTER_ZONE}" == "${zone_not_provided}" ]] && E2E_CLUSTER_ZONE=""
 
-      local geoflag="--gcp-region=${E2E_CLUSTER_REGION}"
-      [[ -n "${E2E_CLUSTER_ZONE}" ]] && geoflag="--gcp-zone=${E2E_CLUSTER_REGION}-${E2E_CLUSTER_ZONE}"
+      local cluster_creation_zone="${E2E_CLUSTER_REGION}"
+      [[ -n "${E2E_CLUSTER_ZONE}" ]] && cluster_creation_zone="${E2E_CLUSTER_REGION}-${E2E_CLUSTER_ZONE}"
+      local geoflag="--gcp-region=${cluster_creation_zone}"
+      # Kubetest --extract expects version string in the format of v[major].[minor].[patch], get latest and convert to this format
+      # if E2E_CLUSTER_VERSION is in the form of [major] or [major].[minor]
+      if [[ "${E2E_CLUSTER_VERSION}" =~ ^[0-9]+$ || "${E2E_CLUSTER_VERSION}" =~ ^[0-9]+\.[0-9]+$ ]]; then
```
adrcunha: Use ()? in the regex to make the minor version optional.

scripts/e2e-tests.sh
```diff
@@ -246,15 +245,22 @@ function create_test_cluster_with_retries() {
       E2E_CLUSTER_ZONE=${e2e_cluster_zone}
       [[ "${E2E_CLUSTER_ZONE}" == "${zone_not_provided}" ]] && E2E_CLUSTER_ZONE=""
 
-      local geoflag="--gcp-region=${E2E_CLUSTER_REGION}"
-      [[ -n "${E2E_CLUSTER_ZONE}" ]] && geoflag="--gcp-zone=${E2E_CLUSTER_REGION}-${E2E_CLUSTER_ZONE}"
+      local cluster_creation_zone="${E2E_CLUSTER_REGION}"
+      [[ -n "${E2E_CLUSTER_ZONE}" ]] && cluster_creation_zone="${E2E_CLUSTER_REGION}-${E2E_CLUSTER_ZONE}"
+      local geoflag="--gcp-region=${cluster_creation_zone}"
+      # Kubetest --extract expects version string in the format of v[major].[minor].[patch], get latest and convert to this format
```
adrcunha: You can shorten these two comments to something like

```# Convert cluster version to a full GKE version if we're not using latest```

File: scripts/e2e-tests.sh [hunk 5]
```diff
@@ -246,15 +245,22 @@ function create_test_cluster_with_retries() {
...
+      # if E2E_CLUSTER_VERSION is in the form of [major] or [major].[minor]
+      if [[ "${E2E_CLUSTER_VERSION}" =~ ^[0-9]+$ || "${E2E_CLUSTER_VERSION}" =~ ^[0-9]+\.[0-9]+$ ]]; then
+        E2E_CLUSTER_VERSION="v$(get_latest_gke_version ${E2E_CLUSTER_VERSION} ${cluster_creation_zone})"
+        [[ -z "${E2E_CLUSTER_VERSION}" ]] && fail_test "Unsupported cluster version provided"
```
```suggestion
        [[ -z "${E2E_CLUSTER_VERSION}" ]] && fail_test "Unsupported cluster version '${E2E_CLUSTER_VERSION}' provided"
```

scripts/e2e-tests.sh
```diff
@@ -246,15 +245,22 @@ function create_test_cluster_with_retries() {
...
+      fi
 
-      header "Creating test cluster in $E2E_CLUSTER_REGION $E2E_CLUSTER_ZONE"
+      header "Creating test cluster in $E2E_CLUSTER_REGION $E2E_CLUSTER_ZONE with version $E2E_CLUSTER_VERSION"
```
```suggestion
      header "Creating test cluster ${E2E_CLUSTER_VERSION} in ${E2E_CLUSTER_REGION}"
```

scripts/library.sh
```diff
@@ -502,6 +502,22 @@ function get_latest_knative_yaml_source() {
...
+# Parameters: $1 - prefix of gke versions
+#             $2 - gcp zone
+function get_latest_gke_version() {
+  local version_prefix="$1"
+  local re_version_match="${version_prefix/./\\.}"
+  re_version_match="${version_prefix/-/\\-}"
+  local gcp_zone="$2"
+  #local full_gke_version=
```
```suggestion
```

scripts/library.sh
```diff
@@ -502,6 +502,22 @@ function get_latest_knative_yaml_source() {
...
+  echo "$(gcloud container get-server-config --zone=${gcp_zone} --format="yaml(validMasterVersions)" \
```
adrcunha: Why yaml and not value?

scripts/library.sh
adrcunha: Do you mean

```Returns the latest GKE version in the given [major].[minor] Kubernetes version```

?

File: scripts/library.sh [hunk 9]
```diff
@@ -502,6 +502,22 @@ function get_latest_knative_yaml_source() {
...
+      | grep -E "${re_version_match}" \
```
adrcunha: You can use -F here, since you're matching against a fixed string, not a regex.

scripts/library.sh
```diff
@@ -502,6 +502,22 @@ function get_latest_knative_yaml_source() {
...
+      | grep -Eo "[0-9]+\.[0-9]+\.[0-9]+" \
```
adrcunha: I believe that by using value(validMasterVersions) as the format you won't need this line.

scripts/e2e-tests.sh
chaodaiG: done

scripts/e2e-tests.sh
chaodaiG: done

scripts/e2e-tests.sh
chaodaiG: I thought about it too, but it's an empty string if this condition is true

scripts/e2e-tests.sh
chaodaiG: done

scripts/e2e-tests.sh
chaodaiG: done

scripts/library.sh
chaodaiG: done

scripts/library.sh
chaodaiG: value returns a single line instead of an array

scripts/library.sh
chaodaiG: Actually I forgot to add `^`, updated

scripts/library.sh
chaodaiG: It's needed, as kubetest doesn't support direct output such as `1.12.8-gke.7`, this line is to grep `1.12.8` out of it

chaodaiG: This is also why the function `get_latest_gke_version` returns [major].[minor].[patch]

scripts/library.sh
chaodaiG: It's actually intended to return [major].[minor].[patch], see comment below

scripts/library.sh
chaodaiG: removed

scripts/e2e-tests.sh
```diff
@@ -127,6 +127,21 @@ function save_metadata() {
 EOF
 }
 
+# Returns the latest gke version in [major].[minor].[patch] with given [major](.[minor])? Kubernetes version
+# Parameters: $1 - prefix of gke versions
+#             $2 - gcp zone
+function get_latest_gke_version() {
+  local version_prefix="$1"
```
adrcunha: You can replace these five commands with a single one and inline it:

```
$ gcloud container get-server-config --zone=us-central1 --format="value(validMasterVersions)" 2> /dev/null | tr ';' '\n' | grep -F 1.13 | cut -f1 -d- | sort | tail -1
1.13.6
```

chaodaiG: done

scripts/e2e-tests.sh
```diff
@@ -246,15 +245,21 @@ function create_test_cluster_with_retries() {
...
       [[ "${E2E_CLUSTER_ZONE}" == "${zone_not_provided}" ]] && E2E_CLUSTER_ZONE=""
 
-      local geoflag="--gcp-region=${E2E_CLUSTER_REGION}"
-      [[ -n "${E2E_CLUSTER_ZONE}" ]] && geoflag="--gcp-zone=${E2E_CLUSTER_REGION}-${E2E_CLUSTER_ZONE}"
+      local cluster_creation_zone="${E2E_CLUSTER_REGION}"
+      [[ -n "${E2E_CLUSTER_ZONE}" ]] && cluster_creation_zone="${E2E_CLUSTER_REGION}-${E2E_CLUSTER_ZONE}"
+      # Convert cluster version to a full GKE version if we're not using latest
+      if [[ "${E2E_CLUSTER_VERSION}" =~ ^[0-9](\.[0-9]+)?$ ]]; then
+        E2E_CLUSTER_VERSION="v$(gcloud container get-server-config --zone=${cluster_creation_zone} --format="value(validMasterVersions)" 2> /dev/null \
+          | tr ';' '\n' | grep -E ^${E2E_CLUSTER_VERSION} | cut -f1 -d- | sort | tail -1)"
```
adrcunha: You don't need a regex here. But to be sure, -F is fine.

```suggestion
          | tr ';' '\n' | grep -F ${E2E_CLUSTER_VERSION} | cut -f1 -d- | sort | tail -1)"
```

chaodaiG: The reason I used a regex here is to enforce matching the prefix, so that it doesn't happen that `1.19` matching `1.21.19`

adrcunha: Touché. :D

PR #917, opened 2019-06-12
add pubsub labels to continuous prow job

yt3liu:

Add the following labels to the continuous prow jobs:
```
prow.k8s.io/pubsub-project: knative-tests
prow.k8s.io/pubsub-topic: knative-monitoring
prow.k8s.io/pubsub-runID: <job-name>
```

Fixes #

### DIFF
--- a/ci/prow/make_config.go
+++ b/ci/prow/make_config.go
@@ -763,9 +763,9 @@ func generateGoCoveragePeriodic(title string, repoName string, _ yaml.MapSlice) 
 			fmt.Sprintf("--cov-threshold-percentage=%d", data.Base.GoCoverageThreshold)}
 		data.Base.ServiceAccount = ""
 		addExtraEnvVarsToJob(&data.Base)
-		addLabelToJob(&data.Base, "prow.k8s.io/pubsub-project", "knative-tests")
-		addLabelToJob(&data.Base, "prow.k8s.io/pubsub-topic", "knative-monitoring")
-		addLabelToJob(&data.Base, "prow.k8s.io/pubsub-runID", data.PeriodicJobName)
+		addLabelToJob(&data.Base, "prow.k8s.io/pubsub.project", "knative-tests")
+		addLabelToJob(&data.Base, "prow.k8s.io/pubsub.topic", "knative-monitoring")
+		addLabelToJob(&data.Base, "prow.k8s.io/pubsub.runID", data.PeriodicJobName)
 		configureServiceAccountForJob(&data.Base)
 		executeJobTemplate("periodic go coverage", readTemplate(periodicCustomJob), title, repoName, data.PeriodicJobName, false, data)
 		return
--- a/ci/prow/make_config.go
+++ b/ci/prow/make_config.go
@@ -575,9 +575,6 @@ func generatePeriodic(title string, repoName string, periodicConfig yaml.MapSlic
 			if len(data.Base.Args) == 0 {
 				data.Base.Args = allPresubmitTests
 			}
-			addLabelToJob(&data.Base, "prow.k8s.io/pubsub-project", "knative-tests")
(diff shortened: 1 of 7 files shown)


srinivashegde86:
do we want to try this on a periodic job that doesnt run every 2 hours before adding it to continous runs?

ci/prow/make_config.go
```diff
@@ -568,6 +575,9 @@ func generatePeriodic(title string, repoName string, periodicConfig yaml.MapSlic
 			if len(data.Base.Args) == 0 {
 				data.Base.Args = allPresubmitTests
 			}
+			addLabelToJob(&data.Base, "prow.k8s.io/pubsub-project", "knative-tests")
+			addLabelToJob(&data.Base, "prow.k8s.io/pubsub-topic", "knative-monitoring")
+			addLabelToJob(&data.Base, "prow.k8s.io/pubsub-runID", fmt.Sprintf("ci-%s-%s", data.Base.RepoNameForJob, jobNameSuffix))
```
steuhs: You are trying to make this string the same as the job here, right? Can you use just take the job name, instead? What if the job name changed to other format than fmt.Sprintf("ci-%s-%s", data.Base.RepoNameForJob, jobNameSuffix)

yt3liu:
> do we want to try this on a periodic job that doesnt run every 2 hours before adding it to continous runs?

@srinivashegde86  That's a good point.  Do you have a job in mind? 

Maybe `nightly`? Or do you prefer a non-periodic job? 
I think there might be a lot of presubmit jobs created during work hours, and may not be ideal for initial testing.

ci/prow/make_config.go
yt3liu: The job name is not populated until much later after it determines the job type. I added the labels here with duplicated job name logic because I don't want to change the code flow for the initial testing as we may expand the scope to label all periodic jobs. 

I can move the name logic if we are not going to label all periodic jobs in the future. Currently, the periodic job types are `continuous`, `nightly`, `branch-ci`, `dot-release`, `auto-release`, `performance`, `performance-mesh`, `latency`, `custom-job`, `cron`, `release`, `webhook-apicoverage`. 

Do you know if there's a job type from the above list that should not be labelled?

steuhs: why don't we just use func getTestGroupName(repoName string, jobName string). In this way if there is any change in the future we don't need to deal with bugs derived from discrepancy.

We actually need the jobname to be returned from pubsub, because we need to store that in the db. If we don't return it through runID here then we need to change crier's implementation to export it.

Sorry, I don't understand your 2nd and 3rd paragraph. You can talk with me offline

srinivashegde86:
> Or do you prefer a non-periodic job?

I am not sure :) How about the go coverage job? - https://github.com/knative/test-infra/blob/master/ci/prow/config.yaml#L2339

ci/prow/make_config.go
yt3liu: Changed the labelled job to go-coverage which is not part of this method. Let's decide on which jobs to label later.

srinivashegde86:
/cc @dushyanthsc 
/cc @chaodaiG 

can one of you please take a look at the config changes here and add the lgtm label? I am not as familiar with them

adrcunha:
/approve

You might want to start one of the coverage jobs manually using run_job.sh and immediately inspect it to ensure it behaves the way you want. I don't see any issues with the new labels.

yt3liu:
> /approve
> 
> You might want to start one of the coverage jobs manually using run_job.sh and immediately inspect it to ensure it behaves the way you want. I don't see any issues with the new labels.

Very good point. Just realized I was diffing the wrong file when I was comparing the generated file content by mkpj. Updated the labels and confirmed pubsub subscriber receive message correctly. 
https://prow.knative.dev/rerun?prowjob=f4426cba-8e0f-11e9-8d84-acde48001122

/hold cancel

Issue #763, opened 2019-05-13
Send monitoring alerts when needed


srinivashegde86:
Once we have the config and errorlogs table defined, we should add the alerting using cloud mail component

PR #918, opened 2019-06-13
update alerts table and send alert email conditionally

### LINKED ISSUE #765: Store alerts in alerts table
/assign @steuhs

steuhs:
Complete Step 6 & 7 in design doc. Except the following details in email message:
- Error message
- Number of occurrences
- Jobs affected along with a link to the gcs artifact directory that contains the build logs
- Pull Requests affected

Will complete those in a follow up PR

Fixes #765,  #763(partly)


shared/mysql/client.go
```diff
@@ -67,7 +67,7 @@ func (c DBConfig) TestConn() error {
 }
 
 func (c DBConfig) Connect() (*sql.DB, error) {
-	conn, err := sql.Open(driverName, c.dataStoreName(c.DatabaseName))
+	conn, err := sql.Open(driverName, c.dataStoreName(c.DatabaseName)+"?parseTime=true")
```
srinivashegde86: can you add a comment as to why we need parsetime?

tools/monitoring/config/config.go
```diff
@@ -74,6 +75,11 @@ func (selected SelectedConfig) applyDefaults() {
 	}
 }
 
+// Duration converts the time period stored as minutes int to a Duration object
+func (selected SelectedConfig) Duration() time.Duration {
```
srinivashegde86:
```suggestion
func (s SelectedConfig) Duration() time.Duration {
```

tools/monitoring/mysql/alert.go
```diff
@@ -0,0 +1,131 @@
...
+	"log"
+	"time"
+
+	"github.com/knative/test-infra/tools/monitoring/config"
+	"github.com/knative/test-infra/tools/monitoring/mail"
+)
+
+const (
+	alertInsertStmt = `
+	INSERT INTO Alerts (
```
srinivashegde86:
```suggestion
	  INSERT INTO Alerts (
```

tools/monitoring/mysql/alert.go
```diff
@@ -0,0 +1,131 @@
...
+	log.Printf("sending alert...")
+	subject := fmt.Sprintf("Error pattern reached alerting threshold: %s", errorPattern)
+	body := fmt.Sprintf(emailTemplate, config.Duration(), errorPattern, config.Hint)
+
+	return mailConfig.Send(recipients, subject, body)
+}
+
+// CheckAlertCondition checks whether the given error pattern meets
+// the alert condition specified in config
+func CheckAlertCondition(errorPattern string, config *config.SelectedConfig, db *sql.DB) (bool, error) {
```
srinivashegde86: this seems like a function on the config

tools/monitoring/mysql/alert.go
```diff
@@ -0,0 +1,131 @@
...
+
+	emailTemplate = `In the past %v, 
+The number of occurrences of the following error pattern reached threshold:
+%s
+
+Hint for diagnose & recovery: %s
+`
+)
+
+func sendAlert(errorPattern string, config *config.SelectedConfig, mailConfig *mail.Config, recipients []string) error {
```
srinivashegde86: should we have an embedded struct for mail config and add this method on that?

```
type MailConfig struct {
  * mail.Config
}

func (m *MailConfig) sendalert()...
```

tools/monitoring/mysql/alert.go
```diff
@@ -0,0 +1,131 @@
...
+// CheckAlertCondition checks whether the given error pattern meets
+// the alert condition specified in config
+func CheckAlertCondition(errorPattern string, config *config.SelectedConfig, db *sql.DB) (bool, error) {
+	// the timestamp we want to start collecting logs
+	startTime := time.Now().Add(config.Duration())
+
+	var nMatches, nJobs, nPRs int
+
+	row := db.QueryRow(`
+		SELECT
```
srinivashegde86: move to const?

tools/monitoring/mysql/alert.go
```diff
@@ -0,0 +1,131 @@
...
+
+	if err := row.Scan(&nMatches, &nJobs, &nPRs); err != nil {
+		return false, err
+	}
+
+	return nMatches >= config.Occurrences && nJobs >= config.JobsAffected && nPRs >= config.PrsAffected, nil
+}
+
+// Alert checks alert condition and alerts table and send alert mail conditionally
+func Alert(errorPattern string, config *config.SelectedConfig, db *sql.DB, mailConfig *mail.Config, recipients []string) (bool, error) {
```
srinivashegde86:
```
func (m *MailConfig) Alert()
```

tools/monitoring/mysql/alert.go
```diff
@@ -0,0 +1,131 @@
...
+	alertConditionMet, err := CheckAlertCondition(errorPattern, config, db)
+	if err != nil {
```
srinivashegde86: combine
```suggestion
	if ok, err := CheckAlertCondition(); err != nil  || !ok {
```

tools/monitoring/mysql/alert.go
```diff
@@ -0,0 +1,131 @@
...
+		return false, err
+	}
+	if !alertConditionMet {
+		return false, nil
+	}
+
+	shouldSendMail, err := checkAlertsTable(errorPattern, config.Duration(), db)
```
srinivashegde86: ditto. combine
```suggestion
	if ok, err := checkAlertsTable(errorPattern, config.Duration(), db); err != nil || !ok
```

tools/monitoring/mysql/alert.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		return false, err
+	}
+	if !shouldSendMail {
+		return false, nil
+	}
+	if err = sendAlert(errorPattern, config, mailConfig, recipients); err != nil {
```
srinivashegde86:
```suggestion
	err = senAlert()
	return err == nil, err
```

tools/monitoring/mysql/alert.go
```diff
@@ -0,0 +1,131 @@
...
+		return false, err
+	}
+	return true, nil
+}
+
+// checkAlertsTable checks alert table and see if it is necessary to send alert email
+// also updates the alerts table if email sent
+func checkAlertsTable(errorPattern string, window time.Duration, db *sql.DB) (bool, error) {
```
srinivashegde86:
```func (m *MailConfig)

File: tools/monitoring/mysql/alert.go [hunk 12]
```diff
@@ -0,0 +1,131 @@
...
+	if !shouldSendMail {
+		return false, nil
+	}
+	if err = sendAlert(errorPattern, config, mailConfig, recipients); err != nil {
+		return false, err
+	}
+	return true, nil
+}
+
+// checkAlertsTable checks alert table and see if it is necessary to send alert email
```
srinivashegde86: this should be renamed, as it does more than check alerts table

tools/monitoring/mysql/alert.go
```diff
@@ -0,0 +1,131 @@
...
+// also updates the alerts table if email sent
+func checkAlertsTable(errorPattern string, window time.Duration, db *sql.DB) (bool, error) {
+	var id int
+	var sent time.Time
+
+	row := db.QueryRow(`
+		SELECT ID, Sent
```
srinivashegde86: const

tools/monitoring/mysql/alert.go
```diff
@@ -0,0 +1,131 @@
...
+	return true, nil
+}
+
+// checkAlertsTable checks alert table and see if it is necessary to send alert email
+// also updates the alerts table if email sent
+func checkAlertsTable(errorPattern string, window time.Duration, db *sql.DB) (bool, error) {
+	var id int
+	var sent time.Time
+
+	row := db.QueryRow(`
```
srinivashegde86: this should be queryrows as we may get >1 rows

tools/monitoring/mysql/alert.go
```diff
@@ -0,0 +1,131 @@
...
+		FROM Alerts
+		WHERE ErrorPattern = ?`,
+		errorPattern)
+
+	if err := row.Scan(&id, &sent); err != nil {
+		if err != sql.ErrNoRows {
+			return false, err
+		}
+
+		_, err := db.Query(alertInsertStmt, errorPattern, time.Now())
```
srinivashegde86: why are we inserting into laerts table when there was some err scanning the row?

tools/monitoring/mysql/alert.go
steuhs: what's the advantage for doing that?

tools/monitoring/mysql/alert.go
steuhs: yes, but I cannot think of a good concise name...

tools/monitoring/mysql/alert.go
steuhs: Actually I don't think we should always put queries in constants, especially for the ones that is only associated with a single function. It makes reading harder, because you have to look at two places to understand what is going on. Furthermore, these queries are not really constant, they have variables (?), it is hard to see what variable has what value when we have them in two places

tools/monitoring/mysql/alert.go
steuhs: maybe processAlertTable?
it's a big vague though

tools/monitoring/mysql/alert.go
steuhs: I don't see any content here

tools/monitoring/mysql/alert.go
steuhs: ditto

tools/monitoring/mysql/alert.go
steuhs: ErrorPattern is a unique field

tools/monitoring/mysql/alert.go
steuhs: this handles the error case of "sql.ErrNoRows" only. This just means no matching record

tools/monitoring/mysql/alert.go
srinivashegde86: Its a property of the monitoring mail config object. It should not be used by any other object

tools/monitoring/mysql/alert.go
srinivashegde86: weird. updated

tools/monitoring/mysql/alert.go
srinivashegde86: maybe break it into 2 functions -> checkAlert and update()

tools/monitoring/mysql/alert.go
srinivashegde86: yes, but there can be multiple alerts for the same pattern. One reason can be that we havent cleaned up yet.

tools/monitoring/mysql/alert.go
srinivashegde86: this will be executed when `Scan().err != nil && err != NoRows` which means you had an err running that query or you had 0 rows returned. Shouldnt this be `Scan().err == nil && err != NoRows()`

tools/monitoring/mysql/alert.go
srinivashegde86: ditto for what?

tools/monitoring/mysql/alert.go
steuhs: Got it. We should use update instead of insert when adding an alert whose pattern already exist. In that way it can be guaranteed at most one record is found

tools/monitoring/mysql/alert.go
steuhs: I explain in another comment that why we don't want to put sql queries in const

tools/monitoring/mysql/alert.go
steuhs: Scan().err != nil && err != NoRows which means you had an err running that query AND that error is "you had 0 rows returned."

Scan().err == nil && err != NoRows() doesn't make sense. If the err is nil why do we want to check if err != NoRows()

tools/monitoring/mysql/alert.go
```diff
@@ -0,0 +1,119 @@
...
+		INSERT INTO Alerts (
+			Sent, ErrorPattern
+		) VALUES (?,?)`
+
+	alertUpdateStmt = `
+	UPDATE Alerts
+	SET Sent = (?)
+	WHERE ErrorPattern = (?)`
+
+	emailTemplate = `In the past %v,
```
yt3liu: For the email template, it's missing a few fields. From the design, these are the fields that'll be included: 

* Error message pattern
* Error message
* Number of occurrences
* Jobs affected along with a link to the gcs artifact directory that contains the build logs
* Pull Requests affected
* Recovery suggestions. 

You can try using [text/template](https://golang.org/pkg/text/template/) to generate the error message if the format string gets too long.

steuhs: The missing fields are already mentioned in the PR description. This PR is already complicated and we should do that in a separate PR, with some function rewritten to get the needed fields

I'll look into text template

tools/monitoring/mysql/alert.go
srinivashegde86: Agreed. But, this function is different and should not dependent on the update being done correctly. We should validate that we are only getting one row, if we expect that

tools/monitoring/mysql/alert.go
```diff
@@ -0,0 +1,119 @@
...
+	emailTemplate = `In the past %v, 
+The number of occurrences of the following error pattern reached threshold:
+%s
+
+Hint for diagnose & recovery: %s
+`
+)
+
+type MailConfig struct {
+	mailConfig *mail.Config
```
srinivashegde86:
```suggestion
	*mail.Config
```

tools/monitoring/mysql/alert.go
```diff
@@ -0,0 +1,119 @@
...
+	recipients []string
+}
+
+func (m *MailConfig) sendAlert(errorPattern string, config *config.SelectedConfig) error {
+	log.Printf("sending alert...")
+	subject := fmt.Sprintf("Error pattern reached alerting threshold: %s", errorPattern)
+	body := fmt.Sprintf(emailTemplate, config.Duration(), errorPattern, config.Hint)
+
+	return m.mailConfig.Send(m.recipients, subject, body)
```
srinivashegde86:
```suggestion
	return m.Send(m.recipients, subject, body)
```

tools/monitoring/mysql/alert.go
steuhs: I double checked. It is not possible to have more than one entry with the same UNIQUE field to exist in mysql. There is no way for this to return more than one row

tools/monitoring/mysql/alert.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+
+		// if no record found, instruct to add a record
+		return alertInsertStmt, nil
+	}
+
+	if sent.Add(window).Before(time.Now()) {
+		// if previous alert expires. Instruct to update the timestamp
+		log.Printf("previous alert timestamp=%v expired, alert window size=%v", sent, window)
+		return alertUpdateStmt, nil
```
yt3liu: I don't think we need two different SQL statements for insert/update. We can change the SQL query to handle duplicate and get rid of the need to pass the statement. 

https://dev.mysql.com/doc/refman/8.0/en/insert-on-duplicate.html

tools/monitoring/mysql/alert.go
```diff
@@ -0,0 +1,119 @@
...
+	if err != nil || queryTemplate == "" {
+		return false, err
+	}
+
+	if err := updateAlertsTable(queryTemplate, errorPattern, db); err != nil {
+		return false, err
+	}
+
+	err = m.sendAlert(errorPattern, config)
+	return err == nil, err
```
yt3liu: Why return `err == nil`? For all cases where it returns err, it always return false. I think you can change the signature to `func (m *MailConfig) Alert(errorPattern string, config *config.SelectedConfig, db *sql.DB) error` and get rid of the pair-value return.

tools/monitoring/mysql/alert.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+// checkAlertsTable checks alert table and see if it is necessary to send alert email
+// returns a sql statement that contains db update action to perform. An empty string indicates
+// no db operations to perform and no alert email needs to be sent
+func checkAlertsTable(errorPattern string, window time.Duration, db *sql.DB) (string, error) {
```
srinivashegde86: why are returning the SQL query? We should just return err. Check Joyce's comment about using ON DUPLICATE clause

tools/monitoring/config/config.go
```diff
@@ -59,21 +61,50 @@ type SelectedConfig struct {
...
+}
+
+// CheckAlertCondition checks whether the given error pattern meets
+// the alert condition specified in config
+func (s *SelectedConfig) CheckAlertCondition(errorPattern string, db *sql.DB) (bool, error) {
+	// the timestamp we want to start collecting logs
+	startTime := time.Now().Add(s.Duration())
+
+	var nMatches, nJobs, nPRs int
+
```
srinivashegde86: remove extra line

tools/monitoring/mysql/alert.go
steuhs: the bool is returns whether an alert is sent. It can be false even when there is no error

yt3liu: What is the bool value used for after it's returned from `Alert`? 

Also, I don't think it's a good idea to send the email from mysql package. I think the mail message generation can be moved to `mail` package and the actual sending email can be moved to the main app.

For the sql package, it can return the bool value on sending or not sending the alert.

steuhs: I think it is important to expose whether an alert is actually sent instead of keeping it within the function
I agree with your 2nd paragraph. In fact I don't think we need an mysql package at all, because we have sql calls in other packages as well and we already moved the common sql functionalities to the shared folder. Furthermore, I think we can move the mail package to shared folder as well, because none of the content their is coupled with monitoring
I prefer to get this checked in first before we do more refactoring. I have refactoring done locally, but if I upload it to this PR there will be a lot of change.

PR #922, opened 2019-06-13
e2e test meta data works with non latest version

chaodaiG:
Forgot to add in PR #905 . Since cluster versions support non `latest`,  `save_metadata` should also change to correctly reflect the version

### DIFF
--- a/scripts/e2e-tests.sh
+++ b/scripts/e2e-tests.sh
@@ -115,7 +115,9 @@ function save_metadata() {
     geo_key="Zone"
     geo_value="${E2E_CLUSTER_REGION}-${E2E_CLUSTER_ZONE}"
   fi
-  local cluster_version="$(gcloud container clusters list --project=${E2E_PROJECT_ID} --format='value(currentMasterVersion)')"
+  local cluster_version="${E2E_CLUSTER_VERSION}"
+  [[ "${cluster_version}" == "gke-latest" ]] && \
+    cluster_version="$(gcloud container clusters list --project=${E2E_PROJECT_ID} --format='value(currentMasterVersion)')"
   cat << EOF > ${ARTIFACTS}/metadata.json
 {
   "E2E:${geo_key}": "${geo_value}",


scripts/e2e-tests.sh
```diff
@@ -115,7 +115,9 @@ function save_metadata() {
     geo_key="Zone"
     geo_value="${E2E_CLUSTER_REGION}-${E2E_CLUSTER_ZONE}"
   fi
-  local cluster_version="$(gcloud container clusters list --project=${E2E_PROJECT_ID} --format='value(currentMasterVersion)')"
+  local cluster_version="${E2E_CLUSTER_VERSION}"
+  [[ "${cluster_version}" == "gke-latest" ]] && \
```
adrcunha: This gcloud command should correctly report the full version of the test cluster, and your PR is just saving such gcloud command (at the expense of not reporting the gke variant), which is not a performance issue. What am I missing?

chaodaiG: Sorry I didn't quite get what you meant. This PR itself doesn't alter the behavior when `cluster_version` is not passed in as parameter. I just checked one of the job on gubernator that the `E2E:Version` is `1.13.6-gke.6`. And this PR would make this value `1.13.6` instead of full gke variant. I will explore how to get this fixed in this PR

adrcunha: That's my question: why restricting to 1.13.6 instead of the full gke variant?

chaodaiG: That's because kubetest doesn't take this [extract_k8s.go](https://github.com/kubernetes/test-infra/blob/d460c4a1a46fd8528a3a8f8f79b78d073207f9c8/kubetest/extract_k8s.go#L82), I have tried to update it to accept full gke variant, but couldn't make it work due to `--extract` doesn't work with gke variant.

adrcunha: Looks like we're talking about different things. Let's sync up offline.

chaodaiG: @adrcunha , I just did a test ran and you're right, closing this PR.

Issue #428, opened 2019-01-28
Show summary of build failures in gubernator


adrcunha:
Currently the build tests show the errors in the log, but nothing in the summary on gubernator. It's easy for devs to be without a clue as to what failed in the build tests. Now that we have support for generating junit XML files with test status, we can build a simple tool to properly summarize build failures so it's reported by gubernator.

adrcunha:
This is specially true for markdown linting errors, which are not displayed at all in gubernator's summary.

PR #923, opened 2019-06-13
Surface build errors on Gubernator

adrcunha:
Part of #428.


adrcunha:
WAI when forcing a build error for testing purposes:

image

PR #933, opened 2019-06-14
Increase boskos janitor count to match number of boskos projects.

coryrc:
We had stockout. If they happen again, it will have to be a different reason.

/lint
/assign chaodaiG

### DIFF
--- a/ci/prow/boskos/config.yaml
+++ b/ci/prow/boskos/config.yaml
@@ -104,7 +104,7 @@ metadata:
     app: boskos-janitor
   namespace: test-pods
 spec:
-  replicas: 20  # Number distributed janitor instances
+  replicas: 20  # Number of distributed janitor instances
   template:
     metadata:
       labels:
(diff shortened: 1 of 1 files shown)


ci/prow/boskos/config.yaml
```diff
@@ -104,7 +104,7 @@ metadata:
     app: boskos-janitor
   namespace: test-pods
 spec:
-  replicas: 12  # 12 distributed janitor instances
+  replicas: 40  # This number distributed janitor instances. Should keep around the same number as number of projects.
```
srinivashegde86: I believe 40 is high number. Do we know how long the resource contention happened for? We need the higher number of boskos projects as they run for ~30 min in presubmits and >1 hr on periodic (perf tests run for 2 hrs). But, the janitors dont run as long, and I dont think we need one per project. Maybe try with half, 20?

adrcunha:
Nevermind.

/lgtm
/approve

ci/prow/boskos/config.yaml
coryrc: Set to 20.

PR #948, opened 2019-06-17
Add an argument --knative-version for users to specify the knative version

### LINKED ISSUE #897: Allow the version of the Knative installation be configurable

houshengbo:
We are only able to install Knative based on the latest source code, and it will provide great help to other Knative modules for testing, if we can customize which previous release of Knative we can install.

Fixes #897

### DIFF
--- a/scripts/e2e-tests.sh
+++ b/scripts/e2e-tests.sh
@@ -393,6 +393,7 @@ EMIT_METRICS=0
 SKIP_KNATIVE_SETUP=0
 SKIP_ISTIO_ADDON=0
 GCP_PROJECT=""
+KNATIVE_VERSION="latest"
 E2E_SCRIPT=""
 E2E_CLUSTER_VERSION=""
 GKE_ADDONS=""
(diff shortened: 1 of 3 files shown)


adrcunha:
Can you please describe your strategy? You're creating a flag and setting 2 environment variables that are not used anywhere. What's your plan?

/hold

houshengbo:
@adrcunha 
I am fixing the issue in knative client: https://github.com/knative/client/issues/168
regarding running the integration against latest knative serving and a previous release.

I need to make sure at least  the version is configurable, so that we can specify which version of knative we can install.

adrcunha:
You don't have to make changes here. If you do, they should be self contained. I suggest you add a custom flag for knative/client, like https://github.com/knative/serving/blob/master/test/e2e-common.sh#L42

Also notice that your changes won't produce the effect you intend, as your new env vars are not used anywhere. And setting $KNATIVE_SERVING_VERSION in the job won't work either, as it will be overwritten by library.sh. Better use the custom flag handling in knative/client as I suggested, and then use that to proper call start_release_knative_serving().

houshengbo:
Thanks, I did not realize there is a magic parse_flags.

PR #953, opened 2019-06-17
Add jobs testing different k8s versions

chaodaiG:
Testing different versions combinations of k8s and istio

part of: #871

### DIFF
--- a/ci/prow/config_knative.yaml
+++ b/ci/prow/config_knative.yaml
@@ -127,6 +127,14 @@ periodics:
       full-command: "./test/e2e-tests.sh --istio-version 1.1.7 --mesh"
     - custom-job: istio-1.1.7-no-mesh
       full-command: "./test/e2e-tests.sh --istio-version 1.1.7 --no-mesh"
+    - custom-job: k8s-1.12-istio-1.1
+      full-command: "./test/e2e-tests.sh --cluster-version \"1.12\" --istio-version 1.1-latest"
+    - custom-job: k8s-1.12-istio-1.0
+      full-command: "./test/e2e-tests.sh --cluster-version \"1.12\" --istio-version 1.0-latest"
+    - custom-job: k8s-1.11-istio-1.1
+      full-command: "./test/e2e-tests.sh --cluster-version \"1.11\" --istio-version 1.1-latest"
+    - custom-job: k8s-1.11-istio-1.0
+      full-command: "./test/e2e-tests.sh --cluster-version \"1.11\" --istio-version 1.0-latest"
     - nightly: true
     - dot-release: true
     - auto-release: true
--- a/ci/prow/config_knative.yaml
+++ b/ci/prow/config_knative.yaml
@@ -128,13 +128,14 @@ periodics:
     - custom-job: istio-1.1.7-no-mesh
       full-command: "./test/e2e-tests.sh --istio-version 1.1.7 --no-mesh"
     - custom-job: k8s-1.12-istio-1.1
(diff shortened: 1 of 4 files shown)


chaodaiG:
/hold

wait until https://github.com/knative/serving/pull/4372 is merged

ci/testgrid/config.yaml
```diff
@@ -255,6 +271,18 @@ dashboards:
   - name: istio-1.1.7-no-mesh
     test_group_name: ci-knative-serving-istio-1.1.7-no-mesh
     base_options: "sort-by-name="
+  - name: k8s-1.12-istio-1.1
```
adrcunha: I think we should start thinking about grouping all these variations in a separate tab/group...

ci/prow/config_knative.yaml
```diff
@@ -127,6 +127,14 @@ periodics:
       full-command: "./test/e2e-tests.sh --istio-version 1.1.7 --mesh"
     - custom-job: istio-1.1.7-no-mesh
       full-command: "./test/e2e-tests.sh --istio-version 1.1.7 --no-mesh"
+    - custom-job: k8s-1.12-istio-1.1
+      full-command: "./test/e2e-tests.sh --cluster-version \"1.12\" --istio-version 1.1-latest"
```
adrcunha: You can drop the escaped quotes for the cluster version, it will increase readability. Here and everywhere.

ci/prow/make_config.go
```diff
@@ -1214,7 +1214,8 @@ func generateTestGroup(projName string, repoName string, jobNames []string) {
 			extras["short_text_metric"] = "coverage"
 			// Do not alert on coverage failures (i.e., coverage below threshold)
 			extras["num_failures_to_alert"] = "9999"
-		case "istio-1.0.7-mesh", "istio-1.0.7-no-mesh", "istio-1.1.7-mesh", "istio-1.1.7-no-mesh":
+		case "istio-1.0.7-mesh", "istio-1.0.7-no-mesh", "istio-1.1.7-mesh", "istio-1.1.7-no-mesh",
```
adrcunha: This is getting worse, we should definitely address this sooner than later. :)

chaodaiG: Totally agree. As we discussed earlier, we can reuse the data already parsed while generating Prow config to generate testgrid config, which requires non trivial amount of refactoring

ci/prow/config_knative.yaml
adrcunha: If `isNum` returns true for "./test/e2e-tests.sh --cluster-version \"1.12\" --istio-version 1.1-latest", it's buggy.

ci/prow/make_config.go
adrcunha: Indeed.

ci/prow/config_knative.yaml
chaodaiG: It returns true for "1.12"

adrcunha: Ah. It's because of https://github.com/knative/test-infra/blob/master/ci/prow/make_config.go#L523

This is tricky. I'm not sure if single quotes would work here, but definitely add a comment about the need of quoting the value.

chaodaiG: Single quote doesn't work as is, it would work if change this line to also return if contains single quote: https://github.com/knative/test-infra/blob/d8e74c06fd5a0d54fe246d8c51329b3b103ba135/ci/prow/make_config.go#L907

I can try to make it work

chaodaiG: done, please take a look

adrcunha:
/approve
/lgtm

Thanks

PR #956, opened 2019-06-18
Update dependencies

### LINKED ISSUE #957: Issue for two projects using different version of dependencies
When I worked with `eventing` project, I met a dependency issue, but I'm not sure what is the best practice to solve it. So I create an issue here to keep a record.

The problem and steps to reproduce are described as below:
Both `eventing` and `test-infra` depend on `cloud.google.com/go`, but the versions are different - `v0.39.0` for `eventing` and `v0.33.1` for `test-infra`.
In `test-infra`, only code under `shared` directory is referencing code in `cloud.google.com/go`, so it worked fine if `eventing` do not use code in `test-infra/shared`.
But then I wanted to use `test-infra/shared` in `eventing`, both projects would have to depend on `cloud.google.com/go` `v0.39.0`. However, this version is not compatible with the transitive dependencies `test-infra` relies on - `google.golang.org/api` and `google.golang.org/grpc`, and things become more complicated as the two are also transitive dependencies of `eventing`, but different versions.

This issue happened when I tried to add `test-infra/shared` as a dependency in `eventing`, so I think it should be solved in `test-infra`. I created #956 as an attempt, but I'm not sure if it's the best practice.

This issue leads me to think about a scenario that seems even more annoying:
What if `test-infra` uses a function that exists in `cloud.google.com/go` `v0.33.1` but not in `v0.39.0`? Then `eventing` will not be able to use `test-infra/shared` anymore. And `serving` will not be able to update `cloud.google.com/go` to `v0.39.0`, as it's currently using `test-infra/shared`.

Fredy-Z:
1. Update `cloud.google.com/go` to `v0.39.0` as this is the version that `eventing` relies on.
2. Update other transitive dependencies, including `google.golang.org/api` and `google.golang.org/grpc` to solve build error.

Fixes #957


Gopkg.toml
```diff
@@ -15,3 +15,7 @@ required = [
 [[prune.project]]
   name = "github.com/google/licenseclassifier"
   non-go = false
+
+[[constraint]]
+  name = "cloud.google.com/go"
+  version = "0.39.0"
```
chaodaiG: I'm not seeing this version in eventing https://github.com/knative/eventing/blob/master/Gopkg.toml
Is it the latest release version?

Fredy-Z: The latest version was `v0.39.0` when eventing updated `cloud.google.com/go` a couple of days ago, but it's now `v0.40.0`. I added this constraint to ensure the consistency. But again, as I mentioned in https://github.com/knative/test-infra/issues/957, I'm not sure if it's the best practice, and I'm even not sure if we should update dependencies here in this repo. 
Any thoughts or better ideas for this?

adrcunha: You shouldn't need to update it here, unless you're using some new feature in eventing that's conflicting with code here (or avoiding a particular version). Is it the only solution?

Fredy-Z: I also tried to directly update dependencies in `eventing` and it looked fine, see https://github.com/knative/eventing/pull/1411. 
So maybe this PR is not needed?
Such issues are unavoidable if we have different version of dependencies in each project, unless there is one single place managing the dependencies for all repos under Knative projects, but it seems impossible under the current dependency model.

adrcunha: If you solved the issue in eventing, so I don't see why unnecessarily restricting a version here.

Fredy-Z: Yeah that's true, I'll close this PR.

Issue #926, opened 2019-06-13
Ensure jobname and tab mapping in testgrid package are up-to-date


chaodaiG:
[testgrid package](https://github.com/knative/test-infra/tree/master/shared/testgrid) contains hardcoded mapping of jobname: tab, which could become outdated when someone updates testgrid config yaml forgot to update in testgrid package. This needs to be enforced as it could fail tools depending testgrid package. 

Suggesting unit test instead of dynamically parsing it at runtime, since this package belong to `shared`, and if it's vendored in another repo it may not have access to testgrid config yaml.

PR #958, opened 2019-06-18
Unit test against testgrid package inconsistent with testgrid config …

chaodaiG:
Inside `share/testgrid` package, there are hardcoded testgroup name with testgrid tab URL, this can be broken by updating of either side but not both sides. Add unit test to guard against this

Fix: #926

### DIFF
--- a/shared/common/common.go
+++ b/shared/common/common.go
@@ -19,6 +19,8 @@ package common
 import (
 	"fmt"
 	"os"
+	"os/exec"
+	"strings"
 )
 
 const allUsersFullPermission = 0777
@@ -38,4 +40,23 @@ func CreateDirWithFileMode(dirPath string, perm os.FileMode) error {
 		}
 	}
 	return nil
+}
+
+// GetRootDir gets directory of git root
+func GetRootDir() (string, error) {
+	cmd := exec.Command("git", "rev-parse", "--show-toplevel")
+	output, err := cmd.Output()
+	if err != nil {
+		return "", err
+	}
+	return strings.TrimSpace(string(output)), nil
+}
+
+// CDToRootDir change directory to git root dir
+func CDToRootDir() error {
+	d, err := GetRootDir()
+	if nil != err {
+		return err
+	}
+	return os.Chdir(d)
 }
--- a/shared/testgrid/yaml.go
+++ b/shared/testgrid/yaml.go
@@ -27,8 +27,8 @@ import (
 
 const configPath = "ci/testgrid/config.yaml"
 
-// AllConfig is entire testgrid config
-type AllConfig struct {
+// Config is entire testgrid config
+type Config struct {
 	Dashboards []Dashboard `yaml:"dashboards"`
 }
 
@@ -45,7 +45,7 @@ type Tab struct {
 }
 
 // NewConfig loads from default config
-func NewConfig() (*AllConfig, error) {
+func NewConfig() (*Config, error) {
 	root, err := common.GetRootDir()
 	if nil != err {
 		return nil, err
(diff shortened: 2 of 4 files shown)


shared/common/common.go
```diff
@@ -39,3 +41,22 @@ func CreateDirWithFileMode(dirPath string, perm os.FileMode) error {
...
+	cmd := exec.Command("git", "rev-parse", "--show-toplevel")
+	output, err := cmd.Output()
+	if err != nil {
+		return "", err
+	}
+	return strings.TrimSpace(string(output)), nil
+}
+
+// CDToRootDir change directory to git root dir
+func CDToRootDir() error {
```
srinivashegde86: changeToRootDir or MoveToRootDir?

shared/testgrid/yaml.go
```diff
@@ -0,0 +1,77 @@
...
+	"path"
+
+	"github.com/knative/test-infra/shared/common"
+	"gopkg.in/yaml.v2"
+)
+
+const configPath = "ci/testgrid/config.yaml"
+
+// AllConfig is entire testgrid config
+type AllConfig struct {
```
srinivashegde86:
```suggestion
type Config struct {
```

shared/testgrid/yaml.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+// NewConfigFromFile loads config from file
+func NewConfigFromFile(fp string) (*AllConfig, error) {
+	ac := &AllConfig{}
+	contents, err := ioutil.ReadFile(fp)
+	if nil == err {
+		err = yaml.Unmarshal(contents, ac)
+	}
+	return ac, err
```
srinivashegde86:
```suggestion
	return nil, err
```

shared/testgrid/yaml.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+	return NewConfigFromFile(path.Join(root, configPath))
+}
+
+// NewConfigFromFile loads config from file
+func NewConfigFromFile(fp string) (*AllConfig, error) {
+	ac := &AllConfig{}
+	contents, err := ioutil.ReadFile(fp)
+	if nil == err {
+		err = yaml.Unmarshal(contents, ac)
```
srinivashegde86: ac can be inlined

shared/testgrid/yaml.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+	return ac, err
+}
+
+// GetTabRelURL finds URL relative to testgrid home URL from testgroup name
+// (generally this is prow job name)
+func (ac *AllConfig) GetTabRelURL(tgName string) (string, error) {
```
srinivashegde86:
```suggestion
func (ac *AllConfig) GetTabURL(tgName string) (string, error) {
```

shared/testgrid/yaml.go
chaodaiG: done

shared/testgrid/yaml.go
chaodaiG: how can it be returned if it's inlined?

shared/testgrid/yaml.go
chaodaiG: same as above, need to return Config

shared/testgrid/yaml.go
chaodaiG: I would prefer the old name since it is really a relative URL, we may need to add GetTabURL function in the future if we need the full URL

shared/common/common.go
chaodaiG: I know CD is not a good verb in a function, but among the others, CD feels more natural to me to be honest

shared/testgrid/yaml.go
srinivashegde86: you need to return a pointer to Config. Why cant that be nil? If there was an err reading the file, I will expect err != nill && config == nil. What is the point of returning an empty config?

srinivashegde86: ah. sorry, I misread this

chaodaiG: actually I think you are right, pushed update

PR #969, opened 2019-06-19
mock default_build_test_runner so unit test doesn't run build test again

chaodaiG:


test/unit/presubmit-tests.sh
```diff
@@ -42,6 +42,10 @@ function init_test_env() {
   function check_links_in_markdown() {
     return 0
   }
+
+  function default_build_test_runner() {
+    return 0
```
adrcunha: You're not checking that the default runner was executed...

chaodaiG: Added another test for it

PR #972, opened 2019-06-19
Moving all PR Serving and Client tests to us-west1

dushyanthsc:
We are observing latency issues and stockout issue with us-central1. This change moves some of the PR tests to use us-west1 region.


chaodaiG:
this is going to fail build test, such a change should be done in template file and generated with `make config`

dushyanthsc:
we discussed this further, it looks like round-robin makes more sense, Closing this PR and making that change in a separate PR.

PR #979, opened 2019-06-20
Reverting kubekins-e2e image

dushyanthsc:
This CL is reverting the change made here: https://github.com/knative/test-infra/commit/9320a359edc4e18c1332e133395bc16230a1f6c3

Kubectl 1.14 is not needed at the moment. So reverting the change as we have seen couple of issues with the updated base image.


adrcunha:
/approve
/lgtm
/hold

Let's test this first.

PR #982, opened 2019-06-20
Add perf testing job for eventing

Fredy-Z:
Add a performance testing job for Eventing.

As part of https://github.com/knative/eventing/issues/939

### DIFF
--- a/ci/prow/make_config.go
+++ b/ci/prow/make_config.go
@@ -676,9 +676,16 @@ func generatePeriodic(title string, repoName string, periodicConfig yaml.MapSlic
 			jobType = getString(item.Key)
 			jobNameSuffix = getString(item.Key)
 			data.Base.Command = performanceScript
-			// We need a larger cluster of at least 16 nodes for perf tests
-			addEnvToJob(&data.Base, "E2E_MIN_CLUSTER_NODES", "16")
-			addEnvToJob(&data.Base, "E2E_MAX_CLUSTER_NODES", "16")
+			// TODO(Fredy-Z): eventing is in the early stage of performance testing, so we use only one node to save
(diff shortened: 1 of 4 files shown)


ci/prow/make_config.go
```diff
@@ -676,9 +676,16 @@ func generatePeriodic(title string, repoName string, periodicConfig yaml.MapSlic
 			jobType = getString(item.Key)
 			jobNameSuffix = getString(item.Key)
 			data.Base.Command = performanceScript
-			// We need a larger cluster of at least 16 nodes for perf tests
-			addEnvToJob(&data.Base, "E2E_MIN_CLUSTER_NODES", "16")
-			addEnvToJob(&data.Base, "E2E_MAX_CLUSTER_NODES", "16")
+			// TODO(Fredy-Z): eventing is in the early stage of performance testing, so we use only one node to save
+			//                resource. Once it has more complex testing scenarios, this separate logic can be removed.
```
adrcunha: Unless 16 nodes break your test, this is unnecessary and just adds complexity to the code.

Fredy-Z: 16 nodes is a waste of resource for the current stage of Eventing performance testing.
And according to my meeting with developers, they want to minimize the network delay in this very first test case, adding more nodes will bring up more side effects.
Also, this piece of code will not live long here as in the next release we will have more complex test scenarios.

adrcunha: Then "waste of resources" is not the main reason here, network delay is. To simplify this, override the min/max values in your script instead of here.

Fredy-Z: Done. And opened https://github.com/knative/eventing/pull/1443 in eventing.

Fredy-Z:
It had conflicts with another PR from Chao. Please lgtm again, thanks.

PR #988, opened 2019-06-20
fixing continuous error

chaodaiG:
test-infra continuous flow was enabled and failed due to flag `--emitmetrics` not defined. This PR fixes this error

example: https://gubernator.knative.dev/build/knative-prow/logs/ci-knative-test-infra-continuous/1141748184683384833/

### DIFF
--- a/test/e2e/flag.go
+++ b/test/e2e/flag.go
@@ -1,0 +1,38 @@
+// +build e2e
+
+/*
+Copyright 2019 The Knative Authors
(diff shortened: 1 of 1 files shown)


adrcunha:
Why not just defining such flag as a no-op instead of unnecessarily adding stuff we'll probably won't need?

chaodaiG:
> Why not just defining such flag as a no-op instead of unnecessarily adding stuff we'll probably won't need?

updated

PR #992, opened 2019-06-20
add shared JSON report library and client in flaky-test-reporter

### LINKED ISSUE #984: flaky re-trier: Add concise JSON logging to flaky-test-reporter
This is the first step in the implementation of the flaky test re-trier. 

We need to report flaky tests to a location where we can programmatically analyze the results, for the implementation of the actual tool. This will be done using the existing artifacts directory for storing pre-processed results.

This should be done in a shared library, so we can reuse it on the re-trier side.

TrevorFarrelly:
Step 1 for implementation of the flaky test re-trier. Adds a shared `jsonreport` library for reading and writing flaky data to the artifacts directory, and implements a flaky-test-reporter client using this library.

Fixes #984 

I have not written unit tests for the library because I am not sure the right way to test a tool which reads/writes files to local directories.

### DIFF
--- a/shared/jsonreport/jsonreport.go
+++ b/shared/jsonreport/jsonreport.go
@@ -1,0 +1,70 @@
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package jsonreport
+
+import (
+	"encoding/json"
+	"io/ioutil"
+	"path"
+
+	"github.com/knative/test-infra/shared/common"
+	"github.com/knative/test-infra/shared/prow"
+)
+
+const (
+	filename = "flaky-tests.json"
+)
+
+type Report struct {
+	Repo  string   `json:"-"`
+	Flaky []string `json:"flaky"`
+}
+
+func (r *Report) WriteToArtifactsDir() error {
+	artifactsDir := prow.GetLocalArtifactsDir()
+	err := common.CreateDir(path.Join(artifactsDir, r.Repo))
+	if nil != err {
+		return err
+	}
+	outFilePath := path.Join(artifactsDir, r.Repo, filename)
+	contents, err := json.Marshal(r)
+	if nil != err {
+		return err
+	}
+	return ioutil.WriteFile(outFilePath, contents, 0644)
+}
+
+func GetReportForRepo(repo string) (*Report, error) {
+	var report Report
+	artifactsDir := prow.GetLocalArtifactsDir()
+	content, err := ioutil.ReadFile(path.Join(artifactsDir, repo, filename))
+	if err != nil {
+		return nil, err
+	}
+	err = json.Unmarshal(content, &report)
+	if err != nil {
+		return nil, err
+	}
+	return &report, nil
+}
+
+func NewReport(repo string, flaky []string) *Report {
+	return &Report{
+		Repo:  repo,
+		Flaky: flaky,
+	}
+}
--- a/shared/jsonreport/jsonreport.go
+++ b/shared/jsonreport/jsonreport.go
@@ -49,7 +49,9 @@ func (r *Report) WriteToArtifactsDir() error {
 }
 
 func GetReportForRepo(repo string) (*Report, error) {
-	var report Report
+	report := &Report{
+    Repo: repo,
+  }
 	artifactsDir := prow.GetLocalArtifactsDir()
 	content, err := ioutil.ReadFile(path.Join(artifactsDir, repo, filename))
 	if err != nil {
@@ -59,7 +61,7 @@ func GetReportForRepo(repo string) (*Report, error) {
 	if err != nil {
 		return nil, err
 	}
-	return &report, nil
+	return report, nil
 }
 
 func NewReport(repo string, flaky []string) *Report {
--- a/tools/flaky-test-reporter/json_reporter.go
+++ b/tools/flaky-test-reporter/json_reporter.go
@@ -21,7 +21,7 @@ import (
 	"log"
 	"sync"
 
-	"github.com/TrevorFarrelly/test-infra/shared/jsonreport"
+	"github.com/knative/test-infra/shared/jsonreport"
 )
 
 func getFlakyTestSet(repoDataAll []*RepoData) map[string]map[string]bool {
--- a/tools/flaky-test-reporter/json_reporter.go
+++ b/tools/flaky-test-reporter/json_reporter.go
@@ -24,8 +24,11 @@ import (
 	"github.com/knative/test-infra/tools/flaky-test-reporter/jsonreport"
 )
 
+// when reporting on all flaky tests in a repo, we want to eliminate the "job" layer, compressing all flaky
+// tests in that repo into a single list. There can be duplicate tests across jobs, though, so we store tests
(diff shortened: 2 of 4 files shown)


shared/jsonreport/jsonreport.go
```diff
@@ -0,0 +1,72 @@
+/*
```
chaodaiG: Sorry I didn't make it clear. I think this package should be under flaky-test-reporter (this is fine in go, i.e. https://github.com/knative/test-infra/tree/master/tools/monitoring), as the logic is so close.

shared/jsonreport/jsonreport.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	outFilePath := path.Join(artifactsDir, r.Repo, filename)
+	contents, err := json.Marshal(r)
+	if nil != err {
+		return err
+	}
+	return ioutil.WriteFile(outFilePath, contents, 0644)
+}
+
+func GetReportForRepo(repo string) (*Report, error) {
```
chaodaiG: I like this api

shared/jsonreport/jsonreport.go
```diff
@@ -0,0 +1,72 @@
...
+const (
+	filename = "flaky-tests.json"
+)
+
+type Report struct {
+	Repo  string   `json:"-"`
+	Flaky []string `json:"flaky"`
+}
+
+func (r *Report) WriteToArtifactsDir() error {
```
chaodaiG: this should be private?

shared/jsonreport/jsonreport.go
```diff
@@ -0,0 +1,72 @@
...
+		return nil, err
+	}
+	err = json.Unmarshal(content, &report)
+	if err != nil {
+		return nil, err
+	}
+	return report, nil
+}
+
+func NewReport(repo string, flaky []string) *Report {
```
chaodaiG: Maybe a `CreateReportForRepo` api?

shared/jsonreport/jsonreport.go
chaodaiG: I think this function should take repo name(org name can be emitted for now), buildID(default to latest), and read from GCS via Prow lib instead of read from local

tools/flaky-test-reporter/json_reporter.go
```diff
@@ -0,0 +1,73 @@
...
+
+import (
+	"fmt"
+	"log"
+	"sync"
+
+	"github.com/knative/test-infra/tools/flaky-test-reporter/jsonreport"
+)
+
+func getFlakyTestSet(repoDataAll []*RepoData) map[string]map[string]bool {
```
srinivashegde86: can you add comment to this func and the expected output

tools/flaky-test-reporter/jsonreport/jsonreport.go
```diff
@@ -0,0 +1,94 @@
...
+	return report, nil
+}
+
+func CreateReportForRepo(repo string, flaky []string, writeFile bool) (*Report, error) {
+	report := &Report{
+		Repo:  repo,
+		Flaky: flaky,
+	}
+	if writeFile {
+		if err := report.writeToArtifactsDir(); err != nil {
```
srinivashegde86:
```suggestion
		return report, report.writeToArtifactsDir()
```

tools/flaky-test-reporter/jsonreport/jsonreport.go
```diff
@@ -49,46 +52,44 @@ func (r *Report) writeToArtifactsDir() error {
...
-	} else {
-		return job.NewBuild(buildID), nil
-	}
+func getJSONForBuild(repo string, buildID int) ([]byte, error) {
+  relPath := path.Join(prow.ArtifactsDir, repo, filename)
+  job := prow.NewJob(jobName, prow.PeriodicJob, "", 0)
+  if buildID != -1 { // if buildID is specified, use it regardless of if it exists
+    return job.NewBuild(buildID).ReadFile(relPath)
+  }
+  for build := range job.GetLatestBuilds(buildCount) { // otherwise find most recent build with a report
```
chaodaiG: Suggest you use `GetLatestBuildNumber` instead, which is much more performant, see this example: https://github.com/knative/test-infra/blob/88a9875993cb7b24de5598265e42edf82521fe36/tools/flaky-test-reporter/result.go#L249

tools/flaky-test-reporter/jsonreport/jsonreport.go
```diff
@@ -0,0 +1,119 @@
...
+
+// inexpensively sort and find the most recent JSON file. Assumes build IDs are in sequential order.
+func getLatestValidJSON(job *prow.Job, path string) ([]byte, error) {
+	maxElapsedTime, _ := time.ParseDuration(fmt.Sprintf("%dh", maxAge*24))
+	buildIDs := job.GetBuildIDs()
+	sort.Sort(sort.Reverse(sort.IntSlice(buildIDs)))
+	for _, buildID := range buildIDs {
+		build := job.NewBuild(buildID)
+		// check for JSON
+		if contents, err := build.ReadFile(path); err != nil {
```
chaodaiG: I know it's gonna work, would prefer a var named differently from an imported lib

tools/flaky-test-reporter/jsonreport/jsonreport.go
```diff
@@ -0,0 +1,119 @@
...
+			continue
+		} else {
+			// JSON exists, check timestamp
+			startTimeInt, err := build.GetStartTime()
+			if err != nil {
+				continue
+			}
+			startTime := time.Unix(startTimeInt, 0)
+			if time.Since(startTime) < maxElapsedTime {
```
chaodaiG: is there a time.Before or time.After?

chaodaiG:
/approve

2 nits, feel free decide deal with them or not

tools/flaky-test-reporter/jsonreport/jsonreport.go
TrevorFarrelly: time.After() is a timer, signalling a channel when that amount of time has passed. I could do something like `if cutoffTime.Before(startTime)` where cutoffTime is time.Now() - maxAge. If that's clearer I will change it, otherwise it should work as-is

chaodaiG:
looks good to me, refer to @srinivashegde86 lgtm it

tools/flaky-test-reporter/json_reporter.go
```diff
@@ -0,0 +1,76 @@
...
+*/
+
+package main
+
+import (
+	"fmt"
+	"log"
+	"sync"
+
+	"github.com/TrevorFarrelly/test-infra/tools/flaky-test-reporter/jsonreport"
```
srinivashegde86: github.com/knative/test-infra?

tools/flaky-test-reporter/jsonreport/jsonreport.go
```diff
@@ -0,0 +1,119 @@
...
+
+// Report contains concise information about current flaky tests
+type Report struct {
+	Repo  string   `json:"-"`
+	Flaky []string `json:"flaky"`
+}
+
+func (r *Report) writeToArtifactsDir() error {
+	artifactsDir := prow.GetLocalArtifactsDir()
+	err := common.CreateDir(path.Join(artifactsDir, r.Repo))
```
srinivashegde86: can be combined

tools/flaky-test-reporter/jsonreport/jsonreport.go
```diff
@@ -0,0 +1,118 @@
...
+)
+
+// Report contains concise information about current flaky tests
+type Report struct {
+	Repo  string   `json:"-"`
+	Flaky []string `json:"flaky"`
+}
+
+func (r *Report) writeToArtifactsDir() error {
+	err := common.CreateDir(path.Join(prow.GetLocalArtifactsDir(), r.Repo))
```
srinivashegde86: you need the artifactsDir var as you are using it on L48. 

I meant combine the if condition below and this line

TrevorFarrelly: sorry, the line preview window confused me. It should be good to go now

PR #993, opened 2019-06-20
Use path_alias for sample-controller repo

chaodaiG:
As part of knative code using `knative.dev` work, prow job also needs some update so that tests work with `knative.dev` instead of `github.com`. This PR is meant to work with an experimenting migration of sample-controller repo in this PR: https://github.com/knative/sample-controller/pull/24

Idea came from here:
https://github.com/kubernetes/test-infra/blob/b3a9e5ba03ff1d38d3dc3a0fae2880874de1db83/config/jobs/kubernetes/sig-gcp/sig-gcp-windows.yaml#L67-L72

Note:
- As tested, postsubmit and presubmit job would fail if provide extra ref with same repo, so removing them

Example run:
https://gubernator.knative.dev/build/knative-prow/pr-logs/pull/knative_sample-controller/24/pull-knative-sample-controller-build-tests/1141822426414321664/

### DIFF
--- a/ci/prow/make_config.go
+++ b/ci/prow/make_config.go
@@ -93,6 +93,7 @@ type baseProwJobTemplateData struct {
 	Image               string
 	Year                int
 	Labels              []string
+	PathAlias           string
 }
 
 // ####################################################################################################
@@ -421,7 +422,7 @@ func newbaseProwJobTemplateData(repo string) baseProwJobTemplateData {
 	data.Volumes = make([]string, 0)
 	data.VolumeMounts = make([]string, 0)
 	data.Env = make([]string, 0)
-	data.ExtraRefs = []string{"- org: " + data.OrgName, "  repo: " + data.RepoName, "  base_ref: master", "  clone_uri: " + data.CloneURI}
+	data.ExtraRefs = []string{"- org: " + data.OrgName, "  repo: " + data.RepoName, "  base_ref: master"}
 	data.Labels = make([]string, 0)
 	return data
 }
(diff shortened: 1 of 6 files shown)


mattmoor:
We're going to use `knative.dev`, not `code.knative.dev` as the issue title suggests :)

chaodaiG:
> We're going to use `knative.dev`, not `code.knative.dev` as the issue title suggests :)

updated

ci/prow/make_config.go
```diff
@@ -531,6 +532,9 @@ func parseBasicJobConfigOverrides(data *baseProwJobTemplateData, config yaml.Map
 			}
 		case "always_run":
 			(*data).AlwaysRun = getBool(item.Value)
+		case "dot-dev":
+			(*data).PathAlias = "path_alias: knative.dev/" + (*data).RepoName
+			(*data).ExtraRefs = append((*data).ExtraRefs, " "+(*data).PathAlias)
```
cjwagner: ExtraRefs don't necessarily have the same repo as the implicit ref.
Also what if there are multiple ExtraRefs (not multiple lines, but multiple ref entries)? Instead of using a list of strings as the raw config lines, why not import the actual types?

chaodaiG: that's a good point, however we haven't ran into that situation yet, I think we'll redo this part when needed

cjwagner: Do all of your jobs have an ExtraRef? This looks like it might add a bare `path_alias: knative.dev/foo` if there is no ExtraRef specified.

chaodaiG: Currently ExtraRef is defined only at a single place, which makes sure all jobs get `repo`, `org` etc, and you're right that in the future this might happen, but we can always catch it during code review

PR #1000, opened 2019-06-21
Cache bazel version in prow-images

adrcunha:
Show it instead of running bazel. This should reduce the memory pressure on Prow nodes by not starting bazel server for just showing its version.

### DIFF
--- a/images/prow-tests/Dockerfile
+++ b/images/prow-tests/Dockerfile
@@ -37,6 +37,10 @@ RUN rm -rf /usr/local/go && \
     tar xzf "${GO_TARBALL}" -C /usr/local && \
     rm "${GO_TARBALL}"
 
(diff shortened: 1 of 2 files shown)


images/prow-tests/Dockerfile
```diff
@@ -37,6 +37,10 @@ RUN rm -rf /usr/local/go && \
     tar xzf "${GO_TARBALL}" -C /usr/local && \
     rm "${GO_TARBALL}"
 
+# Extract bazel version
+RUN bazel version > /bazel_version
```
chaodaiG: nit: do we need to add 2> /dev/null?

adrcunha: No. We want stderr during the build.

PR #1003, opened 2019-06-21
Limit each job to use only 2Gb cpu and 9Gb memory

chaodaiG:


adrcunha:
/hold

I'm not sure this will work as you expect. This would limit CPU/mem, but the jobs requiring it would still be slow.

srinivashegde86:
I think we can close this now as we have added the limits in https://github.com/knative/test-infra/pull/1004

chaodaiG:
They do different things, this PR limits request, while the other limits limit

adrcunha:
Let's wait a bit and see how the cluster behaves during the weekend. Then we can check the graphs and make more sense of this PR.

/hold

chaodaiG:
sgtm

PR #1008, opened 2019-06-21
Also show ko version on presubmit

adrcunha:


scripts/presubmit-tests.sh
```diff
@@ -296,6 +296,8 @@ function main() {
     kubectl version --client
     echo ">> go version"
     go version
+    echo ">> ko version"
+    ko version
```
chaodaiG: this command somehow doesn't work on my machine

adrcunha: You must use the latest ko.

adrcunha:
Prow is not using the latest prow-tests image, thus `ko version` doesn't work.

adrcunha:
The version not being displayed is actually caused by an intenal error in ko:

https://github.com/google/ko/blob/master/pkg/commands/version.go#L46

adrcunha:
`ko version` actually displays the hash of its HEAD commit in the local git tree, which is NOT what we want.
We need a better way.
/close

PR #1009, opened 2019-06-21
actual config yaml used for monitoring

### LINKED ISSUE #974: [monitoring] create config yaml

steuhs:

Fixes #974 


steuhs:
> 1. Since only ci-code-coverage tests are labelled now, let's make `prs-affected` to be 0 for now so we can actually test the whole flow without labelling additional jobs
> 2. I think the config validation unit test still points to the sample one. Update the url in the unit test to validate the new config. https://github.com/knative/test-infra/blob/master/tools/monitoring/config/config_test.go#L201

Done

PR #888, opened 2019-06-10
Add rbac  permission for plank on prowjobs

chaodaiG:
k8s Prow has updates on plank rbac in April 2019, added `get` permission for prowjobs:  https://github.com/kubernetes/test-infra/commit/faf86a5614b39db9633e26be3a265be24aa8eb35#diff-06da85d3bc6241ba7c85d456254c6819

This might be the cause of us getting lots of `403 forbidden` in Prow logs, i.e.: [example log](https://pantheon.corp.google.com/logs/viewer?project=knative-tests&minLogLevel=0&expandAll=false&timestamp=2019-06-10T15%3A30%3A10.997000000Z&customFacets=jsonPayload.component%2CjsonPayload.level&limitCustomFacetWidth=false&dateRangeStart=2019-05-28T16%3A03%3A04.944Z&interval=CUSTOM&resource=container%2Fcluster_name%2Fprow%2Fnamespace_id%2Fdefault&scrollTimestamp=2019-05-28T16%3A05%3A52.000000000Z&dateRangeEnd=2019-05-28T17%3A03%3A04.944Z&advancedFilter=resource.type%3D%22container%22%0Aresource.labels.container_name%3D%22plank%22%0Aresource.labels.namespace_id%3D%22default%22%0Aresource.labels.instance_id%3D%22868475937429195195%22%0Aresource.labels.zone%3D%22us-central1-f%22%0Aresource.labels.pod_id%3D%22plank-8495fbc6ff-r7rtb%22%0Aresource.labels.project_id%3D%22knative-tests%22%0Aresource.labels.cluster_name%3D%22prow%22%0Atimestamp%3D%222019-05-28T16%3A03%3A53.000000000Z%22%0AinsertId%3D%22bnnqphg1mcfknq%22)

PR #1012, opened 2019-06-22
Have a single target to fully update the Prow cluster

adrcunha:


ci/prow/Makefile
```diff
@@ -41,6 +41,8 @@ update-boskos-config: get-cluster-credentials
 update-cluster: get-cluster-credentials
 	kubectl apply -f cluster.yaml
 
+update-full: get-cluster-credentials update-cluster update-config update-boskos
```
chaodaiG: should we consider add a target for update config-start.yaml?

adrcunha: Ah..  I didn't notice that before. LimitRange stuff should go in cluster.yaml, not in cluster_start.yaml.

chaodaiG: there is still a need for applying config_start.yaml, like this one: https://github.com/knative/test-infra/pull/888

chaodaiG:
/lgtm
/hold
in case you are interested in my nit

ci/prow/Makefile
adrcunha: This is a rare case. I prefer not having a rule to apply config_start.yaml, so whoever has to do it really knows that it's needed (instead of someone blindly applying the rule).

/hold cancel

PR #1013, opened 2019-06-22
fix(dep-collector): Rebuild dep-collector unconditionally

rhuss:
Don't reuse an existing dep-collector in /go/bin as this might be broken
and might be created by a different job which requires a different
version.
Instead rebuild dep-collector for each run afresh, either by
picking it up from a vendored dependency,
or, if toosl/dep-collector is not vendored in the project under test,
then install the latest version directly over the air.

The `go get` for licenseclassifier has been removed as this will be
performed implicitely by the build of dep-collector anyway.

I would recommend to switch over for other go tools, too, to *not*
pick up existing binaries especially when there are different versions
used in different projects (which is why there is support for installing
from the vendor dir, right ?). So `run_go_tool` looks broken to me anyway.

### DIFF
--- a/scripts/library.sh
+++ b/scripts/library.sh
@@ -376,16 +376,25 @@ function update_licenses() {
   cd ${REPO_ROOT_DIR} || return 1
   local dst=$1
   shift
-  run_go_tool ./vendor/github.com/knative/test-infra/tools/dep-collector dep-collector $@ > ./${dst}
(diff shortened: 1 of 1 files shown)


rhuss:
This will fix the issue we have in kn-client with all PR builds, e.g. https://github.com/knative/client/pull/134

adrcunha:
Closing in favor of #1023.

PR #1014, opened 2019-06-22
Add make target to update config-start.yaml

dushyanthsc:


adrcunha:
Please don't. config_start.yaml is intended to be used only when creating the cluster. The LimitRange stuff should be in cluster.yaml.

/close

adrcunha:
Please see https://github.com/knative/test-infra/pull/1016

PR #1016, opened 2019-06-23
Fixes for Prow configs

adrcunha:
* unset kubectl context after updating Prow cluster to avoid further interactions with the cluster by mistake
* move LimitRange to cluster.yaml

### DIFF
--- a/ci/prow/Makefile
+++ b/ci/prow/Makefile
@@ -19,27 +19,36 @@ JOB_NAMESPACE ?= test-pods
 
 PROW_DIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
 TESTGRID_DIR := $(PROW_DIR)/../testgrid
+UNSET_CONTEXT := kubectl config unset current-context
 
 get-cluster-credentials:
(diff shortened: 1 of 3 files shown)


ci/prow/Makefile
```diff
@@ -19,27 +19,36 @@ JOB_NAMESPACE ?= test-pods
 
 PROW_DIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
 TESTGRID_DIR := $(PROW_DIR)/../testgrid
+UNSET_CONTEXT := kubectl config unset current-context
 
 get-cluster-credentials:
 	gcloud container clusters get-credentials "$(CLUSTER)" --project="$(PROJECT)" --zone="$(ZONE)"
 
+unset-cluster-credentials:
```
chaodaiG: nit, would be nice to reuse this target instead of `UNSET_CONTEXT`

chaodaiG:
/lgtm
/hold
in case you are interested in the nit I left

ci/prow/Makefile
adrcunha: I find `make unset-cluster-credentials` output too verbose and hard to read, thus the trick.

PR #1020, opened 2019-06-23
Display ko version on presubmit

adrcunha:
ko version is the commit hash it was built from.

This is based on ko's own version command:

https://github.com/google/ko/blob/master/pkg/commands/version.go#L44


adrcunha:
This is based on ko's own `version` command:

https://github.com/google/ko/blob/master/pkg/commands/version.go#L44

PR #1025, opened 2019-06-24
Report node/pod names in presubmit tests

chaodaiG:
node/pod names are missing after migrating from bootstrap to podutil. Report it with the same logic as in bootstrap, until we figure out a way of reporting it in podutil
see https://github.com/kubernetes/test-infra/blob/09bd4c6709dc64308406443f8996f90cf3b40ed1/jenkins/bootstrap.py#L588

### DIFF
--- a/scripts/presubmit-tests.sh
+++ b/scripts/presubmit-tests.sh
@@ -304,6 +304,15 @@ function main() {
       echo ">> docker version"
       docker version
     fi
+    # node/pod names are important for debugging purposes, but they are missing
+    # after migrating from bootstrap to podutil.
+    # Report it here with the same logic as in bootstrap until it is fixed.
+    # (https://github.com/kubernetes/test-infra/blob/09bd4c6709dc64308406443f8996f90cf3b40ed1/jenkins/bootstrap.py#L588)
+    # TODO(chaodaiG): follow up on https://github.com/kubernetes/test-infra/blob/0fabd2ea816daa8c15d410c77a0c93c0550b283f/prow/initupload/run.go#L49
+    echo ">> node name"
+    echo "$(curl -H "Metadata-Flavor: Google" 'http://169.254.169.254/computeMetadata/v1/instance/name' 2> /dev/null)"
+    echo ">> pod name"
+    echo ${HOSTNAME}
   fi
 
   [[ -z $1 ]] && set -- "--all-tests"


scripts/presubmit-tests.sh
```diff
@@ -304,6 +304,15 @@ function main() {
...
     fi
+    # TODO(chaodaiG): node/pod names are missing after migrating from bootstrap to podutil,
+    # report it here with the same logic as in bootstrap,
+    # until we figure out a way of reporting it in podutil
+    # see https://github.com/kubernetes/test-infra/blob/09bd4c6709dc64308406443f8996f90cf3b40ed1/jenkins/bootstrap.py#L588
+    echo ">> node name"
+    echo "$(curl -H "Metadata-Flavor: Google" 'http://169.254.169.254/computeMetadata/v1/instance/name' 2> /dev/null)"
+    echo ">> pod name"
+    echo ${HOSTNAME}
+    git version
```
adrcunha:
```suggestion
```

scripts/presubmit-tests.sh
```diff
@@ -304,6 +304,15 @@ function main() {
       echo ">> docker version"
       docker version
     fi
+    # TODO(chaodaiG): node/pod names are missing after migrating from bootstrap to podutil,
```
adrcunha: Better explain the reason, then add a TODO pointing to an issue that, once resolved, allow us to remove this.

scripts/presubmit-tests.sh
```diff
@@ -304,6 +304,15 @@ function main() {
...
+    # report it here with the same logic as in bootstrap,
+    # until we figure out a way of reporting it in podutil
+    # see https://github.com/kubernetes/test-infra/blob/09bd4c6709dc64308406443f8996f90cf3b40ed1/jenkins/bootstrap.py#L588
+    echo ">> node name"
+    echo "$(curl -H "Metadata-Flavor: Google" 'http://169.254.169.254/computeMetadata/v1/instance/name' 2> /dev/null)"
```
adrcunha: Where does the IP come from? Will it change without notice and break the log?

chaodaiG: It comes from bootstrap.py, which was how node name got injected before. The link above the line points directly to that line

scripts/presubmit-tests.sh
chaodaiG: done, it's a `TODO` in k8s/test-infra code instead of issue

scripts/presubmit-tests.sh
adrcunha: The line is still here...

scripts/presubmit-tests.sh
```diff
@@ -304,6 +304,15 @@ function main() {
       echo ">> docker version"
       docker version
     fi
+    # TODO(chaodaiG follow up https://github.com/kubernetes/test-infra/blob/0fabd2ea816daa8c15d410c77a0c93c0550b283f/prow/initupload/run.go#L49)
```
adrcunha: I suggest:

```
# node/pod names are important for debugging purposes, but they are missing
# after migrating from bootstrap to podutil.
# Report it here with the same logic as in bootstrap until it is fixed.
# (https://github.com/kubernetes/test-infra/blob/09bd4c6709dc64308406443f8996f90cf3b40ed1/jenkins/bootstrap.py#L588)
# TODO(chaodaiG): follow up on https://github.com/kubernetes/test-infra/blob/0fabd2ea816daa8c15d410c77a0c93c0550b283f/prow/initupload/run.go#L49)
```

chaodaiG: done, thanks

scripts/presubmit-tests.sh
chaodaiG: my bad force push, removed

PR #1032, opened 2019-06-25
Always report build test status as a JUnit XML

adrcunha:
Otherwise for CI flows you'll get blank status for passing runs.
Also update the test name so it properly fits the TestGrid display.


chaodaiG:
This is great!
/lgtm
with nit

Issue #1041, opened 2019-06-25
Run perf tests more frequently and in more regions


srinivashegde86:
currently we only run serving perf tests in us-central1 and once a day. We need to get more samples

srinivashegde86:
We are using  a different perf infra and this is not used anymore.

PR #1043, opened 2019-06-25
Run perf jobs every 3 hours instead of once a day

srinivashegde86:
Part of #1041 

The perf jobs take ~2 hours to run. Running them multiple times a day provides us with more metrics to account for the variance in dealing with the cluster used to run the tests

Also, remove the sql secret as we dont want to store the metrics there anymore.

### DIFF
--- a/ci/prow/config.yaml
+++ b/ci/prow/config.yaml
@@ -2264,7 +2264,7 @@ periodics:
     - name: test-account
       secret:
         secretName: test-account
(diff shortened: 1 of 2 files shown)


adrcunha:
/lgtm
/approve

@Fredy-Z any concerns on eventing side? If not, please unhold.
/hold

Fredy-Z:
There is no concern for eventing.
/hold cancel

PR #1057, opened 2019-06-27
Add ssh key for prow updater job

chaodaiG:
Add missing ssh key for prow auto updater job, which is required for `git push` command


ci/prow/periodic_config.go
```diff
@@ -246,6 +246,7 @@ func generateVersionBumpertoolPeriodicJob() {
 	addExtraEnvVarsToJob(&data.Base)
 	configureServiceAccountForJob(&data.Base)
 	addVolumeToJob(&data.Base, "/etc/prow-auto-bumper-github-token", "prow-auto-bumper-github-token", true)
+	addVolumeToJob(&data.Base, "/root/.ssh", "prow-updater-robot-ssh-key", true)
```
Fredy-Z: Has this key been documented in our internal Playbook?

chaodaiG: good point and done

PR #1067, opened 2019-06-27
Don't fail Flaky test report if failed collecting results for a singl…

chaodaiG:
When we add a new repo/job to be analyzed by Flaky test reporter, it might happen that there is no valid runs in that repo/job, which currently fails the entire run. Print a warning and let it continue instead, then report job status as failure at the end

### DIFF
--- a/tools/flaky-test-reporter/main.go
+++ b/tools/flaky-test-reporter/main.go
@@ -22,6 +22,7 @@ package main
 
 import (
 	"flag"
+	"fmt"
 	"log"
 	"os"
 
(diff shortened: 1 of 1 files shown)


tools/flaky-test-reporter/main.go
```diff
@@ -57,13 +58,21 @@ func main() {
...
 	}
+	var jobErrs []error
 	for repoName, jobList := range jobConfigs {
 		for _, jc := range jobList {
 			jc.Repo = repoName
 			log.Printf("collecting results for job '%s' in repo '%s'\n", jc.Name, jc.Repo)
 			rd, err := collectTestResultsForRepo(jc)
 			if nil != err {
-				log.Fatalf("Error collecting results for job '%s' in repo '%s': %v", jc.Name, jc.Repo, err)
+				errMsg := fmt.Sprintf("WARNING: failed collecting results, skipping '%s' in repo '%s': %v", jc.Name, jc.Repo, err)
```
srinivashegde86: Instead of string, can we just make it an err?
```
e := fmt.Errorf("WARNING...")
log.Printf("%v", e)
err = append(err, e)
```

chaodaiG: good idea, done

PR #1069, opened 2019-06-27
Fix update-full target

chaodaiG:
Transitive dependencies in Makefile are not executed, so update-full got context unset after the first target, and failed on second. Use direct invocation instead


adrcunha:
Thanks, it was on my TODO list but I didn't have time to do it. Better than using phony rules.

/lgtm
/approve

Issue #812, opened 2019-05-21
The flaky reporter should take in the config as an input(maybe yaml?) and not hard-code


srinivashegde86:
The config is currently hard-coded - https://github.com/knative/test-infra/blob/master/tools/flaky-test-reporter/config.go#L38

We will be expanding this in the future to handle multiple repos and responding to various slack channels. Maybe we should separate the config definition and store it in a yaml and can update only that separately.

PR #1070, opened 2019-06-28
Use yaml config for Flaky test reporter

chaodaiG:
Refactor flaky test reporter to be yaml config based instead of inline config in go code.
Move config to a package so flaky test re-tryer and reuse it

Fixes: #812 

/cc @srinivashegde86 
/cc @TrevorFarrelly


chaodaiG:
FYI @TrevorFarrelly

tools/flaky-test-reporter/config/config.go
```diff
@@ -0,0 +1,65 @@
...
+package config
+
+import (
+	"io/ioutil"
+
+	"gopkg.in/yaml.v2"
+)
+
+const (
+	defaultConfig = "config.yaml"
```
srinivashegde86:
```suggestion
	const defaultConfig = "config.yaml"
```

tools/flaky-test-reporter/config/config.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+// Config contains all job configs for flaky tests reporting
+type Config struct {
+	JobConfigs []JobConfig `yaml:"jobConfigs"`
```
srinivashegde86:
```suggestion
	Jobs []Job `yaml:"jobs"`
```

tools/flaky-test-reporter/config/config.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// JobConfig is initial configuration for a given repo, defines which job to scan
+type JobConfig struct {
```
srinivashegde86:
```suggestion
type Job struct {
```

tools/flaky-test-reporter/config/config.go
```diff
@@ -0,0 +1,65 @@
...
+	Name            string         `yaml:"name"` // name of job to analyze
+	Repo            string         `yaml:"repo"` // repository to test job on
+	Type            string         `yaml:"type"`
+	GithubIssueRepo string         `yaml:"githubIssueRepo,omitempty"`
```
srinivashegde86:
```suggestion
	IssueRepo string         `yaml:"issueRepo,omitempty"`
```

tools/flaky-test-reporter/config/config.go
```diff
@@ -0,0 +1,65 @@
...
+
+// SlackChannel contains Slack channels info
+type SlackChannel struct {
+	Name     string `yaml:"name"`
+	Identity string `yaml:"identity"`
+}
+
+// NewConfig parses config from configFile, default to "config.yaml" if configFile is empty string
+func NewConfig(configFile string) (*Config, error) {
+	if "" == configFile {
```
srinivashegde86:
```suggestion
	if len(configFile) == 0 {
```

tools/flaky-test-reporter/config/config.yaml
```diff
@@ -0,0 +1,44 @@
+jobConfigs:
```
srinivashegde86: doc header

tools/flaky-test-reporter/config/config.yaml
```diff
@@ -0,0 +1,44 @@
...
+  - name: ci-knative-serving-continuous
+    repo: serving
+    type: postsubmit
+    githubIssueRepo: serving
+    slackChannels:
+      - name: api
```
srinivashegde86: should this be serving-api?

tools/flaky-test-reporter/github_issue.go
```diff
@@ -423,12 +423,12 @@ func (gi *GithubIssue) getFlakyIssues() (map[string][]*flakyIssue, error) {
...
 // these can later on be printed as summary at the end of run
-func (gi *GithubIssue) processGithubIssueForRepo(rd *RepoData, flakyIssuesMap map[string][]*flakyIssue, repoForIssue string, dryrun bool) ([]string, error) {
+func (gi *GithubIssue) processGithubIssueForRepo(rd *RepoData, flakyIssuesMap map[string][]*flakyIssue, dryrun bool) ([]string, error) {
 	var messages []string
 	var errs []error
 
-	// if flag is not set, do not create an issue at all
-	if rd.Config.SkipGithubIssue {
+	repoForIssue := rd.Config.GithubIssueRepo
+	if "" == repoForIssue {
```
srinivashegde86:
```suggestion
	if len(repoForIssue) == 0 {
```

tools/flaky-test-reporter/github_issue_test.go
```diff
@@ -23,6 +23,7 @@ import (
 
 	"github.com/google/go-github/github"
 	"github.com/knative/test-infra/shared/ghutil/fakeghutil"
+	ftrConfig "github.com/knative/test-infra/tools/flaky-test-reporter/config"
```
srinivashegde86: do we need an alias here? We can just use config. ftrConfig is confusing

tools/flaky-test-reporter/main.go
```diff
@@ -26,15 +26,22 @@ import (
...
 
 	"github.com/knative/test-infra/shared/prow"
+	ftrConfig "github.com/knative/test-infra/tools/flaky-test-reporter/config"
 )
 
 func main() {
 	serviceAccount := flag.String("service-account", os.Getenv("GOOGLE_APPLICATION_CREDENTIALS"), "JSON key file for GCS service account")
 	githubAccount := flag.String("github-account", "", "Token file for Github authentication")
 	slackAccount := flag.String("slack-account", "", "slack secret file for authenticating with Slack")
+	configPath := flag.String("configfile", "", "Config file for overriding default config file")
```
srinivashegde86: can we add the path to the default config here instead of empty?

tools/flaky-test-reporter/main.go
```diff
@@ -26,15 +26,22 @@ import (
...
 	dryrun := flag.Bool("dry-run", false, "dry run switch")
 	flag.Parse()
 
+	config, err := ftrConfig.NewConfig(*configPath)
+	if nil != err {
+		log.Fatalf("config file '%s' not found '%v'", *configPath, err)
```
srinivashegde86: Re-word to config cannot be created? There can be other failure reasons than not found

tools/flaky-test-reporter/github_issue_test.go
chaodaiG: I felt naming this package as `config` probably is a bad idea. But naming things is so hard, tried

tools/flaky-test-reporter/main.go
chaodaiG: this is done purposely, so that when another tool wants to use the `config` package it doesn't need to understand where default config is

tools/flaky-test-reporter/main.go
chaodaiG: good idea, done

tools/flaky-test-reporter/config/config.yaml
chaodaiG: done

tools/flaky-test-reporter/github_issue_test.go
chaodaiG: alias removed

tools/flaky-test-reporter/main.go
srinivashegde86: This default is for the ftr tool since this is in main.go and not the config pkg? I agree that the config pkg should be generic, but why cant we add default here when defineing the flag?

chaodaiG: agreed and done

PR #1073, opened 2019-06-28
[monitoring] remove legacy implementation of CheckAlertCondition

steuhs:

Fixes #

### DIFF
--- a/tools/monitoring/config/config.go
+++ b/tools/monitoring/config/config.go
@@ -20,7 +20,6 @@ limitations under the License.
 package config
 
 import (
-	"database/sql"
 	"fmt"
 	"io/ioutil"
 	"log"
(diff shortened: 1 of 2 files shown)


tools/monitoring/config/config.go
```diff
@@ -90,9 +90,9 @@ func (s *SelectedConfig) CheckAlertCondition(errorPattern string, db *sql.DB) (b
 
 	row := db.QueryRow(`
```
yt3liu: Do you know where is `CheckAlertCondition` currently used, except in main.go for testing this function? 
I am working on the clearing alert workflow which also has a modified version of this query. I'm still in the process for testing this query. But I don't think this method is used in the alerting flow currently. 

https://github.com/yt3liu/test-infra/blob/20d63b9b65436ffd624af918eafec5c0b176155b/tools/monitoring/mysql/database.go#L189

steuhs: this is the legacy implementation we no longer need. I am removing it.

PR #1084, opened 2019-07-01
added the clear-alerts prow configs

### LINKED ISSUE #766: Clear alerts when alert condition fails
/assign @yt3liu

yt3liu:

Add the clear alerts prow job 

Fixes #766

### DIFF
--- a/ci/prow/periodic_config.go
+++ b/ci/prow/periodic_config.go
@@ -311,7 +311,7 @@ func generateClearAlertsPeriodicJob() {
 	var data periodicJobTemplateData
 	data.Base = newbaseProwJobTemplateData("knative/test-infra")
 	data.Base.Image = clearalertsDockerImage
-	data.PeriodicJobName = "ci-knative-monitoring-clear-alerts"
+	data.PeriodicJobName = "ci-knative-test-infra-monitoring-clear-alerts"
 	data.CronString = clearAlertsPeriodicJobCron
 	data.Base.Command = "/clearalerts"
 	addVolumeToJob(&data.Base, "/secrets/cloudsql/monitoringdb", "monitoring-db-credentials", true)
--- a/ci/prow/periodic_config.go
+++ b/ci/prow/periodic_config.go
@@ -34,12 +34,13 @@ const (
 	periodicCustomJob = "prow_periodic_custom_job.yaml"
 
 	// Cron strings for key jobs
-	goCoveragePeriodicJobCron        = "0 1 * * *"   // Run at 01:00 every day
-	cleanupPeriodicJobCron           = "0 19 * * 1"  // Run at 11:00PST/12:00PST every Monday (19:00 UTC)
-	flakesReporterPeriodicJobCron    = "0 12 * * *"  // Run at 4:00PST/5:00PST every day (12:00 UTC)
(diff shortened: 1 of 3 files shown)


yt3liu:
/hold
Do not merge until: 
1. https://github.com/knative/test-infra/pull/1083/ is merged. 
1. the `monitoring-db-credentials` is updated to include database host information

ci/prow/periodic_config.go
```diff
@@ -297,3 +298,15 @@ func generateGoCoveragePeriodic(title string, repoName string, _ yaml.MapSlice)
 		return
 	}
 }
+
+// generateClearAlertsPeriodicJob generates the monitoring clear alerts job config.
+func generateClearAlertsPeriodicJob() {
+	var data periodicJobTemplateData
+	data.Base = newbaseProwJobTemplateData("knative/test-infra")
+	data.Base.Image = clearalertsDockerImage
+	data.PeriodicJobName = "ci-knative-monitoring-clear-alerts"
```
srinivashegde86: ci-knative-test-infra-clear-alerts? Should we use the repo-name?

yt3liu: For the clear alerts job, I think one job that runs every 30 minutes is sufficient. If we use the repo name in the job name, I looks to me like it's expecting n repo jobs to be created. 

Also, the clear alerts job is repo agnostic since it clear alerts based on job name regex and error pattern. We may be able to deduce the repo name from the job name, but I don't think we need this extra complexity to support clearing alerts for one specific repo only.

srinivashegde86: I think the repo name is for where the job should be associated with. This is related to test-infra and I think it should be included in the name. We should follow the current pattern of naming.

yt3liu: Thanks for clarifying it. Updated the ci job name to include test-infra.

srinivashegde86:
do you still need the hold here?

yt3liu:
Yes, the secrets on the prow cluster needs to be updated before updating the prow config. But since it's the day before long weekend. Let's not touch the prow cluster today and do it next Monday.

ci/prow/periodic_config.go
```diff
@@ -304,3 +305,15 @@ func generateGoCoveragePeriodic(title string, repoName string, _ yaml.MapSlice)
...
 }
+
+// generateClearAlertsPeriodicJob generates the monitoring clear alerts job config.
+func generateClearAlertsPeriodicJob() {
+	var data periodicJobTemplateData
+	data.Base = newbaseProwJobTemplateData("knative/test-infra")
+	data.Base.Image = clearalertsDockerImage
+	data.PeriodicJobName = "ci-knative-test-infra-monitoring-clear-alerts"
+	data.CronString = clearAlertsPeriodicJobCron
+	data.Base.Command = "/clearalerts"
```
chaodaiG:
```suggestion
	data.Base.Command = "/clearalerts"
	data.Base.ExtraRefs = append(data.Base.ExtraRefs, "  base_ref: "+data.Base.RepoBranch)
```

chaodaiG: please commit this change and run `make config` again

PR #1091, opened 2019-07-01
add Initialize() func to jsonreport

TrevorFarrelly:
Allows us to initialize Prow from the jsonreport library, bypassing a panic that occurs when `golang.org/x/net/trace.init()` is called from multiple contexts. Necessary for further development of the flaky-test-retryer.

Fixes #

### DIFF
--- a/tools/flaky-test-reporter/jsonreport/jsonreport.go
+++ b/tools/flaky-test-reporter/jsonreport/jsonreport.go
@@ -40,6 +40,7 @@ type Report struct {
 	Flaky []string `json:"flaky"`
 }
 
+// Initialize wraps prow's init, which must be called before any other prow functions are used.
 func Initialize(credentials string) error {
 	return prow.Initialize(credentials)
 }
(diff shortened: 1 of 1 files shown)


tools/flaky-test-reporter/jsonreport/jsonreport.go
```diff
@@ -40,6 +40,10 @@ type Report struct {
 	Flaky []string `json:"flaky"`
 }
 
+func Initialize(credentials string) error {
+	return prow.Initialize(credentials)
```
chaodaiG: suggest also creates a struct for handling everything jsonreport related, so the caller doesn't receive `prow` struct when call `Initialize`

TrevorFarrelly: I don't believe there is a struct or a client in this case, unlike other setup functions we use. prow.Initialize() is just a wrapper for gcs.Authenticate(), and all structs are private within that package.

tools/flaky-test-reporter/jsonreport/jsonreport.go
```diff
@@ -40,6 +40,7 @@ type Report struct {
 	Flaky []string `json:"flaky"`
 }
 
+// Initialize wraps prow's init, which must be called before any other prow functions are used.
 func Initialize(credentials string) error {
```
chaodaiG: please rename `credentials` to `serviceAccount` for consistency, this is actually a file path afaik

PR #1098, opened 2019-07-02
Update section about expanding Boskos pool

adrcunha:


ci/prow_setup.md
```diff
@@ -38,12 +38,12 @@
 1. Run `make update-boskos-config` to update the Boskos config.
 
 1. Increase the compute CPU quota for the project to 200. Go to
```
chaodaiG: it's not related to this PR, but just noticed this step should swap with the step above so there is no surprise of the new project picks up performance test before quota increase kicks in

adrcunha: Thanks. Done.

Issue #1099, opened 2019-07-02
Make on-call page easier to remember/access


adrcunha:
https://storage.googleapis.com/knative-infra-oncall/oncall.html is not intuitive. Even though it's pinned in Slack.

adrcunha:
https://knative.github.io/test-infra/ is the new oncall display page.

Added link to knative/test-infra repo root.
Added link as topic in #productivity and #test in Slack.
Updated pinned announcement about oncall in #general, #productivity and #testing in Slack.

PR #1100, opened 2019-07-02
Setup GitHub Pages for displaying oncall rotation

adrcunha:
Curious? See https://adrcunha.github.io/test-infra/

Part of #1099

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,3 +1,3 @@
-# Current Oncall 
+# Current Oncall
 
 <iframe src="https://storage.googleapis.com/knative-infra-oncall/oncall.html" width="100%" height="384"></iframe>
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,3 @@
+# Current Oncall 
+
+<iframe src="https://storage.googleapis.com/knative-infra-oncall/oncall.html" width="100%" height="384"></iframe>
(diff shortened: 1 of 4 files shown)


docs/README.md
```diff
@@ -0,0 +1,3 @@
+# Current Oncall 
+
+<iframe src="https://storage.googleapis.com/knative-infra-oncall/oncall.html" width="100%" height="384"></iframe>
```
chaodaiG: this html tag does seem to work in markdown file

adrcunha: It does. Our linter doesn't like it:

```
docs/README.md:1: MD009 Trailing spaces
docs/README.md:3: MD033 Inline HTML

A detailed description of the rules is available at https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md
```

I'll disable the rule for our repo.

chaodaiG: It still shows raw html in your fork, would it be different in knative?
https://github.com/adrcunha/test-infra/blob/onca/docs/README.md

adrcunha: That's expected, as this markdown file is meant to be viewed through GitHub Pages: https://adrcunha.github.io/test-infra/

chaodaiG: thanks, good to know

PR #925, opened 2019-06-13
flaky-test-reporter: fix a few small issues introduced recently

TrevorFarrelly:
/lint

**What this PR does, why we need it**:
* Report on istio-1.1.7, after upgrade in #873 
* Update TestGrid links to new format, from #882 
* Add TestGrid links for Istio jobs.
* Name artifact after job instead of repo. (since #809 "serving.json" was getting overwritten five times)

**Which issue(s) this PR fixes**:
Fixes #

**Special notes to reviewers**:

**User-visible changes in this PR**:

PR #1105, opened 2019-07-02
refactor jsonreport to better support multi-repo searching

TrevorFarrelly:
With this refactoring, we can get JSON reports via build ID and/or repository, meaning we now support getting all reports from all repos at once

Fixes #

Depends on #1110, /hold until that is merged

### DIFF
--- a/tools/flaky-test-reporter/jsonreport/jsonreport.go
+++ b/tools/flaky-test-reporter/jsonreport/jsonreport.go
@@ -22,6 +22,7 @@ import (
 	"io/ioutil"
 	"path"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/knative/test-infra/shared/common"
@@ -108,6 +109,23 @@ func GetReportForRepo(repo string, buildID int) (*Report, error) {
 		return nil, err
 	}
 	return report, nil
+}
+
+//GetReposWithReports gets all the repositories that are currently being reported on
+func GetReposWithReports() ([]string, error) {
+	job := prow.NewJob(jobName, prow.PeriodicJob, "", 0)
+	buildNum, err := job.GetLatestBuildNumber()
+	if err != nil {
+		return nil, err
+	}
+	build := job.NewBuild(buildNum)
+	artifactPaths := build.GetArtifacts()
+	var repos []string
+	for _, path := range artifactPaths {
+		splitPath := strings.Split(path, "/")
+		repos = append(repos, splitPath[len(splitPath)-1])
+	}
+	return repos, nil
 }
 
 // CreateReportForRepo generates a Report struct and optionally writes it to disk
--- a/tools/flaky-test-reporter/jsonreport/jsonreport.go
+++ b/tools/flaky-test-reporter/jsonreport/jsonreport.go
@@ -69,12 +69,12 @@ func (r *RepoReport) writeToArtifactsDir() error {
 // most recent report
 func (r *Report) GetFlakyTestReport(repo string, buildID int) error {
 	job := prow.NewJob(jobName, prow.PeriodicJob, "", 0)
+	var err error
 	if buildID == -1 {
-		latestBuild, err := getLatestValidBuild(job, repo)
+		buildID, err = getLatestValidBuild(job, repo)
 		if err != nil {
 			return err
 		}
-		return r.GetFlakyTestReport(repo, latestBuild)
 	}
 	build := job.NewBuild(buildID)
 	for _, filepath := range getReportPaths(build, repo) {
--- a/tools/flaky-test-reporter/jsonreport/jsonreport.go
+++ b/tools/flaky-test-reporter/jsonreport/jsonreport.go
@@ -35,12 +35,8 @@ const (
 	maxAge   = 4                            // maximum age in days that JSON data is valid
 )
 
-// Report contains concise information about current flaky tests
+// Report contains concise information about current flaky tests in a given repo
 type Report struct {
-	Reports []RepoReport
-}
-
-type RepoReport struct {
 	Repo  string   `json:"repo"`
 	Flaky []string `json:"flaky"`
 }
@@ -51,7 +47,7 @@ func Initialize(serviceAccount string) error {
 }
 
 // writeToArtifactsDir writes the flaky test data for this repo to disk.
-func (r *RepoReport) writeToArtifactsDir() error {
+func (r *Report) writeToArtifactsDir() error {
 	artifactsDir := prow.GetLocalArtifactsDir()
 	if err := common.CreateDir(path.Join(artifactsDir, r.Repo)); nil != err {
 		return err
@@ -65,26 +61,27 @@ func (r *RepoReport) writeToArtifactsDir() error {
 }
 
 // GetFlakyTestReport collects flaky test reports from the given buildID and repo.
-//Use repo = "" to get reports from all repositories, and buildID = -1 to get the
+// Use repo = "" to get reports from all repositories, and buildID = -1 to get the
 // most recent report
-func (r *Report) GetFlakyTestReport(repo string, buildID int) error {
+func GetFlakyTestReport(repo string, buildID int) ([]Report, error) {
 	job := prow.NewJob(jobName, prow.PeriodicJob, "", 0)
 	var err error
 	if buildID == -1 {
 		buildID, err = getLatestValidBuild(job, repo)
 		if err != nil {
-			return err
+			return nil, err
 		}
 	}
 	build := job.NewBuild(buildID)
+	var reports []Report
 	for _, filepath := range getReportPaths(build, repo) {
 		report, err := readJSONReport(build, filepath)
 		if err != nil {
-			return err
+			return nil, err
 		}
-		r.Reports = append(r.Reports, *report)
+		reports = append(reports, *report)
 	}
-	return nil
+	return reports, nil
 }
 
 // getLatestValidBuild inexpensively sorts and finds the most recent JSON report.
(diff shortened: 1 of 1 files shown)


tools/flaky-test-reporter/jsonreport/jsonreport.go
```diff
@@ -110,6 +111,23 @@ func GetReportForRepo(repo string, buildID int) (*Report, error) {
...
+	job := prow.NewJob(jobName, prow.PeriodicJob, "", 0)
+	buildNum, err := job.GetLatestBuildNumber()
+	if err != nil {
+		return nil, err
+	}
+	build := job.NewBuild(buildNum)
+	artifactPaths := build.GetArtifacts()
+	var repos []string
+	for _, path := range artifactPaths {
+		splitPath := strings.Split(path, "/")
```
srinivashegde86: why are we looking at all artifacts here?

TrevorFarrelly: the reporter writes artifacts to `artifacts/<repo>/<artifact files>`, and in my testing GetArtifacts() only returns the paths one level deep in the artifacts dir, i.e. `artifacts/<repo>`. I added this directory structure in #925 to stop test suites from overwriting each other.

tools/flaky-test-reporter/jsonreport/jsonreport.go
```diff
@@ -110,6 +111,23 @@ func GetReportForRepo(repo string, buildID int) (*Report, error) {
...
 
+//GetReposWithReports gets all the repositories that are currently being reported on
+func GetReposWithReports() ([]string, error) {
+	job := prow.NewJob(jobName, prow.PeriodicJob, "", 0)
+	buildNum, err := job.GetLatestBuildNumber()
+	if err != nil {
+		return nil, err
+	}
+	build := job.NewBuild(buildNum)
+	artifactPaths := build.GetArtifacts()
```
chaodaiG: doesn't this method return an error?

tools/flaky-test-reporter/jsonreport/jsonreport.go
chaodaiG: This is actually a bug, GetArtifacts should return all files instead, I can fix it. You should assume it's returning all files for now

tools/flaky-test-reporter/jsonreport/jsonreport.go
```diff
@@ -110,6 +111,23 @@ func GetReportForRepo(repo string, buildID int) (*Report, error) {
 	return report, nil
 }
 
+//GetReposWithReports gets all the repositories that are currently being reported on
+func GetReposWithReports() ([]string, error) {
```
chaodaiG: this logic should be combined with `getJSONForBuild` above, since they should be looking at the same build

tools/flaky-test-reporter/jsonreport/jsonreport.go
chaodaiG: https://github.com/knative/test-infra/pull/1110

srinivashegde86: My point was there may be artifacts other than the ones added by the jsonreporter. For eg prow adds an empty junit_xml as well. We should be checking only the ones we are interested in.

tools/flaky-test-reporter/jsonreport/jsonreport.go
```diff
@@ -36,7 +37,11 @@ const (
 
 // Report contains concise information about current flaky tests
 type Report struct {
-	Repo  string   `json:"-"`
+	AllReports []RepoReport
```
srinivashegde86:
```suggestion
	Reports []RepoReport
```

tools/flaky-test-reporter/jsonreport/jsonreport.go
```diff
@@ -58,66 +77,78 @@ func (r *Report) writeToArtifactsDir() error {
...
-// CreateReportForRepo generates a Report struct and optionally writes it to disk
-func CreateReportForRepo(repo string, flaky []string, writeFile bool) (*Report, error) {
-	report := &Report{
-		Repo:  repo,
-		Flaky: flaky,
-	}
-	if writeFile {
-		return report, report.writeToArtifactsDir()
+// getLatestValidBuild inexpensively sorts and finds the most recent JSON report.
+// Assumes build IDs are in sequential order.
```
srinivashegde86: why is this an assumption? You are sorting it anyways?

TrevorFarrelly:
/hold

new changes depend on #1110

tools/flaky-test-reporter/jsonreport/jsonreport.go
```diff
@@ -58,66 +77,78 @@ func (r *Report) writeToArtifactsDir() error {
...
+func getLatestValidBuild(job *prow.Job, repo string) (int, error) {
+	maxElapsedTime, _ := time.ParseDuration(fmt.Sprintf("%dh", maxAge*24))
```
srinivashegde86: should we do this if latest is valid? Maybe we shoudl short-circuit this?

tools/flaky-test-reporter/jsonreport/jsonreport.go
TrevorFarrelly: I worded that comment poorly. Assume sequential build IDs are sequential in time, i.e. if latest is ID 3, the previous build is 2, the one before that is 1, etc. Will fix.

tools/flaky-test-reporter/jsonreport/jsonreport.go
```diff
@@ -58,61 +64,94 @@ func (r *Report) writeToArtifactsDir() error {
...
+		}
+		// check if this report is too old
+		startTimeInt, err := build.GetStartTime()
+		if err != nil {
 			continue
+		}
+		startTime := time.Unix(startTimeInt, 0)
+		if time.Since(startTime) < maxElapsedTime {
+			return buildID, nil
 		} else {
```
chaodaiG: else is not needed

tools/flaky-test-reporter/jsonreport/jsonreport.go
```diff
@@ -58,61 +64,94 @@ func (r *Report) writeToArtifactsDir() error {
...
-			if time.Since(startTime) < maxElapsedTime {
-				return contents, nil
-			} else {
-				return nil, fmt.Errorf("latest JSON log is outdated: %.2f days old", time.Since(startTime).Hours()/24)
-			}
+			return 0, fmt.Errorf("latest JSON log is outdated: %.2f days old", time.Since(startTime).Hours()/24)
 		}
 	}
-	return nil, fmt.Errorf("no JSON logs found in recent builds")
+	return 0, fmt.Errorf("no JSON logs found in recent builds")
```
chaodaiG:
```suggestion
	return 0, fmt.Error("no JSON logs found in recent builds")
```

tools/flaky-test-reporter/jsonreport/jsonreport.go
```diff
@@ -58,61 +64,94 @@ func (r *Report) writeToArtifactsDir() error {
...
+
+// readJSONReport builds a repo-specific report object from a given json file path.
+func readJSONReport(build *prow.Build, filename string) (*RepoReport, error) {
+	report := &RepoReport{}
+	contents, err := build.ReadFile(filename)
 	if err != nil {
 		return nil, err
 	}
-	if err = json.Unmarshal(contents, report); err != nil {
+	if err := json.Unmarshal(contents, report); err != nil {
```
chaodaiG: doesn't it complain? err is already defined above

TrevorFarrelly: surprisingly, no. Seems like it defines a new err within the scope of the if statement. I'll change it for consistency anyways.

TrevorFarrelly:
/hold cancel

#1110 merged

tools/flaky-test-reporter/jsonreport/jsonreport.go
```diff
@@ -45,62 +50,94 @@ func Initialize(serviceAccount string) error {
...
-	report := &Report{
-		Repo: repo,
+// getReportPaths searches build artifacts for reports from the given repo, returning
+// the path to any matching files. Use repo = "" to get all reports from all repos.
+func getReportPaths(build *prow.Build, repo string) []string {
+	var matches []string
+	suffix := path.Join(repo, filename)
+	for _, artifact := range build.GetArtifacts() {
+		if strings.HasSuffix(artifact, suffix) {
+			matches = append(matches, artifact)
```
chaodaiG: can there be multiple matches?

TrevorFarrelly: there will be if we use repo = "" to get all reports. If we specify a repo there'll only be one match, but I intend to get all repo data when I call this function

chaodaiG: make sense

chaodaiG:
/approve
@srinivashegde86 please lgtm if you think all your comments are resolved

tools/flaky-test-reporter/jsonreport/jsonreport.go
```diff
@@ -36,7 +37,11 @@ const (
 
 // Report contains concise information about current flaky tests
 type Report struct {
-	Repo  string   `json:"-"`
+	Reports []RepoReport
```
srinivashegde86: whats the point of this struct? Doesnt look like we are using it here anywhere?

TrevorFarrelly: GetFlakyTestReport() should return one, looks like I wrote it to populate the report you call it on. Will fix.

srinivashegde86: why cant GetFlakyTestReport() return []Report instead of this struct?

TrevorFarrelly: very good point, I have no plans to use the Report struct in other places. I can fix this.

Issue #1108, opened 2019-07-03
Make rolling back Prow config possible


chaodaiG:
Prow config is updated by `kubectl` command, which doesn't provide a good way of tracking when Prow config was last updated, nor which commit it based on. We should have a way of tracking last Prow config as well as easily restoring to previous working config if bad things happened

PR #1109, opened 2019-07-03
Make restore working Prow config easier

chaodaiG:
Currently we don't have a good way of restoring to previous working Prow config, this PR provides a way of tracking previous working config, as well as commit hash, which should make restoring easier

Fix: https://github.com/knative/test-infra/issues/1108

/cc @adrcunha 
/cc @Fredy-Z

### DIFF
--- a/ci/prow/Makefile
+++ b/ci/prow/Makefile
@@ -33,8 +33,16 @@ config:
 
 update-config:
 	$(SET_CONTEXT)
-	kubectl create configmap config --from-file=config.yaml=config.yaml --dry-run -o yaml | kubectl replace configmap config -f -
+	$(eval OLD_YAML_CONFIG := $(shell mktemp))
+	$(eval NEW_YAML_CONFIG := $(shell mktemp))
+	$(eval GCS_DEST := gs://knative-prow/configs/config-$(shell date '+%Y_%m_%d_%H:%M:%S').yaml)
+	@kubectl get configmap config -o jsonpath="{.data}" 2>/dev/null > "${OLD_YAML_CONFIG}"
+	@gsutil cp "${OLD_YAML_CONFIG}" "${GCS_DEST}" > /dev/null
+	@sed -e "s/((COMMIT_HASH_TOKEN))/$(shell git rev-parse HEAD)/" "config.yaml" > "${NEW_YAML_CONFIG}"
+	kubectl create configmap config --from-file=config.yaml=${NEW_YAML_CONFIG} --dry-run -o yaml | kubectl replace configmap config -f -
 	$(UNSET_CONTEXT)
+	@echo "Inspect uploaded config file at: ${NEW_YAML_CONFIG}"
+	@echo "Old config file saved at: ${GCS_DEST}"
 
 update-plugins:
 	$(SET_CONTEXT)
--- a/ci/prow/config.yaml
+++ b/ci/prow/config.yaml
@@ -17,6 +17,7 @@
 # ####     USE "make config" TO REGENERATE THIS FILE.     ####
 # ####                                                    ####
 # ############################################################
+# FROM COMMIT: ((COMMIT_HASH_TOKEN))
 plank:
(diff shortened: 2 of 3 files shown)


ci/prow/Makefile
```diff
@@ -33,8 +33,13 @@ config:
 
 update-config:
 	$(SET_CONTEXT)
+	$(eval TMP_YAML_CONFIG := $(shell mktemp))
+	kubectl get configmap config -o jsonpath="{.data}" 2>/dev/null >"${TMP_YAML_CONFIG}"
+	sed -i -e "s/((COMMIT_HASH_TOKEN))/$(shell git rev-parse HEAD)/" "config.yaml"
```
adrcunha: Copy config.yaml to a temp file so you don't have to git checkout it (and also eliminate the dependency on git).

ci/prow/Makefile
```diff
@@ -33,8 +33,13 @@ config:
...
 	kubectl create configmap config --from-file=config.yaml=config.yaml --dry-run -o yaml | kubectl replace configmap config -f -
+	git checkout config.yaml
 	$(UNSET_CONTEXT)
+	@echo "Old config file saved at: ${TMP_YAML_CONFIG}"
```
adrcunha: What about saving it to a GCS bucket with a timestamp, instead of a local file on a workstation?

ci/prow/Makefile
chaodaiG: good idea, done. Unfortunately can't completely eliminate dependency on git since we need the git commit

ci/prow/Makefile
chaodaiG: good idea, done

ci/prow/Makefile
```diff
@@ -33,13 +33,16 @@ config:
...
 update-config:
 	$(SET_CONTEXT)
-	$(eval TMP_YAML_CONFIG := $(shell mktemp))
-	kubectl get configmap config -o jsonpath="{.data}" 2>/dev/null >"${TMP_YAML_CONFIG}"
-	sed -i -e "s/((COMMIT_HASH_TOKEN))/$(shell git rev-parse HEAD)/" "config.yaml"
-	kubectl create configmap config --from-file=config.yaml=config.yaml --dry-run -o yaml | kubectl replace configmap config -f -
-	git checkout config.yaml
+	$(eval OLD_YAML_CONFIG := $(shell mktemp))
+	$(eval NEW_YAML_CONFIG := $(shell mktemp))
+	$(eval GCS_DEST := gs://knative-prow/configs/prow_job_config/$(shell date +%s))
```
adrcunha: %s is impossible to read, better format it as "YYYY-MM-DD HH:MM:SS" or similar. You could store them inside /configs as "config_YYYY-MM-DD_HH:MM:SS.yaml" for example.

ci/prow/Makefile
```diff
@@ -33,8 +33,16 @@ config:
 
 update-config:
 	$(SET_CONTEXT)
-	kubectl create configmap config --from-file=config.yaml=config.yaml --dry-run -o yaml | kubectl replace configmap config -f -
+	$(eval OLD_YAML_CONFIG := $(shell mktemp))
+	$(eval NEW_YAML_CONFIG := $(shell mktemp))
+	$(eval GCS_DEST := gs://knative-prow/configs/prow_job_config/$(shell date +%s))
+	@kubectl get configmap config -o jsonpath="{.data}" 2>/dev/null >"${OLD_YAML_CONFIG}"
```
adrcunha:
```suggestion
	@kubectl get configmap config -o jsonpath="{.data}" 2>/dev/null >" ${OLD_YAML_CONFIG}"
```

ci/prow/Makefile
```diff
@@ -33,8 +33,16 @@ config:
...
+	@gsutil cp "${OLD_YAML_CONFIG}" "${GCS_DEST}" >/dev/null
```
adrcunha:
```suggestion
	@gsutil cp "${OLD_YAML_CONFIG}" "${GCS_DEST}" > /dev/null
```

ci/prow/Makefile
chaodaiG: actually I thought about that, decide not to use `YYYY-MM-DD_HH:MM:SS` as it's already a metadata for each GCS file

adrcunha: Yes, it is. But it's way quicker/easier to figure that out when listing the files (through UI or CLI) instead of the extra metadata step, right? And what if I'm looking for a specific file, say from X days ago? I would have to check the metadata of a bunch of files to finally find the one I'm looking for.

chaodaiG: The UI is fine as the metadata is right there, while CLI is a legit use case, updated

PR #1114, opened 2019-07-03
Enhance the test-infra status page

adrcunha:
Curious? https://adrcunha.github.io/test-infra/

* No more iframes
* Add Prow and TestGrid links
* Easier to read, one-line oncall display


docs/cors-json-file.json
```diff
@@ -0,0 +1,6 @@
+[
+  {
+    "origin": ["https://adrcunha.github.io"],
```
chaodaiG: should this be `knative.github.io`?

adrcunha: Good catch, thanks.

PR #1116, opened 2019-07-08
Also make testgrid URL one-line

Fredy-Z:
A very small improvement for https://knative.github.io/test-infra/:
On Mac chrome, the URL of testgrid is shown two lines because it uses a different font. Set `max-width` to `700px` to avoid line break, see https://fredy-z.github.io/test-infra/

/cc @adrcunha 
/cc @chaodaiG

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,3 +1,8 @@
+<style>
+/* Increase content width from 640px to 800px for systems with larger font to avoid line break. */
+.inner {max-width: 800px;}
+</style>
+
 # Prow: <https://prow.knative.dev>
 
 # TestGrid: <https://testgrid.knative.dev>


adrcunha:
It makes no difference to me in my screen...

Fredy-Z:
> It makes no difference to me in my screen...

Do you mean https://fredy-z.github.io/test-infra/? If yes that would be strange, since I double checked, it should only be two-liners if the screen width is under 700px, which is a phone's size.

adrcunha:
I didn't check on a phone, but on my laptop and desktop it already shows as one line.

Fredy-Z:
Sorry maybe my PR name was not clear, what I meant was the change as below.
Screen Shot 2019-07-08 at 1 45 49 PM
Screen Shot 2019-07-08 at 1 46 39 PM

adrcunha:
I understand what you're doing and the problem you're trying to solve, but I can't reproduce it in my laptop or desktop. How did you face it?

Fredy-Z:
> I understand what you're doing and the problem you're trying to solve, but I can't reproduce it in my laptop or desktop. How did you face it?

Ahh, I checked with @coryrc and he did not had this problem either. But he helped me to find the root cause, it was because the default font used on Mac is different, which caused automatic line break. And he told me another better solution by adding `&nbsp;` to avoid automatic line break, see https://stackoverflow.com/questions/1357078/whats-the-difference-between-nbsp-and.

Thanks!

adrcunha:
Thanks for clarifying. However, now you're breaking the flow for small screens, forcing the text to be in a single line. I'm not sure why Macs would break the line, if there's enough horizontal space for the text. But if that's indeed the problem, I would try reducing the margins for the theme's main content.

Fredy-Z:
It seems it's still better to overwrite the `max-width` to a larger value (700px is enough), as the margin is `auto`.

docs/README.md
```diff
@@ -1,3 +1,7 @@
+<style>
+#main_content {max-width: 700px;}
```
adrcunha: I was able to reproduce. On a Mac with Chrome. What you really want is to override `max-width` of `.inner` to, say, 800px because currently it's 640px. Not great, but...

Fredy-Z: Done.

docs/README.md
```diff
@@ -1,3 +1,7 @@
+<style>
+.inner {max-width: 800px;}
```
adrcunha: Can we add a one-line comment here explaining why? Something like:

`// Increase content width from 640px to 800px for large screens.`

Fredy-Z: Sure. Done.

PR #1125, opened 2019-07-10
support adding error rates as metrics in perf tests

Fredy-Z:
1. Add `ErrorsPercentageOverall` to get the overall `ErrorRate` from the performance testing result.
2. Rename the `perfPropertyName` to make it less confusing.

/cc @srinivashegde86

### DIFF
--- a/shared/loadgenerator/loadgenerator.go
+++ b/shared/loadgenerator/loadgenerator.go
@@ -200,30 +200,29 @@ func (gr *GeneratorResults) ErrorsPercentage(idx int) float64 {
 		return 0
 	}
 
-	successes, errors := gr.errorSuccessCounts(idx)
-	return float64(errors*100) / float64(errors+successes)
+	errors, total := gr.errorAndTotalCounts(idx)
+	return float64(errors*100) / float64(total)
 }
 
 // ErrorsPercentageOverall returns the error percentage of the result based on response codes.
 // Any non 200 response will be counted as errors.
 func (gr *GeneratorResults) ErrorsPercentageOverall() float64 {
-	var successes, errors int64
+	var errors, total int64
 	for i := 0; i < len(gr.Result); i++ {
-		subSuccesses, subErrors := gr.errorSuccessCounts(i)
-		successes += subSuccesses
+		subErrors, subTotal := gr.errorAndTotalCounts(i)
 		errors += subErrors
+		total += subTotal
 	}
-	return float64(errors*100) / float64(errors+successes)
+	return float64(errors*100) / float64(total)
 }
 
-func (gr *GeneratorResults) errorSuccessCounts(idx int) (int64, int64) {
-	var successes, errors int64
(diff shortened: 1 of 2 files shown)


shared/loadgenerator/loadgenerator.go
```diff
@@ -208,5 +225,5 @@ func (gr *GeneratorResults) ErrorsPercentage(idx int) float64 {
 			errors = errors + count
 		}
 	}
-	return float64(errors*100) / float64(errors+successes)
+	return successes, errors
```
srinivashegde86: we can return errs and total? So, that we dont have to do an additional calc of total = succ + err above

Fredy-Z: Good idea. Done.

shared/performance/performance.go
```diff
@@ -37,3 +39,14 @@ func CreatePerfTestCase(metricValue float32, metricName, testName string) junit.
 
 	return tc
 }
+
+// CreatePerfErrorRateTestCase creates a perf error rate test case with the provided value
+func CreatePerfErrorRateTestCase(metricValue float32, testName string) junit.TestCase {
```
srinivashegde86: why are we creating an err property to the test case? We should treat it like a metric if the test wants to store this?

Fredy-Z: I've removed this function, and changed the variable name to `perfPropertyName` to make it less confusing.
It seems unnecessary to change the actual name since the previous result will not be able to show correctly on testgrid.

shared/loadgenerator/loadgenerator.go
```diff
@@ -194,19 +194,35 @@ func (gr *GeneratorResults) SaveJSON() error {
 }
 
 // ErrorsPercentage returns the error percentage of the result[idx] based on response codes.
-// Any non 200 response will provide a value > 0.0
+// Any non 200 response will be counted as errors.
 func (gr *GeneratorResults) ErrorsPercentage(idx int) float64 {
```
srinivashegde86: This is a breaking change as a few methods in serving that use this and this will break the auto-updater.

srinivashegde86:
@chaodaiG what is the process for making breaking changes in test-infra? The changes in this PR will break the serving dep auto-updater

shared/loadgenerator/loadgenerator.go
Fredy-Z: It will not break the auto-updater. Either the function name or interface is not changed. I just added one extra function `ErrorsPercentageOverall`.

Fredy-Z:
> @chaodaiG what is the process for making breaking changes in test-infra? The changes in this PR will break the serving dep auto-updater

Please see the comment above.

shared/loadgenerator/loadgenerator.go
srinivashegde86: yeah. Dont think this will break it.

PR #1126, opened 2019-07-10
add flaky-test-reporter's config.yaml to its image

TrevorFarrelly:
add the reporter's new yaml config format to its Dockerfile in the expected default location.

If it makes more sense somewhere else in the image, we will need to add a flag in `ci/prow/periodic_config.go`.

Fixes #

### DIFF
--- a/images/flaky-test-reporter/Dockerfile
+++ b/images/flaky-test-reporter/Dockerfile
@@ -25,6 +25,7 @@ ADD . $TEMP_REPO_DIR
 # Build flaky test reporter tool in the container
 RUN make -C $TEMP_REPO_DIR/tools/$TOOL_NAME/
 RUN cp $TEMP_REPO_DIR/tools/$TOOL_NAME/$TOOL_NAME /$TOOL_NAME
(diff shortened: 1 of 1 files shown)


images/flaky-test-reporter/Dockerfile
```diff
@@ -25,6 +25,7 @@ ADD . $TEMP_REPO_DIR
 # Build flaky test reporter tool in the container
 RUN make -C $TEMP_REPO_DIR/tools/$TOOL_NAME/
 RUN cp $TEMP_REPO_DIR/tools/$TOOL_NAME/$TOOL_NAME /$TOOL_NAME
+RUN cp $TEMP_REPO_DIR/tools/$TOOL_NAME/config.yaml config.yaml
```
srinivashegde86: why dont we just copy everything under $TOOL_NAME?

TrevorFarrelly: That would add source files, the README, and other things right? Certainly doable, it adds a bit of bloat though.

srinivashegde86: yes. but its just a few files. Other images add all the files in the repo :)

I am fine either way, but seems like everytime you want something in the image, you will have to keep adding it.

PR #1145, opened 2019-07-16
clean up the log

yt3liu:

Do not print out the build log content to the log. 

Fixes #

/cc @srinivashegde86 
/cc @steuhs

### DIFF
--- a/tools/monitoring/alert/alert.go
+++ b/tools/monitoring/alert/alert.go
@@ -59,9 +59,9 @@ func (c *Client) RunAlerting() {
 }
 
 func (c *Client) handleReportMessage(rmsg *prowapi.ReportMessage) {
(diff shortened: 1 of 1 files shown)


tools/monitoring/alert/alert.go
```diff
@@ -59,6 +59,8 @@ func (c *Client) RunAlerting() {
 }
 
 func (c *Client) handleReportMessage(rmsg *prowapi.ReportMessage) {
+	log.Printf("Received Pubsub message %v\n", rmsg)
```
srinivashegde86: should this be insied if condition? Else, we will be logging for every job every status change.

yt3liu: Yes, you're right. Updated.

PR #1147, opened 2019-07-17
[monitoring] Add a failure type that happens recently.

steuhs:

Fixes #

### DIFF
--- a/tools/monitoring/config/config.yaml
+++ b/tools/monitoring/config/config.yaml
@@ -25,7 +25,7 @@ spec:
         occurrences: 1
         jobs-affected: 1
         prs-affected: 0
-        period: 1440  
+        period: 1440
 
   - error-pattern: 'Something went wrong: failed to prepare test environment: --provider=.* boskos failed to acquire project: resources not found'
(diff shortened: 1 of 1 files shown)


tools/monitoring/config/config.yaml
```diff
@@ -41,6 +41,20 @@ spec:
         prs-affected: 0
         period: 1440
 
+  - error-pattern: 'ERROR: timeout waiting for pods to come up'
+    hint: 'This may lead to Knative setup failure'
```
yt3liu: For the hint, I think these are action items the oncall can take in order to fix the infrastructure issue. Could you please add more into how to fix this issue if you know potential issue/fix? 
This may be more of a case-by-case issue that the pods timeout. If you don't have enough info to come up with an oncall action item, then the current hint is good for now.

steuhs: I am not familiar with how to deal with pod not coming up. But adding the hint here helps build a connection with the Knative setup failure. So when the oncall sees the two failures together he got the hint of such a connection and investigate further. I think after an oncall investigated a problem he can give out a better recovery hint.

PR #1149, opened 2019-07-17
disable the clear-alerts prow job

yt3liu:

Disabling the clear-alerts prow job. 

Reason: 
The prow cluster does not have ip-alias enabled. So it cannot connect to the cloud sql instance through private ip. Currently, there's no safe, easy way to convert an existing cluster to be a vpc-native cluster. 

The current clear-alerts job checks the same config as the alerting workflow. So clear-alerts workflow deletes the Alert from the database only when alerting workflow consider those alerts as expired. So even with the prow job enabled, the clear-alerts job does not clear the alert faster. This job will be more useful if when the configs are separated for those two workflows. 

Fixes #

/cc @srinivashegde86 
/cc @steuhs

### DIFF
--- a/ci/prow/make_config.go
+++ b/ci/prow/make_config.go
@@ -144,7 +144,6 @@ var (
 	prowversionbumperDockerImage string
 	githubCommenterDockerImage   string
 	coverageDockerImage          string
-	clearalertsDockerImage       string
 	prowTestsDockerImage         string
 	presubmitScript              string
 	releaseScript                string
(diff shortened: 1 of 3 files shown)


ci/prow/make_config.go
```diff
@@ -1087,7 +1087,6 @@ func main() {
 			return !repo.Processed && repo.EnableGoCoverage
 		}, generateGoCoveragePeriodic)
 		generateCleanupPeriodicJob()
-		generateClearAlertsPeriodicJob()
```
srinivashegde86: we should also delete this method implemenetation?

PR #1150, opened 2019-07-18
add prow job for running Knative-Serving e2e against Gloo

ilackarms:

 https://github.com/knative/serving/pull/4704 introduced the ability to run the Serving e2e test suite using Gloo as the clusteringress rather than Istio. 

The changes in this PR add a Prow job to enable running that flavor of the tests in CI.

### DIFF
--- a/ci/prow/config_knative.yaml
+++ b/ci/prow/config_knative.yaml
@@ -176,6 +176,9 @@ periodics:
     - custom-job: istio-1.2-no-mesh
       full-command: "./test/e2e-tests.sh --istio-version 1.2-latest --no-mesh"
       dot-dev: true
+    - custom-job: gloo-0.17.1
+      full-command: "./test/e2e-tests.sh --gloo-version 0.17.1"
+      dot-dev: true
     - nightly: true
       dot-dev: true
     - dot-release: true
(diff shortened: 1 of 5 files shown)


adrcunha:
/ok-to-test
/lgtm
/approve

Please note that the current oncall (@dushyanthsc) needs to update the Prow cluster with the new config (once it's merged) so the new job will start running. @ilackarms please sync up with him.

adrcunha:
Looks like you have to regenerate `config.yaml` using `make config`?

ci/prow/config_knative.yaml
```diff
@@ -166,6 +166,9 @@ periodics:
     - custom-job: istio-1.2-no-mesh
       full-command: "./test/e2e-tests.sh --istio-version 1.2-latest --no-mesh"
       dot-dev: true
+    - custom-job: gloo-0.17.1
```
chaodaiG: you may want to remove patch version just like `istio`, so there is no need to update everywhere with patch releases

ilackarms: this would require another PR for `serving`. if you don't mind, i'll leave it on `0.17.1` for now and address this in a follow up?

chaodaiG: sgtm

chaodaiG:
/lgtm
/approve
Please work with @Fredy-Z applying these changes since he is oncall this week

PR #1153, opened 2019-07-18
removed redundant test and unneeded code

yt3liu:

* removed redundant test and unneeded code
* enhanced the logging for the clear-alerts job 

Fixes #

/cc @srinivashegde86 
/cc @steuhs

### DIFF
--- a/tools/monitoring/clearalerts/main.go
+++ b/tools/monitoring/clearalerts/main.go
@@ -54,6 +54,7 @@ func main() {
 		log.Fatalf("Failed to list the alerts: %v", err)
 	}
 
+	nDel := 0
 	for _, a := range alerts {
 		isAlerting := false
 		isExpired := false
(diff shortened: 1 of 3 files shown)


tools/monitoring/clearalerts/main.go
```diff
@@ -54,6 +54,7 @@ func main() {
 		log.Fatalf("Failed to list the alerts: %v", err)
 	}
 
+	nDel := 0
```
srinivashegde86: I dont think we need this var. We have the log statements for every pattern being deleted

srinivashegde86:
small nit.

tools/monitoring/clearalerts/main.go
yt3liu: I can add a log statement `Finished clearing alerts in the database` at the end of the function instead of printing out the number of jobs deleted. This is mainly for printing something in the log so we know the job has completed successfully when there's no error pattern to delete.

srinivashegde86: we can keep this then.

PR #1157, opened 2019-07-18
Fix the mess in the /docs directory

adrcunha:
* use index.html instead of README.md
* move documentation from README.txt to README.md
* explain this dir and where to find docs in README.md


docs/README.md
```diff
@@ -1,37 +1,23 @@
...
+
+Instead, this is the source of the GitHub Pages for knative/test-infra. See https://pages.github.com/ for details.
+
+Documentation can be found in the several directories of the repository. For example:
+
+* [Main README](../README.md)
+* [Documentation on the CI/CD system](../ci/README.md)
+* [Documentation on the helper scripts](../scripts/README.md)
+
+## Updating the test-infra GitHub page
```
chaodaiG: I think this should be single `#`

adrcunha: This is not a title.

PR #1159, opened 2019-07-18
Add monitoring for DNS flakiness

chaodaiG:


tools/monitoring/config/config.yaml
```diff
@@ -152,3 +152,17 @@ spec:
         jobs-affected: 1
         prs-affected: 0
         period: 1440
+
+  - error-pattern: 'Could not resolve host: github.com'
```
yt3liu: Move this error to the top of the file. 
@steuhs Correct me if I'm wrong, but I believe the error matching logic only scans through the config top down and only select the first config that it finds.

chaodaiG: had an in person discussion with @steuhs , this doesn't apply for this pattern

PR #1165, opened 2019-07-19
flaky-test-retryer: add github integration

### LINKED ISSUE #987: flaky re-trier: Add GitHub integration
Part 4 of implementation of the flaky test re-trier

When our job is triggered by a failing test on a new PR, we will want to re-run those flaky tests until they succeed. We will do this by using the existing GitHub bot to comment `/retest` and a brief message as to why we are doing this

If the PR's failed tests include both flaky and non-flaky tests, we will also post a comment saying why we will not retest until the non-flaky tests are fixed.

TrevorFarrelly:
This PR adds GitHub commenting functionality to the flaky-test-retryer. The comments feature a list of jobs, the number of retries, and either a list of tests that prevent the job from restarting or the command to restart the test. [Example](https://github.com/TrevorFarrelly/knative-testing/pull/30)

When we update a comment, we have to delete and re-post it so knative-prow robot detects a new `/test <job>` command. This does cause some GitHub email notification spam, and unfortunately I could not find another easy solution.

If any Knative devs have ideas for different comment formatting, or other ways to restart the jobs, I'd love for some input. The retryer is meant to help you after all, not spam your inbox any more than necessary :)

Fixes #987 

### DIFF
--- a/shared/ghutil/fakeghutil/fakeghutil.go
+++ b/shared/ghutil/fakeghutil/fakeghutil.go
@@ -160,6 +160,17 @@ func (fgc *FakeGithubClient) EditComment(org, repo string, commentID int64, comm
 	return nil
 }
 
+// DeleteComment deletes comment from issue
+func (gc *GithubClient) DeleteComment(org, repo string, commentID int64) error {
+	for _, comments := range fgc.Comments {
+		if comment, ok := comments[commentID]; ok {
+			delete(comments, commentID)
+			return nil
+		}
+	}
+	return fmt.Errorf("cannot find comment")
+}
+
 // AddLabelsToIssue adds label on issue
 func (fgc *FakeGithubClient) AddLabelsToIssue(org, repo string, issueNumber int, labels []string) error {
 	targetIssue := fgc.Issues[repo][issueNumber]
--- a/shared/ghutil/fakeghutil/fakeghutil.go
+++ b/shared/ghutil/fakeghutil/fakeghutil.go
@@ -163,7 +163,7 @@ func (fgc *FakeGithubClient) EditComment(org, repo string, commentID int64, comm
 // DeleteComment deletes comment from issue
 func (fgc *FakeGithubClient) DeleteComment(org, repo string, commentID int64) error {
 	for _, comments := range fgc.Comments {
-		if comment, ok := comments[commentID]; ok {
+		if _, ok := comments[commentID]; ok {
 			delete(comments, commentID)
 			return nil
 		}
--- a/shared/ghutil/fakeghutil/fakeghutil.go
+++ b/shared/ghutil/fakeghutil/fakeghutil.go
@@ -161,7 +161,7 @@ func (fgc *FakeGithubClient) EditComment(org, repo string, commentID int64, comm
 }
 
 // DeleteComment deletes comment from issue
-func (gc *GithubClient) DeleteComment(org, repo string, commentID int64) error {
+func (fgc *FakeGithubClient) DeleteComment(org, repo string, commentID int64) error {
 	for _, comments := range fgc.Comments {
 		if comment, ok := comments[commentID]; ok {
 			delete(comments, commentID)
--- a/shared/ghutil/issue.go
+++ b/shared/ghutil/issue.go
@@ -199,6 +199,19 @@ func (gc *GithubClient) EditComment(org, repo string, commentID int64, commentBo
 	return err
 }
 
+// DeleteComment deletes comment from issue
+func (gc *GithubClient) DeleteComment(org, repo string, commentID int64) error {
+	_, err := gc.retry(
+		fmt.Sprintf("deleting comment '%s %s %d'", org, repo, commentID),
+		maxRetryCount,
+		func() (*github.Response, error) {
+			resp, err := gc.Client.Issues.DeleteComment(ctx, org, repo, commentID)
+			return resp, err
+		},
+	)
+	return err
+}
+
 // AddLabelsToIssue adds label on issue
 func (gc *GithubClient) AddLabelsToIssue(org, repo string, issueNumber int, labels []string) error {
 	_, err := gc.retry(
--- a/tools/flaky-test-retryer/github_commenter.go
+++ b/tools/flaky-test-retryer/github_commenter.go
@@ -21,7 +21,23 @@ limitations under the License.
 package main
 
 import (
+	"fmt"
+	"log"
+	"regexp"
+	"strconv"
+	"strings"
+
+	"github.com/google/go-github/github"
 	"github.com/knative/test-infra/shared/ghutil"
+)
+
+const (
+	maxRetries = 3
+)
+
+var (
+	commentTemplate = "The following tests failed due to flakiness:\n\n" +
+		"Test name | Retries\n--- | ---\n%s\n\n%s"
 )
 
 // GithubClient wraps the ghutil Github client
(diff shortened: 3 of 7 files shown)


chaodaiG:
/cc @adrcunha 
He gave me pretty good suggestions on Flaky test reporter tool

shared/ghutil/issue.go
```diff
@@ -199,6 +199,19 @@ func (gc *GithubClient) EditComment(org, repo string, commentID int64, commentBo
 	return err
 }
 
+// DeleteComment deletes comment from issue
+func (gc *GithubClient) DeleteComment(org, repo string, commentID int64) error {
```
chaodaiG: I think you need to add this to the Interface to make unit test work

tools/flaky-test-retryer/github_commenter.go
```diff
@@ -37,3 +53,107 @@ func NewGithubClient(githubAccount string) (*GithubClient, error) {
...
+	}
+	return fmt.Sprintf(commentTemplate, strings.Join(entryString, "\n"), cmd), entries[jd.JobName] <= maxRetries
+}
+
+// buildRetryString increments the retry counter and generates a /test string if we have
+// more retries available.
+func buildRetryString(job string, entries map[string]int) string {
+	entries[job]++
+	if entries[job] <= maxRetries {
+		return fmt.Sprintf("Automatically retrying...\n/test %s", job)
```
chaodaiG: How about listing tests that are flaky here? This can make the logic of `buildRetryString` and `buildNoRetryString` share majority of the logic

tools/flaky-test-retryer/github_commenter.go
```diff
@@ -37,3 +53,107 @@ func NewGithubClient(githubAccount string) (*GithubClient, error) {
...
+	}
+	return ""
+}
+
+// buildNoRetryString formats the tests that prevent us from retrying into a drop-down
+// menu underneath the table of retries.
+func buildNoRetryString(job string, outliers []string) string {
+	dropdownFmt := "<details>\n<summary>Non-flaky failing tests preventing automatic retry of %s:</summary>\n<br>\n<code>%s</code>\n</details>"
```
chaodaiG: I'm not sure about hiding them is the best idea. If the purpose is to avoid displaying too long message, I would recommend doing the first 10 tests then add another line saying `X more tests also failed`

tools/flaky-test-retryer/github_commenter.go
```diff
@@ -37,3 +53,107 @@ func NewGithubClient(githubAccount string) (*GithubClient, error) {
...
+}
+
+// getOldComment queries the GitHub PR specified and gets the comment made by us. If no comment
+// is found, we do not error, since we will be creating a new one anyways.
+func (gc *GithubClient) getOldComment(org, repo string, pull int) (*github.IssueComment, error) {
+	comments, err := gc.ListComments(org, repo, pull)
+	if err != nil {
+		return nil, err
+	}
+	user, err := gc.GetGithubUser()
```
chaodaiG: I would prefer collect this information up ahead of time, and store it in gc

tools/flaky-test-retryer/github_commenter.go
```diff
@@ -37,3 +53,107 @@ func NewGithubClient(githubAccount string) (*GithubClient, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	for _, comment := range comments {
```
chaodaiG: a comment here would be helpful, like `this robot should only leave one comment in a PR`. And a guard here might be even better

tools/flaky-test-retryer/handler.go
```diff
@@ -94,16 +96,16 @@ func (hc *HandlerClient) HandleJob(jd *JobData) {
 	}
 	logWithPrefix(jd, "got %d flaky tests from today's report\n", len(flakyTests))
 
-	if outliers := getNonFlakyTests(failedTests, flakyTests); len(outliers) > 0 {
+	outliers := getNonFlakyTests(failedTests, flakyTests)
+	if len(outliers) > 0 {
```
chaodaiG: this if condition is done twice, does this logic have to be here?

tools/flaky-test-retryer/main.go
```diff
@@ -34,13 +34,14 @@ const (
...
 	githubAccount := flag.String("github-account", "", "Token file for Github authentication")
+	dryrun := flag.Bool("dry-run", false, "dry run switch")
 	flag.Parse()
 
 	if err := InitLogParser(*serviceAccount); nil != err {
 		log.Fatalf("Failed authenticating GCS: '%v'", err)
 	}
 
-	handler, err := NewHandlerClient(*githubAccount)
+	handler, err := NewHandlerClient(*githubAccount, *dryrun)
```
chaodaiG: I feel like this can be done slightly different, like https://github.com/knative/pkg/blob/e4bc08cc8deda3eac40a8d53984d09c9d0e1ad48/test/e2e_flags.go#L48

shared/ghutil/fakeghutil/fakeghutil.go
```diff
@@ -160,6 +160,17 @@ func (fgc *FakeGithubClient) EditComment(org, repo string, commentID int64, comm
 	return nil
 }
 
+// DeleteComment deletes comment from issue
+func (gc *GithubClient) DeleteComment(org, repo string, commentID int64) error {
```
srinivashegde86:
```suggestion
func (fgc *FakeGithubClient) DeleteComment(org, repo string, commentID int64) error {
```

tools/flaky-test-retryer/github_commenter.go
```diff
@@ -21,9 +21,25 @@ limitations under the License.
...
+	"log"
+	"regexp"
+	"strconv"
+	"strings"
+
+	"github.com/google/go-github/github"
 	"github.com/knative/test-infra/shared/ghutil"
 )
 
+const (
```
srinivashegde86:
```suggestion
const maxRetries = 3
```

tools/flaky-test-retryer/github_commenter.go
```diff
@@ -21,9 +21,25 @@ limitations under the License.
...
+	maxRetries = 3
+)
+
+var (
+	commentTemplate = "The following tests failed due to flakiness:\n\n" +
```
srinivashegde86:
```suggestion
	commentTemplate = "The following tests are currently flaky. Running them again to verify...:\n\n" +
```

tools/flaky-test-retryer/github_commenter.go
```diff
@@ -37,3 +53,107 @@ func NewGithubClient(githubAccount string) (*GithubClient, error) {
...
+		if *comment.GetUser().Login == *user.Login {
+			return comment, nil
+		}
+	}
+	return nil, nil
+}
+
+// parseEntries collects retry information from the given comment, so we can reuse it in
+// a new comment.
+func parseEntries(comment *github.IssueComment) (map[string]int, error) {
```
srinivashegde86: I would like to see some unit tests around this.

tools/flaky-test-retryer/handler.go
TrevorFarrelly: this was so I could log how many tests are not flaky and whether or not we triggered a retry. I could move the logic, or forgo it altogether as it's mostly for testing purposes

chaodaiG: This would be helpful debugging message, just think it could be combined with the other `if else` if you agree with what I suggested over there

tools/flaky-test-retryer/github_commenter.go
TrevorFarrelly: updated format, see the new examples at the same link above.

tools/flaky-test-retryer/github_commenter.go
TrevorFarrelly: added Login field to gc, populated in the constructor

tools/flaky-test-retryer/github_commenter.go
```diff
@@ -37,3 +53,107 @@ func NewGithubClient(githubAccount string) (*GithubClient, error) {
...
+			return err
+		}
+	}
+	_, err = gc.CreateComment(jd.Refs[0].Org, jd.Refs[0].Repo, jd.Refs[0].Pulls[0].Number, newComment)
+	return err
+}
+
+// getOldComment queries the GitHub PR specified and gets the comment made by us. If no comment
+// is found, we do not error, since we will be creating a new one anyways.
+func (gc *GithubClient) getOldComment(org, repo string, pull int) (*github.IssueComment, error) {
```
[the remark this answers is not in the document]
TrevorFarrelly: fixed, added a tag to the beginning of the comment in addition to username

tools/flaky-test-retryer/github_commenter.go
TrevorFarrelly: fixed

shared/ghutil/fakeghutil/fakeghutil.go
TrevorFarrelly: fixed

tools/flaky-test-retryer/handler.go
TrevorFarrelly: done

tools/flaky-test-retryer/github_commenter.go
TrevorFarrelly: commented, I will add a check in another commit

shared/ghutil/issue.go
TrevorFarrelly: fixed

tools/flaky-test-retryer/github_commenter.go
TrevorFarrelly: added a new const in other changes

tools/flaky-test-retryer/github_commenter.go
TrevorFarrelly: As it stands, I don't collect the flaky tests that affect this job, only the outliers, so this will require some more logic elsewhere. Do you think it would be helpful from a development perspective, seeing which flaky tests affected that job?

chaodaiG: No one probably would even pay attention to that, but from my point of view, it would be useful for checking if this tool works properly

tools/flaky-test-retryer/main.go
TrevorFarrelly: moved all flags to that format, not necessary but figured I'd do it for consistency

tools/flaky-test-retryer/github_commenter.go
TrevorFarrelly: added a check and solution if there's more than one comment, it's a little strange though so I'd like your input.

tools/flaky-test-retryer/github_commenter.go
```diff
@@ -35,5 +54,152 @@ func NewGithubClient(githubAccount string) (*GithubClient, error) {
...
+		return comments[i].GetCreatedAt().Unix() < comments[i].GetCreatedAt().Unix()
+	})
+
+	switch len(matches) {
+	case 0:
+		return nil, nil
+	case 1:
+		return matches[0], nil
+	default:
+		if Flags.DryRun {
```
chaodaiG: I think relying on a global var would make unit test a bit harder, add it to `gc` might be an option

tools/flaky-test-retryer/github_commenter.go
chaodaiG: yep, this is strange, I would prefer it to fail since this is not suppose to happen. We can discuss how to surface this kind of problem

tools/flaky-test-retryer/github_commenter.go
TrevorFarrelly: working on a more comprehensive unit test for all of github_commenter, I'll get that in ASAP

TrevorFarrelly: alright, github_commenter_test.go is up. I will continue to add more test cases to the rest of the retryer over time.

tools/flaky-test-retryer/github_commenter.go
TrevorFarrelly: done

tools/flaky-test-retryer/github_commenter_test.go
```diff
@@ -0,0 +1,191 @@
...
+func TestGetOldComment(t *testing.T) {
+	gc := getFakeGithubClient()
+	cases := []struct {
+		client      *GithubClient
+		org, repo   string
+		pull        int
+		wantComment *github.IssueComment
+		wantErr     error
+	}{
+		{gc, fakeOrg, fakeRepo, fakePullID, nil, nil},
```
chaodaiG: nit, this test is different from what most table-driven tests, since the order matters, you may want to add a comment here

chaodaiG:
/ok-to-test
/approve

feel free fix/ignore with the nit I left. will let @srinivashegde86 lgtm

tools/flaky-test-retryer/github_commenter.go
```diff
@@ -21,19 +21,175 @@ limitations under the License.
...
+	// this robot should only leave one comment in a given PR. Collect everything with our identifier,
+	// just in case.
+	var matches []*github.IssueComment
+	for _, comment := range comments {
+		match, err := regexp.Match(identifier, []byte(comment.GetBody()))
+		if err != nil {
+			return nil, err
+		}
+		if match && *comment.GetUser().Login == gc.Login {
+			matches = append(matches, comment)
```
srinivashegde86: whats the point of using an array, if this should never have >1 ? Why not just have a var. If we really want to be sure that we are not doing this twice, we can err out when the condition is true and var is set.

tools/flaky-test-retryer/github_commenter.go
```diff
@@ -21,19 +21,175 @@ limitations under the License.
...
+		}
+	}
+
+	switch len(matches) {
```
srinivashegde86: I dont think this switch is necessary. Have just one var and return var, err

tools/flaky-test-retryer/github_commenter_test.go
TrevorFarrelly: gotcha, added a comment

tools/flaky-test-retryer/github_commenter_test.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+	for _, data := range cases {
+		actual, _ := parseEntries(data.input)
+		if !reflect.DeepEqual(actual, data.want) {
+			t.Fatalf("parse entries: got '%v', want '%v'", actual, data.want)
+		}
+	}
+}
+
+func TestBuildNewComment(t *testing.T) {
```
srinivashegde86: add one for testParseentries as well

tools/flaky-test-retryer/github_commenter.go
```diff
@@ -21,19 +21,175 @@ limitations under the License.
...
+	oldEntries, err := parseEntries(oldComment)
+	if err != nil {
+		return err
+	}
+	if _, ok := oldEntries[jd.JobName]; !ok {
+		oldEntries[jd.JobName] = 0
+	}
+	newComment, canRetry := buildNewComment(jd, oldEntries, outliers)
+	if !canRetry {
+		return fmt.Errorf("expended all %d retries", maxRetries)
```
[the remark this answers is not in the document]
TrevorFarrelly: The old comment would say 3/3 retries in this case so I figured it was implied. I can post a comment explicitly stating that if it makes sense.

tools/flaky-test-retryer/github_commenter.go
TrevorFarrelly: done

tools/flaky-test-retryer/github_commenter.go
TrevorFarrelly: done

tools/flaky-test-retryer/github_commenter_test.go
TrevorFarrelly: do you mean a case in TestBuildNewComment or a TestParseEntries? because TestParseEntries is written [here](https://github.com/knative/test-infra/pull/1165/files/e31bff94147128bed54d2d9fb2eb59a6be704801#diff-00576b64e8131e8f783bc75ba0c4d685R154).

tools/flaky-test-retryer/github_commenter.go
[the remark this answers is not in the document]
srinivashegde86: yes. But, the comment will say attempt #3/3. But, there will be no result from that attempt. Having a different message like `Attempted 3 times and failed` is more explicit.

srinivashegde86:
/lgtm
/hold
There is one nit about the last attempt comment. Cancel hold with `/hold cancel` if you think its not needed.

tools/flaky-test-retryer/github_commenter.go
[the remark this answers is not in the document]
TrevorFarrelly: done

PR #1171, opened 2019-07-20
move -race option to go_test_e2e and default_unit_test_runner

Fredy-Z:
When vegeta generate test loads, it spawns workers without limit to achieve the desired request rate, see https://github.com/tsenart/vegeta/issues/344.
It causes our performance tests to fail with for `race: limit on 8128 simultaneously alive goroutines is exceeded, dying`, see https://prow.knative.dev/view/gcs/knative-prow/logs/ci-knative-serving-performance/1152413085785395200. So we'll need to remove the `-race` option from the `run test` command.

This PR moves the `-race` option from `report_go_test` to `go_test_e2e` and `default_unit_test_runner`, so that in `perf-test.sh` we can use `report_go_test` without `-race`.

Please LMK if you think there are other better solutions. 

/cc @adrcunha 
/cc @srinivashegde86

### DIFF
--- a/scripts/library.sh
+++ b/scripts/library.sh
@@ -340,7 +340,7 @@ function report_go_test() {
   # Run tests in verbose mode to capture details.
   # go doesn't like repeating -v, so remove if passed.
   local args=" $@ "
-  local go_test="go test -race -v ${args/ -v / }"
+  local go_test="go test -v ${args/ -v / }"
   # Just run regular go tests if not on Prow.
   echo "Running tests with '${go_test}'"
   local report="$(mktemp)"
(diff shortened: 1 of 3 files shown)


scripts/library.sh
```diff
@@ -340,7 +340,7 @@ function report_go_test() {
   # Run tests in verbose mode to capture details.
   # go doesn't like repeating -v, so remove if passed.
   local args=" $@ "
-  local go_test="go test -race -v ${args/ -v / }"
+  local go_test="go test -v ${args/ -v / }"
```
adrcunha: This is a breaking change. Unit tests won't run with `-race` anymore and it will allow race conditions to slip into code. If -race is a problem for the load tests, I suggest making this a helper function or handling a custom `-norace` parameter.

adrcunha: Actually, I checked and no one is using `report_go_test` directly. It's safe to remove `-race` from here and add it to `default_unit_test_runner()` in `presubmit-tests.sh` and `go_test_e2e()` here.

Fredy-Z: Yeah I was also checking it :-)
Done.

Issue #1164, opened 2019-07-19
Separate each service/deployment config into it's own file


chaodaiG:
Currently we have Prow config files that contains multiple services/deployments, for example `boskos`, `janitor`, `reaper` in https://github.com/knative/test-infra/blob/master/ci/prow/boskos/config.yaml. This approach makes it not easy to roll back from bad config applied by `kubectl apply -f config.yaml`, as `kubectl delete -f config.yaml` would delete everything in the config, while we might only need to delete one of them. 
Proposing splitting them up to have a single service/deployment in each file, like in this example: https://github.com/knative/test-infra/blob/master/ci/prow/boskos/config.yaml

[BONUS] Have a proper strategy of applying them in batches

adrcunha:
/assign @chaodaiG 

The config can be split and put into a separate subdirectory. A `make` target can make it easier to apply all at once. Other operations (like delete), full or individual, should be done by hand to avoid mistakes.

PR #1174, opened 2019-07-23
Break down each deployment to it's own config file

chaodaiG:
Fix: #1164 
/cc @adrcunha 
/cc @coryrc


srinivashegde86:
[nit] Should the files be called <type>_deployment.yaml? or they can be just called <type>.yaml considering they are under deployments dir

chaodaiG:
> [nit] Should the files be called `<type>_deployment.yaml`? or they can be just called `<type>.yaml` considering they are under deployments dir

I think it's a good idea, done

coryrc:
I don't think we should change the Makefile names, because then k8s.io docs will no longer be that useful. Or, if still doing so, the docs in the repo need to be updated.

coryrc:
Oh, looks like we don't have our own copy of docs except in playbook. I don't think the Makefile names should change then.

ci/prow/deployments/horologuim.yaml
```diff
@@ -0,0 +1,40 @@
+# Copyright 2019 The Knative Authors
```
adrcunha: typo: this file should be named horologium.yaml

ci/prow/deployments/prowjob-crd.yaml
```diff
@@ -0,0 +1,64 @@
+# Copyright 2019 The Knative Authors
```
adrcunha: Maybe just crds.yaml?

ci/prow/Makefile
```diff
@@ -45,27 +46,42 @@ update-config:
...
 	$(SET_CONTEXT)
 	kubectl create configmap plugins --from-file=plugins.yaml=plugins.yaml --dry-run -o yaml | kubectl replace configmap plugins -f -
 	$(UNSET_CONTEXT)
 
-update-boskos:
+# Update all deployments of boskos
+update-boskos-deployments:
 	$(SET_CONTEXT)
-	kubectl apply -f boskos/config.yaml
+	@for f in $(shell ls boskos/deployments/*.yaml); do kubectl apply -f $${f}; done
```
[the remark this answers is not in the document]
chaodaiG: that's what I initially tried, it didn't work

ci/prow/Makefile
```diff
@@ -45,27 +46,42 @@ update-config:
...
 	$(UNSET_CONTEXT)
 
-update-boskos-config:
+# Update single deployment of boskos, expect passing in ${NAME} like `make update-boskos-deployment NAME=janitor_deployment`
```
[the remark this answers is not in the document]
chaodaiG: done, add `single` and `all` in target names for single and multiple deployments

ci/prow/deployments/horologuim.yaml
chaodaiG: good catch, done

ci/prow/deployments/prowjob-crd.yaml
chaodaiG: makes sense, done

ci/prow/Makefile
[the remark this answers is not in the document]
chaodaiG: we don't seem to have a section in any README for this Makefile, but I didn't notice some other places in `*.md` files that needs to be updated

PR #1177, opened 2019-07-23
Wait for Boskos instead of fail

### LINKED ISSUE #1101: Kubetest waits for Boskos instead of fail when Boskos ran out
Change Kubetest from k8s/test-infra to allow it taking an additional parameter, which let Kubetest wait for certain amount of time on next available Boskos resource. Knative jobs using Kubetest should use this parameter instruct how long it could wait on Boskos becoming available

chaodaiG:
Currently e2e tests rely on Kubetest acquiring Boskos projects, which would fail if there is no free Boskos project at the moment. The support was recently enabled in https://github.com/kubernetes/test-infra/pull/13217. This PR lets e2e tests waiting for Boskos instead of failing
/hold
This probably won't work until the changes in PR #1176 applied on Prow cluster

Fixes #1101

### DIFF
--- a/scripts/e2e-tests.sh
+++ b/scripts/e2e-tests.sh
@@ -214,8 +214,11 @@ function create_test_cluster() {
   [[ -n "${GCP_PROJECT}" ]] && test_cmd_args+=" --gcp-project ${GCP_PROJECT}"
   [[ -n "${E2E_SCRIPT_CUSTOM_FLAGS[@]}" ]] && test_cmd_args+=" ${E2E_SCRIPT_CUSTOM_FLAGS[@]}"
   local extra_flags=()
-  # If using boskos, save time and let it tear down the cluster
-  (( ! IS_BOSKOS )) && extra_flags+=(--down)
+  if (( IS_BOSKOS )); then # Add arbitrary duration, wait for Boskos projects acquisition before error out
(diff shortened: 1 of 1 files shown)


scripts/e2e-tests.sh
```diff
@@ -214,8 +214,11 @@ function create_test_cluster() {
   [[ -n "${GCP_PROJECT}" ]] && test_cmd_args+=" --gcp-project ${GCP_PROJECT}"
   [[ -n "${E2E_SCRIPT_CUSTOM_FLAGS[@]}" ]] && test_cmd_args+=" ${E2E_SCRIPT_CUSTOM_FLAGS[@]}"
   local extra_flags=()
-  # If using boskos, save time and let it tear down the cluster
-  (( ! IS_BOSKOS )) && extra_flags+=(--down)
+  if (( IS_BOSKOS )); then # Add arbitrary duration, wait for Boskos projects acquisition before error out
+    extra_flags+=(--boskos-wait-duration=20m)
+  else # If using boskos, save time and let it tear down the cluster
```
Fredy-Z: Should be "If not using boskos"?

chaodaiG: makes sense, updated

chaodaiG:
Tested and worked with new image
/hold cancel

PR #1198, opened 2019-07-26
Enable Boskos Metrics server 

### LINKED ISSUE #1160: Enable boskos Metrics reporter
https://github.com/kubernetes/test-infra/tree/master/boskos#other-components

coryrc:
(really just a translation layer for Prometheus metrics collection).

Fix comment example.

Will be needed when Prow supports

Fixes #1160

### DIFF
--- a/ci/prow/boskos/deployments/metrics.yaml
+++ b/ci/prow/boskos/deployments/metrics.yaml
@@ -1,0 +1,40 @@
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: boskos-metrics
+  namespace: test-pods
+spec:
+  selector:
+    app: boskos-metrics
+  ports:
+  - name: default
+    protocol: TCP
+    port: 80
+    targetPort: 8080
+---
+# Metrics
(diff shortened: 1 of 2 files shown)


ci/prow/boskos/deployments/metrics.yaml
```diff
@@ -0,0 +1,40 @@
+---
+apiVersion: v1
+kind: Service
```
srinivashegde86: this will be only exposed cluster clocally. Is that intended? If we want to read the values here from externally, we need to enable the loadbalancer type and with probably a static IP

coryrc: Oh, that's what that was supposed to do. I love the naming.

coryrc: No, I don't think it's necessary, but could be added later if desired.

coryrc:
Also, this config has already been applied.

coryrc:
@srinivashegde86 need approve

PR #1218, opened 2019-07-31
make the subscriber client reusable

yt3liu:
Reusing the pubsub client as needed. 

From [pubsub doc](https://godoc.org/cloud.google.com/go/pubsub#Client): 
```
// Client is a Google Pub/Sub client scoped to a single project.
//
// Clients should be reused rather than being created as needed.
// A Client may be shared by multiple goroutines.
```

Fixes #


TrevorFarrelly:
thanks!

PR #1234, opened 2019-08-06
use ID instead of login for flaky-test-retryer

TrevorFarrelly:
identify previous comments by user ID rather than login, so we can handle updates to the bot's display name and login info. Tested locally.

Fixes #

### DIFF
--- a/tools/flaky-test-retryer/github_commenter.go
+++ b/tools/flaky-test-retryer/github_commenter.go
@@ -45,7 +45,7 @@ var (
 // GithubClient wraps the ghutil Github client
 type GithubClient struct {
 	ghutil.GithubOperations
-	Login  string
+	ID     int64
 	Dryrun bool
 }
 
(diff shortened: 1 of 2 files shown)


tools/flaky-test-retryer/github_commenter.go
```diff
@@ -59,7 +59,7 @@ func NewGithubClient(githubAccount string, dryrun bool) (*GithubClient, error) {
 	if err != nil {
 		return nil, err
 	}
-	return &GithubClient{ghc, *user.Login, dryrun}, nil
+	return &GithubClient{ghc, user.GetID(), dryrun}, nil
```
chaodaiG: GetID return int64 instead of ptr, right?

chaodaiG:
/ok-to-test
/lgtm
/approve

/hold
feel free unhold it if the answer to my question is yes

tools/flaky-test-retryer/github_commenter.go
TrevorFarrelly: Yes, seems the getter funcs strip the pointer off for us: https://godoc.org/github.com/google/go-github/github#User.GetID

PR #1235, opened 2019-08-06
flaky test retryer uses verioned tag instead of latest

chaodaiG:
See best practice here: https://kubernetes.io/docs/concepts/configuration/overview/#container-images

### DIFF
--- a/tools/flaky-test-retryer/gke_deployment/retryer_service.yaml
+++ b/tools/flaky-test-retryer/gke_deployment/retryer_service.yaml
@@ -37,7 +37,7 @@ spec:
     spec:
       containers:
       - name: flaky-test-retry-bot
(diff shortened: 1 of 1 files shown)


chaodaiG:
/cc @srinivashegde86 
FYI @TrevorFarrelly

srinivashegde86:
Are we going to verify the image once we upload it somehow, before using the tagged image? If not, we are adding a manual step doing what latest will do for us automatically.

chaodaiG:
The problem of `latest`, is that it makes it hard to know which image is really running on the cluster, or whether the image has been successfully updated since it's always shown as `latest`. So I think we should move away from using `latest` in this service, just like mentioned in the `best practice` link in the PR body.

chaodaiG:
Alternatively, we can keep `latest` tag in the yaml file, but always update the service with specific tag

PR #1237, opened 2019-08-06
remove duplicates in ARTIFACTS_TO_PUBLISH

### LINKED ISSUE #1233: serving 0.8 release failed publishing
The first run of dot_release for serving 0.8 release branch failed. All tests passed, but failed pu
blishing:
```
Error uploading release asset: Unprocessable Entity (HTTP 422)
Duplicate value for "name"
Error publishing the release, retrying in 15s...
Error creating release: Unprocessable Entity (HTTP 422)
Duplicate value for "tag_name"
Error publishing the release, retrying in 15s...
Error creating release: Unprocessable Entity (HTTP 422)
Duplicate value for "tag_name"
error: Cannot publish release to GitHub
```

The job is here: https://prow.knative.dev/view/gcs/knative-prow/logs/ci-knative-serving-auto-release/1158748345653530625

steuhs:
To fix the issue here: https://github.com/knative/test-infra/issues/1233

Fixes #1233

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -472,6 +472,7 @@ function run_validation_tests() {
 # Parameters: $1..$n - files to add to the release.
 function publish_artifacts() {
   (( ! PUBLISH_RELEASE )) && return
+  awk '!seen[$0]++' ARTIFACTS_TO_PUBLISH > ARTIFACTS_TO_PUBLISH
   tag_images_in_yamls ${ARTIFACTS_TO_PUBLISH}
   publish_to_gcs ${ARTIFACTS_TO_PUBLISH}
   publish_to_github ${ARTIFACTS_TO_PUBLISH}


scripts/release.sh
```diff
@@ -472,6 +472,7 @@ function run_validation_tests() {
 # Parameters: $1..$n - files to add to the release.
 function publish_artifacts() {
   (( ! PUBLISH_RELEASE )) && return
+  awk '!seen[$0]++' ARTIFACTS_TO_PUBLISH > ARTIFACTS_TO_PUBLISH
```
chaodaiG: can we use `sort | uniq` instead? `awk` is pretty hard to understand and maintain

chaodaiG: also, I think it should be at line 522

Fredy-Z:
/hold
This change does not solve this problem, as the value of `ARTIFACTS_TO_PUBLISH` is a list of file names separated by spaces, like `a.yaml b.yaml c.yaml a.yaml`. 
https://unix.stackexchange.com/questions/353321/remove-all-duplicate-word-from-string-using-shell-script for reference.

scripts/release.sh
steuhs: I think we can just use uniq. 
I think we should remove the duplicate in early stage as possible because it is creates redundancy and may introduce other unexpected problems in the future

steuhs:
Addressed the comments in this PR and opened a new one in https://github.com/knative/serving/pull/5080 to replace this

PR #1244, opened 2019-08-08
[coverage] add additional health check for a post-submit build

### LINKED ISSUE #1242: Coverage shouldn't fail if the baseline it compares to failed
Currently presubmit coverage test looks for latest master coverage result and compares against it, the latest master coverage result is produced by postsubmit coverage test. Normally this flow works, but when there is one postsubmit coverage test failed, presubmit coverage test would fail due to not able to find baseline. An example is here: https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/knative_eventing/1658/pull-knative-eventing-go-coverage/1159243486059302912

Currently it's fine as coverage is optional, so PRs can still merge and keep generating new baselines. But it's not hard to imagine this could block all PRs from merging once it becomes required. We should either retry postsubmit or allow presubmit comparing with earlier runs, to make this flow more resilient

steuhs:
Each pre-submit coverage job looks for coverage profile for the latest healthy master build. We add the check of profile existence to the health check so that master build without a coverage profile will be ignored.

Fixes #1242

### DIFF
--- a/tools/coverage/gcs/gcsPostsubmit.go
+++ b/tools/coverage/gcs/gcsPostsubmit.go
@@ -95,7 +95,8 @@ func (p *PostSubmit) dirOfCompletionMarker(build int) (result string) {
 
 func (p *PostSubmit) isBuildHealthy(build int) bool {
 	return p.StorageClient.DoesObjectExist(p.Ctx, p.Bucket,
-		p.dirOfCompletionMarker(build))
+		p.dirOfCompletionMarker(build)) &&
+		p.StorageClient.DoesObjectExist(p.Ctx, p.Bucket, path.Join(p.dirOfArtifacts(build), p.covProfileName))
 }
 
 func (p *PostSubmit) pathToGoodCoverageArtifacts() (result string) {


adrcunha:
Please elaborate on the PR description.

tools/coverage/gcs/gcsPostsubmit.go
```diff
@@ -95,7 +95,8 @@ func (p *PostSubmit) dirOfCompletionMarker(build int) (result string) {
 
 func (p *PostSubmit) isBuildHealthy(build int) bool {
 	return p.StorageClient.DoesObjectExist(p.Ctx, p.Bucket,
-		p.dirOfCompletionMarker(build))
+		p.dirOfCompletionMarker(build)) &&
+		p.StorageClient.DoesObjectExist(p.Ctx, p.Bucket, path.Join(p.dirOfArtifacts(build), p.covProfileName))
```
chaodaiG: the logic is searching backward until find a good one, there is a potential risk of comparing with something really old

steuhs: In practice that won't happen. The additional condition is false only if all unit tests fail thus no coverage profile is produced at all. When that happen, the failure should only affect post-submit jobs for a very limited time - unit test failures that affect all jobs will not stay in the master branch for long.

adrcunha: Isn't this also currently true? So, what's the point of this check? Even if the baseline is not that old, you will (even for a very limited time -- how long exactly is "limited"?) get a wrong report, no?

chaodaiG: this is also currently true. Ideally the baseline should be the commit that the pull merge based on, anything else should be a wrong report. Currently the criteria is already lowered to rely on latest discoverable coverage report in master, and if we're ok with this logic, then it should be clear how far back we allow it go

steuhs:
> Please elaborate on the PR description.

Done

tools/coverage/gcs/gcsPostsubmit.go
adrcunha: I checked the code and the function that lists the baseline builds does sort them from newest to oldest:

https://github.com/knative/test-infra/blob/b3c110910057090babb081ce0b1d1b1d326649c0/tools/coverage/gcs/gcsPostsubmit.go#L74

So I think we're good here.

chaodaiG: I think we should at least have a limit on the exhaustive for loop: https://github.com/knative/test-infra/blob/b3c110910057090babb081ce0b1d1b1d326649c0/tools/coverage/gcs/gcsPostsubmit.go#L112
An extreme case is postsubmit stopped working one day without anyone notice, and all coverage reports are wrong

adrcunha: Makes sense. Can we use monitoring for catching that?

chaodaiG: catching postsubmit failures? I think we can, just could be noisy, as I have noticed them failed a few times for various reasons

steuhs: We should rather monitor on post-submit failures

steuhs: Instead of having a limit on the loop, I think we should fail the post-submit job with failed test and then use monitoring on it.

@chaodaiG Regarding whether to use the master branch as the baseline instead of what you proposed - it was a design decision made at the beginning of the project and the advantages were discussed before it was finalized. It was NOT of lowered criteria.

chaodaiG: I'm ok with the decision, my point is we should prevent it from comparing with "TOO OLD" result in master, which could be very inaccurate

chaodaiG: What would be the action item if monitoring found out about postsubmit failure?

steuhs: @chaodaiG if monitoring find out about postsubmit failures, oncall will be notified and they can look into  why presubmit failed multiple times

chaodaiG: This isn't a blocking issue for now, since the report contains absolute coverage data. We should definitely revisit this when there is need using regression ratio for PR gating

chaodaiG: But please add the monitoring stuff for postsubmit as we discussed

chaodaiG:
/approve

Wait for @adrcunha  applying the lgtm

tools/coverage/gcs/gcsPostsubmit.go
adrcunha: I'll LGTM once the monitoring is in place, either here or in a different PR.

steuhs: @adrcunha Here is the PR that implements monitoring https://github.com/knative/test-infra/pull/1324

adrcunha: Friendly ping. 1324 is still open, and this PR pretty dusty. :P

chaodaiG:
/unassign

cc me if needed

shinigambit:
/close
It's been stuck for a while so I don't think this will get in

PR #1252, opened 2019-08-09
added boskoscli to the prow-test image

yt3liu:

Fixes #


srinivashegde86:
can you add some info on why we need this or link to an issue?

yt3liu:
Closing this issue per offline discussion.

srinivashegde86:
boskoscli adds the ability to use boskos using the cli. But, we are not using it as of now and test-infra 2.0 will provide an api to hook into boskos directly. We can add this later, if we need a cli to interact with boskos.

PR #1255, opened 2019-08-12
Add prow jobs to update clusters running serving benchmarks

Fredy-Z:
Add prow jobs to update clusters running serving benchmarks, to replace the cronjobs we are running now:
https://github.com/knative/serving/blob/master/test/performance/tools/update-serving/update.yaml
https://github.com/knative/serving/blob/master/test/performance/tools/recreate-serving/recreate.yaml

This PR serves as the first step to migrate the performance testing infra in `serving` to `test-infra`, a few following steps will be:
1. Migrate common part in https://github.com/knative/serving/tree/master/test/performance/tools to `test-infra` (also rewrite them with `Go` for test-infra 2.0)
2. Use the above common code to setup the environment in `serving`, `eventing`, etc., and install project-related components for running the benchmarks

As a few design decisions related to `Mako` haven't been made, we haven't come out an end-to-end solution for performance testing infra. We should be able to have something new in next week's meeting. After the decision is made, I will write a detailed design proposal for it.

/cc @srinivashegde86 
/cc @adrcunha 
/cc @chaodaiG

### DIFF
--- a/ci/prow/config.yaml
+++ b/ci/prow/config.yaml
@@ -5630,6 +5630,54 @@ periodics:
     - name: housekeeping-github-token
       secret:
         secretName: housekeeping-github-token
+- cron: "30 07 * * *"
+  name: ci-knative-serving-recreate-clusters
+  agent: kubernetes
+  decorate: true
+  extra_refs:
+  - org: knative
+    repo: test-infra
+  spec:
+    containers:
+    - image: gcr.io/knative-performance/update-serving:v20190809-c290f606fc20
+      imagePullPolicy: Always
+      command:
+      - "/workspace/tools/recreate-serving/recreate.sh"
+      volumeMounts:
+      - name: performance-test
+        mountPath: /etc/performance-test
+        readOnly: true
+      env:
+      - name: GOOGLE_APPLICATION_CREDENTIALS
+        value: /etc/performance-test/service-account.json
+    volumes:
+    - name: performance-test
+      secret:
+        secretName: performance-test
+- cron: "0 * * * *"
+  name: ci-knative-serving-update-clusters
+  agent: kubernetes
+  decorate: true
+  extra_refs:
+  - org: knative
+    repo: test-infra
+  spec:
+    containers:
+    - image: gcr.io/knative-performance/update-serving:v20190809-c290f606fc20
+      imagePullPolicy: Always
+      command:
+      - "/workspace/tools/update-serving/update.sh"
+      volumeMounts:
+      - name: performance-test
+        mountPath: /etc/performance-test
+        readOnly: true
+      env:
+      - name: GOOGLE_APPLICATION_CREDENTIALS
+        value: /etc/performance-test/service-account.json
+    volumes:
+    - name: performance-test
+      secret:
+        secretName: performance-test
 postsubmits:
   knative/serving:
   - name: post-knative-serving-go-coverage
--- a/ci/prow/config.yaml
+++ b/ci/prow/config.yaml
@@ -5519,6 +5519,7 @@ periodics:
   extra_refs:
   - org: knative
     repo: test-infra
+    base_ref: master
   spec:
     containers:
     - image: gcr.io/k8s-prow/commenter:v20190731-e3f7b9853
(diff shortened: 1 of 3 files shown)


Fredy-Z:
/hold
Let me test it first.

chaodaiG:
sounds like a good opportunity to move the entire performance common package to be under `knative/pkg`, like what's done in this PR: https://github.com/knative/pkg/pull/555

ci/prow/config.yaml
```diff
@@ -5630,6 +5630,54 @@ periodics:
...
+  decorate: true
+  extra_refs:
+  - org: knative
+    repo: test-infra
+  spec:
+    containers:
+    - image: gcr.io/knative-performance/update-serving:v20190809-c290f606fc20
+      imagePullPolicy: Always
+      command:
+      - "/workspace/tools/recreate-serving/recreate.sh"
```
chaodaiG: where does this script live now?

ci/prow/config.yaml
```diff
@@ -5630,6 +5630,54 @@ periodics:
...
+  decorate: true
+  extra_refs:
+  - org: knative
+    repo: test-infra
+  spec:
+    containers:
+    - image: gcr.io/knative-performance/update-serving:v20190809-c290f606fc20
+      imagePullPolicy: Always
+      command:
+      - "/workspace/tools/update-serving/update.sh"
```
adrcunha: Same question as Chao above.

ci/prow/config.yaml
```diff
@@ -5630,6 +5630,54 @@ periodics:
...
+      volumeMounts:
+      - name: performance-test
+        mountPath: /etc/performance-test
```
adrcunha: I assume this is a secret already present in the Prow cluster?

ci/prow/periodic_config.go
```diff
@@ -398,3 +402,38 @@ func generateIssueTrackerPeriodicJob(jobName, labelFilter, updatedTime, comment
 	addVolumeToJob(&data.Base, "/etc/housekeeping-github-token", "housekeeping-github-token", true, "")
 	executeJobTemplate(jobName, readTemplate(periodicCustomJob), "presubmits", "", data.PeriodicJobName, false, data)
 }
+
+// generateServingClusterUpdatePeriodicJobs generates periodic jobs to update serving clusters
+// that run performance testing benchmarks
+func generateServingClusterUpdatePeriodicJobs() {
+	recreateServingClustersJobName := "ci-knative-serving-recreate-clusters"
```
adrcunha: You're not reusing the strings, why declare them and then have a single usage instead of embedding into the function call?

ci/prow/config.yaml
srinivashegde86: we should not be using this image. We can directly run the scripts in serving like we do with perf-tests.sh https://github.com/knative/serving/blob/master/test/performance/tools/update-serving/update.sh

Fredy-Z:
Wait for https://github.com/knative/serving/pull/5142 to be merged.

ci/prow/config.yaml
Fredy-Z: Done.

ci/prow/periodic_config.go
Fredy-Z: Fixed.

ci/prow/config.yaml
Fredy-Z: Yes, I created it on the Prow cluster.

ci/prow/config.yaml
Fredy-Z: It's now pointing to one script in https://github.com/knative/serving/pull/5142, so we'll need to wait for that PR to be merged first.

ci/prow/config.yaml
Fredy-Z: Ditto.

ci/prow/config.yaml
```diff
@@ -5618,6 +5618,78 @@ periodics:
...
+      env:
+      - name: GOOGLE_APPLICATION_CREDENTIALS
+        value: /etc/test-account/service-account.json
+      - name: PERF_TEST_GOOGLE_APPLICATION_CREDENTIALS
+        value: /etc/performance-test/service-account.json
+    volumes:
+    - name: test-account
+      secret:
+        secretName: test-account
+    - name: performance-test
```
srinivashegde86: this secret is in the default ns. This needs to be in test-pods ns

Fredy-Z: Thanks for catching this! I realized it but forgot to make the update. 
Added.

Fredy-Z:
https://github.com/knative/serving/pull/5142 has been merged.
/hold cancel

chaodaiG:
/approve
wait for the other reviewers do /lgtm

PR #1276, opened 2019-08-16
google/knative-gcp mimics all repo level settings in knative

chaodaiG:


adrcunha:
/hold

Please double-check that this is indeed desirable by the repo admins.

ci/prow/templates/prow_config_header.yaml
```diff
@@ -100,6 +100,7 @@ tide:
     - do-not-merge/invalid-owners-file
   merge_method:
     knative: squash
+    google: squash
```
adrcunha: If `google/knative-gcp` can be used instead, it would be a better choice.

chaodaiG:
/hold
wait for @n3wscott 's comment

adrcunha:
/lgtm

@chaodaiG unhold if you're still OK with these changes.

ci/prow/templates/prow_config_header.yaml
chaodaiG: done

PR #1280, opened 2019-08-19
Convenience Makefile credential rules for monitoring cluster

adrcunha:


tools/monitoring/Makefile
```diff
@@ -12,5 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+CLUSTER       ?= monitoring
```
chaodaiG: should this be `CLUSTER = monitoring` to avoid accident?

adrcunha: 1. This is intended to give flexibility.
1. I don't think "make CLUSTER=prow" would configure an accident. :)

chaodaiG: gotcha

PR #1284, opened 2019-08-20
Add a new github method - GetPullRequestByCommitID

Fredy-Z:
1. Add a new github method - `GetPullRequestByCommitID`, so we can find a PR with the commit ID generated from "git squash"
2. `ListPullRequestsWithCommit` is a new method in github client, so I'll need to update the github dependency.

/cc @adrcunha 
/cc @chaodaiG


shared/ghutil/pullrequest.go
```diff
@@ -133,6 +133,35 @@ func (gc *GithubClient) GetPullRequest(org, repo string, ID int) (*github.PullRe
 	return res, err
 }
 
+// GetPullRequestByCommitID gets PullRequest by commit ID
+func (gc *GithubClient) GetPullRequestByCommitID(org, repo string, commitID string) (*github.PullRequest, error) {
```
chaodaiG: Your may also want to add it to fake client so it's easier for you writing unit tests later

srinivashegde86:
```suggestion
func (gc *GithubClient) GetPullRequestByCommitID(org, repo, commitID string) (*github.PullRequest, error) {
```

shared/ghutil/pullrequest.go
```diff
@@ -133,6 +133,35 @@ func (gc *GithubClient) GetPullRequest(org, repo string, ID int) (*github.PullRe
...
+	var res []*github.PullRequest
+	_, err := gc.retry(
```
srinivashegde86: use combined if?

shared/ghutil/pullrequest.go
Fredy-Z: Added.

shared/ghutil/pullrequest.go
Fredy-Z: Done. Thanks!

chaodaiG:
/hold
add the method in this interface as well:
https://github.com/knative/test-infra/blob/master/shared/ghutil/client.go

Fredy-Z:
> /hold
> add the method in this interface as well:
> https://github.com/knative/test-infra/blob/master/shared/ghutil/client.go

Added.

PR #1290, opened 2019-08-20
[coverage] update string mapper to accommodate knative repo structure change

steuhs:
The change of repo structure from github.com/$REPO_OWNER/$REPO_NAME to knative.dev/$REPO_NAME/ breaks the function FilePathProfileToGithub, because the function is based on the repo structure. This change is to fix the issue.

Fixes #

### DIFF
--- a/tools/coverage/githubUtil/github.go
+++ b/tools/coverage/githubUtil/github.go
@@ -21,13 +21,13 @@ import (
 )
 
 // convert filepath from profile format to github format
-// (github.com/$REPO_OWNER/$REPO_NAME/pkg/... -> pkg/...)
+// (knative.dev/$REPO_NAME/pkg/... -> pkg/...)
 func FilePathProfileToGithub(filePath string) string {
-	slice := strings.SplitN(filePath, "/", 4)
-	if len(slice) < 4 {
-		log.Printf("FilePath string cannot be splitted into 4 parts: [sep=%s] %s; "+
+	slice := strings.SplitN(filePath, "/", 3)
+	if len(slice) < 3 {
+		log.Printf("FilePath string cannot be splitted into 3 parts: [sep=%s] %s; "+
 			"Original string is returned\n", "/", filePath)
 		return filePath
 	}
-	return slice[3]
+	return slice[2]
 }


tools/coverage/githubUtil/github.go
```diff
@@ -21,13 +21,13 @@ import (
 )
 
 // convert filepath from profile format to github format
-// (github.com/$REPO_OWNER/$REPO_NAME/pkg/... -> pkg/...)
+// (knative.dev/$REPO_NAME/pkg/... -> pkg/...)
```
chaodaiG: not all repos are on `knative.dev` yet, this will break some other repos

JRBANCEL:
Fix the corresponding test.
Will it brake the repos that are not using knative.dev/... ?
/lgtm

tools/coverage/githubUtil/github.go
steuhs: I'll change it to handle different logic

tools/coverage/githubUtil/github.go
```diff
@@ -16,18 +16,45 @@ limitations under the License.
...
+	if result == file {
+		log.Fatalf("repo path (%s) is not a prefix of filepath (%s):", repoPath, file)
+	}
+	return result
+}
+
+// GetRepoPath gets repository path relative to GOPATH/src
+func GetRepoPath() (string, error) {
+	out, err := exec.Command("git", "rev-parse", "--show-toplevel").Output()
+	if nil != err {
```
JRBANCEL: nit: 'err != nil' is the Go way :-)

tools/coverage/githubUtil/github.go
```diff
@@ -16,18 +16,45 @@ limitations under the License.
...
+		return "", fmt.Errorf("failed git rev-parse --show-toplevel: '%v'", err)
+	}
+	gopath := os.Getenv("GOPATH")
+	if gopath == "" {
+		return "", errors.New("GOPATH is empty")
+	}
+	relPath, err := filepath.Rel(gopath+"/src", string(out))
```
JRBANCEL: path.Join(gopath, "src") is safer, you don't have to deal with '/'

tools/coverage/githubUtil/github.go
```diff
@@ -16,18 +16,45 @@ limitations under the License.
...
+func GetRepoPath() (string, error) {
+	out, err := exec.Command("git", "rev-parse", "--show-toplevel").Output()
+	if err != nil {
+		return "", fmt.Errorf("failed git rev-parse --show-toplevel: '%v'", err)
+	}
+	gopath := os.Getenv("GOPATH")
+	if gopath == "" {
+		return "", errors.New("GOPATH is empty")
+	}
+	relPath, err := filepath.Rel(gopath+"/src", string(out))
```
yt3liu: For appending the path, please use path.Join as this prevents the need of adding `/` in front of `src` and remove the need to worry about extra "/"s in the path. 
https://golang.org/pkg/path/#Join

tools/coverage/githubUtil/github_test.go
```diff
@@ -20,7 +20,7 @@ import (
 )
 
 func TestFilePathProfileToGithub(t *testing.T) {
-	input := "github.com/myRepoOwner/myRepoName/pkg/ab/cde"
+	input := "knative.dev/test-infra/pkg/ab/cde"
```
JRBANCEL: It would be better to test both scenarios to ensure there is no regression.

steuhs: I don't think we can test both scenarios. GetRepoPath is making system calls it uses actual system env vars and directory.

adrcunha: Not a problem. You can override GOPATH or mock GetRepoPath.

tools/coverage/githubUtil/github_test.go
```diff
@@ -16,15 +16,35 @@ limitations under the License.
...
+		getRepoRoot = func() (string, error) {
+			return repoRoot, nil
+		}
+		gopath := os.Getenv("GOPATH")
+		os.Setenv("GOPATH", "/d1/d2/d3/gopath")
+		defer os.Setenv("GOPATH", gopath)
+		actualOutput := FilePathProfileToGithub(input)
+
+		if actualOutput != expectedOutput {
+			t.Fatalf("input=%s; expected output=%s; actual output=%s", input, expectedOutput,
```
yt3liu: Generally, use t.Error and not t.Fatal to compare properties of a function's output.

tools/coverage/githubUtil/github_test.go
```diff
@@ -16,15 +16,35 @@ limitations under the License.
...
-	if actualOutput != expectedOutput {
-		t.Fatalf("input=%s; expected output=%s; actual output=%s", input, expectedOutput,
-			actualOutput)
-	}
+	t.Run("repo=knative.dev", func(t *testing.T) {
+		input := "knative.dev/test-infra/pkg/ab/cde"
+		expectedOutput := "pkg/ab/cde"
+		actualOutput := FilePathProfileToGithub(input)
+		if actualOutput != expectedOutput {
+			t.Fatalf("input=%s; expected output=%s; actual output=%s", input, expectedOutput,
```
yt3liu: Could you please change the output to follow this format: 
```
YourFunc(%v) = %v, want %v"
```

tools/coverage/githubUtil/github_test.go
```diff
@@ -16,15 +16,35 @@ limitations under the License.
...
 
 func TestFilePathProfileToGithub(t *testing.T) {
-	input := "github.com/myRepoOwner/myRepoName/pkg/ab/cde"
-	expectedOutput := "pkg/ab/cde"
-	actualOutput := FilePathProfileToGithub(input)
-	if actualOutput != expectedOutput {
-		t.Fatalf("input=%s; expected output=%s; actual output=%s", input, expectedOutput,
-			actualOutput)
-	}
+	t.Run("repo=knative.dev", func(t *testing.T) {
```
yt3liu: Please change this to use table-driven test.

tools/coverage/githubUtil/github_test.go
```diff
@@ -20,31 +20,31 @@ import (
...
-			return repoRoot, nil
-		}
-		gopath := os.Getenv("GOPATH")
-		os.Setenv("GOPATH", "/d1/d2/d3/gopath")
-		defer os.Setenv("GOPATH", gopath)
-		actualOutput := FilePathProfileToGithub(input)
+	input := "github.com/myRepoOwner/myRepoName/pkg/ab/cde"
+	expectedOutput := "pkg/ab/cde"
+	repoRoot := "/d1/d2/d3/gopath/src/github.com/myRepoOwner/myRepoName"
+	getRepoRoot = func() (string, error) {
```
adrcunha: Restore the original function once your test is done.

tools/coverage/githubUtil/github.go
```diff
@@ -16,18 +16,51 @@ limitations under the License.
...
 	}
-	return slice[3]
+	result := strings.TrimPrefix(file, repoPath+"/")
+	if result == file {
+		log.Fatalf("repo path (%s) is not a prefix of filepath (%s):", repoPath, file)
+	}
+	return result
+}
+
+var getRepoRoot = func() (string, error) {
```
adrcunha: Add a comment stating what this does and why it's a var instead of a function.

PR #1295, opened 2019-08-20
Remove update-test-infra.sh since sockpuppet already does it

chaodaiG:
Sockpuppet already does the same thing every day, there is no need to keep this script


srinivashegde86:
sockpuppet already does the updates every morning. Do we need this script as well?

chaodaiG:
Initially sockpuppet only does serving, it seems that it's doing it everywhere now, I'll turn this PR into a delete then

PR #1303, opened 2019-08-21
Move slack operations to shared

Fredy-Z:
Move slack operations from `flaky-test-reporter` to `shared`, so we can use it in other repos.

/cc @chaodaiG

### DIFF
--- a/shared/slackutil/messaging.go
+++ b/shared/slackutil/messaging.go
@@ -28,36 +28,34 @@ import (
 	"net/url"
 )
 
-const (
-	slackChatPostMessageURL = "https://slack.com/api/chat.postMessage"
-)
+const postMessageURL = "https://slack.com/api/chat.postMessage"
 
-// SlackOperations defines the operations that can be done to Slack
-type SlackOperations interface {
-	PostMessageToChannel(text, channel string) error
+// Operations defines the operations that can be done to Slack
+type Operations interface {
+	Post(text, channel string) error
 }
 
-// slackClient contains Slack bot related information
-type slackClient struct {
+// client contains Slack bot related information
+type client struct {
 	userName  string
 	tokenStr  string
 	iconEmoji *string
 }
 
-// NewSlackClient reads token file and stores it for later authentication
-func NewSlackClient(userName, slackTokenPath string) (SlackOperations, error) {
-	b, err := ioutil.ReadFile(slackTokenPath)
+// NewClient reads token file and stores it for later authentication
+func NewClient(userName, tokenPath string) (Operations, error) {
+	b, err := ioutil.ReadFile(tokenPath)
 	if err != nil {
 		return nil, err
 	}
-	return &slackClient{
+	return &client{
 		userName: userName,
 		tokenStr: strings.TrimSpace(string(b)),
 	}, nil
 }
 
-// PostMessageToChannel posts the given text to channel
-func (c *slackClient) PostMessageToChannel(text, channel string) error {
+// Post posts the given text to channel
+func (c *client) Post(text, channel string) error {
 	uv := url.Values{}
 	uv.Add("username", c.userName)
 	uv.Add("token", c.tokenStr)
@@ -71,8 +69,8 @@ func (c *slackClient) PostMessageToChannel(text, channel string) error {
 }
 
 // postMessage does http post
-func (c *slackClient) postMessage(uv url.Values) error {
-	resp, err := http.PostForm(slackChatPostMessageURL, uv)
+func (c *client) postMessage(uv url.Values) error {
+	resp, err := http.PostForm(postMessageURL, uv)
 	if err != nil {
 		return err
 	}
--- a/shared/slackutil/messaging.go
+++ b/shared/slackutil/messaging.go
@@ -58,7 +58,7 @@ func NewSlackClient(userName, slackTokenPath string) (SlackOperations, error) {
 
 // PostMessageToChannel posts the given text to channel
 func (c *slackClient) PostMessageToChannel(text, channel string) error {
-	uv := &url.Values{}
+	uv := url.Values{}
 	uv.Add("username", c.userName)
 	uv.Add("token", c.tokenStr)
 	if nil != c.iconEmoji {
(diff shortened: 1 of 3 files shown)


shared/slackutil/messaging.go
```diff
@@ -0,0 +1,92 @@
...
+		uv.Add("icon_emoji", *c.iconEmoji)
+	}
+	uv.Add("channel", channel)
+	uv.Add("text", text)
+
+	return c.postMessage(uv)
+}
+
+// postMessage does http post
+func (c *slackClient) postMessage(uv *url.Values) error {
```
srinivashegde86:
```suggestion
func (c *slackClient) postMessage(uv url.Values) error {
```

srinivashegde86: Dont move pointers around, if not needed. We are just converting it back and forth

shared/slackutil/messaging.go
```diff
@@ -0,0 +1,92 @@
...
+	resp, err := http.PostForm(slackChatPostMessageURL, *uv)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+	t, _ := ioutil.ReadAll(resp.Body)
```
srinivashegde86: do this after the if

shared/slackutil/messaging.go
```diff
@@ -0,0 +1,92 @@
...
+	if resp.StatusCode != 200 {
```
srinivashegde86: use http const StatusOK instead of 200

shared/slackutil/messaging.go
```diff
@@ -0,0 +1,92 @@
...
+		return fmt.Errorf("http response code is not 200 '%s'", string(t))
+	}
+	// response code could also be 200 if channel doesn't exist, parse response body to find out
+	var b struct {
```
srinivashegde86: why do we need to check this as well?

shared/slackutil/messaging.go
Fredy-Z: Done.

shared/slackutil/messaging.go
Fredy-Z: Done.

shared/slackutil/messaging.go
Fredy-Z: The comment explained it.

shared/slackutil/messaging.go
Fredy-Z: It's used in the following error message, I think it's better to keep it for debugging purpose.

shared/slackutil/messaging.go
```diff
@@ -0,0 +1,92 @@
...
+	slackChatPostMessageURL = "https://slack.com/api/chat.postMessage"
+)
+
+// SlackOperations defines the operations that can be done to Slack
+type SlackOperations interface {
+	PostMessageToChannel(text, channel string) error
+}
+
+// slackClient contains Slack bot related information
+type slackClient struct {
```
srinivashegde86: should this be called just client? Else, this will be slackutil.slackClient, which adds the stutter.

shared/slackutil/messaging.go
```diff
@@ -0,0 +1,92 @@
...
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"strings"
+
+	"net/http"
+	"net/url"
+)
+
+const (
```
srinivashegde86: [nit] dont need the (). Can be const postMessageUrl = ".."

shared/slackutil/messaging.go
```diff
@@ -0,0 +1,92 @@
...
+	slackChatPostMessageURL = "https://slack.com/api/chat.postMessage"
```
srinivashegde86:
```suggestion
	postMessageURL = "https://slack.com/api/chat.postMessage"
```

shared/slackutil/messaging.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+// SlackOperations defines the operations that can be done to Slack
+type SlackOperations interface {
+	PostMessageToChannel(text, channel string) error
```
srinivashegde86:
```suggestion
	Post(text, channel string) error
```

shared/slackutil/messaging.go
Fredy-Z: Removed.

shared/slackutil/messaging.go
Fredy-Z: Done.

shared/slackutil/messaging.go
```diff
@@ -0,0 +1,90 @@
...
+
+const postMessageURL = "https://slack.com/api/chat.postMessage"
+
+// Operations defines the operations that can be done to Slack
+type Operations interface {
+	Post(text, channel string) error
+}
+
+// client contains Slack bot related information
+type client struct {
```
chaodaiG: I think this needs to be exposed

shared/slackutil/messaging.go
```diff
@@ -0,0 +1,90 @@
...
+	userName  string
+	tokenStr  string
+	iconEmoji *string
+}
+
+// NewClient reads token file and stores it for later authentication
+func NewClient(userName, tokenPath string) (Operations, error) {
```
chaodaiG: shouldn't it return Client instead?

shared/slackutil/messaging.go
Fredy-Z: It doesn't have to, see `Interface factories` in https://www.sohamkamani.com/blog/golang/2018-06-20-golang-factory-patterns/.

The benefit of writing like this is, we do not expose it to the user, so user will only be able to construct the `client` with `NewClient`, in which we can enforce validations for the input params.

shared/slackutil/messaging.go
Fredy-Z: Ditto.

srinivashegde86:
will wait for chao to lgtm

shared/slackutil/messaging.go
chaodaiG: I see, that's good to learn, the downside I can think about this approach is user will not be able to cast, but it's ok in this case, since it's just a simple wrapper

PR #1300, opened 2019-08-21
build-test: do not run e2e tests in presubmit.

n3wscott:
See: https://storage.googleapis.com/knative-prow/pr-logs/pull/google_knative-gcp/200/pull-google-knative-gcp-build-tests/1164200726688174082/build-log.txt

PR #1305, opened 2019-08-22
Set serving build test requests/limits both to 16Gb memory

chaodaiG:
Serving build tests are killed OOM sometimes, increase the limit from 8Gb to 16Gb to help alleviate the problem. 
Example: https://prow.knative.dev/view/gcs/knative-prow/logs/ci-knative-serving-continuous/1163966682885001216

/cc @mattmoor 
/cc @adrcunha 
/cc @yt3liu 

/hold
May not be needed if #1300 solves the problem by building sequentially

### DIFF
--- a/ci/prow/make_config.go
+++ b/ci/prow/make_config.go
@@ -90,6 +90,7 @@ type baseProwJobTemplateData struct {
 	Env                 []string
 	Volumes             []string
 	VolumeMounts        []string
+	Resources           []string
 	Timeout             int
 	AlwaysRun           bool
 	LogsDir             string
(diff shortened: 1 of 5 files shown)


ci/prow/make_config.go
```diff
@@ -462,6 +464,17 @@ func setupDockerInDockerForJob(data *baseProwJobTemplateData) {
 	(*data).SecurityContext = []string{"privileged: true"}
 }
 
+// setReourcesReqForJob sets resource requirement for job
+func setReourcesReqForJob(res yaml.MapSlice, data *baseProwJobTemplateData) {
+	for _, val := range res {
+		foo := getString(val.Key)
```
adrcunha: foo? Really? :)

chaodaiG: what an accident :)

adrcunha:
/lgtm
/approve

I do think that checking if 1300 helps this issue is indeed worth.

ci/prow/make_config.go
chaodaiG: fixed. Thanks!

chaodaiG:
It's still happening in serving: https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/knative_serving/5202/pull-knative-serving-build-tests/1164953939439783936

/hold cancel

PR #1306, opened 2019-08-22
Fix flaky-test-retryer: scan current run instead of latest

chaodaiG:
Retryer skipped retrying one of failed tests in serving, the log says there is no test failed. The reason could be it's scanning latest build, and this is not reliable if someone manually triggered rerun. Using RunID instead of latest

Example: https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/knative_serving/5249/pull-knative-serving-integration-tests/1164571049820426240

/cc @srinivashegde86 

/hold
let me test it first

### DIFF
--- a/tools/flaky-test-retryer/log_parser.go
+++ b/tools/flaky-test-retryer/log_parser.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"log"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"knative.dev/test-infra/shared/junit"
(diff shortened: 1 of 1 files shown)


chaodaiG:
Actually I just double thought about it, @TrevorFarrelly might have done this on purpose. If someone already started retest, then retryer should just be silent. @srinivashegde86 , what do you think?

srinivashegde86:
Yeah. I think we should wait on the job to finish and not start another one and only look at the latest results.

chaodaiG:
/close
since we agree that this is intended behavior. I will start another one with comment on this line, so the others also know why doing it this way

PR #1318, opened 2019-08-28
Make temp files creation work across OSes

antoineco:

The `file_XXXXX` pattern only works with files that do not have an extension on macOS, so temporal XML files are currently created with the literal and conflicting name `junit_XXXXXXXX.xml`.


antoineco:
/assign @srinivashegde86
(I hope the bot is not giving absurd advises 🙂)

PR #1320, opened 2019-08-28
fixing paths for config.yaml since it's moved

chaodaiG:
The config.yaml file was moved in #1313 , update the path so it still works

/cc @srinivashegde86


srinivashegde86:
/lgtm
thanks :)

PR #1322, opened 2019-08-29
A few bug fixes in flaky-test-reporter

chaodaiG:
Fixing a few bugs with recent updates in flaky-test-reporter, this has been tested in dryrun mode and it worked

/cc @srinivashegde86

### DIFF
--- a/tools/flaky-test-reporter/github_issue.go
+++ b/tools/flaky-test-reporter/github_issue.go
@@ -397,12 +397,16 @@ func (gih *GithubIssueHandler) getFlakyIssues() (map[string][]*flakyIssue, error
 			return nil, err
 		}
 		for _, issue := range issues {
-			flakyIssue, err := gih.githubToFlakyIssue(issue)
+			fi, err := gih.githubToFlakyIssue(issue)
 			if err != nil {
 				return nil, err
 			}
-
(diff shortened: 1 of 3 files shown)


tools/flaky-test-reporter/github_issue.go
```diff
@@ -551,10 +555,19 @@ func (gih *GithubIssueHandler) processGithubIssues(repoDataAll []RepoData, dryru
 
 	for _, rd := range repoDataAll {
 		issues, messages, err := gih.processGithubIssuesForRepo(rd, flakyGHIssuesMap, dryrun)
+		if _, ok := messagesMap[rd.Config.Repo]; !ok {
+			messagesMap[rd.Config.Repo] = make(map[string][]string)
```
srinivashegde86: is there an adv to using make? It's empty by default and will add it as the first item

chaodaiG: I see, thanks for the playground demo. actually this is commented on a wrong line, initializing map is still needed. I have removed the initialization of slices up and below

srinivashegde86: you dont need to check if a key exists to check the map init. We can use `len(messagesMap) == 0` or `messagesMap == nil` - https://play.golang.org/p/nNlgTAJan5r

PR #1324, opened 2019-08-31
Added monitoring for code coverage post-submit

steuhs:
`go test -coverprofile` returns a non-zero return code if any of the unit test fails.
While occasional such failure may be caused by flakiness, consistent failures in post-submit workflow of code coverage is likely caused by bad configuration or running environment. We want to be able to monitor on these failures with pre-set thresholds.
1. make post-submit job fail at the end, if the `go test` has a non-zero return code
2. Add the error pattern to monitoring so that alert will be sent if threshold is met

Fixes #

### DIFF
--- a/tools/coverage/artifacts/profiling.go
+++ b/tools/coverage/artifacts/profiling.go
@@ -16,6 +16,7 @@ limitations under the License.
 package artifacts
 
 import (
+	"fmt"
 	"io"
 	"log"
 	"os"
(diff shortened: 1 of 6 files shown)


tools/coverage/artifacts/profiling.go
```diff
@@ -34,7 +35,7 @@ func NewProfileReader(reader io.ReadCloser) *ProfileReader {
 
 // runProfiling writes coverage profile (&its stdout) by running go test on
 // target package
-func runProfiling(covTargets []string, localArts *LocalArtifacts) {
+func runProfiling(covTargets []string, localArts *LocalArtifacts) (err error) {
```
adrcunha: The error return is now important, that should be a test for that.

yt3liu:
You also need to add pubsub labels to all the post-submit code coverage jobs for monitoring to be aware of those jobs. (can be done in a separate PR) 

/cc @yt3liu

chaodaiG:
/uncc
add me back if needed

PR #1327, opened 2019-09-03
limit flaky report issue comment length

### LINKED ISSUE #1326: Flaky test reporter failed updating Github issue comment due to exceeding characters limit
Flaky test reporter keeps on appending current results on top of old results without trimming old results. After a while they would fail as a comment can only allow 65535 characters. See example here: https://prow.knative.dev/view/gcs/knative-prow/logs/ci-knative-flakes-reporter/1168856445676949508

chaodaiG:
Flaky test reporter started to fail updating github comments due to exceeding 65535 characters limit, fixing it by reformatting it as well as limiting numbers of builds to display
Fixes #1326 

/cc @srinivashegde86 
/cc @adrcunha 
/cc @Fredy-Z

### DIFF
--- a/tools/flaky-test-reporter/github_issue.go
+++ b/tools/flaky-test-reporter/github_issue.go
@@ -192,10 +192,12 @@ func (gih *GithubIssueHandler) createHistoryUnicode(rd RepoData, comment, testFu
 	sort.Slice(records, func(i, j int) bool {
 		return records[i] > records[j]
 	})
-	// Truncate to keep only 100 records, as there is a 65535 characters limit,
-	// see https://github.com/knative/test-infra/issues/1326
-	if len(records) > 100 {
-		records = records[:100]
+	// There is a 65535 characters limit for Github comment. As tested
+	// (https://github.com/chaodaiG/test-github-api/issues/129#issuecomment-527972076),
+	// one comment can at least contain 120 records, keep only 60 records to be
+	// safe. See https://github.com/knative/test-infra/issues/1326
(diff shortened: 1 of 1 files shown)


Fredy-Z:
Can you give an example of what the comment will be like? Will the format of comments like https://github.com/knative/serving/issues/5203#issuecomment-522539693 be changed?

chaodaiG:
Good point, an example is here: https://github.com/chaodaiG/test-github-api/issues/129#issuecomment-527652270

tools/flaky-test-reporter/github_issue.go
```diff
@@ -174,30 +179,35 @@ func (gih *GithubIssueHandler) createHistoryUnicode(rd RepoData, comment, testFu
...
+		uniqHistoryEntries.Insert(strings.ReplaceAll(hist[0], afterHistoryToken, ""))
+	}
+	records := append([]string{currentUnicode}, uniqHistoryEntries.List()...)
+	sort.Slice(records, func(i, j int) bool {
+		return records[i] > records[j]
+	})
+	// Truncate to keep only 100 records, as there is a 65535 characters limit,
+	// see https://github.com/knative/test-infra/issues/1326
+	if len(records) > 100 {
+		records = records[:100]
```
adrcunha: Are you positively sure each record is shorter than 655 characters, plus any header/footer for a 100-item entry? If so, better make it clear here.

chaodaiG: good point, updated

PR #1328, opened 2019-09-03
Remove code from test-infra and use code from pkg

Fredy-Z:
As part of the work to remove the loop dependencies between `pkg` and `test-infra`, remove `prometheus`, `ghutil` and `slackutil` from `test-infra/shared`, and use the code from `pkg/test`.

/cc @adrcunha 
/cc @chaodaiG 
/cc @srinivashegde86


adrcunha:
Looks like you still have issues with dependencies...

Fredy-Z:
> Looks like you still have issues with dependencies...

Yeah I'm still looking into it...

/hold

adrcunha:
This PR doesn't remove `ghutil` or `slackutil` as stated in the description. Is it intended?

Fredy-Z:
I have a PR that moved `ghutil` and `slackutil` to `pkg/test` - https://github.com/knative/pkg/pull/601. This PR removes the `ghutil` and `slackutil` from `test-infra`, and points the references to `pkg/test`.

The reason for the migration is `ghutil` and `slackutil` are used in the `Mako` library in `pkg/test` to alert for performance regression, and we don't want `pkg` to depend on `test-infra`. 

Does the above make sense?

Fredy-Z:
> This PR doesn't remove `ghutil` or `slackutil` as stated in the description. Is it intended?

I have a PR that moved `ghutil` and `slackutil` to `pkg/test` - https://github.com/knative/pkg/pull/601. This PR removes the `ghutil` and `slackutil` from `test-infra`, and points the references to `pkg/test`.

The reason for the migration is `ghutil` and `slackutil` are used in the `Mako` library in `pkg/test` to alert for performance regression, and we don't want `pkg` to depend on `test-infra`. 

Does the above make sense?

adrcunha:
> > This PR doesn't remove `ghutil` or `slackutil` as stated in the description. Is it intended?
> 
> I had a PR that moved `ghutil` and `slackutil` to `pkg/test` - [knative/pkg#601](https://github.com/knative/pkg/pull/601). This PR removes `ghutil` and `slackutil` from `test-infra`, and points the references to `pkg/test`.
> 
> The reason for the migration is `ghutil` and `slackutil` are used in the `Mako` library in `pkg/test` to alert for performance regression, and we don't want `pkg` to depend on `test-infra`.
> 
> Does the above make sense?

You misunderstood my comment. I was not asking for a justification of these changes (they do make sense), I just stated that the packages were not removed in this PR as you mentioned in the description. Anyway, I re-checked the (huge) list of changes and finally found them: GitHub report them as "moved to pkg" instead of deleted.

/lgtm
/approve

PR #1331, opened 2019-09-05
Fix latency and coverage Makefiles

adrcunha:
Build the tools in the container for compatibility sake.

Remove the support from building a binary in the shared Makefile, since this is now unused.

### DIFF
--- a/images/coverage/Dockerfile
+++ b/images/coverage/Dockerfile
@@ -32,4 +32,4 @@ RUN cp $TEMP_REPO_DIR/tools/$TOOL_NAME/$TOOL_NAME /$TOOL_NAME
 # Remove test-infra from the container
 RUN rm -fr $TEMP_REPO_DIR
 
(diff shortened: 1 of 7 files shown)


images/coverage/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+ADD . $TEMP_REPO_DIR
+
+# Build tool in the container
+RUN make -C $TEMP_REPO_DIR/tools/$TOOL_NAME/
+RUN cp $TEMP_REPO_DIR/tools/$TOOL_NAME/$TOOL_NAME /$TOOL_NAME
+
+# Remove test-infra from the container
+RUN rm -fr $TEMP_REPO_DIR
+
+ENTRYPOINT ["/metrics"]
```
chaodaiG:
```suggestion
ENTRYPOINT ["/coverage"]
```

PR #1335, opened 2019-09-09
Hide error when gcp isn't being used

duglin:
When running `hack/release.sh --skip-tests` I see this:
```
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
@@@@ Mon Sep  9 10:02:40 PDT 2019
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
hack/../vendor/knative.dev/test-infra/scripts/release.sh: line 488: gcloud: command not found
- gcloud user:
- Go path: /home/travis/gopath
- Repository root: /home/travis/gopath/src/knative.dev/eventing
- Destination GCR: ko.local
- Tests will NOT be run
- Artifacts WILL NOT be tagged
- Release will not be published
- Release will be BUILT FROM SOURCE
```
Notice the error after the banner.

This will hide that.

Signed-off-by: Doug Davis <dug@us.ibm.com>

/lint

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -485,7 +485,9 @@ function main() {
   parse_flags $@
   # Log what will be done and where.
   banner "Release configuration"
-  echo "- gcloud user: $(gcloud config get-value core/account)"
+  if which gcloud &>/dev/null ; then
+    echo "- gcloud user: $(gcloud config get-value core/account)"
+  fi
   echo "- Go path: ${GOPATH}"
   echo "- Repository root: ${REPO_ROOT_DIR}"
   echo "- Destination GCR: ${KO_DOCKER_REPO}"


scripts/release.sh
```diff
@@ -485,7 +485,6 @@ function main() {
   parse_flags $@
   # Log what will be done and where.
   banner "Release configuration"
-  echo "- gcloud user: $(gcloud config get-value core/account)"
```
adrcunha: Please, don't remove it, as this is useful for debugging. I'm totally OK with dealing with gcloud not being installed though.

duglin:
Not thrilled with cloud specific stuff in there but if you insist on keeping it then I added a place for other people to add their logic too.

adrcunha:
> Not thrilled with cloud specific stuff in there but if you insist on keeping it then I added a place for other people to add their logic too.

This script is mainly intended to run on Prow, which uses GCP and gcloud, thus that debugging is important.

scripts/release.sh
adrcunha: What's the use of ibmcloud user here? For gcloud, it's a useful debugging piece because of the Prow job user. AFAICU ibmcloud isn't used anywhere.

duglin: I have no idea how gcloud is used in these so I just was guessing at what's going on.  Is the prow user not listed any place else in the output? Like in the prow debug stuff?

duglin:
> > Not thrilled with cloud specific stuff in there but if you insist on keeping it then I added a place for other people to add their logic too.
> 
> This script is mainly intended to run on Prow, which uses GCP and gcloud, thus that debugging is important.

People need to be able to run the scripts that build the release artifacts from their own environments.

scripts/release.sh
adrcunha: gcloud is used when publishing releases. Prow has no debug stuff for jobs, it's up to the job doing that.

adrcunha:
> People need to be able to run the scripts that build the release artifacts from their own environments.

Sure. That's why this PR is welcome. The only thing I don't want is to loose debug coverage when this runs as a Prow job.

scripts/release.sh
```diff
@@ -485,7 +485,9 @@ function main() {
...
+  if which gcloud &>/dev/null ; then
+    echo "- Cloud user: $(gcloud config get-value core/account)"
```
adrcunha:
```suggestion
    echo "- gcloud user: $(gcloud config get-value core/account)"
```

adrcunha:
/approve
/lgtm

Thanks!

PR #1338, opened 2019-09-09
Using updated slackutil in knative/pkg

Fredy-Z:
- Use updated slackutil in https://github.com/knative/pkg/pull/645.
- Use `helpers.Run` and `helpers.CombineErrors` in https://github.com/knative/pkg/pull/645.

### DIFF
--- a/tools/flaky-test-reporter/json_reporter.go
+++ b/tools/flaky-test-reporter/json_reporter.go
@@ -21,6 +21,7 @@ import (
 	"log"
 	"sync"
 
+	"knative.dev/pkg/test/helpers"
 	"knative.dev/test-infra/tools/flaky-test-reporter/jsonreport"
 )
 
@@ -55,7 +56,7 @@ func writeFlakyTestsToJSON(repoDataAll []RepoData, dryrun bool) error {
 			for test := range testSet {
 				testList = append(testList, test)
 			}
-			if err := run(
+			if err := helpers.Run(
 				fmt.Sprintf("writing JSON report for repo '%s'", repo),
 				func() error {
 					_, err := client.CreateReport(repo, testList, true)
@@ -74,5 +75,5 @@ func writeFlakyTestsToJSON(repoDataAll []RepoData, dryrun bool) error {
 	}
 	wg.Wait()
 	close(ch)
-	return combineErrors(allErrs)
+	return helpers.CombineErrors(allErrs)
 }
--- a/tools/flaky-test-reporter/slack_notification.go
+++ b/tools/flaky-test-reporter/slack_notification.go
@@ -24,6 +24,7 @@ import (
 	"sync"
 	"time"
 
+	"knative.dev/pkg/test/helpers"
 	"knative.dev/pkg/test/slackutil"
 	"knative.dev/test-infra/shared/testgrid"
 )
@@ -87,7 +88,7 @@ func sendSlackNotifications(repoDataAll []RepoData, c slackutil.WriteOperations,
 			go func() {
 				defer wg.Done()
 				message := createSlackMessageForRepo(rd, flakyIssues)
-				if err := run(
+				if err := helpers.Run(
 					fmt.Sprintf("post Slack message for job '%s' from repo '%s' in channel '%s'", rd.Config.Name, rd.Config.Repo, channel.Name),
 					func() error {
 						return c.Post(message, channel.Identity)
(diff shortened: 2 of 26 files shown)


chaodaiG:
It surprised me that the build-test passed when there are changes under vendor directly, seems like a bug

srinivashegde86:
Use the helpers for `https://github.com/knative/test-infra/blob/master/tools/flaky-test-reporter/error.go and https://github.com/knative/test-infra/blob/master/tools/flaky-test-reporter/run.go` as well.

Fredy-Z:
> It surprised me that the build-test passed when there are changes under vendor directly, seems like a bug

Yeah we run `./hack/update-deps.sh` on it, but we don't `diff` the result like what's done in `/hack/verify-codegen.sh`.

Fredy-Z:
> Use the helpers for `https://github.com/knative/test-infra/blob/master/tools/flaky-test-reporter/error.go and https://github.com/knative/test-infra/blob/master/tools/flaky-test-reporter/run.go` as well.

Thanks for the reminder! Will do it.

Fredy-Z:
> It surprised me that the build-test passed when there are changes under vendor directly, seems like a bug

Yeah it's not preventing such change, since the `default_build_test_runner` does not run `./hack/update-deps.sh`.

tools/flaky-test-reporter/slack_notification.go
```diff
@@ -84,7 +84,8 @@ func sendSlackNotifications(repoDataAll []RepoData, c slackutil.WriteOperations,
 		for i := range channels {
 			wg.Add(1)
 			channel := channels[i]
-			go func(wg *sync.WaitGroup) {
+			go func() {
```
chaodaiG: are you sure this works without passing wg?

chaodaiG:
> > It surprised me that the build-test passed when there are changes under vendor directly, seems like a bug
> 
> Yeah it's not preventing such change, since the `default_build_test_runner` does not run `./hack/update-deps.sh`.

it's because we don't have `verify-codegen.sh` in test-infra

tools/flaky-test-reporter/slack_notification.go
Fredy-Z: Yes, one example here - https://golang.org/src/sync/example_test.go

Fredy-Z:
> > > It surprised me that the build-test passed when there are changes under vendor directly, seems like a bug
> > 
> > 
> > Yeah it's not preventing such change, since the `default_build_test_runner` does not run `./hack/update-deps.sh`.
> 
> it's because we don't have `verify-codegen.sh` in test-infra

`verify-codegen.sh` pairs with `update-codegen.sh`, we don't have any code to generate, so..

chaodaiG:
> > > > It surprised me that the build-test passed when there are changes under vendor directly, seems like a bug
> > > 
> > > 
> > > Yeah it's not preventing such change, since the `default_build_test_runner` does not run `./hack/update-deps.sh`.
> > 
> > 
> > it's because we don't have `verify-codegen.sh` in test-infra
> 
> `verify-codegen.sh` pairs with `update-codegen.sh`, we don't have any code to generate, so..

not entirely true, at least we have this to generate: https://github.com/knative/serving/blob/24bc31214c6140a9d03570974b414082d004fa9b/hack/update-codegen.sh#L83

Fredy-Z:
> > > > > It surprised me that the build-test passed when there are changes under vendor directly, seems like a bug
> > > > 
> > > > 
> > > > Yeah it's not preventing such change, since the `default_build_test_runner` does not run `./hack/update-deps.sh`.
> > > 
> > > 
> > > it's because we don't have `verify-codegen.sh` in test-infra
> > 
> > 
> > `verify-codegen.sh` pairs with `update-codegen.sh`, we don't have any code to generate, so..
> 
> not entirely true, at least we have this to generate: https://github.com/knative/serving/blob/24bc31214c6140a9d03570974b414082d004fa9b/hack/update-codegen.sh#L83

Ah yeah, that's true, didn't notice that..

tools/flaky-test-reporter/json_reporter.go
```diff
@@ -55,7 +56,7 @@ func writeFlakyTestsToJSON(repoDataAll []RepoData, dryrun bool) error {
 			for test := range testSet {
 				testList = append(testList, test)
 			}
-			if err := run(
+			if err := helpers.Run(
```
srinivashegde86: Since you are using the helpers, can we delete these two files from here?

Fredy-Z: I did delete them. Github diff "intelligently" considered them as renames :-(

srinivashegde86: oh. okies

PR #1339, opened 2019-09-10
Acquire cluster admin role for only GKE

Fredy-Z:
According to the documents under https://knative.dev/docs/install/, cluster admin role is only required for GKE, so in `e2e-tests.sh`, only acquire cluster admin role for GKE clusters.

This can potentially make `e2e-tests.sh --run-tests` works with other Kubernetes vendors as long as the clusters and other environments are properly set up.

/cc @adrcunha 
/cc @chaodaiG

### DIFF
--- a/scripts/e2e-tests.sh
+++ b/scripts/e2e-tests.sh
@@ -326,7 +326,6 @@ function setup_test_cluster() {
   # Setup the admin role and also KO_DOCKER_REPO if it is a GKE cluster
   if [[ -z "$(kubectl get clusterrolebinding cluster-admin-binding 2> /dev/null)" && "${k8s_cluster}" =~ ^gke_.* ]]; then
     acquire_cluster_admin_role ${k8s_user} ${E2E_CLUSTER_NAME} ${E2E_CLUSTER_REGION} ${E2E_CLUSTER_ZONE}
-    kubectl config set-context ${k8s_cluster} --namespace=default
     # Incorporate an element of randomness to ensure that each run properly publishes images.
     export KO_DOCKER_REPO=gcr.io/${E2E_PROJECT_ID}/${E2E_BASE_NAME}-e2e-img/${RANDOM}
   fi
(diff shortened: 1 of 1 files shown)


scripts/e2e-tests.sh
```diff
@@ -323,8 +323,8 @@ function setup_test_cluster() {
     abort "kubeconfig context set to ${k8s_cluster}, which is forbidden"
 
   # If cluster admin role isn't set, this is a brand new cluster
-  # Setup the admin role and also KO_DOCKER_REPO
-  if [[ -z "$(kubectl get clusterrolebinding cluster-admin-binding 2> /dev/null)" ]]; then
+  # Setup the admin role and also KO_DOCKER_REPO if it is a GKE cluster
+  if [[ -z "$(kubectl get clusterrolebinding cluster-admin-binding 2> /dev/null)" && "${k8s_cluster}" =~ ^gke_.* ]]; then
```
chaodaiG: I think we should also move definition of `k8s_user` down here too, it just prints irrelevant info if not gke

adrcunha: I prefer making it explicit that this is gcloud user, as this PR doesn't remove the dependency on gcloud. In line 339, use "gcloud user is" instead of "User is".

chaodaiG: Works for me

scripts/e2e-tests.sh
```diff
@@ -323,8 +323,8 @@ function setup_test_cluster() {
...
     acquire_cluster_admin_role ${k8s_user} ${E2E_CLUSTER_NAME} ${E2E_CLUSTER_REGION} ${E2E_CLUSTER_ZONE}
     kubectl config set-context ${k8s_cluster} --namespace=default
```
chaodaiG: would this be needed for other providers?

adrcunha: Makes sense. And KO_DOCKER_REPO should be set outside this `if` as well.

chaodaiG: yeah, but then we'll need to discuss what the docker registry should be since user might not have gcloud access

adrcunha: On a second though, nevermind moving KO_DOCKER_REPO. Already existing GCP environments and non-GCP environment will have to set it anyway.

Fredy-Z: I'm even not sure if `kubectl config set-context ${k8s_cluster} --namespace=default` is needed for `gcloud`. In the previous logic, this line will only be executed on a brand new cluster, which must already have namespace set to `default`.
But I think it makes sense to use the `default` namespace  as the default one for the current context in e2e tests, so I moved it outside the `if`.

PR #1348, opened 2019-09-13
Convert all 'nil =='style to '== nil' style

chaodaiG:
Currently there is a mixture of `nil == err` and `err == nil` styles, and the former one was mainly introduced by me, the primary purpose was to avoid accidental variable assignment like
```
if err = nil {
}
```
Which is always true in some other languages. This is not the case in go as it would complain `syntax error: assignment err = nil used as value`. This change convert all `nil == err` usages to `err == nil` to ensure consistency

### DIFF
--- a/shared/gcs/gcs.go
+++ b/shared/gcs/gcs.go
@@ -48,13 +48,13 @@ func Authenticate(ctx context.Context, serviceAccount string) error {
 func Exists(ctx context.Context, bucketName, storagePath string) bool {
 	// Check if this is a file
 	handle := createStorageObject(bucketName, storagePath)
-	if _, err := handle.Attrs(ctx); nil == err {
+	if _, err := handle.Attrs(ctx); err == nil {
 		return true
 	}
 	// Check if this is a directory,
 	// gcs directory paths are virtual paths, they automatically got deleted if there is no child file
 	_, err := getObjectsIter(ctx, bucketName, strings.TrimRight(storagePath, " /")+"/", "").Next()
-	return nil == err
+	return err == nil
 }
 
 // ListChildrenFiles recursively lists all children files.
@@ -82,7 +82,7 @@ func Copy(ctx context.Context, srcBucketName, srcPath, dstBucketName, dstPath st
 // Download file from gcs
 func Download(ctx context.Context, bucketName, srcPath, dstPath string) error {
 	handle := createStorageObject(bucketName, srcPath)
-	if _, err := handle.Attrs(ctx); nil != err {
+	if _, err := handle.Attrs(ctx); err != nil {
 		return err
 	}
 
@@ -95,24 +95,20 @@ func Download(ctx context.Context, bucketName, srcPath, dstPath string) error {
 		return err
 	}
 	defer src.Close()
-	if _, err = io.Copy(dst, src); nil != err {
-		return err
-	}
-	return nil
+	_, err = io.Copy(dst, src)
+	return err
 }
 
 // Upload file to gcs
 func Upload(ctx context.Context, bucketName, dstPath, srcPath string) error {
 	src, err := os.Open(srcPath)
-	if nil != err {
+	if err != nil {
 		return err
 	}
 	dst := createStorageObject(bucketName, dstPath).NewWriter(ctx)
-	if _, err = io.Copy(dst, src); nil != err {
-		return err
-	}
-	dst.Close()
-	return nil
+	defer dst.Close()
+	_, err = io.Copy(dst, src)
+	return err
 }
 
 // Read reads the specified file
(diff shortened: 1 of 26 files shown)


shared/gcs/gcs.go
```diff
@@ -48,13 +48,13 @@ func Authenticate(ctx context.Context, serviceAccount string) error {
 func Exists(ctx context.Context, bucketName, storagePath string) bool {
 	// Check if this is a file
 	handle := createStorageObject(bucketName, storagePath)
-	if _, err := handle.Attrs(ctx); nil == err {
+	if _, err := handle.Attrs(ctx); err == nil {
```
srinivashegde86: these can be combined
```
if _, err := handle.Attrs(); err != nil {
  _, err = getObjectsIter().Next()
  return err == nil
}
return true
```

shared/gcs/gcs.go
```diff
@@ -95,7 +95,7 @@ func Download(ctx context.Context, bucketName, srcPath, dstPath string) error {
 		return err
 	}
 	defer src.Close()
-	if _, err = io.Copy(dst, src); nil != err {
+	if _, err = io.Copy(dst, src); err != nil {
```
srinivashegde86: no need of if
```
_, err = io.Copy()
return err
```

shared/gcs/gcs.go
```diff
@@ -104,11 +104,11 @@ func Download(ctx context.Context, bucketName, srcPath, dstPath string) error {
 // Upload file to gcs
 func Upload(ctx context.Context, bucketName, dstPath, srcPath string) error {
 	src, err := os.Open(srcPath)
-	if nil != err {
+	if err != nil {
 		return err
 	}
 	dst := createStorageObject(bucketName, dstPath).NewWriter(ctx)
-	if _, err = io.Copy(dst, src); nil != err {
+	if _, err = io.Copy(dst, src); err != nil {
```
srinivashegde86: ditto. No need of if
```
defer dst.Close()
_, err = io.Copy()
return err
```

shared/prow/prow.go
```diff
@@ -195,7 +195,7 @@ func (j *Job) GetFinishedBuilds() []Build {
 	var finishedBuilds []Build
 	builds := j.GetBuilds()
 	for _, build := range builds {
-		if true == build.IsFinished() {
+		if build.IsFinished() == true {
```
srinivashegde86:
```suggestion
		if build.IsFinished() {
```

shared/prow/prow.go
```diff
@@ -328,10 +328,10 @@ func getBuildIDFromBuildPath(buildPath string) (int, error) {
 // v must be an arbitrary struct, slice, or string.
 func unmarshalJSONFile(storagePath string, v interface{}) error {
 	contents, err := gcs.Read(ctx, BucketName, storagePath)
-	if nil != err {
+	if err != nil {
 		return err
 	}
-	if err := json.Unmarshal(contents, v); nil != err {
+	if err := json.Unmarshal(contents, v); err != nil {
```
srinivashegde86: ditto. no need of if
```
return json.Unmarshal()
```

shared/gcs/gcs.go
chaodaiG: I would prefer current logic: return true if it's a file, else check if it's a folder.

shared/gcs/gcs.go
chaodaiG: agree. Done

tools/flaky-test-reporter/result.go
```diff
@@ -219,9 +219,9 @@ func (rd *RepoData) getResultSliceForTest(testName string) []junit.TestStatusEnu
 	ts := rd.TestStats[testName]
 	for i, buildID := range rd.BuildIDs {
 		switch {
-		case true == intSliceContains(ts.Failed, buildID):
+		case intSliceContains(ts.Failed, buildID) == true:
```
srinivashegde86:
```suggestion
		case intSliceContains(ts.Failed, buildID):
```

tools/flaky-test-reporter/result.go
```diff
@@ -219,9 +219,9 @@ func (rd *RepoData) getResultSliceForTest(testName string) []junit.TestStatusEnu
...
 			res[i] = junit.Failed
-		case true == intSliceContains(ts.Passed, buildID):
+		case intSliceContains(ts.Passed, buildID) == true:
```
srinivashegde86:
```suggestion
		case intSliceContains(ts.Passed, buildID):
```

tools/prow-auto-bumper/main.go
```diff
@@ -34,12 +34,12 @@ func main() {
 	dryrun := flag.Bool("dry-run", false, "dry run switch")
 	flag.Parse()
 
-	if nil != dryrun && true == *dryrun {
+	if dryrun != nil && *dryrun == true {
```
srinivashegde86:
```suggestion
	if dryrun != nil && *dryrun {
```

shared/gcs/gcs.go
chaodaiG: done

shared/prow/prow.go
chaodaiG: done

PR #1350, opened 2019-09-16
Make test-infra more mac friendly.

n3wscott:
/lint

Do not assume `/bin/bash`. For these scripts to work on a mac, the script needs to ask the env for the version of bash that is set as the default for the system, please use `#!/usr/bin/env bash` always. This lets the mac bash use bash4

### DIFF
--- a/scripts/library.sh
+++ b/scripts/library.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # Copyright 2018 The Knative Authors
 #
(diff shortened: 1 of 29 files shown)


scripts/library.sh
```diff
@@ -283,9 +283,9 @@ function acquire_cluster_admin_role() {
     local key=$(mktemp)
     echo "Certificate in ${cert}, key in ${key}"
     gcloud --format="value(masterAuth.clientCertificate)" \
-      container clusters describe $2 ${geoflag} | base64 -d > ${cert}
```
n3wscott: This will fix https://github.com/knative/eventing-contrib/issues/595

srinivashegde86:
/lgtm

Changes LGTM, will wait on Chao or Adriano to approve.

PR #1355, opened 2019-09-17
Remove prow jobs for Knative build

Fredy-Z:
Knative `build` has been deprecated, we should be able to remove the jobs for it.

/cc @adrcunha 
/cc @chaodaiG


adrcunha:
RiP Knative Build

/cc @ImJasonH 
/lgtm
/approve

PR #1360, opened 2019-09-18
Update flaky-test-retryer to use latest image

chaodaiG:
/cc @srinivashegde86 
/cc @Fredy-Z 

/hold
The image was already pushed from master, will unhold once it's verified to work


chaodaiG:
/hold cancel
I believe it's working properly now

PR #1361, opened 2019-09-18
testgrid: Adding API Coverage tab for knative-serving testgrid dashboard

dushyanthsc:
Adding API Coverage tab for knative-serving testgrid dashboard


adrcunha:
Please do not update the raw configs directly, it won't work. Changes must be either done in the templates, in the config generator, or in the config_knative.yaml file, as usual.

Fredy-Z:
Done in https://github.com/knative/test-infra/pull/1381
/close

PR #1377, opened 2019-09-20
Sort release versions/tags properly so we can release 0.10 and beyond

chaodaiG:
current implementation would sort "0.2, 0.1, 0.10" as "0.1, 0.10, 0.2", which would fail release if 0.10 is cut. Fix it with proper sorting

/cc @adrcunha 
/cc @Fredy-Z


adrcunha:
Good catch. :)

/lgtm

PR #1390, opened 2019-09-24
Dump all Kubernetes resources in dedicated file

JRBANCEL:
Follow up on #1352:
* `kubectl get all` doesn't dump the CRDs and doesn't even dump all Kubernetes resources, so it is replaced by a manual loop dumping everything in a clean fashion (high level listing first and then full dump)
* Removed `kubectl get ...` calls that are now redundant
* Dump all of this in a dedicated file to keep `build.log` clean instead of immensely bloated as it is today

### DIFF
--- a/scripts/e2e-tests.sh
+++ b/scripts/e2e-tests.sh
@@ -93,15 +93,32 @@ function dump_cluster_state() {
   echo "***         E2E TEST FAILED         ***"
   echo "***    Start of information dump    ***"
   echo "***************************************"
-  echo ">>> All resources:"
-  kubectl get all --all-namespaces --show-labels -o wide
-  echo ">>> Services:"
-  kubectl get services --all-namespaces --show-labels -o wide
-  echo ">>> Events:"
-  kubectl get events --all-namespaces --show-labels -o wide
-  function_exists dump_extra_cluster_state && dump_extra_cluster_state
-  echo ">>> Full dump: ${ARTIFACTS}/k8s.dump.txt"
-  kubectl get all --all-namespaces -o yaml > ${ARTIFACTS}/k8s.dump.txt
+
+  local output="${ARTIFACTS}/k8s.dump.txt"
+  echo ">>> The dump is located at ${output}"
+
+  for crd in $(kubectl api-resources --verbs=list -o name | sort); do
+    local count="$(kubectl get $crd --all-namespaces --no-headers 2>/dev/null | wc -l)"
+    echo ">>> ${crd} (${count} objects)"
+    if [[ "${count}" > "0" ]]; then
+      echo ">>> ${crd} (${count} objects)" >> ${output}
+
+      echo ">>> Listing" >> ${output}
+      kubectl get ${crd} --all-namespaces >> ${output}
+
+      echo ">>> Details" >> ${output}
+      if [[ "${crd}" == "secrets" ]]; then
+        echo "Secrets are ignored for security reasons" >> ${output}
+      else
+        kubectl get ${crd} --all-namespaces -o yaml >> ${output}
+      fi
+    fi
+  done
+
+  if function_exists dump_extra_cluster_state; then
+    echo ">>> Extra dump" >> ${output}
+    dump_extra_cluster_state >> ${output}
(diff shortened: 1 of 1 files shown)


scripts/e2e-tests.sh
```diff
@@ -93,15 +93,30 @@ function dump_cluster_state() {
...
-  echo ">>> Events:"
-  kubectl get events --all-namespaces --show-labels -o wide
-  function_exists dump_extra_cluster_state && dump_extra_cluster_state
-  echo ">>> Full dump: ${ARTIFACTS}/k8s.dump.txt"
-  kubectl get all --all-namespaces -o yaml > ${ARTIFACTS}/k8s.dump.txt
+
+  local output="${ARTIFACTS}/tmp/k8s.dump.txt"
+  echo ">>> The dump is located at ${output}"
+
+  for crd in `kubectl api-resources --verbs=list -o name | sort`; do
```
adrcunha: $() is preferred over `` for readability sake.

```suggestion
  for crd in $(kubectl api-resources --verbs=list -o name | sort); do
```

scripts/e2e-tests.sh
```diff
@@ -93,15 +93,30 @@ function dump_cluster_state() {
...
+    count=$(kubectl get $crd --all-namespaces --no-headers 2>/dev/null | wc -l)
```
```suggestion
    local count="$(kubectl get $crd --all-namespaces --no-headers 2>/dev/null | wc -l)"
```

scripts/e2e-tests.sh
```diff
@@ -93,15 +93,30 @@ function dump_cluster_state() {
...
-  kubectl get all --all-namespaces --show-labels -o wide
-  echo ">>> Services:"
-  kubectl get services --all-namespaces --show-labels -o wide
-  echo ">>> Events:"
-  kubectl get events --all-namespaces --show-labels -o wide
-  function_exists dump_extra_cluster_state && dump_extra_cluster_state
-  echo ">>> Full dump: ${ARTIFACTS}/k8s.dump.txt"
-  kubectl get all --all-namespaces -o yaml > ${ARTIFACTS}/k8s.dump.txt
+
+  local output="${ARTIFACTS}/tmp/k8s.dump.txt"
```
chaodaiG: I love this idea of not bloating e2e logs!
Don't you need to create the `${ARTIFACTS}/tmp` dir before dumping logs?

JRBANCEL: /tmp is a bug, is is an artifact from my local debugging.
Good catch, will fix.

adrcunha: Maybe applying this suggestion is easier? :)

```suggestion
  local output="${ARTIFACTS}/k8s.dump.txt"
```

scripts/e2e-tests.sh
```diff
@@ -93,15 +93,30 @@ function dump_cluster_state() {
...
+      if [[ "${crd}" == "secrets" ]]; then
+        echo "Secrets are ignored for security reasons" >> ${output}
+      else
+        kubectl get ${crd} --all-namespaces -o yaml >> ${output}
+      fi
+    fi
+  done
+
+  echo ">>> Extra dump" >> ${output}
+  function_exists dump_extra_cluster_state && dump_extra_cluster_state >> ${output}
```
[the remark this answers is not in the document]
JRBANCEL: Makes sense. Done.

scripts/e2e-tests.sh
```diff
@@ -93,15 +93,29 @@ function dump_cluster_state() {
...
+      echo ">>> Details" >> ${output}
+      if [[ "${crd}" == "secrets" ]]; then
+        echo "Secrets are ignored for security reasons" >> ${output}
+      else
+        kubectl get ${crd} --all-namespaces -o yaml >> ${output}
+      fi
+    fi
+  done
+
+  function_exists dump_extra_cluster_state && echo ">>> Extra dump" >> ${output} && dump_extra_cluster_state >> ${output}
```
adrcunha: Please use `if` here for readability.

adrcunha:
Thanks!

/approve
/lgtm

PR #1401, opened 2019-09-25
Make Prow config more easily configurable from CLI

adrcunha:


ci/prow/Makefile
```diff
@@ -12,14 +12,26 @@
...
-
-PROW_DIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
-TESTGRID_DIR := $(PROW_DIR)/../testgrid
-SET_CONTEXT := gcloud container clusters get-credentials "$(CLUSTER)" --project="$(PROJECT)" --zone="$(ZONE)"
+CLUSTER            ?= prow
+PROJECT            ?= knative-tests
+ZONE               ?= us-central1-f
+JOB_NAMESPACE      ?= test-pods
+KNATIVE_CONFIG     ?= config_knative.yaml
+SKIP_CONFIG_BACKUP ?=
```
chaodaiG: what's the use case of skipping config backup? can't we pass in `PROW_CONFIG_GCS` if the concern is to avoid collision?

adrcunha: I just want to have as many knobs as possible. Sample use case: a staging/temp cluster where I don't care about preserving the config.

chaodaiG: sgtm

PR #1406, opened 2019-09-26
Make knative-dev a viewer of Boskos projects

adrcunha:
* Also refactor `set_permissions.sh` to make updates simpler.

### DIFF
--- a/ci/prow/boskos/set_permissions.sh
+++ b/ci/prow/boskos/set_permissions.sh
@@ -47,7 +47,7 @@ readonly RESOURCES=(
 
 # Loop through the list of resources and add them.
 
-# Hold the current role used to add accounts.
+# Start with a non-existing role, so gcloud clearly fails if resources are set incorrectly.
 role="unknown"
 for res in ${RESOURCES[@]}; do
   if [[ ${res} == roles/* ]]; then
--- a/ci/prow/boskos/set_permissions.sh
+++ b/ci/prow/boskos/set_permissions.sh
@@ -17,38 +17,44 @@
 set -e
 
 readonly PROJECT=${1:?"First argument must be the boskos project name."}
-readonly PROJECT_OWNERS=("prime-engprod-sea@google.com")
-readonly PROJECT_GROUPS=("knative-productivity-admins@googlegroups.com")
-readonly PROJECT_SAS=(
+
+# APIs, Permissions and accounts to be set.
+readonly RESOURCES=(
+    "roles/owner"
+    "prime-engprod-sea@google.com"
+
+    "roles/editor"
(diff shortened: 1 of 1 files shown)


ci/prow/boskos/set_permissions.sh
```diff
@@ -17,38 +17,44 @@
...
-  echo "NOTE: Adding service account ${sa}"
-  gcloud projects add-iam-policy-binding ${PROJECT} --member serviceAccount:${sa} --role roles/editor
-done
+# Loop through the list of resources and add them.
 
-# Enable APIS
-for api in ${PROJECT_APIS[@]}; do
-  echo "NOTE: Enabling API ${api}"
-  gcloud services enable ${api} --project=${PROJECT}
+role="unknown"
```
chaodaiG: can we have a comment for this?

ci/prow/boskos/set_permissions.sh
```diff
@@ -17,38 +17,44 @@
...
+for res in ${RESOURCES[@]}; do
+  if [[ ${res} == roles/* ]]; then
+    role=${res}
+    continue
+  fi
+  if [[ ${res} == *.googleapis.com ]]; then
```
chaodaiG: can we use `else if` here, then `else` after this part?

adrcunha: Why?

ci/prow/boskos/set_permissions.sh
adrcunha: For the `role` var or for the loop?

ci/prow/boskos/set_permissions.sh
chaodaiG: so there is no need to continue twice

ci/prow/boskos/set_permissions.sh
chaodaiG: the role var

ci/prow/boskos/set_permissions.sh
adrcunha: You're sacrificing readability of the flow for just a repeated `continue`?

chaodaiG: I should have marked it as a nit, let's ignore it

ci/prow/boskos/set_permissions.sh
chaodaiG: sorry I should have made it clearer. The reason I asked for a comment was because I'm not sure what will happen if `role` is not initialized as `unknown`. My guess is something bad could happen if role is empty when running `gcloud` command

adrcunha: It would fail with "missing role" instead of "unknown role bla". Updated the comment.

PR #1410, opened 2019-09-27
Parameterize Prow host and TestGrid bucket

adrcunha:
Allow these to be set through command-line flags in `make_config.go`.

### DIFF
--- a/ci/prow/testgrid_config.go
+++ b/ci/prow/testgrid_config.go
@@ -67,8 +67,12 @@ var (
 // baseTestgridTemplateData contains basic data about the testgrid config file.
 // TODO(Fredy-Z): remove this structure and use baseProwJobTemplateData instead
 type baseTestgridTemplateData struct {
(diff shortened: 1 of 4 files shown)


ci/prow/testgrid_config.go
```diff
@@ -67,8 +67,12 @@ var (
 // baseTestgridTemplateData contains basic data about the testgrid config file.
 // TODO(Fredy-Z): remove this structure and use baseProwJobTemplateData instead
 type baseTestgridTemplateData struct {
-	TestGroupName string
-	Year          int
+	ProwHost          string
+	TestGridHost      string
+	GubernatorHost    string
```
chaodaiG: Seems like testgrid only uses `ProwHost`, are the other 3 newly added vars included just for consistency?

adrcunha: Yes.

PR #1417, opened 2019-09-30
Fix creating boskos projects from scratch

adrcunha:


ci/prow/boskos/create_projects.sh
```diff
@@ -28,8 +28,9 @@ fi
 
 # Get the index of the last boskos project from the resources file
 LAST_INDEX=$(grep "knative-boskos-" ${RESOURCE_FILE} | grep -o "[0-9]\+" | sort -nr | head -1)
+[[ -z "${LAST_INDEX}" ]] && LAST_INDEX=0
```
chizhg: If we assume there can be no existing boskos projects when we run this script, the `sed` command below will fail since it won't be able to find the place to append.

adrcunha: Good catch, thanks. Fixed.

chizhg:
/lgtm
Thanks!

PR #1418, opened 2019-09-30
Fail script if application default credentials are not set

adrcunha:


ci/prow/boskos/set_permissions.sh
```diff
@@ -78,7 +83,6 @@ done
 echo "Activating GCS service account"
 curl -s -X GET -H "Authorization: Bearer ${ACCESS_TOKEN}" "https://www.googleapis.com/storage/v1/projects/${PROJECT}/serviceAccount"
 GCS_SERVICE_ACCOUNT="service-${PROJECT_NUMBER}@gs-project-accounts.iam.gserviceaccount.com"
-echo "GCS service account is ${GCS_SERVICE_ACCOUNT}"
```
adrcunha: You're probably wondering why I removed this line: the account is already displayed by the output of `curl` above, so this is just a redundant log.

ci/prow/boskos/set_permissions.sh
```diff
@@ -78,7 +83,6 @@ done
 echo "Activating GCS service account"
 curl -s -X GET -H "Authorization: Bearer ${ACCESS_TOKEN}" "https://www.googleapis.com/storage/v1/projects/${PROJECT}/serviceAccount"
 GCS_SERVICE_ACCOUNT="service-${PROJECT_NUMBER}@gs-project-accounts.iam.gserviceaccount.com"
```
chizhg: Then also remove this line?

adrcunha:
```suggestion
```

adrcunha: I messed up my branch, this line was added back and I missed it. :P 
Removed, thanks!

ci/prow/boskos/set_permissions.sh
chizhg: /lgtm

PR #1419, opened 2019-09-30
Parameterize the Prow log bucket when creating the config

adrcunha:
Bonus: have a harmless default rule to avoid unintended changes when simply running `make`.


chaodaiG:
The more parameterization I see, the more I'm feeling that we should have a centralized location for storing these constants, like `PROW_PROJECT`, `PROW_CLUSTER`, `PROW_GCS`, `TESTGRID_GCS`, etc. these configs can be protected by checking if git is currently checked out directly from knative/test-infra. And there can be another staging config which is consumed by forks

chaodaiG:
/lgtm
Since the previous comment shouldn't block this PR

PR #1420, opened 2019-09-30
Parameterize the registry of docker images used in Prow config

adrcunha:


adrcunha:
/hold

This doesn't work this way.

ci/prow/make_config.go
```diff
@@ -1309,6 +1313,14 @@ func main() {
 	if len(flag.Args()) != 1 {
 		log.Fatal("Pass the config file as parameter")
 	}
+
+	flakesreporterDockerImage = *dockerImagesBase + "/" + *flakesreporterDockerImageName
```
chaodaiG: nit, use path.Join instead of string concatenation since `*dockerImagesBase` could end with `/`

chaodaiG:
Left one nit, will lgtm once merge conflict is resolved

chaodaiG:
/lgtm
/hold
feel free unhold if you decide to ignore my nit

ci/prow/make_config.go
adrcunha: Done.

PR #1426, opened 2019-10-02
Add missing roles to boskos initial config

adrcunha:
A new instance, if created from scratch, won't start since these roles will be missing.

### DIFF
--- a/ci/prow/boskos/config_start.yaml
+++ b/ci/prow/boskos/config_start.yaml
@@ -21,34 +21,3 @@ metadata:
   namespace: test-pods
 data:
   resources: ""
----
-kind: ClusterRole
-apiVersion: rbac.authorization.k8s.io/v1
-metadata:
-  name: boskos
-rules:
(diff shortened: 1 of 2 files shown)


chizhg:
Can you add some explanation for this PR? :-)

adrcunha:
Done.

ci/prow/boskos/config_start.yaml
```diff
@@ -21,3 +21,34 @@ metadata:
   namespace: test-pods
 data:
   resources: ""
+---
+kind: ClusterRole
```
yt3liu: It looks like there's similar role binding in the general config_start.yaml. Do those need to be removed? 
https://github.com/knative/test-infra/blob/master/ci/prow/config_start.yaml#L56

adrcunha: Good catch, thanks. Updated the right section in the right file.

PR #1428, opened 2019-10-02
Move CI/CD Makefiles configuration to a single place

adrcunha:
Bonuses:
* a few more comments
* document reasons and solutions for auth errors in update-config
* parameterize TestGrid settings

### DIFF
--- a/ci/Makefile.base
+++ b/ci/Makefile.base
@@ -1,4 +1,4 @@
-# Copyright 2018 The Knative Authors
+# Copyright 2019 The Knative Authors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


ci/Makefile.base
```diff
@@ -0,0 +1,55 @@
+# Copyright 2018 The Knative Authors
```
chaodaiG: 2019

ci/Makefile.base
```diff
@@ -0,0 +1,55 @@
...
+# Default settings for the CI/CD system.
+
+CLUSTER            ?= prow
+PROJECT            ?= knative-tests
+ZONE               ?= us-central1-f
+JOB_NAMESPACE      ?= test-pods
+KNATIVE_CONFIG     ?= config_knative.yaml
+SKIP_CONFIG_BACKUP ?=
+
+CI_DIR           ?= $(realpath $(dir $(abspath $(lastword $(MAKEFILE_LIST)))))
```
chaodaiG: feel like I can use a comment here

ci/Makefile.base
adrcunha: Done.

ci/Makefile.base
adrcunha: Why now, but not in the original code? Just because of realpath()? :) Done.

chaodaiG: thanks, just think it would be useful for someone who is not super familiar with all of these commands

PR #1429, opened 2019-10-02
Update and add missing roles to sinker initial config

adrcunha:
A new instance, if created from scratch, won't start since these roles will be wrong or missing.

Shamelessly copied from k8s Prow config.

### DIFF
--- a/ci/prow/config_start.yaml
+++ b/ci/prow/config_start.yaml
@@ -236,6 +236,39 @@ metadata:
 kind: Role
 apiVersion: rbac.authorization.k8s.io/v1beta1
 metadata:
+  namespace: default
+  name: "sinker"
+rules:
+  - apiGroups:
+      - "prow.k8s.io"
+    resources:
+      - prowjobs
+    verbs:
+      - delete
+      - list
+      - watch
+      - get
+  - apiGroups:
+      - ""
+    resources:
+      - configmaps
+    resourceNames:
+      - prow-sinker-leaderlock
+    verbs:
+      - get
+      - update
+  - apiGroups:
+      - ""
+    resources:
+      - configmaps
+      - events
+    verbs:
+      - create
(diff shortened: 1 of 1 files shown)


ci/prow/config_start.yaml
```diff
@@ -236,26 +236,66 @@ metadata:
...
+      - watch
+      - get
   - apiGroups:
-      - "prow.k8s.io"
+      - ""
     resources:
-      - prowjobs
+      - configmaps
+    resourceNames:
+      - prow-sinker-leaderlock
```
yt3liu: Which resource is named prow-sinker-leaderlock? (Maybe I'm missing some context here)

ci/prow/config_start.yaml
```diff
@@ -236,26 +236,66 @@ metadata:
...
+      - create
+---
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  namespace: test-pods
+  name: "sinker"
+rules:
+  - apiGroups:
+      - ""
```
yt3liu: Does it make sense to add a comment "" indicates the core API group. I find it not-intuitive and only figure out what it does after i read the doc. (could just be because i'm not a k8s expert :( 
https://kubernetes.io/docs/reference/access-authn-authz/rbac/#role-and-clusterrole

ci/prow/config_start.yaml
adrcunha: I didn't investigate what that is, I simply copied from the main Prow config. I'm really doubtful about deeply understanding and documenting all the stuff copied from the main Prow config, since we don't follow its development too closely.

ci/prow/config_start.yaml
adrcunha: Ditto.

PR #1430, opened 2019-10-02
Merge Boskos initial setup into main initial setup

adrcunha:
Have a single initial setup, it's simpler.


ci/prow/config_start.yaml
```diff
@@ -31,6 +31,13 @@ metadata:
 data:
   config: ""
 ---
+apiVersion: v1
```
chaodaiG: I think we need a comment here mentioning it's for Boskos configmap

adrcunha: Done.

PR #1432, opened 2019-10-03
Remove test account for performance test jobs

chizhg:
We do not need the test account for performance test jobs. This PR removes it, and set `GOOGLE_APPLICATION_CREDENTIALS` to the service account key that is specific to that project.

This change will need to be coordinated with other changes.

/hold


chizhg:
The change in this PR was included in https://github.com/knative/test-infra/pull/1478 and has been merged.
/close

PR #1438, opened 2019-10-04
Parameterize Boskos projects creation/setup

adrcunha:
* allow custom resource file through $BOSKOS_RESOURCE_FILE
* allow custom project prefix through $BOSKOS_PROJECT_PREFIX
* allow extra custom roles through command line in both scripts

### DIFF
--- a/ci/prow/set_boskos_permissions.sh
+++ b/ci/prow/set_boskos_permissions.sh
@@ -17,6 +17,10 @@
 set -e
 
 readonly PROJECT=${1:?"First argument must be the boskos project name."}
+# Remaining arguments are resources to be added, just like the RESOURCES array below.
+# For example, this command add an extra editor to the project:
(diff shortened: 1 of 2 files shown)


ci/prow/set_boskos_permissions.sh
```diff
@@ -17,6 +17,10 @@
 set -e
 
 readonly PROJECT=${1:?"First argument must be the boskos project name."}
+# Remaining arguments are resources to be added, just like the RESOURCES array below.
```
chaodaiG: It might be not able to achieve easily, but still want to see if it's possible to enforce the first argument to be project

adrcunha: If it's not, gcloud in line 40 will fail and the script aborted. Ideally we would use flags to avoid confusion, but that would make this script too complicated, it's not worth it.

PR #1442, opened 2019-10-04
Clean up images and clusters in parallel

### LINKED ISSUE #1399: Improve efficiency of test resources cleanup tool
Left over test resources such as test images are cleaned up with a [cleanup tool](https://github.com/knative/test-infra/tree/master/tools/cleanup) that runs weekly, which has been having trouble catching up cleaning up as we have expanded Boskos pool from 20 to 75 projects since the tool was implemented. 

The performance of the tool is not ideal as it's running cleaning up in sequential order, which means it `recursively` lists images and deletes them one by one. We should improve the tool to allow it doing these jobs in parallel so that it can do the job much faster

chizhg:
Fix #1399 

Accidentally saw this issue, and looks it does not require much change. Hope it does not overlap with your work @adrcunha 

/cc @chaodaiG 
/cc @adrcunha

### DIFF
--- a/ci/prow/config.yaml
+++ b/ci/prow/config.yaml
@@ -6070,7 +6070,7 @@ periodics:
       command:
       - "./tools/cleanup/cleanup.sh"
       args:
-      - "--project-resource-yaml ci/prow/boskos/resources.yaml"
+      - "--project-resource-yaml ci/prow/boskos_resources.yaml"
       - "--days-to-keep-images 30"
       - "--hours-to-keep-clusters 24"
       - "--service-account /etc/test-account/service-account.json"
--- a/ci/prow/periodic_config.go
+++ b/ci/prow/periodic_config.go
@@ -220,7 +220,7 @@ func generateCleanupPeriodicJob() {
 	data.Base.DecorationConfig = []string{"timeout: 432000000000000"} // 120 hours
 	data.Base.Command = cleanupScript
 	data.Base.Args = []string{
-		"--project-resource-yaml ci/prow/boskos/resources.yaml",
+		"--project-resource-yaml ci/prow/boskos_resources.yaml",
 		"--days-to-keep-images 30",
 		"--hours-to-keep-clusters 24",
 		"--service-account " + data.Base.ServiceAccount,
--- a/test/unit/cleanup-tests.sh
+++ b/test/unit/cleanup-tests.sh
@@ -19,7 +19,7 @@ source $(dirname $0)/../../tools/cleanup/cleanup-functions.sh
 
 readonly _FAKE_NIGHTLY_PROJECT_NAME="gcr.io/knative-nightly"
 readonly _FAKE_BOSKOS_PROJECT_NAME="gcr.io/fake-boskos-project"
-readonly _PROJECT_RESOURCE_YAML="ci/prow/boskos/resources.yaml"
+readonly _PROJECT_RESOURCE_YAML="ci/prow/boskos_resources.yaml"
 readonly _RE_PROJECT_NAME="knative-boskos-[a-zA-Z0-9]+"
 
 # Call "cleanup.sh" function with given parameters
--- a/tools/cleanup/README.md
+++ b/tools/cleanup/README.md
@@ -34,5 +34,5 @@ more than 24 hours ago.
 
 There is a weekly prow job that triggers this tool runs at 11:00/12:00PM(Day
 light saving) PST every Monday. This tool scans all gcr projects defined in
-[ci/prow/boskos/resources.yaml](/ci/prow/boskos/resources.yaml) and deletes
+[ci/prow/boskos_resources.yaml](/ci/prow/boskos/resources.yaml) and deletes
 images older than 90 days and clusters older than 24 hours.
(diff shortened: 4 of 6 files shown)


adrcunha:
I have two concerns about this PR:
1. Did you test it, to ensure gcloud doesn't fail when the two delete operations run in parallel?
1. Spawning multiple shell processes in background makes logging and debugging a pain. There's no better option?

chaodaiG:
Same here, I was actually imagining a go tool using routines, interested in practicing concurrent knowledge here? :)

chizhg:
> I have two concerns about this PR:
> 
> 1. Did you test it, to ensure gcloud doesn't fail when the two delete operations run in parallel?
> 2. Spawning multiple shell processes in background makes logging and debugging a pain. There's no better option?

I tested it - https://prow.knative.dev/log?job=ci-knative-cleanup&id=1180247075057569792, though it's still running... 
I tried to use `xargs`, but it was always failing, I believe it's because of the recursive function calls in `delete_old_images_from_gcr`. Someone also suggested to use [parallel](https://www.slashroot.in/how-run-multiple-commands-parallel-linux), but I believe it will also fail due to the same reason.

chizhg:
> Same here, I was actually imagining a go tool using routines, interested in practicing concurrent knowledge here? :)

We'll still have the logging and debugging issue as mentioned above, as we can see from our e2e tests.

chizhg:
Seems the job is already broken due to the boskos resource file path change. Do you guys want me to submit a PR to fix that first?

adrcunha:
> I tested it - https://prow.knative.dev/log?job=ci-knative-cleanup&id=1180247075057569792, though it's still running...

I finished after 6h20m:

https://storage.googleapis.com/knative-prow/logs/ci-knative-cleanup/1180247075057569792/build-log.txt

It's a good example of how messed the log becomes when starting multiple shell commans in background at once. And there are a few nasty errors too:

```
ERROR: (gcloud.container.images.list) Bad status during token exchange: 429
Quota Exceeded.
Checking gcr.io/knative-boskos-61/keventing-contrib-e2e-img/23484/controller-17550e04ebf9af897aa4b799ad074e13 for removal
ERROR: gcloud crashed (BadStateException): Unexpected response pinging the registry: 429
Body: Quota Exceeded.
```

> I tried to use `xargs`, but it was always failing, I believe it's because of the recursive function calls in `delete_old_images_from_gcr`. Someone also suggested to use [parallel](https://www.slashroot.in/how-run-multiple-commands-parallel-linux), but it should also fail due to the same reason.

I suggest a more simple approach: run the tool more often (say, every 6h) and limit the number of images/projects to something that can be processed in less than 6h (pass them through flags). This way you'll have a constant "cleanup flow" that shouldn't take a huge time to cleanup the projects anymore (but do expect the first runs to be fully booked though).

chaodaiG:
@chizhg , still working on this?

chizhg:
> > I tested it - https://prow.knative.dev/log?job=ci-knative-cleanup&id=1180247075057569792, though it's still running...
> 
> I finished after 6h20m:
> 
> https://storage.googleapis.com/knative-prow/logs/ci-knative-cleanup/1180247075057569792/build-log.txt
> 
> It's a good example of how messed the log becomes when starting multiple shell commans in background at once. And there are a few nasty errors too:
> 
> ```
> ERROR: (gcloud.container.images.list) Bad status during token exchange: 429
> Quota Exceeded.
> Checking gcr.io/knative-boskos-61/keventing-contrib-e2e-img/23484/controller-17550e04ebf9af897aa4b799ad074e13 for removal
> ERROR: gcloud crashed (BadStateException): Unexpected response pinging the registry: 429
> Body: Quota Exceeded.
> ```
> 
> > I tried to use `xargs`, but it was always failing, I believe it's because of the recursive function calls in `delete_old_images_from_gcr`. Someone also suggested to use [parallel](https://www.slashroot.in/how-run-multiple-commands-parallel-linux), but it should also fail due to the same reason.
> 
> I suggest a more simple approach: run the tool more often (say, every 6h) and limit the number of images/projects to something that can be processed in less than 6h (pass them through flags). This way you'll have a constant "cleanup flow" that shouldn't take a huge time to cleanup the projects anymore (but do expect the first runs to be fully booked though).

The problem described in #1399 is our current cleanup job is unable to delete the stale images and GKE clusters (it's running once per week and its timeout is 5 days). The suggested approach will still have at most one job running at the same time, which does not make much difference IMO.

We may need further discussion on how it should be solved.

chizhg:
> @chizhg , still working on this?

Sorry I don't have enough bandwidth for working on this. Feel free to pick it up.

/close

adrcunha:
> The problem described in #1399 is our current cleanup job is unable to finish cleaning up all the stale images and GKE clusters (it's running once per week and has timeout to be 5 days). The suggested approach will still have at most one job running at the same time, which does not make much difference IMO.

This is because of the huge backlog, thus I suggested changing the periodicity. Once you have a small backlog the cleanup should go quickly.

chaodaiG:
Increasing frequency should solve the problem for now, I would suggest we create a feature request solving this properly in the long run

PR #1448, opened 2019-10-08
coverage tool uses existing sets rather than inventing another one

chaodaiG:
Noticed an implementation of `sets` while working on #1447 , delete it and use an already existing lib instead


chaodaiG:
/hold
found more

chaodaiG:
done
/hold cancel

PR #1449, opened 2019-10-08
Use clustermanager lib for managing cluster lifecycle

chaodaiG:
This PR utilizes clustermanger lib for all GKE and Boskos related logics, aiming at retiring kubetest.

This change is meant to be applied to test-infra only at the moment, let it bake for a while before enabling it in other repo(s). So instead of replacing kubetest logic in `scripts/e2e-tests.sh`, the changes are in `test/e2e-tests.sh`, by overriding `create_test_cluster` and `setup_test_cluster`. When we feel good about it we can move these 2 functions to `scripts/e2e-tests.sh` for the migration.

Note for reviewers:

- For easier overall testing and evaluation, I have included some utils written for this purpose in this PR as well, some if not all of them shall be ported to pkg. These utils include:
    - `metahelper`: read/write metadata which is used by Prow
    - `prow-cluster-operation`: a CLI tool which can be invoked easily by bash script
- For `prow-cluster-operation`, my personal feeling is that it's more natural to be a CLI tool, as this will make us easier to update, the only catch is we may need to version it in case of breaking changes. However, all of these are open to discussion

/cc @adrcunha 
/cc @chizhg


chaodaiG:
Modifying vendor in this PR for better testing, please don't review yet

test/prow-cluster-operation/README.md
```diff
@@ -0,0 +1,57 @@
...
+1. Acquiring cluster if kubeconfig already points to it
+1. Get GCP project name if not provided as a parameter:
+    - [In Prow] Acquire from Boskos
+    - [Not in Prow] Read from gcloud config
+
+    Failed obtaining project name will fail the tool
+1. Get default cluster name if not provided as a parameter
+1. Delete cluster if cluster with same name and location already exists in GKE
+1. Create cluster
+1. Write cluster metadata to `${ARTIFACT}/meta.json`
```
chizhg: Just curious, how is the file being used in the following steps?

chaodaiG: This is replacing https://github.com/knative/test-infra/blob/adedf32eed790736a3a1404a6c3e46d3fef8ad88/scripts/e2e-tests.sh#L138

Prow uses it for displaying metadata, and we also use it for setting up everything that's consumed by downstream cluster setup function

chaodaiG:
Waiting on https://github.com/knative/pkg/pull/755

chaodaiG:
/close
In favor of #1460

Issue #1454, opened 2019-10-10
Tide is not working for knative/website anymore


adrcunha:
See https://github.com/knative/website/pull/97#issuecomment-540265926

adrcunha:
The reason is that after https://github.com/knative/test-infra/pull/1412 only repos with presubmits have tide enabled.

PR #1455, opened 2019-10-10
Configure knative/website for tide

adrcunha:
knative/website only uses tide, but a presubmit entry is required to generate the config. Thus, add a presubmit entry with no presubmit tests enabled.

Part of #1454


chaodaiG:
don't you need to update template file?

chaodaiG:
nvm
/lgtm

PR #1453, opened 2019-10-09
Explicitly pass flags to serving E2E tests

JRBANCEL:
**Which issue(s) this PR fixes**:
Part of #1398.

Corresponding `knative.dev/serving` PR: https://github.com/knative/serving/pull/5782

PR #1456, opened 2019-10-10
Fix Prow

JRBANCEL:
https://github.com/knative/test-infra/pull/1453 broke Prow because we either have to use `args` or `full-command`.

### DIFF
--- a/ci/prow/config_knative.yaml
+++ b/ci/prow/config_knative.yaml
@@ -35,27 +35,20 @@ presubmits:
           memory: 16Gi
     - unit-tests: true
     - integration-tests: true
-      args:
-      - "--run-test"
-      - "./test/e2e-tests.sh --istio-version 1.2-latest --no-mesh"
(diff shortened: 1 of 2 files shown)


ci/prow/config_knative.yaml
```diff
@@ -35,27 +35,20 @@ presubmits:
           memory: 16Gi
     - unit-tests: true
     - integration-tests: true
-      args:
-      - "--run-test"
-      - "./test/e2e-tests.sh --istio-version 1.2-latest --no-mesh"
+      full-command: "./test/e2e-tests.sh --istio-version 1.2-latest --no-mesh"
```
adrcunha: This will work, but you're loosing all the presubmit checks by supplying the full command.

I suggest rolling back the previous PR if the jobs are broken, then properly passing the flags. I think this is unsupported by presubmit-tests.sh, so it will need a small update.

JRBANCEL:
/close
Will revert previous change.

PR #1465, opened 2019-10-15
Automatically generate the Prow plugins config

adrcunha:
Repos with plugins enabled are those with tide enabled.

The autogenerated plugins config file will be longer now because each repo is specified explicitly.

Currently all repos have the same set of plugins enabled.

There are no changes in which plugins are enabled for any repos, compared to the previous static config.

### DIFF
--- a/ci/prow/templates/prow_plugins.yaml
+++ b/ci/prow/templates/prow_plugins.yaml
@@ -1,0 +1,77 @@
+# Copyright 2019 The Knative Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+approve:
+- repos:
+  [[indent_array 2 .TideRepos]]
+  require_self_approval: false
+  ignore_review_state: false
+  
+heart:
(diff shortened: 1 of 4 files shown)


ci/prow/templates/prow_plugins.yaml
```diff
@@ -0,0 +1,77 @@
...
+    knative:
+      org_maintainers_team_id: 2652083 # knative-admin
+      # TODO(Fredy-Z): Enable this plugin for other repos when needed.
+      project_repo_configs:
+        eventing:
+          # TODO(grantr): replace with a new team eventing-project-maintainers
+          repo_maintainers_team_id: 3012514 # knative-milestone-maintainers
+          repo_default_column_map:
+            perf/measurement:
+              To do
```
yt3liu: Does this need to be moved as a comment? Is this a special place holder?

ci/prow/templates/prow_plugins.yaml
```diff
@@ -0,0 +1,77 @@
...
+    # your team responsible for maintaining the milestones:
+    # curl -H "Authorization: token <token>" https://api.github.com/orgs/<org-name>/teams
+    maintainers_id: 3012514
+    maintainers_team: knative-milestone-maintainers
+
+project_config:
+  project_org_configs:
+    knative:
+      org_maintainers_team_id: 2652083 # knative-admin
+      # TODO(Fredy-Z): Enable this plugin for other repos when needed.
```
yt3liu: Which plugin is the TODO referring to?

adrcunha: The project plugin.

ci/prow/templates/prow_plugins.yaml
adrcunha: No, that's a string.

adrcunha:
Please note that prow_plugins.yaml is just the previous plugins.yaml with a few changes to make it a template.

PR #1470, opened 2019-10-16
Add lenses config for spyglass

chizhg:
Add lenses config for spyglass, as per https://github.com/kubernetes/test-infra/tree/master/prow/spyglass.
The lenses adds all the current default regexes for highlight, and a new regex that highlights the go test timeout, to make it more noticeable. With this in place, we can highlight more contents that need our attention.

/cc @chaodaiG 
/cc @adrcunha

### DIFF
--- a/ci/prow/config.yaml
+++ b/ci/prow/config.yaml
@@ -56,7 +56,7 @@ deck:
         name: build-log
         config:
           highlight_regexes:
-          - 'panic: test timed out after'
+          - 'panic: test timed out after.*$'
           - timed out
           - 'ERROR:'
           - (\s|^)(FAIL|Failure \[)\b
--- a/ci/prow/config.yaml
+++ b/ci/prow/config.yaml
@@ -29,10 +29,10 @@ plank:
     grace_period: 15000000000 # 15s
     utility_images:
       # Update these versions when updating plank version in cluster.yaml
-      clonerefs: "gcr.io/k8s-prow/clonerefs:v20190925-d7e1d9b17"
-      initupload: "gcr.io/k8s-prow/initupload:v20190925-d7e1d9b17"
-      entrypoint: "gcr.io/k8s-prow/entrypoint:v20190925-d7e1d9b17"
-      sidecar: "gcr.io/k8s-prow/sidecar:v20190925-d7e1d9b17"
+      clonerefs: "gcr.io/k8s-prow/clonerefs:v20191009-0492f1d6e"
+      initupload: "gcr.io/k8s-prow/initupload:v20191009-0492f1d6e"
+      entrypoint: "gcr.io/k8s-prow/entrypoint:v20191009-0492f1d6e"
+      sidecar: "gcr.io/k8s-prow/sidecar:v20191009-0492f1d6e"
     gcs_configuration:
       bucket: "knative-prow"
       path_strategy: "explicit"
--- a/ci/prow/config.yaml
+++ b/ci/prow/config.yaml
@@ -43,17 +43,14 @@ deck:
     gcs_browser_prefix: https://console.cloud.google.com/storage/browser/
     testgrid_config: gs://knative-testgrid/config
     testgrid_root: https://testgrid.knative.dev/
-    viewers:
-      "started.json|finished.json":
-      - "metadata"
-      "build-log.txt":
-      - "buildlog"
-      "artifacts/junit.*\\.xml":
-      - "junit"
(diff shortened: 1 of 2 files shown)


ci/prow/config.yaml
```diff
@@ -51,6 +51,19 @@ deck:
       "artifacts/junit.*\\.xml":
       - "junit"
     announcement: "The old job viewer, Gubernator, has been deprecated in favour of this page, Spyglass.{{if .ArtifactPath}} For now, the old page is <a href='https://gubernator.knative.dev/build/{{.ArtifactPath}}'>still available</a>.{{end}} Please send feedback to Knative productivity."
+    lenses:
+    - lens:
+        name: build-log
+        config:
+          highlight_regexes:
+          - 'panic: test timed out after'
+          - timed out
```
adrcunha: I wonder why this entry doesn't capture the one above it. Could it be the missing single quotes?

chizhg: It can capture it, but it's very easy to be ignored, see the screenshot below:
image

I believe if the whole sentence is highlighted, it will be more noticeable to the developers.

chizhg: Modified a bit to highlight the whole line.

chizhg:
The config does not seem to work, and it's causing some rendering problem. And Kubernetes test-infra is not actually using this config.. I'm gonna check with the Spyglass developer, and close this issue for now.

chizhg:
The reason is because `viewer` config has been deprecated in spyglass. Removed the config and tested with the new config, it works as expected , see https://prow.knative.dev/view/gcs/knative-prow/logs/ci-knative-serving-istio-1.3-mesh/1184270415527153664

chizhg:
Also, I feel developers should only care if a specific test case fails, so I want to change the `FAIL` regex to `FAIL: Test\w+`, WDYT? @chaodaiG @adrcunha

adrcunha:
> Also, I feel developers should only care if a specific test case fails, things like `FAIL FAIL knative.dev/serving/test/conformance/api/v1alpha1 669.044s` are just noise, so I want to change the `FAIL` regex to `FAIL: Test\w+`, WDYT? @chaodaiG @adrcunha

+1. Loose "error" and "fail" highlights are just noise, and usually one's concerned about a test case and not about the location of its package.

chizhg:
Modified the regexes a bit, I believe it must be at least a little better than before, we can make further changes if there are more logs we need attention.

/hold cancel

chaodaiG:
I might have read it wrong, but it seems to me the regex for `Error` is still `Error:` but not `Error: Test`

ci/prow/config.yaml
```diff
@@ -43,14 +43,27 @@ deck:
...
+    - lens:
+        name: metadata
+      required_files:
+      - started.json|finished.json
+    - lens:
+        name: buildlog
+        config:
+          highlight_regexes:
+          - 'panic: test timed out after.*$'
+          - timed out
```
chaodaiG: can we make the `timed out` a little more specific? Maybe we should ask serving developers whether they want to see this line highlighted or not:
```
istio-system                                                                                               5m36s       Warning   FailedMount                                                            pod/istio-ingressgateway-78c998c64b-c62bd                                                                                            kubelet, gke-kserving-e2e-cls1184-default-pool-9be9e001-2x7d                  MountVolume.SetUp failed for volume "ingressgateway-certs" : couldn't propagate object cache: timed out waiting for the condition                                                                                                                                                                                                                                                                                                                                                            5m36s        1       istio-ingressgateway-78c998c64b-c62bd.15cdfca29dd60f7d   
```

link here(it loads slow): https://prow.knative.dev/view/gcs/knative-prow/logs/ci-knative-serving-istio-1.3-mesh/1184270415527153664#1:build-log.txt%3A21306

chizhg:
> I might have read it wrong, but it seems to me the regex for `Error` is still `Error:` but not `Error: Test`

It's actually `FAIL:` :)
I revisited the page, I think `FAIL:` is enough to highlight the test cases and eliminate the noise. I didn't make it matching the test case name because the RED text are making me feel dazzled... But just my personal feeling :)

ci/prow/config.yaml
chizhg: `timed out` is okay I think, in https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/knative_serving/5760/pull-knative-serving-integration-tests/1184527739000983552, the `timed out` seems to be the real reason for the test failure.

chaodaiG:
Sgtm
/lgtm

chizhg:
https://prow.knative.dev/view/gcs/knative-prow/logs/ci-knative-serving-istio-1.3-mesh/1184270415527153664 looks much nicer 😄

PR #1473, opened 2019-10-17
Add regex to match boskos error to highlight on spyglass

chizhg:
The boskos error is happening a bit frequently recently, and it's not shown clearly on the log page, e.g. https://prow.knative.dev/view/gcs/knative-prow/logs/ci-knative-eventing-continuous/1184869372296957952

Add a new regex to match it.

/cc @chaodaiG 
/cc @adrcunha

### DIFF
--- a/ci/prow/config.yaml
+++ b/ci/prow/config.yaml
@@ -53,11 +53,15 @@ deck:
         name: buildlog
         config:
           highlight_regexes:
-          - 'panic: test timed out after.*$'
           - timed out
           - 'ERROR:'
-          - 'FAIL:'
           - (\s|^)panic\b
+          # test case failure
+          - 'FAIL:'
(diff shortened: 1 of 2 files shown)


ci/prow/config.yaml
```diff
@@ -53,11 +53,15 @@ deck:
         name: buildlog
         config:
           highlight_regexes:
-          - 'panic: test timed out after.*$'
           - timed out
```
adrcunha: Since you're here, what about removing the loose "timed out" and "panic" entries, and moving "ERROR:" to next to "FAILL:"?

chizhg: As per discussion offline, `timed out` is useful under certain situations so we still keep it, also separate the regexes into two groups to make them more clear.

PR #1482, opened 2019-10-22
Add pj-on-kind.sh script for running ProwJobs locally.

cjwagner:
Allows for locally testing any ProwJob by running it in a Kind cluster.  Any volumes on the job that are not emptydir or hostpath will trigger a prompt allowing for 1) the volumesource to be used as is (assume it exists in the kind cluster), 2) a hostpath substitution, 3) an emptydir substitution.
Instead of uploading results to GCS, job logs, metadata, and artifacts are copied to a local directory.
See full docs here: https://github.com/kubernetes/test-infra/blob/master/prow/build_test_update.md#using-pj-on-kindsh

It looks like most jobs require some low-privilege credentials to run. It would be good to provide documentation or a script explaining how devs can generate their own credentials and how to use them with pj-on-kind.sh https://github.com/knative/test-infra/blob/e9c240a0a1c460184daea05d4bf32ab1fedd52d0/ci/prow/config.yaml#L2622-L2628

/assign @chaodaiG

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ bazel-*
 .vscode/
 .DS_Store
 *.swp
+pj.yaml
+pod.yaml
--- a/ci/prow/pj-on-kind.sh
+++ b/ci/prow/pj-on-kind.sh
@@ -1,0 +1,30 @@
+#!/usr/bin/env bash
+# Copyright 2019 The Knative Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Runs prow/pj-on-kind.sh with config arguments specific to the prow.knative.dev instance.
+# Requries go, docker, and kubectl.
+
+# Documentation: https://github.com/kubernetes/test-infra/blob/master/prow/build_test_update.md#using-pj-on-kindsh
+# Example usage:
+# ./pj-on-kind.sh pull-knative-test-infra-unit-tests
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+prowdir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd -P)"
+export CONFIG_PATH="${prowdir}/config.yaml"
+
+bash <(curl -sSfL https://raw.githubusercontent.com/kubernetes/test-infra/master/prow/pj-on-kind.sh) "$@"


ci/prow/pj-on-kind.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+# Copyright 2019 The Kubernetes Authors.
```
chaodaiG: not sure if this is the copyright header we want, @adrcunha ?

.gitignore
```diff
@@ -4,3 +4,5 @@ bazel-*
 .vscode/
 .DS_Store
 *.swp
+pj.yaml
```
chaodaiG: these 2 could be helpful too, is there any example `pj.yaml` and `pod.yaml` that can help us understand it better?

ci/prow/pj-on-kind.sh
chaodaiG: Not sure if this is the desired copyright, @adrcunha any idea?

.gitignore
cjwagner: These are files are generated as intermediate steps of the pj-on-kind.sh script and are not removed to facilitate with debugging (see https://github.com/kubernetes/test-infra/blob/master/prow/build_test_update.md#modifying-pj-on-kindsh-for-special-scenarios).

I added them to the .gitignore so that they aren't accidentally checked in. I don't think there is much value in checking in example contents.

chaodaiG: ah. I see, I read it wrong, thought these 2 files are used to override some configs before running the jobs. Now it all makes sense to me. Thanks!

ci/prow/pj-on-kind.sh
```diff
@@ -0,0 +1,29 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Runs prow/pj-on-kind.sh with config arguments specific to the prow.knative.dev instance.
+# Requries go, docker, and kubectl.
+
+# Example usage:
```
chaodaiG: can we have a link to the README here?

cjwagner: Done.

ci/prow/pj-on-kind.sh
adrcunha: Indeed, this is the wrong boilerplate.

cjwagner: Fixed. FWIW we catch boilerplate inconsistencies with a reusable presubmit check: https://github.com/kubernetes/repo-infra/blob/master/hack/verify_boilerplate.py
https://github.com/kubernetes/repo-infra#using-this-repository

.gitignore
adrcunha: Consider storing pj.yaml in /tmp, or allowing the caller of your script determining where if should live.

ci/prow/pj-on-kind.sh
adrcunha: Interesting. We should make Matt's bot recognize bad boilerplates and suggest adding them (if it doesn't do it already).

PR #1488, opened 2019-10-24
Add new test job for serving operator

chizhg:
As requested by @houshengbo , add a new custom job for serving-operator to run test against the latest serving. 

Bonus:
Use default setting, instead of `fatal` in testgrd config generator, which makes us easier to add a custom job with any name.

/cc @adrcunha 
/cc @chaodaiG 
/cc @houshengbo

### DIFF
--- a/ci/prow/config_knative.yaml
+++ b/ci/prow/config_knative.yaml
@@ -175,7 +175,8 @@ presubmits:
     - custom-test: integration-tests-on-latest-serving
       always_run: false
       optional: true
-      command:
+      args:
+      - "--run-test"
       - "./test/e2e-tests-latest-serving.sh"
 
   knative/eventing-operator:
@@ -363,7 +364,8 @@ periodics:
       dot-dev: true
       go113: true
     - custom-job: integration-tests-on-latest-serving
-      command:
+      args:
+      - "--run-test"
       - "./test/e2e-tests-latest-serving.sh"
       dot-dev: true
       go113: true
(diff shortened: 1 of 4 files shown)


ci/prow/config_knative.yaml
```diff
@@ -172,6 +172,10 @@ presubmits:
       dot-dev: true
     - go-coverage: true
       dot-dev: true
+    - custom-test: integration-tests-on-latest-serving
```
houshengbo: We do not need this custom job for PR build so far, only the one in the periodic job is needed. Thx.

ci/prow/config_knative.yaml
```diff
@@ -172,6 +172,10 @@ presubmits:
...
+      always_run: false
+      optional: true
+      full-command: "./test/e2e-tests-latest-serving.sh"
```
adrcunha: Don't use full-command, you're skipping all the presubmit checks.

chizhg: Done, changed to command

adrcunha: It should be args, including a `--run-tests`, just like serving does.

chizhg: Changed presubmit job to that format. Also tried to change the periodic job, but seems it's not supported.

ci/prow/config_knative.yaml
```diff
@@ -357,6 +363,14 @@ periodics:
     - auto-release: true
       dot-dev: true
       go113: true
+    - custom-job: integration-tests-on-latest-serving
+      command:
+      - "./test/presubmit-tests.sh"
```
adrcunha: This is not a presubmit job, don't use presubmit-tests.sh.

chizhg: Sorry got the two messed, fixed.

Issue #1494, opened 2019-10-27
Monitoring should only handle failed jobs or be configurable


chizhg:
According to the logic in https://github.com/knative/test-infra/blob/027b4ff36fdb3a671f12e4fb1d386a742ee12183/tools/monitoring/alert/alert.go#L61, monitoring is handling all Prow jobs (parse the build log, match the error pattern, decide if alert needs to be sent, etc.), it can easily overload the service. 

We should only handle failed jobs. If we do need to handle some normal jobs for specific reason, we need to make it configurable in the `config.yaml`, and skip all jobs that should not be handled.

/kind bug

steuhs:
Fixed by https://github.com/knative/test-infra/pull/1501. 
Need to build and use new image after the change

steuhs:
Image updated and deployed https://github.com/knative/test-infra/pull/1504

PR #1501, opened 2019-10-29
[monitoring] ignore build logs of success jobs

steuhs:
Processing build from success jobs costs computing power. More importantly, with all inclusive patterns ".*" now defined in monitoring's config yaml, even success runs will be included in error log. Such behavior is undesirable.

Also improved the logging a bit

Fixes #

### DIFF
--- a/tools/monitoring/alert/alert.go
+++ b/tools/monitoring/alert/alert.go
@@ -58,8 +58,8 @@ func (c *Client) RunAlerting() {
 }
 
 func (c *Client) handleReportMessage(rmsg *prowapi.ReportMessage) {
-	if rmsg.Status == prowapi.SuccessState || rmsg.Status == prowapi.FailureState || rmsg.Status == prowapi.AbortedState {
-		log.Printf("Received Pubsub message %v\n", rmsg)
+	if rmsg.Status == prowapi.FailureState || rmsg.Status == prowapi.AbortedState {
+		log.Printf("Received Pubsub message in %s state: %v\n", rmsg.Status, rmsg)
 
 		config, err := config.ParseDefaultConfig()
 		if err != nil {


chizhg:
/lgtm
/cc @chaodaiG 
/cc @adrcunha 
This is related to https://github.com/knative/test-infra/issues/1494

tools/monitoring/alert/alert.go
```diff
@@ -58,8 +58,8 @@ func (c *Client) RunAlerting() {
 }
 
 func (c *Client) handleReportMessage(rmsg *prowapi.ReportMessage) {
-	if rmsg.Status == prowapi.SuccessState || rmsg.Status == prowapi.FailureState || rmsg.Status == prowapi.AbortedState {
-		log.Printf("Received Pubsub message %v\n", rmsg)
+	if rmsg.Status == prowapi.FailureState || rmsg.Status == prowapi.AbortedState {
```
adrcunha: Wouldn't it be safer to just use

```suggestion
	if rmsg.Status != prowapi.SuccessState {
```

?

adrcunha:
/lgtm
/approve

/hold
because of my suggestion

tools/monitoring/alert/alert.go
chizhg: Not sure if the status can be other states, like `Pending` or `Triggered`. If yes, probably we should not make this change. Maybe @yt3liu can confirm?

yt3liu: The status can definitely be `Pending` and `Triggered`. For the purpose of monitoring, it'd only care about final states.

tools/monitoring/alert/alert.go
```diff
@@ -58,8 +58,8 @@ func (c *Client) RunAlerting() {
...
+		log.Printf("Received Pubsub message in %s state: %v\n", rmsg.Status, rmsg)
```
yt3liu: FYI, dumping rmsg also includes the states as well. But this is ok if you'd like to be more explicit on the state in the log.
```
Received Pubsub message &{knative-tests knative-monitoring ci-knative-client-continuous success https://prow.knative.dev/view/gcs/knative-prow/logs/ci-knative-client-continuous/1188909808967749632 gs://knative-prow/logs/ci-knative-client-continuous/1188909808967749632 [{knative client  master   [] knative.dev/client  false 0}] periodic ci-knative-client-continuous}
```

tools/monitoring/alert/alert.go
steuhs: @chaodaiG an example would be resource not ready/found for long and the job times out

tools/monitoring/alert/alert.go
steuhs: Let's be more explicit so that we can filter message with what we defined here

tools/monitoring/alert/alert.go
steuhs: @adrcunha Joyce is right. We should only check final state

adrcunha: Thanks for clarifying.

/hold cancel

PR #1503, opened 2019-10-29
[monitoring] update build and deployment doc of monitoring tool

steuhs:

Fixes #

### DIFF
--- a/tools/monitoring/README.md
+++ b/tools/monitoring/README.md
@@ -31,6 +31,8 @@ monitoring_service file to make any updates.
 1. Run `make push` in `images/monitoring/` to build and deploy the monitoring
    images.
 
+1. Update `tools/monitoring/gke_deployment/monitoring_service.yaml` to use the image tag created in previous step
+
 1. Update to use the latest image on GKE
 
    ```bash
--- a/tools/monitoring/README.md
+++ b/tools/monitoring/README.md
@@ -31,23 +31,15 @@ monitoring_service file to make any updates.
 1. Build and deploy the monitoring images.
    ```bash
    cd images/monitoring/
-   make build
-   make push_versioned
+   make push
    ```
 
 1. Update `tools/monitoring/gke_deployment/monitoring_service.yaml` to use the image tag created in previous step
 
-1. Update to use the latest image on GKE
+1. Apply the changes to use the new image in production
 
    ```bash
-   kubectl rollout restart deployment.apps/monitoring-app
-   ```
-   Note that if the "restart" sub-command is not found, update your kubectl installation.
-
-   Check the rollout status
-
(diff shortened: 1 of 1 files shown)


tools/monitoring/README.md
```diff
@@ -28,14 +28,21 @@ monitoring_service file to make any updates.
 
 ### Update Image
 
-1. `images/monitoring/Makefile` Commands to build and deploy the monitoring
-   images.
+1. Build and deploy the monitoring images.
+   ```bash
+   cd images/monitoring/
+   make build
+   make push_versioned
```
yt3liu: `make push` is sufficient
https://github.com/knative/test-infra/blob/master/shared/Makefile.simple-image

tools/monitoring/README.md
```diff
@@ -28,14 +28,21 @@ monitoring_service file to make any updates.
...
+   ```
+
+1. Update `tools/monitoring/gke_deployment/monitoring_service.yaml` to use the image tag created in previous step
 
 1. Update to use the latest image on GKE
 
    ```bash
    kubectl rollout restart deployment.apps/monitoring-app
```
yt3liu: The rollout restart would work only if the tag remains the same. (also i think the command may be deprecated)
To use the latest image, create a new PR updating [monitoring.yaml](https://github.com/knative/test-infra/blob/master/tools/monitoring/gke_deployment/monitoring_service.yaml#L43), and run `kubectl apply -f <monitoring_service.yaml>` on the monitoring cluster after that's merged.

tools/monitoring/README.md
```diff
@@ -28,14 +28,21 @@ monitoring_service file to make any updates.
...
    ```
+   Note that if the "restart" sub-command is not found, update your kubectl installation.
```
yt3liu: Since we don't need to use `kubectl rollout` anymore, a simple check on the status of the components should be enough i.e. `kubectl get all` or `kubectl get pods`.

tools/monitoring/README.md
steuhs: Merging is not required to test out the result with `kubectl apply -f <monitoring_service.yaml>`, right?

yt3liu: do you mean you'd like to run `kubectl apply -f <monitoring_service.yaml>` without the change merged into the service? 

If that's the case, I would say merging is still required. Otherwise, you'll be saying merging changes in made in the monitoring_service.yaml can be changed without going through proper code reviews. 

If you're interested in fully testing changes, spin up a cluster and deploy the service on your sandbox project.

steuhs: I see. That's a practice which holds for other projects as well. I guess we don't need to include it here.

PR #1514, opened 2019-11-01
Increase alert period

chizhg:
/cc @chaodaiG 
/cc @steuhs


steuhs:
/hold

this is not the cause of the missing alert problem. I'll upload a PR soon to address it

steuhs:
see https://github.com/knative/test-infra/pull/1515

chaodaiG:
/approve
I think this is orthogonal with #1515

PR #1515, opened 2019-11-01
Fix config issue that potentially caused some alert not sent

### LINKED ISSUE #1513: Alert was not sent after the number of job failures pass the threshold
We have configured monitoring for performance testing related Prow jobs, but when it failed for multiple times (more than the threshold set in the configuration file),  no alert was sent. 

Job history: https://prow.knative.dev/?job=ci-knative-eventing-update-clusters&state=failure
Config: https://github.com/knative/test-infra/blob/89cde81f6a2bba89d07b85fec67086cafac6ce26/tools/monitoring/config/config.yaml#L181

Screenshot of the related log message is in the attachment.
image

steuhs:
Same error pattern should not appear more than once in the config.
The monitoring tool only looks at the first matching error pattern in the config.

The mis-config likely caused the issue https://github.com/knative/test-infra/issues/1513

Fixes #1513 

### DIFF
--- a/tools/monitoring/config/config.yaml
+++ b/tools/monitoring/config/config.yaml
@@ -161,6 +161,7 @@ spec:
         jobs-affected: 1
         prs-affected: 0
         period: 1440
+
   - error-pattern: 'step failures'
     hint: 'Verify that these are not from dry-runs'
     alerts:
@@ -168,6 +169,7 @@ spec:
         occurrences: 2
         jobs-affected: 1
         period: 2880
+
   - error-pattern: '.*' # We need to check for all errors
     hint: 'Look at the errors in the log'
     alerts:
@@ -175,9 +177,6 @@ spec:
         occurrences: 1
         jobs-affected: 1
         period: 1440
-  - error-pattern: '.*' # Check all errors
-    hint: 'Look at the errors in build log'
-    alerts:
       - job-name-regex: 'ci-knative-.*-clusters' # Performance recycle clusters jobs
         occurrences: 3
         jobs-affected: 1


tools/monitoring/config/config.yaml
```diff
@@ -161,23 +161,22 @@ spec:
...
   - error-pattern: '.*' # We need to check for all errors
     hint: 'Look at the errors in the log'
     alerts:
       - job-name-regex: 'ci-knative-cleanup' # Cleanup test images for all knative
         occurrences: 1
         jobs-affected: 1
         period: 1440
-  - error-pattern: '.*' # Check all errors
-    hint: 'Look at the errors in build log'
-    alerts:
```
yt3liu: Could you please explain why this config change would fix the issue? Currently, there are at least 3 error jobs matching the job-name-regex 'ci-knative-.*-clusters' in the database. That should satisfy the last config requirement without making this change.

chaodaiG:
If this change fixed the error, then I think the logic of only look at first matching error pattern is a bug, needs to be fixed
/hold

steuhs:
> If this change fixed the error, then I think the logic of only look at first matching error pattern is a bug, needs to be fixed
> /hold

Can you give an example why this would be a bug? @chaodaiG

tools/monitoring/config/config.yaml
steuhs: Without this fix, the entire configuration here will not be used at all, because the error pattern already appeared before in the config file.

`  - error-pattern: '.*' # Check all errors
    hint: 'Look at the errors in build log'
    alerts:
      - job-name-regex: 'ci-knative-.*-clusters' # Performance recycle clusters jobs
        occurrences: 3
        jobs-affected: 1
        period: 1440`

I explained it in the PR message and please sync offline if there is further questions

chaodaiG:
> > If this change fixed the error, then I think the logic of only look at first matching error pattern is a bug, needs to be fixed
> > /hold
> 
> Can you give an example why this would be a bug? @chaodaiG

The logic to me should be process each `error pattern` and `jobname`, only dedup if the same combination of these 2 have been defined above

steuhs:
> > > If this change fixed the error, then I think the logic of only look at first matching error pattern is a bug, needs to be fixed
> > > /hold
> > 
> > 
> > Can you give an example why this would be a bug? @chaodaiG
> 
> The logic to me should be process each `error pattern` and `jobname`, only dedup if the same combination of these 2 have been defined above

Makes sense. Will fix the bug

chaodaiG:
Had an in person discussion, let's have the fix in and defer the logic cleanup

PR #1519, opened 2019-11-03
Initial config parser for supervisor

adrcunha:
It unmarshals a YAML defined in https://docs.google.com/document/d/1PQ6DLL8kqMSSmciTSrR9RLaJt8sBgaziLEkwkP9TKec/edit# into the equivalent structs.

Part of #1183.

### DIFF
--- a/ci/supervisor/config/config.go
+++ b/ci/supervisor/config/config.go
@@ -1,0 +1,329 @@
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package config contains functions related to config files.
+package config
+
+import (
+	"fmt"
+	yaml "gopkg.in/yaml.v2"
+)
+
+// yamlMap is a generic type for YAML files
+type yamlMap = map[interface{}]interface{}
+
+// EnvVar holds a environment variable (name and value)
+type EnvVar struct {
+	Name, Value string
+}
+
+// ResourcesRequest holds resource requests for a task or job
+type ResourcesRequest struct {
+	CPU    int
+	Memory string
+}
+
+// ResourcesConfig holds the config of resources used by a task or job
+type ResourcesConfig struct {
+	Requests ResourcesRequest
+}
+
+// DefaultSettings holds the default settings to be used by the supervisor
+type DefaultSettings struct {
+	MergeMethod       string `yaml:"merge-method"`
+	Optional          bool
+	ParseGoTestOutput bool   `yaml:"parse-go-test-output"`
+	ReportToGitHub    bool   `yaml:"report-to-github"`
+	EnableDocker      bool   `yaml:"enable-docker"`
+	PresubmitFilter   string `yaml:"presumit-filter"`
+	Image             string
+	ImageArgs         []string `yaml:"image-args"`
+	Env               []EnvVar
+	Resources         ResourcesConfig
+	TestGridConfig    string `yaml:"testgrid-config"`
+}
+
+// CodeCoverageSettings holds the settings for the code coverage task
+type CodeCoverageSettings struct {
+	Threshold      string
+	FailsPresubmit bool `yaml:"fails-presubmit"`
+	Image          string
+	ImageArgs      []string `yaml:"image-args"`
+	Env            []EnvVar
+	Resources      ResourcesConfig
+}
+
+// Task holds the basic information for a task
+type Task struct {
+	Command           string
+	ParseGoTestOutput bool `yaml:"parse-go-test-output"`
+	EnableDocker      bool `yaml:"enable-docker"`
+	Image             string
+	ImageArgs         []string `yaml:"image-args"`
+	Env               []EnvVar
+	Resources         ResourcesConfig
+}
+
+// ProwPlugin holds the parameters of a Prow plugin
+type ProwPlugin struct {
+	Enabled    bool
+	Parameters yamlMap
+}
+
+// PresubmitTask holds the parameters of a presubmit task
+type PresubmitTask struct {
+	TaskParameters  Task
+	PresubmitFilter string `yaml:"presubmit-filter"`
+	Optional        bool
+	ReportToGitHub  bool `yaml:"report-to-github"`
+}
+
+// PostsubmitTask holds the parameters of a postsubmit task
+type PostsubmitTask struct {
+	TaskParameters Task
+}
+
+// PeriodicTask holds the parameters or a periodic task
+type PeriodicTask struct {
+	TaskParameters          Task
+	Periodicity             string
+	MinorReleasePeriodicity string `yaml:"minor-release-periodicity"`
+	MinorReleaseArg         string `yaml:"minor-release-arg"`
+}
+
+// Defaults holds the default settings for a repository
+type Defaults struct {
+	Settings     DefaultSettings
+	ProwPlugins  map[string]ProwPlugin
+	CodeCoverage CodeCoverageSettings
+}
+
+// RepoConfig holds all the configuration for a repository
+type RepoConfig struct {
+	Defaults        Defaults
+	PresubmitTasks  map[string]PresubmitTask
+	PostsubmitTasks map[string]PostsubmitTask
+	PeriodicTasks   map[string]PeriodicTask
+}
+
+// NightlyReleaseSettings holds the default configuration of the nightly release task
+type NightlyReleaseSettings struct {
+	Window string
+}
+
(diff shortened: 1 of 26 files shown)


chaodaiG:
I assumed this tool will parse yaml config written from developers from each WG, and they will rely on GoDoc for API usages lookup as listed here: https://godoc.org/?q=knative%2Ftest-infra. If this is the case then I think most if not all exported fields will need comments like this example: https://github.com/googleapis/google-api-go-client/blob/d708bdaa52e7a6d12efa371d241639933f1455fb/container/v1/container-gen.go#L657

adrcunha:
> I assumed this tool will parse yaml config written from developers from each WG, and they will rely on GoDoc for API usages lookup as listed here: https://godoc.org/?q=knative%2Ftest-infra. If this is the case then I think most if not all exported fields will need comments like this example: https://github.com/googleapis/google-api-go-client/blob/d708bdaa52e7a6d12efa371d241639933f1455fb/container/v1/container-gen.go#L657

I prefer writing proper documentation instead of relying on GoDoc for two major reasons:

* this is not an API, the only relevant information is the YAML file format
* the YAML fields are scattered around the code due to the required custom parsing

chaodaiG:
> > I assumed this tool will parse yaml config written from developers from each WG, and they will rely on GoDoc for API usages lookup as listed here: https://godoc.org/?q=knative%2Ftest-infra. If this is the case then I think most if not all exported fields will need comments like this example: https://github.com/googleapis/google-api-go-client/blob/d708bdaa52e7a6d12efa371d241639933f1455fb/container/v1/container-gen.go#L657
> 
> I prefer writing proper documentation instead of relying on GoDoc for two major reasons:
> 
> * this is not an API, the only relevant information is the YAML file format
> * the YAML fields are scattered around the code due to the required custom parsing

Agree that this is not API, there is not too much difference documenting in the code and documenting elsewhere. But it's harder to do code review without documentation, as it takes time to understand what kind of assumptions are taken

adrcunha:
> Agree that this is not API, there is not too much difference documenting in the code and documenting elsewhere. But it's harder to do code review without documentation, as it takes time to understand what kind of assumptions are taken

If you skim through the code you'll notice that this is a pretty simple YAML parser without any complex assumptions. It unmarshals a YAML defined in https://docs.google.com/document/d/1PQ6DLL8kqMSSmciTSrR9RLaJt8sBgaziLEkwkP9TKec/edit# into the equivalent structs, thus there are only 2 exported functions. Updated the description, let me know if this is still unclear.

ci/supervisor/config/config_test.go
```diff
@@ -0,0 +1,472 @@
...
+  foo-tests:
+    thisisanunrezognizedkey: "foo"`,
+			RepoConfig{
+				PresubmitTasks:  map[string]PresubmitTask{},
+				PeriodicTasks:   map[string]PeriodicTask{},
+				PostsubmitTasks: map[string]PostsubmitTask{"foo-tests": {}},
+			},
+			nil,
+		}, { // full YAML
+			`defaults:
```
chaodaiG: might be my personal preference, I feel it better readable having this kind of test data in separated files, so that dev's favorite editor can have proper syntax highlights for yaml. I have seen an example here:
https://github.com/jstemmer/go-junit-report/tree/master/testdata

ci/supervisor/config/config.go
```diff
@@ -0,0 +1,329 @@
...
+	"fmt"
+	yaml "gopkg.in/yaml.v2"
+)
+
+// yamlMap is a generic type for YAML files
+type yamlMap = map[interface{}]interface{}
+
+// EnvVar holds a environment variable (name and value)
+type EnvVar struct {
+	Name, Value string
```
chaodaiG: I think this would work, but isn't it better to explicitly encode their yaml keys?

ci/supervisor/config/config.go
```diff
@@ -0,0 +1,329 @@
...
+		}
+	}
+
+	return r, nil
+}
+
+// UnmarshalYAML unmarshals a SupervisorDefaults struct, correctly unmarshalling the Prow plugin section
+func (sd *SupervisorDefaults) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	// Expected fields
+	var structuredData struct {
```
chaodaiG: Why not just use sd itself? would it be easier if this method becomes a global helper method like this:
```
func UnmarshalYAMLToSuperVisorDefaults(unmarshal func(interface{}) error) (*SupervisorDefaults, error) {
```

ci/supervisor/config/config.go
```diff
@@ -0,0 +1,329 @@
...
+				return badSetting
+			}
+		}
+		pluginMap[pluginName] = plugin
+	}
+	return nil
+}
+
+// UnmarshalYAML unmarshals a Defaults struct, correctly unmarshalling the Prow plugin section
+func (d *Defaults) UnmarshalYAML(unmarshal func(interface{}) error) error {
```
chaodaiG: sorry reviewed backwards, please see comment on the next `UnmarshalYAML`

ci/supervisor/config/config.go
```diff
@@ -0,0 +1,329 @@
...
+
+// parsePlugins unmarshals a Prow plugin section into the given map
+func parsePlugins(pluginMap map[string]ProwPlugin, rawData yamlMap) error {
+	for p := range rawData {
+		pluginName, _ := p.(string)
+		plugin := ProwPlugin{Enabled: true}
+		badSetting := fmt.Errorf("valid setting for plugin %q is one of [disabled|no|false|<configuration>]", pluginName)
+		// Setting is a simple enable/disable string
+		if v, ok := rawData[pluginName].(string); ok {
+			if v == "disabled" || v == "no" || v == "false" {
```
chaodaiG: feel like these should be constants

ci/supervisor/config/config.go
```diff
@@ -0,0 +1,329 @@
...
+	Defaults     SupervisorDefaults
+	Repositories map[string][]string
+}
+
+// parsePlugins unmarshals a Prow plugin section into the given map
+func parsePlugins(pluginMap map[string]ProwPlugin, rawData yamlMap) error {
+	for p := range rawData {
+		pluginName, _ := p.(string)
+		plugin := ProwPlugin{Enabled: true}
+		badSetting := fmt.Errorf("valid setting for plugin %q is one of [disabled|no|false|<configuration>]", pluginName)
```
chaodaiG: nit, rename to `invalidSettingErr`

ci/supervisor/config/config.go
```diff
@@ -0,0 +1,329 @@
...
+	r.PostsubmitTasks = make(map[string]PostsubmitTask, 0)
+
+	// Extract the raw YAML
+	var rawConfig map[string]interface{}
+	if err := yaml.Unmarshal(content, &rawConfig); err != nil {
+		return RepoConfig{}, err
+	}
+
+	// Unmarshal the defaults section
+	if d, err := yaml.Marshal(rawConfig["defaults"]); err == nil {
```
chaodaiG: constant?

ci/supervisor/config/config.go
```diff
@@ -0,0 +1,329 @@
...
+		if err := yaml.Unmarshal(d, &r.Defaults); err != nil {
+			return RepoConfig{}, err
+		}
+	} else {
+		return RepoConfig{}, err
+	}
+
+	// Unmarshal the presubmit-tasks section
+	if value, ok := rawConfig["presubmit-tasks"].(yamlMap); ok {
```
chaodaiG: constant?

chaodaiG: might just be my personal preference, I feel like this block and following 2 blocks can be something like this:
```
r.LoadPresubmitTasksFromRawConfigOrFail(rawConfig)
r.LoadPeriodicTasksFromRawConfigOrFail(rawConfig)
r.LoadPostsubmitTasksFromRawConfigOrFail(rawConfig)
```

ci/supervisor/config/config.go
```diff
@@ -0,0 +1,329 @@
...
+			taskName, _ := a.(string)
+			// Setting can be simply the command to run
+			if v, ok := value[taskName].(string); ok {
+				r.PresubmitTasks[taskName] = PresubmitTask{TaskParameters: Task{Command: v}}
+				// Otherwise it's the complete set of parameters
+			} else {
+				if d, err := yaml.Marshal(value[taskName]); err == nil {
+					pt := PresubmitTask{}
+					err = yaml.Unmarshal(d, &pt)
+					err = yaml.Unmarshal(d, &pt.TaskParameters)
```
chaodaiG: this err is unused, guess you meant check err instead of else below. And this err doesn't seem to be covered by unit test

ci/supervisor/config/config.go
```diff
@@ -0,0 +1,329 @@
...
+	s.Repositories = make(map[string][]string, 0)
+
+	// Extract the raw YAML
+	var rawConfig map[string]interface{}
+	if err := yaml.Unmarshal(content, &rawConfig); err != nil {
+		return SupervisorConfig{}, err
+	}
+
+	// Unmarshal the defaults section
+	if d, err := yaml.Marshal(rawConfig["defaults"]); err == nil {
```
chaodaiG: An idea to save 2 returns
```
	d, err := yaml.Marshal(rawConfig["defaults"])
	if err == nil {
		err = yaml.Unmarshal(d, &s.Defaults)
	} 
	if err != nil {
		return SupervisorConfig{}, err
	}
```

Issue #1526, opened 2019-11-06
Prevent monitoring release from being broken


chizhg:
Monitoring nightly release job released a broken artifact on Nov. 4, 2019, which broke eventing e2e tests and blocked all PRs:
```
error: error validating "https://storage.googleapis.com/knative-nightly/serving/latest/monitoring.yaml": error validating data: ValidationError(DaemonSet.spec): missing required field "selector" in io.k8s.api.apps.v1.DaemonSetSpec; if you choose to ignore these errors, turn validation off with --validate=false
ERROR: Knative setup failed
```

Though we don't have any e2e tests for monitoring, we can at least `kubectl apply -f ${MONITORING_YAML}` to make sure it can be installed.

There are two things we need to do for this task:
- [ ] Similar as `start_knative_serving`, add a function `start_knative_monitoring` in `scripts/library.sh`
- [ ] In `e2e-smoke-tests.sh`, also call `start_knative_monitoring` to make sure it can be started.

PR #1527, opened 2019-11-06
Add a new function to start_knative_monitoring

chizhg:
Add a new function to start_knative_monitoring, which is the first step of #1526 

/cc @adrcunha 
/cc @chaodaiG

### DIFF
--- a/scripts/library.sh
+++ b/scripts/library.sh
@@ -411,9 +411,9 @@ function start_knative_serving() {
   header "Starting Knative Serving"
   subheader "Installing Knative Serving"
   echo "Installing Serving CRDs from $1"
-  kubectl apply --selector knative.dev/crd-install=true -f "$1"
+  kubectl apply --selector knative.dev/crd-install=true -f "$1" || return 1
   echo "Installing the rest of serving components from $1"
-  kubectl apply -f "$1"
+  kubectl apply -f "$1" || return 1
   wait_until_pods_running knative-serving || return 1
 }
 
@@ -423,12 +423,13 @@ function start_knative_monitoring() {
   header "Starting Knative Monitoring"
   subheader "Installing Knative Monitoring"
   # namespace istio-system needs to be created first, due to the comment
-  # mentioned in https://github.com/knative/serving/blob/4202efc0dc12052edc0630515b101cbf8068a609/config/monitoring/tracing/zipkin/100-zipkin.yaml#L21
-  kubectl create namespace istio-system
+  # mentioned in
+  # https://github.com/knative/serving/blob/4202efc0dc12052edc0630515b101cbf8068a609/config/monitoring/tracing/zipkin/100-zipkin.yaml#L21
+  kubectl create namespace istio-system 2>/dev/null
   echo "Installing Monitoring CRDs from $1"
-  kubectl apply --selector knative.dev/crd-install=true -f "$1"
+  kubectl apply --selector knative.dev/crd-install=true -f "$1" || return 1
   echo "Installing the rest of monitoring components from $1"
-  kubectl apply -f "$1"
+  kubectl apply -f "$1" || return 1
   wait_until_pods_running knative-monitoring || return 1
   wait_until_pods_running istio-system || return 1
 }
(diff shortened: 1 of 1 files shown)


adrcunha:
This doesn't sound like a generic function. Instead, it looks like a specific flow that should be added to serving's e2e-tests.sh instead.

chizhg:
> This doesn't sound like a generic function. Instead, it looks like a specific flow that should be added to serving's e2e-tests.sh instead.

Eventing is installing monitoring as well, see https://github.com/knative/eventing/blob/4edf2fd5f592310c4734f7c0da6f44aa54f04c06/test/e2e-common.sh#L44. 
If we add it here, eventing can also reuse it.

scripts/library.sh
```diff
@@ -417,6 +417,22 @@ function start_knative_serving() {
...
 }
 
+# Install Knative Monitoring in the current cluster.
+# Parameters: $1 - Knative Monitoring manifest.
+function start_knative_monitoring() {
+  header "Starting Knative Monitoring"
+  subheader "Installing Knative Monitoring"
+  # namespace istio-system needs to be created first, due to the comment
+  # mentioned in https://github.com/knative/serving/blob/4202efc0dc12052edc0630515b101cbf8068a609/config/monitoring/tracing/zipkin/100-zipkin.yaml#L21
+  kubectl create namespace istio-system
```
adrcunha: Please either fail the function or hide the error of the namespace already exists.

scripts/library.sh
```diff
@@ -417,6 +417,22 @@ function start_knative_serving() {
...
+  echo "Installing Monitoring CRDs from $1"
+  kubectl apply --selector knative.dev/crd-install=true -f "$1"
```
adrcunha: Fail function if you can't install?

scripts/library.sh
```diff
@@ -417,6 +417,22 @@ function start_knative_serving() {
...
+  echo "Installing the rest of monitoring components from $1"
+  kubectl apply -f "$1"
```
adrcunha: Ditto.

scripts/library.sh
chizhg: Done.

scripts/library.sh
chizhg: Done.

scripts/library.sh
chizhg: Done.

scripts/library.sh
```diff
@@ -411,12 +411,29 @@ function start_knative_serving() {
   header "Starting Knative Serving"
   subheader "Installing Knative Serving"
   echo "Installing Serving CRDs from $1"
-  kubectl apply --selector knative.dev/crd-install=true -f "$1"
+  kubectl apply --selector knative.dev/crd-install=true -f "$1" || return 1
```
adrcunha: Hmmm... I understand your change here, but I would prefer not touching `start_knative_serving` in this PR unless we had evidence that this is a bug. Otherwise I'm concerned that you might inadvertently cause issues in the next test-infra update.

scripts/library.sh
```diff
@@ -411,12 +411,29 @@ function start_knative_serving() {
...
+# Parameters: $1 - Knative Monitoring manifest.
+function start_knative_monitoring() {
+  header "Starting Knative Monitoring"
+  subheader "Installing Knative Monitoring"
+  # namespace istio-system needs to be created first, due to the comment
+  # mentioned in
+  # https://github.com/knative/serving/blob/4202efc0dc12052edc0630515b101cbf8068a609/config/monitoring/tracing/zipkin/100-zipkin.yaml#L21
+  kubectl create namespace istio-system 2>/dev/null
+  echo "Installing Monitoring CRDs from $1"
+  kubectl apply --selector knative.dev/crd-install=true -f "$1" || return 1
```
chaodaiG: seems like this function is fast fail, can we use errexit here instead of return 1 everywhere?

adrcunha: You don't want to do that, it's disruptive. We want the caller to decide what to do in case of failure.

scripts/library.sh
chizhg: Removed.

PR #1529, opened 2019-11-06
Increase boskos projects to 100

chizhg:
We occasionally get `boskos failed to acquire project` error since janitor is cleaning the projects slowly, increase boskos projects to 100.

I've already created the additional 25 projects, do you guys think we still need this change? Or we just ignore these errors for now?
/cc @adrcunha 
/cc @chaodaiG

### DIFF
--- a/ci/prow/boskos_resources.yaml
+++ b/ci/prow/boskos_resources.yaml
@@ -89,5 +89,30 @@ resources:
   - knative-boskos-73
   - knative-boskos-74
   - knative-boskos-75
+  - knative-boskos-76
+  - knative-boskos-77
+  - knative-boskos-78
+  - knative-boskos-79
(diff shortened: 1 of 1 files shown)


adrcunha:
I would prefer that you had not created another 25 projects, as I think the pool is way past the  reasonable threshold regarding maintenance. But since the projects were already created, let's add them here so the pool exhaustion doesn't happen so soon (hopefully) and we keep track of them. I just hope we don't have to make quota request changes anytime soon... :)

adrcunha:
Speaking of maintenance, I assume you did update the quotas for the 25 projects, as explained in our documentation, right? If so, please unhold.

/lgtm
/approve
/hold

chizhg:
Yes, I have already updated the quota.
/hold cancel

PR #1530, opened 2019-11-06
Markdown link checker presubmit for knative/docs is optional

adrcunha:
Otherwise tide will be blocked if the job fails (it's informational, not a merge blocker).


chizhg:
Does it mean dead links will be allowed in the docs?

adrcunha:
> Does it mean dead links will be allowed in the docs?

Technically they already are, as the link checker was disabled in the presubmit tests.

adrcunha:
For more details about this optional, non-blocking check in knative/docs, see https://github.com/knative/docs/pull/1954

chizhg:
> For more details about this optional, non-blocking check in knative/docs, see [knative/docs#1954](https://github.com/knative/docs/pull/1954)

Got it, thanks for the explanation!
/lgtm

PR #1536, opened 2019-11-11
Display nodes of the test cluster before running the test

adrcunha:
Missed duplicating the change to e2e-tests.sh (which is test-infra specific) in 1535.

/cc @chaodaiG


adrcunha:
`bazel build @k8s//prow/cmd/checkconfig` passed locally...

/retest

PR #1544, opened 2019-11-13
Allow for me to set KO_DOCKER_REPO

duglin:
If I set it to a value, I expect the script to do what I asked for.
But if there is no value then let it default to whatever it does today.

Signed-off-by: Doug Davis <dug@us.ibm.com>

<!--
Request Prow to automatically lint any go code in this PR:

/lint

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -99,6 +99,7 @@ YAMLS_TO_PUBLISH=""
 ARTIFACTS_TO_PUBLISH=""
 FROM_NIGHTLY_RELEASE=""
 FROM_NIGHTLY_RELEASE_GCS=""
+USE_KO_DOCKER_REPO=""
 export GITHUB_TOKEN=""
 
 # Convenience function to run the hub tool.
@@ -348,6 +349,7 @@ function parse_flags() {
       --dot-release) is_dot_release=1 ;;
       --auto-release) is_auto_release=1 ;;
       --from-latest-nightly) FROM_NIGHTLY_RELEASE=latest ;;
+      --use-ko-docker-repo) USE_KO_DOCKER_REPO=1 ;;
       *)
         [[ $# -ge 2 ]] || abort "missing parameter after $1"
         shift
@@ -381,6 +383,9 @@ function parse_flags() {
           --from-nightly)
             [[ $1 =~ ^v[0-9]+-[0-9a-f]+$ ]] || abort "nightly tag must be 'vYYYYMMDD-commithash'"
             FROM_NIGHTLY_RELEASE=$1
+            ;;
+          --use-env-ko-docker-repo)
+            USE_KO_DOCKER_REPO=$1
             ;;
           *) abort "unknown option ${parameter}" ;;
         esac
@@ -420,7 +425,11 @@ function parse_flags() {
   if (( ! PUBLISH_RELEASE )); then
     (( has_gcr_flag )) && echo "Not publishing the release, GCR flag is ignored"
     (( has_gcs_flag )) && echo "Not publishing the release, GCS flag is ignored"
-    KO_DOCKER_REPO="${KO_DOCKER_REPO:-ko.local}"
+    if (( USE_KO_DOCKER_REPO )); then
+      KO_DOCKER_REPO="${KO_DOCKER_REPO:-ko.local}"
+    else
+      KO_DOCKER_REPO="ko.local"
+    fi
     KO_FLAGS="${KO_FLAGS}"
     RELEASE_GCS_BUCKET=""
   fi
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -384,9 +384,6 @@ function parse_flags() {
             [[ $1 =~ ^v[0-9]+-[0-9a-f]+$ ]] || abort "nightly tag must be 'vYYYYMMDD-commithash'"
             FROM_NIGHTLY_RELEASE=$1
             ;;
-          --use-env-ko-docker-repo)
-            USE_KO_DOCKER_REPO=$1
-            ;;
           *) abort "unknown option ${parameter}" ;;
         esac
     esac
@@ -421,6 +418,11 @@ function parse_flags() {
     prepare_dot_release
   fi
 
+  if (( ! USE_KO_DOCKER_REPO )); then
+    # If they didn't use the special flag then force it to gcr
+    export KO_DOCKER_REPO="gcr.io/knative-nightly"
+  fi
+
   # Update KO_DOCKER_REPO and KO_FLAGS if we're not publishing.
   if (( ! PUBLISH_RELEASE )); then
     (( has_gcr_flag )) && echo "Not publishing the release, GCR flag is ignored"
@@ -444,9 +446,6 @@ function parse_flags() {
   (( TAG_RELEASE )) && TAG="${BUILD_TAG}"
   [[ -n "${RELEASE_VERSION}" ]] && TAG="v${RELEASE_VERSION}"
   [[ -n "${RELEASE_VERSION}" && -n "${RELEASE_BRANCH}" ]] && (( PUBLISH_RELEASE )) && PUBLISH_TO_GITHUB=1
-
-  # If not already set then go ahead and default to gcr
-  export KO_DOCKER_REPO="${KO_DOCKER_REPO:-gcr.io/knative-nightly}"
 
   readonly BUILD_COMMIT_HASH
   readonly BUILD_YYYYMMDD
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -427,12 +427,11 @@ function parse_flags() {
   if (( ! PUBLISH_RELEASE )); then
     (( has_gcr_flag )) && echo "Not publishing the release, GCR flag is ignored"
     (( has_gcs_flag )) && echo "Not publishing the release, GCS flag is ignored"
-    if (( USE_KO_DOCKER_REPO )); then
-      KO_DOCKER_REPO="${KO_DOCKER_REPO:-ko.local}"
-    else
-      KO_DOCKER_REPO="ko.local"
-    fi
-    KO_FLAGS="${KO_FLAGS}"
+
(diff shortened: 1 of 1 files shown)


coryrc:
I'm guessing --release-gcr option that lets you set KO_DOCKER_REPO does not do what you need?

duglin:
I was afraid of setting that flag :-) 

Seems to me that if the user sets KO_DOCKER_REPO to a value the script should use it

duglin:
/hold

might be another fix needed...

scripts/release.sh
```diff
@@ -421,8 +420,8 @@ function parse_flags() {
   if (( ! PUBLISH_RELEASE )); then
     (( has_gcr_flag )) && echo "Not publishing the release, GCR flag is ignored"
     (( has_gcs_flag )) && echo "Not publishing the release, GCS flag is ignored"
-    KO_DOCKER_REPO="ko.local"
-    KO_FLAGS="-L ${KO_FLAGS}"
+    KO_DOCKER_REPO="${KO_DOCKER_REPO:-ko.local}"
+    KO_FLAGS="${KO_FLAGS}"
```
duglin: removed the -L because it's redundant with the ko.local value - and if set it would override the KO_DOCKER_REPO env var

coryrc:
> I was afraid of setting that flag :-)
> 
> Seems to me that if the user sets KO_DOCKER_REPO to a value the script should use it

I think it is intentional so the release process doesn't become messed up by accident if someone leaves an ENV variable set wrong. I think a better way, if --release-gcr doesn't work, is add something like --ko-docker-repo-override and have it just change KO_DOCKER_REPO.

duglin:
> I think it is intentional so the release process doesn't become messed up by accident if someone leaves an ENV variable set wrong.  I think a better way, if --release-gcr doesn't work, is add something like --ko-docker-repo-override and have it just change KO_DOCKER_REPO.

If a "normal" dev has write access to the release repo then it seems like there are other security issues since any run of `ko` could cause harm then.

The other scripts that I've had to use to do builds appear to use KO_DOCKER_REPO as a way to tell it where to put images (e.g. serving does that). I haven't followed the code enough to fully understand why eventing's process is different, but the consistency of KO_DOCKER_REPO being how someone conveys this information is very nice from a consistency point of view. Adding more flags to re-enforce what someone is asking for just feels redundant.

duglin:
> I think it is intentional so the release process doesn't become messed up by accident if someone leaves an ENV variable set wrong. 

thinking more about this... isn't the opposite more likely? Meaning, someone is pointing to a personal repo when they run the release script? Why would someone be pointing to the official repo "by accident" as part of a non-release-build action?

coryrc:
> > I think it is intentional so the release process doesn't become messed up by accident if someone leaves an ENV variable set wrong.  I think a better way, if --release-gcr doesn't work, is add 
> 
> If a "normal" dev has write access to the release repo then it seems like there are other security issues since any run of `ko` could cause harm then.

It's more like if the release breaks and somebody has to end up running things manually, they can break things (and we have) because humans aren't perfect.

> The other scripts that I've had to use to do builds appear to use KO_DOCKER_REPO as a way to tell it where to put images (e.g. serving does that). 

Everything but release does work that way. But the design is intentional in specifying everything explicitly just to avoid accidents.

> > I think it is intentional so the release process doesn't become messed up by accident if someone leaves an ENV variable set wrong.
> 
> thinking more about this... isn't the opposite more likely? Meaning, someone is pointing to a personal repo when they run the release script? Why would someone be pointing to the official repo "by accident" as part of a non-release-build action?

Correct, but we don't want to mess up anything else either! If we're running release scripts manually, something has probably already gone wrong and I'd rather it make me confirm all the parameters it is going to use than silently take them.


Hey, what about an option --use-env-ko-docker-repo ?

duglin:
> Hey, what about an option --use-env-ko-docker-repo ?

I think it's weird to require the user to add a "yes I know what I'm doing" flag, but at this point I just need a way to do a build so ok.

scripts/release.sh
```diff
@@ -383,6 +384,9 @@ function parse_flags() {
             [[ $1 =~ ^v[0-9]+-[0-9a-f]+$ ]] || abort "nightly tag must be 'vYYYYMMDD-commithash'"
             FROM_NIGHTLY_RELEASE=$1
             ;;
+          --use-env-ko-docker-repo)
+            USE_KO_DOCKER_REPO=$1
```
coryrc: Have to remove this

scripts/release.sh
```diff
@@ -437,6 +445,9 @@ function parse_flags() {
   [[ -n "${RELEASE_VERSION}" ]] && TAG="v${RELEASE_VERSION}"
   [[ -n "${RELEASE_VERSION}" && -n "${RELEASE_BRANCH}" ]] && (( PUBLISH_RELEASE )) && PUBLISH_TO_GITHUB=1
 
+  # If not already set then go ahead and default to gcr
+  export KO_DOCKER_REPO="${KO_DOCKER_REPO:-gcr.io/knative-nightly}"
+
```
coryrc: Have to protect this with if NOT (( USE_KO_DOCKER_REPO )) then export KO_DOCKER_REPO="gcr.io/knative-nightly" done

duglin: I think this needs to be before the check for PUBLISH_RELEASE, so I moved it but check me on that...

scripts/release.sh
```diff
@@ -349,6 +349,7 @@ function parse_flags() {
       --dot-release) is_dot_release=1 ;;
       --auto-release) is_auto_release=1 ;;
       --from-latest-nightly) FROM_NIGHTLY_RELEASE=latest ;;
+      --use-ko-docker-repo) USE_KO_DOCKER_REPO=1 ;;
```
adrcunha: nit: why not just `--docker-repo my-docker-repo.foo`?

duglin: do you mean change the flag name from `use-ok-docker-repo` to `docker-repo` and pass in the repo name there? If so, I guess, but now there are two different ways to specify the location and that seems even more confusing than a flag that means "I really mean it".

scripts/release.sh
adrcunha: Makes sense. But now this line is unnecessary.

scripts/release.sh
```diff
@@ -417,12 +418,21 @@ function parse_flags() {
...
+    export KO_DOCKER_REPO="gcr.io/knative-nightly"
+  fi
+
   # Update KO_DOCKER_REPO and KO_FLAGS if we're not publishing.
   if (( ! PUBLISH_RELEASE )); then
     (( has_gcr_flag )) && echo "Not publishing the release, GCR flag is ignored"
     (( has_gcs_flag )) && echo "Not publishing the release, GCS flag is ignored"
-    KO_DOCKER_REPO="ko.local"
-    KO_FLAGS="-L ${KO_FLAGS}"
+    if (( USE_KO_DOCKER_REPO )); then
```
adrcunha: Move the KO_DOCKER_REPO initialization to line 424, then here you just need to force it to ko.local if USE_KO_DOCKER_REPO is 0, like

```
(( ! USE_KO_DOCKER_REPO )) && KO_DOCKER_REPO="ko.local"
```

scripts/release.sh
adrcunha: Makes sense, ignore my comment.

scripts/release.sh
duglin: okie dokie

scripts/release.sh
duglin: doi :-)

scripts/release.sh
duglin: wasn't sure whether 424 was in the old vs new code.... check to see if I put it where you wanted

scripts/release.sh
adrcunha: My apologies, I should have read this PR more thoroughly. So I see that what you want is to be able to specify a custom location when you build your test release. So you don't need a new flag, you just need to call:

./hack/release.sh --release-gcr gcr.io/my-gcr --release-gcs my-gcs --notag-release --publish

You can add --skip-tests if you don't want to run all tests (unit+e2e) before building the release.

duglin: what do I put for "my-gcs"?  When I replaced the --release-gcr flag with my private repo and left "my-gcs" as is, it didn't seem to do anything, no build, no push.  And yet it also didn't give an error.

adrcunha: my-gcs is any GCS bucket you have write access to. From serving HEAD, if I run the command above (with --skip-tests) I do get the images published to my GCR and the manifests copied to my bucket. In the output, I see the banners "Building the release" (and the expected output from ko) and also "New release published successfully" at the end. I think we're using different scripts. Let's debug this offline on Slack.

duglin: I'm not using GCS - that's part of the point of this PR. We need to allow people to build Knative outside of Google.

duglin: Ideally, I think it would be nice if scripts like this took two input optional parameters:
1 - the registry in which to push the build images (e.g. KO_DOCKER_REPO, default to local docker)
2 - the dir in which to place all files needed to do an install (default to some `install_files` type of dir).

That gives people who want to build knative themselves the easiest path to getting all of the information they need in very well defined (local) locations, with minimal effort.

If after that some wrapper script wanted to push the files to some company specific location (like GCS) they can do so, but I think the core script should be generic and easy for everyone to run.

I'm not against this script knowing how to talk to company specific components (like GCS) but that should require a flag to do so - the default should be the easiest... store the files in a dir.

adrcunha: Please note that this script was created to publish Knative, and that currently relies on GCP. That said, I understand and agree that we can add features to make the script store the manifests locally instead of a GCS bucket, that's simple.

adrcunha: https://github.com/knative/test-infra/pull/1554 should give you the feature you want.

duglin:
@adrcunha correct me if I'm wrong, but I believe that https://github.com/knative/test-infra/pull/1554 should cover this one, right?  I'm gonna close this with that assumption. Let me know if I got it wrong.

PR #1546, opened 2019-11-14
Replace the shell script cleanup tool with a go equivalent

adrcunha:
The go tool cleans up registries in parallel, thus immensely speeding up the process (4h x 30h).

Run the tool every 2 days to avoid quota errors, since now we have 100 boskos projects (https://cloud.google.com/container-registry/quotas)

Basic testing is in place, although the tool was tested manually. Further required unit testing (currently noted as TODOs) will be added in a future PR, as I would like to have this running and cleaning up the projects before something breaks. :)

Part of #1399 

/assign @chaodaiG


adrcunha:
I can't make the build tests pass cleanly, even though they pass locally. Closing this and creating the same PR again as https://github.com/knative/test-infra/pull/1549

/close

PR #1547, opened 2019-11-14
updated the monitor config to remove noise

yt3liu:

Update the config to remove noise. 

Max failure from Mid-October to November for the below errors: 
* Knative setup failed: 33
* test setup failed: 2
* Something went wrong: starting e2e cluster: error creating cluster: 26
* ERROR: timeout waiting for pods to come up: 7 

Fixes #

### DIFF
--- a/tools/monitoring/config/config.yaml
+++ b/tools/monitoring/config/config.yaml
@@ -13,20 +13,6 @@
 # limitations under the License.
 
 spec:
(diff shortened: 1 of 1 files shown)


chizhg:
I believe the biggest noise is from the `.*` error pattern. Can we completely remove it (the last config)? Performance testing jobs will not need it since it's already been monitored by Stackdriver.

PR #1549, opened 2019-11-15
Replace the shell script cleanup tool with a go equivalent

### LINKED ISSUE #1399: Improve efficiency of test resources cleanup tool

adrcunha:
The go tool cleans up registries in parallel, thus immensely speeding up the process (4h x 30h).

Run the tool every 2 days to avoid quota errors, since now we have 100 boskos projects (https://cloud.google.com/container-registry/quotas)

Basic testing is in place, although the tool was tested manually. Further required unit testing (currently noted as TODOs) should be added in a future PR, as I would like to have this running and cleaning up the projects before something breaks. :)

Fixes #1399

/assign @chaodaiG


tools/cleanup/cleanup.go
```diff
@@ -0,0 +1,308 @@
...
+
+	log.Printf("All operations finished in %s", time.Now().Sub(start))
+
+	return nil
+}
+
+// main is the script entry point.
+func main() {
+	flag.Parse()
+	if err := cleanup(); err != nil {
```
chaodaiG: IMHO, to make it easier to implement unit tests, it would be helpful to use something OOP like style, so that the actual functions that need network can be easily mocked, and all functions other than main can have unit tests. something in my mind looks like:
```
imageClient := newImageClient(sa)
imageClient.Run(option. daysToKeepImages...)
...
```

tools/cleanup/cleanup.go
```diff
@@ -0,0 +1,308 @@
...
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/pkg/errors"
+	"google.golang.org/api/option"
+	"knative.dev/test-infra/shared/common"
+
+	"knative.dev/pkg/test/gke"
+)
+
+// resourceDeleter is a function that deletes a particular type of resource, and return how many resources were deleted.
+type resourceDeleter func(string, time.Time) (int, error)
```
chaodaiG: feel like can be a interface

tools/cleanup/cleanup.go
```diff
@@ -0,0 +1,308 @@
...
+	count := 0
+	for _, cluster := range clusters {
+		creation, err := time.Parse(time.RFC3339, cluster.CreateTime)
+		if err != nil {
+			return count, errors.Wrapf(err, "error getting creation time for cluster %q", cluster.Name)
+		}
+		age := int(time.Since(creation).Hours())
+		log.Printf("%s/%s is %d hours old", project, cluster.Name, age)
+		if creation.Before(before) {
+			if *dryRun {
```
chaodaiG: this dryRun logic can be refactored out or maybe use this one: https://github.com/knative/pkg/blob/master/test/helpers/dryrun.go

tools/cleanup/cleanup.go
```diff
@@ -0,0 +1,308 @@
...
+			countLock.Unlock()
+			if err != nil {
+				cause := errors.Cause(err).Error()
+				errsLock.Lock()
+				if _, ok := errs[cause]; !ok {
+					errs[cause] = err
+				}
+				errsLock.Unlock()
+			}
+			<-concurrencyChan
```
chaodaiG: this is not guaranteed to run every time, failures would block the channel, is it intended?

tools/cleanup/cleanup.go
```diff
@@ -0,0 +1,308 @@
...
+	for i := range projects {
+		wg.Add(1)
+		go func(project string) {
+			defer wg.Done()
+			// Block processing if concurrency reached the limit.
+			concurrencyChan <- struct{}{}
+			// Do not process if previous invocations failed. This prevents a large
+			// build-up of failed requests and rate limit exceeding (e.g. bad auth).
+			errsLock.RLock()
+			failed := len(errs) > 0
```
chaodaiG: so a single failure fails all following events?

tools/cleanup/cleanup.go
```diff
@@ -0,0 +1,308 @@
...
+	errStrings := make([]string, 0)
+	for i := range errs {
+		errStrings = append(errStrings, errs[i].Error())
+	}
+	return count, errStrings
+}
+
+// showStats displays basic stats based on the return values from deleteResources.
+func showStats(count int, errs []string) {
+	log.Printf("%d resources deleted", count)
```
chaodaiG: not sure if count is needed if it's just for displaying how many resources deleted, does it worth throttling with another lock with just this information?

tools/cleanup/cleanup.go
```diff
@@ -0,0 +1,308 @@
...
+				return
+			}
+			c, err := deleteFunc(project, before)
+			// Update counter and errors list.
+			countLock.Lock()
+			count += c
+			countLock.Unlock()
+			if err != nil {
+				cause := errors.Cause(err).Error()
+				errsLock.Lock()
```
chaodaiG: how about sending to an unbufferred channel instead of a lock?

tools/cleanup/cleanup.go
```diff
@@ -0,0 +1,308 @@
...
+	defer close(concurrencyChan)
+
+	count := 0
+	for i := range projects {
+		wg.Add(1)
+		go func(project string) {
+			defer wg.Done()
+			// Block processing if concurrency reached the limit.
+			concurrencyChan <- struct{}{}
+			// Do not process if previous invocations failed. This prevents a large
```
chaodaiG: I think there is a race condition here, since this for loop schedule goroutines without any wait time in between, all of the projects probably would start at the same time, and they lined up waiting for `errsLock` then enter the next phase one right after another, so when the first goroutine hit an error some of the other goroutines probably already passed this check. 
How about make it wait for a few seconds before scheduling the next project? Or just wait when `i == 1`?

tools/cleanup/cleanup.go
```diff
@@ -0,0 +1,308 @@
...
+			// build-up of failed requests and rate limit exceeding (e.g. bad auth).
+			errsLock.RLock()
```
chaodaiG: using an errChannel maybe? by this way we can do:
```
for i := range projects {
  // If errChannel is not empty: return
```

tools/cleanup/cleanup.go
adrcunha: Yes. As explained in the comment above, we want to avoid an avalanche of errors (usually the quota error due to the amount of projects).

tools/cleanup/cleanup.go
adrcunha: It's not needed, but it's a nice to have statistics. Deleting a resource is several orders of magnitude slower than a simple locking, the throttling you're concerned about is negligible.

tools/cleanup/cleanup.go
adrcunha: Your suggested wait will throttle much more than the counter lock above. :D I'm not concerned about this case, because the idea is to avoid a flooding of errors, it's OK if a few routines try the resource deletion even if a failure occurred (it won't break the tool or the process).

adrcunha:
I confess I wasn't expecting such a simple tool to have its code being more complex than necessary, but since this is not blocking anything and this is not a high priority for me I'll implement your suggestions as my spare time permits. I'll ping you once this is ready for review. Feel free to add more comments as you see fit though.

tools/cleanup/cleanup.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+	gcrrepo, err := name.NewRepository(repoRoot)
+	if err != nil {
+		return 0, errors.Wrapf(err, "cannot open registry %q", repoRoot)
+	}
+	count := 0
+	// Walk down the registry, checking all images and deleting the old ones.
+	return count, google.Walk(gcrrepo, func(repo name.Repository, tags *google.Tags, err error) error {
+		// If we got an error, just return it, there's nothing to do here.
+		if tags == nil {
```
chizhg: Should it be `if err != nil` ?

tools/cleanup/cleanup.go
```diff
@@ -0,0 +1,308 @@
...
+					errs[cause] = err
+				}
+				errsLock.Unlock()
+			}
+			<-concurrencyChan
+		}(projects[i])
+	}
+	wg.Wait()
+	// Extract the error strings from the map of errors.
+	errStrings := make([]string, 0)
```
chizhg: Change to `make([]string, len(errs))` ? Then line 235 will need to be changed to `errStrings[i] = errs[i].Error()`.

tools/cleanup/cleanup.go
```diff
@@ -0,0 +1,308 @@
...
+	for i := range errs {
```
chizhg: The key is a string, and `i` is used to represent an integer, probably change it to `s`?

tools/cleanup/cleanup.go
```diff
@@ -0,0 +1,308 @@
...
+	// Select projects, perform operations.
+	var projects []string
+	if projects, err = selectProjects(*project, *projectResourceYaml, *reProjectName); err != nil {
+		return err
+	}
+	start := time.Now()
+
+	if *daysToKeepImages >= 0 {
+		log.Println("Removing images that are:")
+		log.Printf("- older than %d days", *daysToKeepImages)
```
chizhg:
```suggestion
		log.Printf("- older than %d days\n", *daysToKeepImages)
```

tools/cleanup/cleanup.go
```diff
@@ -0,0 +1,308 @@
...
+		showStats(deleteResources(projects, start.Add(-24*time.Hour*time.Duration(*daysToKeepImages)), deleteImages))
+	}
+
+	if *hoursToKeepClusters >= 0 {
+		log.Println("Removing clusters that are:")
+		log.Printf("- older than %d hours", *hoursToKeepClusters)
```
chizhg:
```suggestion
		log.Printf("- older than %d hours\n", *hoursToKeepClusters)
```

tools/cleanup/cleanup.go
adrcunha: Unnecessary, log entries are one per line.

tools/cleanup/cleanup.go
adrcunha: Unnecessary, log entries are one per line.

tools/cleanup/cleanup.go
adrcunha: Done.

tools/cleanup/cleanup.go
adrcunha: Good catch, fixed.

tools/cleanup/cleanup.go
adrcunha: Good catch, done.

tools/cleanup/cleanup.go
adrcunha: Actually this code had a bug and the order wasn't constant, fixed.

tools/cleanup/cleanup.go
adrcunha: Indeed. Done.

tools/cleanup/cleanup.go
```diff
@@ -0,0 +1,308 @@
...
+
+	"knative.dev/pkg/test/gke"
+)
+
+// resourceDeleter is a function that deletes a particular type of resource, and return how many resources were deleted.
+type resourceDeleter func(string, time.Time) (int, error)
+
+var (
+	// Command-line flags.
+	projectResourceYaml  = flag.String("project-resource-yaml", "", "Resources file containing the names of the projects to be cleaned up.")
```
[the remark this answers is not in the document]
adrcunha: Passed options as parameters to functions, as using a single struct to pass around unrelated parameters is a bit awkward. That should also benefit testing.

tools/cleanup/cleanup.go
adrcunha: Good idea, it's cleaner. Done.

tools/cleanup/cleanup.go
adrcunha: Done.

tools/cleanup/cleanup.go
```diff
@@ -0,0 +1,308 @@
...
+	return projects, nil
+}
+
+// deleteImage deletes a single GCR image pointed by the given reference.
+func deleteImage(ref string) error {
+	image, err := name.ParseReference(ref)
+	if err != nil {
+		return errors.Wrapf(err, "failed to parse reference %q", ref)
+	}
+	if err := remote.Delete(image, remote.WithAuthFromKeychain(auther)); err != nil {
```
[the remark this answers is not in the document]
adrcunha: This means increasing the complexity of the code and making it harder to read just to make mocking easier. I would prefer using a global variable pointing to the method, so I can override it in tests. WDYT?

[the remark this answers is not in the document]
chaodaiG: that'll work for me, as long as it's easy to write unit test

tools/cleanup/cleanup.go
adrcunha: Handled this in the comment above about using an OOP style.

tools/cleanup/cleanup.go
adrcunha: Made the suggested change to the approach. Honestly I'm not convinced it makes mocking easier since it's just 2 "classes". PTAL and LMK if that's what you had in mind. The code is untested, but I don't want to invest more time on it before making sure the changes do address your feedback. Once they do, I'll make sure the code works and update the tests.

chaodaiG: Thanks  for making the change, this is very close to what I had in mind. Will lgtm once the code is tested

shared/common/common.go
```diff
@@ -42,14 +43,21 @@ func CreateDirWithFileMode(dirPath string, perm os.FileMode) error {
 	return nil
 }
 
+// ExecCommand run the given command and returns the output
+func ExecCommand(cmd string, args ...string) (string, string, error) {
```
chizhg: This can now be replaced with the `cmd` lib we have in `pkg/test` - https://github.com/knative/pkg/blob/master/test/cmd/command.go, if you like.

adrcunha:
> Do you think if it is fine to leave them (it seems there will not be much cost as no traffic will be sent to the topics)? Or shall we also support cleaning them up in this tool?

For the sake of simplicity and productivity (this PR is already 2 month old), and since they don't count against quota, I would say leave it and implement a cleaner if necessity arises.

shared/common/common.go
adrcunha: Thanks. Done.

adrcunha:
@chaodaiG this is finally ready for another review.

chaodaiG:
/approve

Defer to @chizhg apply lgtm since he had question also

chizhg:
/lgtm
Thanks!

PR #1554, opened 2019-11-20
Allow publishing manifests to a local dir

adrcunha:
Use `--release-dir xxx` for that (as opposed to `--release-gcs yyy`).

/cc @duglin

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -93,9 +93,9 @@ RELEASE_VERSION=""
 RELEASE_NOTES=""
 RELEASE_BRANCH=""
 RELEASE_GCS_BUCKET="knative-nightly/${REPO_NAME}"
+RELEASE_DIR=""
 KO_FLAGS="-P"
 VALIDATION_TESTS="./test/presubmit-tests.sh"
-YAMLS_TO_PUBLISH=""
 ARTIFACTS_TO_PUBLISH=""
 FROM_NIGHTLY_RELEASE=""
 FROM_NIGHTLY_RELEASE_GCS=""
@@ -334,6 +334,7 @@ function find_latest_nightly() {
 function parse_flags() {
   local has_gcr_flag=0
   local has_gcs_flag=0
+  local has_dir_flag=0
   local is_dot_release=0
   local is_auto_release=0
 
@@ -365,7 +366,13 @@ function parse_flags() {
             ;;
           --release-gcs)
             RELEASE_GCS_BUCKET=$1
+            RELEASE_DIR=""
             has_gcs_flag=1
+            ;;
+          --release-dir)
+            RELEASE_DIR=$1
+            RELEASE_GCS_BUCKET=""
+            has_dir_flag=1
             ;;
           --version)
             [[ $1 =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]] || abort "version format must be '[0-9].[0-9].[0-9]'"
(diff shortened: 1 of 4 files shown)


scripts/release.sh
```diff
@@ -458,22 +472,24 @@ function parse_flags() {
...
 
-# Publishes the generated artifacts to GCS, GitHub, etc.
+# Publishes the generated artifacts to directory, GCS, GitHub, etc.
 # Parameters: $1..$n - files to add to the release.
 function publish_artifacts() {
   (( ! PUBLISH_RELEASE )) && return
   tag_images_in_yamls ${ARTIFACTS_TO_PUBLISH}
-  publish_to_gcs ${ARTIFACTS_TO_PUBLISH}
+  if [[ -n "${RELEASE_DIR}" ]]; then
+    cp ${ARTIFACTS_TO_PUBLISH} ${RELEASE_DIR} || abort "cannot copy release to '${RELEASE_DIR}'"
```
duglin: where do the files end up by default? I'd prefer if they appeared only in the destination dir so that 1) we don't run the risk of them over-writing existing files, and 2) we don't leave them around by using a `cp` instead of `mv`.  And switching this to a `mv` wouldn't be good since then the net result would be that we erased a pre-existing file with the same name.

duglin:
Does this fully replace https://github.com/knative/test-infra/pull/1544  ?

scripts/release.sh
adrcunha: It's up to the caller of this script to define where the manifests are written to. Specifically for serving, they're written to a temporary directory (https://github.com/knative/serving/blob/bcc67e3b34b2e6d1f9e567d1fb53aeddc4d173db/hack/generate-yamls.sh#L46). So in the end, at least for serving, we're achieving the goal you desire.

adrcunha:
Friendly ping @duglin.

scripts/release.sh
```diff
@@ -422,10 +432,13 @@ function parse_flags() {
     (( has_gcr_flag )) && echo "Not publishing the release, GCR flag is ignored"
     (( has_gcs_flag )) && echo "Not publishing the release, GCS flag is ignored"
     KO_DOCKER_REPO="ko.local"
-    KO_FLAGS="-L ${KO_FLAGS}"
     RELEASE_GCS_BUCKET=""
+    [[ -z "${RELEASE_DIR}" ]] && RELEASE_DIR="${REPO_ROOT_DIR}"
   fi
 
+  [[ -z "${RELEASE_GCS_BUCKET}" && -z "${RELEASE_DIR}" ]] && abort "--release-gcs or --release-dir must be used"
+  [[ -n "${RELEASE_DIR}" ]] && mkdir -p "${RELEASE_DIR}" 2> /dev/null
```
duglin: I think it might be better to remove the `2> /dev/null` since if the dir already exists then `mkdir` remains silent due to the `-p`, so the only time we should see stderr text is when something is really bad, like we don't have permission to create the dir - which should be shown to the user (and probably stop processing)

duglin:
Sorry for the delay... travels and then vacation.

Just one comment about `2>` but aside from that it looks good to me. I haven't had a chance to test it though in our env.  thanks!

scripts/release.sh
adrcunha: Good catch. Fixed.

duglin:
@adrcunha is this one ready to go?

adrcunha:
Yes, it is.

duglin:
cool - wanna merge it?

adrcunha:
Needs LGTM.

duglin:
@adrcunha which flags do I need to set so that it picks up the KO_DOCKER_REPO value and uses that instead of a gcr or ko.local ?

PR #1558, opened 2019-11-25
Use prow-tests image with go1.13 for all eventing repos

chaodaiG:
Use prow-tests image with go1.13 for all eventing repos, so that we can migrate pkg to use go1.13

/cc @chizhg 
FYI @n3wscott @markusthoemmes

### DIFF
--- a/ci/prow/config_knative.yaml
+++ b/ci/prow/config_knative.yaml
@@ -81,6 +81,10 @@ presubmits:
       - "./test/e2e-tests.sh --https"
 
   knative/client:
+    - repo-settings:
+      go112-branches:
+      - release-0.9
+      - release-0.10
     - build-tests: true
       dot-dev: true
     - unit-tests: true
@@ -173,6 +177,10 @@ presubmits:
       dot-dev: true
 
   knative/caching:
+    - repo-settings:
+      go112-branches:
+      - release-0.9
+      - release-0.10
     - build-tests: true
       dot-dev: true
     - unit-tests: true
--- a/ci/prow/config_knative.yaml
+++ b/ci/prow/config_knative.yaml
@@ -101,6 +101,11 @@ presubmits:
       legacy-branches:
       - release-0.7
       - release-0.8
+      go112-branches:
+      - release-0.7
+      - release-0.8
+      - release-0.9
+      - release-0.10
     - build-tests: true
       resources:
         requests:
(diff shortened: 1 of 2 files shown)


chizhg:
/hold

I have two questions regarding this change:
- Do we also need to make changes to Knative test-infra, client and caching? These repos are also vendoring pkg.
- Do we have a way to test this config update, or do we need to?

chaodaiG:
> /hold
> 
> I have two questions regarding this change:
> 
> * Do we also need to make changes to Knative test-infra, client and caching? These repos are also vendoring pkg.
> * Do we have a way to test this config update, or do we need to?

good point. Updated with client and caching, not sure about test-infra, planning to do it at the same time with pkg

We do have a way to test this update, but not a very elegant way

ci/prow/config_knative.yaml
```diff
@@ -173,6 +177,10 @@ presubmits:
       dot-dev: true
 
   knative/caching:
+    - repo-settings:
+      go112-branches:
+      - release-0.9
```
chizhg: caching does not have these two releases. Will leaving it empty work for the config generator?

chaodaiG:
Not that great, it doesn't hurt IMO

chizhg:
I have manually run `test/e2e-tests.sh` for eventing with Go 1.13, and it's working.

But client will fail due to https://github.com/knative/client/blob/e07b5a98b50120c2da7a228705d403f31d1f1a43/test/e2e/e2e_flags.go#L25

chaodaiG:
> I have manually run `test/e2e-tests.sh` for eventing with Go 1.13, and it's working.
> 
> But client will fail due to https://github.com/knative/client/blob/e07b5a98b50120c2da7a228705d403f31d1f1a43/test/e2e/e2e_flags.go#L25 , which is one change I made a long time ago. I'm not sure about the history of this `--emit-metrics` flag or if it's still useful, but it's required by all CI jobs...

thanks for the investigation, let's fix client first
/hold

adrcunha:
> I have manually run `test/e2e-tests.sh` for eventing with Go 1.13, and it's working.
> 
> But client will fail due to https://github.com/knative/client/blob/e07b5a98b50120c2da7a228705d403f31d1f1a43/test/e2e/e2e_flags.go#L25 , which is one change I made a long time ago. I'm not sure about the history of this `--emit-metrics` flag or if it's still useful, but it's required by all CI jobs...

This isn't used anymore. Let's kill it. Filed https://github.com/knative/test-infra/issues/1568

chizhg:
I believe now the Prow jobs can be updated.
/hold cancel
/lgtm
/cc @yt3liu

chaodaiG:
Thanks Chi!

PR #1559, opened 2019-11-26
Enable Istio 1.4 in pull-knative-serving-istio-1.4-* test grids and remove 1.2

nak3:
This is similar request with https://github.com/knative/serving/issues/5656 and https://github.com/knative/test-infra/pull/1394.

This PR has following two changes:

[x] Enable Istio 1.4.0 as it is already included by https://github.com/knative/serving/pull/6097
[x] Remove Istio 1.2.x as it will be EOL on December 13th/2019
[x] Use Istio ~~1.4~~ 1.3 for pull-knative-serving-integration-tests
/lint

### DIFF
--- a/ci/prow/config.yaml
+++ b/ci/prow/config.yaml
@@ -404,7 +404,7 @@ presubmits:
         args:
         - "./test/presubmit-tests.sh"
         - "--run-test"
-        - "./test/e2e-tests.sh --istio-version 1.2-latest --no-mesh"
+        - "./test/e2e-tests.sh --istio-version 1.4-latest --no-mesh"
         volumeMounts:
         - name: repoview-token
           mountPath: /etc/repoview-token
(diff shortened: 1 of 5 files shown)


ci/prow/config.yaml
```diff
@@ -531,7 +531,7 @@ presubmits:
         args:
         - "./test/presubmit-tests.sh"
         - "--run-test"
-        - "./test/e2e-upgrade-tests.sh --istio-version 1.2-latest --no-mesh"
+        - "./test/e2e-upgrade-tests.sh --istio-version 1.4-latest --no-mesh"
```
nak3: I was not sure whether we should use Istio 1.3 or 1.4 for pull-knative-serving-integration-tests.
I think latest (Istio 1.4) is better, but I will change it to 1.3 if we should preserve 1.3.

nak3: Never mind. 1.3 should be used, I think.

chaodaiG:
/lgtm
/approve

I'm oncall this week, will apply them

PR #1562, opened 2019-12-02
Add prow config for knative/sample-source

lberk:
Needed prow config for [knative/sample-source](https://github.com/knative/sample-source)

Fixes #

@chizhg PR as requested on slack.

### DIFF
--- a/ci/prow/config_knative.yaml
+++ b/ci/prow/config_knative.yaml
@@ -177,6 +177,12 @@ presubmits:
     - unit-tests: true
       dot-dev: true
 
+  knative/sample-source:
+    - build-tests: true
+      dot-dev: true
+    - unit-tests: true
+      dot-dev: true
+
   google/knative-gcp:
     - repo-settings:
       performance: true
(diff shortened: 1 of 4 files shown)


ci/prow/config_knative.yaml
```diff
@@ -177,6 +177,12 @@ presubmits:
     - unit-tests: true
       dot-dev: true
 
+  knative/sample-source:
```
chizhg: Please also add CI jobs for it, as what we have for `sample-controller` - https://github.com/knative/test-infra/blob/2c4cd9df33c648a86d95bdd4119386cba04cde56/ci/prow/config_knative.yaml#L416.

It does not hurt for not having it though, but it's better to follow the same pattern.

lberk: Ah sorry, thought that might've been what was meant by the guide saying to not setup presubmit tests just yet.  Will add it now.

lberk: added in a80e114

PR #1577, opened 2019-12-06
Increase cpu request so that we don't overscheduling

chaodaiG:
CPU request was 1 and limit was 4, which results in overscheduling of pods competing for resources, and when a node is busy containers are competing for cpus, this makes some of serving unit test fail due to timeout, this has been repro'ed with KIND on my devbox. Double requested cpu to avoid overscheduling and preventing test from being flaky

Related: https://github.com/knative/serving/issues/6127

/cc @adrcunha 
/cc @chizhg

### DIFF
--- a/ci/prow/deployments/cpu_limitrange.yaml
+++ b/ci/prow/deployments/cpu_limitrange.yaml
@@ -22,5 +22,5 @@ spec:
     - default:
         cpu: 4000m
       defaultRequest:
-        cpu: 1000m
+        cpu: 2000m
       type: Container


chizhg:
> CPU request was 1 and limit was 4, which results in overscheduling of pods competing for resources, and when a node is busy containers are competing for cpus, this makes some of serving unit test fail due to timeout, this has been repro'ed with KIND on my devbox. Double requested cpu to avoid overscheduling and preventing test from being flaky
> 
> Related: [knative/serving#6127](https://github.com/knative/serving/issues/6127)
> 
> /cc @adrcunha
> /cc @chizhg

/hold

This change will double the CPU requests for all Prow jobs. Can we just increase it for serving unit tests job?

chaodaiG:
First of all, it doesn't make sense to has cpu request being 1 and limit of 4, this is too far away, I'm ok with cpu request being 1 and limit being 2, and that probably would work

chizhg:
> First of all, it doesn't make sense to has cpu request being 1 and limit of 4, this is too far away, I'm ok with cpu request being 1 and limit being 2, and that probably would work

"cpu request being 1 and limit being 2" may not be able to solve this problem, as serving unit tests job can probably need CPUs between 2 ~ 4.
"cpu request being 2 and limit being 4" means we may need to increase Prow cluster size.
Can we make this change early next week since it's Friday? 😅

chaodaiG:
haha, totally make sense. Let's do it next week, I agree that we will need more nodes, might double what we have now

Issue #1533, opened 2019-11-08
Intermittently getting "boskos failed to acquire project" error


chizhg:
For the integration tests Prow jobs, we intermittently got "boskos failed to acquire GKE project" error:
```
2019/11/07 02:47:45 failed acquiring GKE cluster: 'failed acquiring boskos project: 'boskos failed to acquire GKE project: status 500 Internal Server Error, status code 500'
```

Expanding boskos pool to 100 (https://github.com/knative/test-infra/pull/1529) does not seem to have solved this problem. 

Based on the discussion, @chaodaiG has a suspicion that it's because there is a race condition for compute resources between boskos pods and test pods since they are living in the same node pool of the Prow cluster, thus separating them into different node pools should help solve this problem. Before working on this solution, we need to collect relevant metrics to justify it.

chaodaiG:
This is getting worse, it's happen pretty frequently now, see example: https://testgrid.knative.dev/serving#istio-1.4-no-mesh&width=20&exclude-non-failed-tests=50&sort-by-failures=

We should prioritize investigating and fixing this. 

/assign @yt3liu

chizhg:
The boskos image update did not fix the issue.

After looking into boskos source code, I'm pretty sure it falls into  https://github.com/kubernetes/test-infra/blob/0c5566a3f019399377910bed349c2c280f5539e8/boskos/ranch/ranch.go#L155 when the error happens. One step further, it is https://github.com/kubernetes/test-infra/blob/0c5566a3f019399377910bed349c2c280f5539e8/boskos/crds/crd_storage.go#L52.

The keep error log is
```
{"error":"Operation cannot be fulfilled on resources.boskos.k8s.io \"knative-boskos-03\": the object has been modified; please apply your changes to the latest version and try again","level":"error","msg":"No available resource","time":"2019-12-11T06:24:20Z"}
```

Let's look it deeper tomorrow.

chaodaiG:
I figured this out too, looking into more info

chaodaiG:
When I check this morning, the Boskos container in Prow cluster was still using the old image. There are 2 observations:
- Boskos is configured as statefulset, which needs more manual step adopting new image. See config here: https://github.com/knative/test-infra/blob/master/ci/prow/deployments/boskos_statefulset.yaml
- Boskos in k8s/test-infra was switched to deployment instead of statefulset in this PR: https://github.com/kubernetes/test-infra/pull/13594

yt3liu:
[...]

PR #1584, opened 2019-12-11
Switching Boskos to deployment, as well as removing persistent volume

chaodaiG:
This change is to make our Boskos deployment consistent with upstream, almost a mirror of https://github.com/kubernetes/test-infra/pull/13594

Also copied from the PR above:
```
kubectl delete -n test-pods statefulset boskos
# wait for boskos-0 pod to be gone in test-pods ns
kubectl delete -n test-pods pv boskos-storage
kubectl apply -f ci/prow/deployments/boskos_deployment.yaml
```

Part of: #1533 

/cc @yt3liu 
/cc @sebastienvas


chaodaiG:
/hold
So that this PR doesn't get merged at a wrong time

ci/prow/deployments/boskos_volumes.yaml
```diff
@@ -1,34 +0,0 @@
...
-#
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-
-# Boskos deployment for Knative Prow instance
-apiVersion: v1
-kind: PersistentVolume
```
adrcunha: If you're removing this file, when creating a new cluster (or recreating boskos), won't it miss the configuration of the persistent volume used by boskos?

adrcunha: Wait... So the new boskos doesn't need persistent storage anymore?

chaodaiG: yeah that's right, I have double checked with the author of the PR in k8s/test-infra, Boskos has stopped using persistent storage about a year ago, and we are already using it. So these are just like cleaning up

adrcunha: Awesome. Let's delete that volume then once this is in. :)

/lgtm
/approve

PR #1585, opened 2019-12-11
Change all Prow jobs running on master to Go 1.13

chizhg:
At this point, we should be able to use Go 1.13 in all Prow jobs targeted to the master branch.

BTW, since release-0.11 has been cut, there should be some further changes regarding that new release branch, but I would prefer to make the change in a followup PR.

/cc @chaodaiG 
/cc @adrcunha 
/cc @yt3liu


adrcunha:
We should cleanup the go113 settings (i.e., make them default) in a follow up PR.

/lgtm

chizhg:
Friendly ping.
Can I get an approve for this PR?

PR #1594, opened 2019-12-17
Add labels "proposal/0.[11-19]" to Prow's /label plugin

### LINKED ISSUE #1580: Add label config to Prow
Currently `/label` understand only the default labels. From its help:

The label plugin provides commands that add or remove certain types of labels. Labels of the following types can be manipulated: 'area/*', 'committee/*', 'kind/*', 'language/*', 'priority/*', 'sig/*', 'triage/*', and 'wg/*'. More labels can be configured to be used via the /label command.

Extra labels can be configured through config:

https://github.com/kubernetes/test-infra/blob/master/prow/plugins/label/label.go#L57

https://github.com/kubernetes/test-infra/blob/master/config/prow/plugins.yaml#L101

For example, eventing would like to use `proposal/*` labels through `/label`.

We can also consider unifying labels among all repos in GitHub like Kubernetes does: https://github.com/kubernetes/test-infra/blob/master/label_sync/labels.md

adrcunha:
Fixes #1580 

/cc @chizhg

### DIFF
--- a/ci/prow/templates/prow_plugins.yaml
+++ b/ci/prow/templates/prow_plugins.yaml
@@ -51,6 +51,19 @@ project_config:
           # TODO(grantr): replace with a new team eventing-contrib-project-maintainers
           repo_maintainers_team_id: 3012514 # knative-milestone-maintainers
 
+# Settings for the "label" plugin.
+label:
+  additional_labels:
+  - proposal/0.11
+  - proposal/0.12
+  - proposal/0.13
+  - proposal/0.14
+  - proposal/0.15
+  - proposal/0.16
+  - proposal/0.17
+  - proposal/0.18
+  - proposal/0.19
+
 # Plugins enabled for each repo.
 
 plugins:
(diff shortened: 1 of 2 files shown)


adrcunha:
image

adrcunha:
The /label plugin help text led me to believe that a generic proposal/* handler could be added. I filed https://github.com/kubernetes/test-infra/issues/15628 noticing this inconsistency, but to avoid blocking this issue I added the proposal labels from 0.11 to 0.19.

ci/prow/templates/prow_plugins.yaml
```diff
@@ -51,6 +51,19 @@ project_config:
           # TODO(grantr): replace with a new team eventing-contrib-project-maintainers
           repo_maintainers_team_id: 3012514 # knative-milestone-maintainers
 
+# Settings for the "label" plugin.
+label:
+  additional_labels:
+  - proposal/0.11
```
chizhg: Can we add a comment here saying "replace with proposal/* once https://github.com/kubernetes/test-infra/issues/15628 is solved"?

adrcunha: I thought about that, but it can be that 15628 could be closed as WAI. In this case, there's nothing to do here.

PR #1598, opened 2019-12-23
Prow job to run serving e2e against ambassador

concaf:
/lint

This PR adds support to run Knative Serving's e2e tests with Ambassador latest (currently 0.86.1)

This PR needs to be merged to run tests at https://github.com/knative/serving/pull/6268

None

### DIFF
--- a/ci/prow/config_knative.yaml
+++ b/ci/prow/config_knative.yaml
@@ -95,6 +95,13 @@ presubmits:
       args:
         - "--run-test"
         - "./test/e2e-tests.sh --contour-version latest"
+    - custom-test: ambassador-latest
+      dot-dev: true
+      always_run: false
+      optional: true
+      args:
+        - "--run-test"
+        - "./test/e2e-tests.sh --ambassador-version 0.86.1"
     - custom-test: https
       dot-dev: true
       always_run: false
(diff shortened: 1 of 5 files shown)


ci/prow/config_knative.yaml
```diff
@@ -79,6 +79,12 @@ presubmits:
       args:
         - "--run-test"
         - "./test/e2e-tests.sh --kourier-version stable"
+    - custom-test: ambassador-latest
+      always_run: false
+      optional: true
+      args:
+        - "--run-test"
+        - "./test/e2e-tests.sh --ambassador-version 0.86.1"
```
adrcunha: Use "latest" instead of 0.86.1, here and below. Currently e2e-tests.sh will ignore the version and install the one hardcoded in the code (so this can generate confusion). Also, you'll have two places to change the version number in the future.

concaf: Makes sense, fixed in separate commit. Thanks!

adrcunha:
It seems that you need to run `make config` again to sync up `config.yaml`.

adrcunha:
6268 is merged, this is now good for sync+merge+deploy.

concaf:
@adrcunha I've updated config.yaml using `make config`. Could you guide what's next? i.e. I want to see run tests in CI. Does that happen in serving's CI once this goes in?

chizhg:
I'll take care of updating the Prow config after this PR goes in.

/lgtm
/approve

PR #1615, opened 2020-01-09
Increase unit test run count to 5 to reduce flakiness

JRBANCEL:
`-count=1` is not enough to catch data races and other sources of flakiness in the unit tests.
This increases the test run count to 5 to increase the likelihood to catch those issues early on.

Ideally, it should be higher than 5, but we don't want to make the execution of the unit tests leg too long.

### DIFF
--- a/scripts/presubmit-tests.sh
+++ b/scripts/presubmit-tests.sh
@@ -231,7 +231,7 @@ function run_unit_tests() {
 
 # Default unit test runner that runs all go tests in the repo.
 function default_unit_test_runner() {
-  report_go_test -race ./...
+  report_go_test -race -count=5 ./...
 }
 
 # Run integration tests. If there's no `integration_tests` function, run the


adrcunha:
This is a global change, I'm not comfortable doing this without agreement from the leads.

/hold

JRBANCEL:
This was briefly discussed here: https://knative.slack.com/archives/CA1DTGZ2N/p1578604442010600

adrcunha:
Thanks for the pointer. I'll let Matt LGTM. Please note that, for this to work, you'll need to update test-infra in the other repos.

/hold cancel
/approve

mattmoor:
I'd like to see us move this direction, but for a global change it needs to be socialized more.  I'll drop it in the leads channel, to solicit feedback.

scripts/presubmit-tests.sh
```diff
@@ -231,7 +231,7 @@ function run_unit_tests() {
 
 # Default unit test runner that runs all go tests in the repo.
 function default_unit_test_runner() {
-  report_go_test -race ./...
+  report_go_test -race -count=5 ./...
```
matzew: :+1:

mattmoor:
Unless something comes up, let’s aim for Tuesday post cut?

JRBANCEL:
Sounds good.
I would like someone from EngProd to validate that a count > 1 won't break the tools (TestGrid, Prow, etc...).

chizhg:
/uncc @adrcunha 
/cc @chaodaiG 
/cc @coryrc 

If this PR goes in and works well, it should be able to help reduce the test flakiness for unit tests.

chaodaiG:
Seems like this discovered a badly written test in test-infra :) I'll take a look

chaodaiG:
It does seem like it's messing up with Prow in terms of tests counting, and it will mess up Testgrid as well. For example the unit test from this PR: https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/knative_test-infra/1615/pull-knative-test-infra-unit-tests/1215399931288752132

The junit result file looks like this:
Screen Shot 2020-02-03 at 2 18 20 PM

coryrc:
Why are we rerunning all the tests and not just the ones that failed?

JRBANCEL:
> Why are we rerunning all the tests and not just the ones that failed?

Because a successful run doesn't mean it is not flaky, just that we got lucky.
While N successful runs gives us more confidence in the non-flakiness of the test.

PR #1620, opened 2020-01-15
Change eventing release file name to eventing.yaml

chizhg:
The file name of eventing release is changed from `release.yaml` to `eventing.yaml` in https://github.com/knative/eventing/pull/2382

/cc @adrcunha 
/cc @chaodaiG

### DIFF
--- a/scripts/library.sh
+++ b/scripts/library.sh
@@ -638,5 +638,5 @@ readonly REPO_NAME_FORMATTED="Knative $(capitalize ${REPO_NAME//-/ })"
 
 # Public latest nightly or release yaml files.
 readonly KNATIVE_SERVING_RELEASE="$(get_latest_knative_yaml_source "serving" "serving")"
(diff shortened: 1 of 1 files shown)


adrcunha:
Split this into 2 PRs. One thing is updating a filename, other is creating a resource in every test project. Also, why can't the knative-gcp test create the app?

chizhg:
> Split this into 2 PRs. One thing is updating a filename, other is creating a resource in every test project. Also, why can't the knative-gcp test create the app?

Moved the other change to https://github.com/knative/test-infra/pull/1621

I think it's better to do it here instead of in the tests of knative-gcp, because:
1. Once an App Engine app is created, it cannot be deleted
2. Each project can only has one App Engine app
So it's basically a one-time work similar as other things we are doing in the `set_boskos_permissions.sh` script (actually I think script name needs to be updated after this change)

PR #1621, opened 2020-01-16
Create App Engine app for boskos projects

chizhg:
As required by step 1 in https://github.com/google/knative-gcp/tree/master/docs/scheduler, in order to use `SchedulerSource`, an App Engine app needs to be created.

I think it's better to do it here instead of in the tests of knative-gcp, because:
1. Once an App Engine app is created, it cannot be deleted
2. Each project can only has one App Engine app

So it's basically a one-time work similar as the other things we are doing in this script.

/cc @adrcunha

### DIFF
--- a/ci/prow/set_up_boskos_project.sh
+++ b/ci/prow/set_up_boskos_project.sh
@@ -106,4 +106,4 @@ gcloud projects add-iam-policy-binding ${PROJECT} \
 #
 # This command will throw an error if the app is already created, but since we expect to run
 # this script idempotently, we always mark this command as succeeded.
-gcloud app create --region=us-central || true
+gcloud app create --region=us-central || echo "AppEngine app probably already exists, ignoring..."
--- a/ci/prow/set_up_boskos_project.sh
+++ b/ci/prow/set_up_boskos_project.sh
@@ -19,7 +19,7 @@ set -e
 readonly PROJECT=${1:?"First argument must be the boskos project name."}
 # Remaining arguments are resources to be added, just like the RESOURCES array below.
 # For example, this command add an extra editor to the project:
-# $ set_boskos_permissions.sh my-boskos roles/editor my@service-account.com
+# $ set_up_boskos_project.sh my-boskos roles/editor my@service-account.com
 shift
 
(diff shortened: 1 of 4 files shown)


ci/prow/set_up_boskos_project.sh
```diff
@@ -99,3 +99,11 @@ curl -s -X GET -H "Authorization: Bearer ${ACCESS_TOKEN}" "https://www.googleapi
...
   --member="serviceAccount:service-${PROJECT_NUMBER}@gs-project-accounts.iam.gserviceaccount.com" \
   --role roles/pubsub.publisher
+
+# As required by step 1 in https://github.com/google/knative-gcp/tree/master/docs/scheduler,
+# create an App Engine app.
+# We use us-central here, but it indeed does not matter which region this app is created in.
+#
+# This command will throw an error if the app is already created, but since we expect to run
+# this script idempotently, we always mark this command as succeeded.
+gcloud app create --region=us-central || true
```
adrcunha:
```suggestion
gcloud app create --region=us-central || echo "AppEngine app probably already exists, ignoring..."
```

ci/prow/update_all_boskos_projects.sh
```diff
@@ -38,6 +38,6 @@ if [[ -z "${BOSKOS_PROJECTS}" ]]; then
 fi
 
 for boskos_project in ${BOSKOS_PROJECTS}; do
-  # Set permissions for this project
-  "./set_boskos_permissions.sh" ${boskos_project} $@
+  # Set up this project
+  "./set_up_boskos_project.sh" ${boskos_project} $@
```
adrcunha:
```suggestion
  ./set_up_boskos_project.sh ${boskos_project} $@
```

ci/prow/set_up_boskos_project.sh
chizhg: Done.

ci/prow/update_all_boskos_projects.sh
chizhg: Done.
It's a bit strange that I cannot apply the changes here through the UI..

PR #1631, opened 2020-01-22
Donation of Wathola test tool

cardil:

Wathola test tool was implemented to assert continuous operation of knative eventing during the upgrade and downgrade. It relies on [cloudevents sdk](https://github.com/cloudevents/sdk-go) and can be used to validate any cloudevents compatible framework.

It consists of three parts: sender, forwarder, and receiver. Sender sends events. Forwarder resend them (this can be scaled from zero). Receiver collects the events, and provide `/report` REST endpoint that
can be used to validate if all events was propagated by network just once.

I would like to put it in here, so the images can be prebuild and published publicly on conteiner repository. That would allow using them in testing of knative/eventing#2271 and also in upgrade tests of [knative/eventing-operator](https://github.com/knative/eventing-operator) without the need of building and adding additional dependency between k/e-o and k/e.

none

There should be configured a Job that pushes the 3 wathala container images.

none


coryrc:
Thank you, that fixes an important hole in testing. I think it would make more sense in knative/pkg; we've put [testing-only things](https://github.com/knative/pkg/tree/master/test/gke) there before and we intend to not have test-infra vendored some day.

adrcunha:
I don't think this tool belongs to test-infra too. If it can be used by both eventing and eventing-contrib, then indeed knative/pkg seems to be a better place, as suggested by Cory.

/hold

cardil:
Right. But, `knative/pkg` is strictly a library that is used by import statement.

This tool is a binary that is packaged to container images and it's used in that way. My PR knative/eventing#2388 uses this tool only as a container image. Now it is published on quay from my personal repo (https://quay.io/repository/cardil/wathola-forwarder)

I'd like to give away this tool to knative or cloudevents communities. This place seams a much more sensible than `knative/pkg`, as it already holds tools and images that are publicly published on gcr.io

Moving it to `knative/pkg` will introduce a lot of unneeded dependencies to all knative projects, and forces each repo to build it each time.

In my mind the best solution would be to just transfer ownership of https://github.com/cardil/wathola to knative or cloudevents as it is, just reconfiguring quay.io builds.

@coryrc @adrcunha what do you think?

adrcunha:
I understand. However, it still seems quite out-of-place putting the tool here the way it is. I like the idea of transferring the ownership of the original tool first, then figuring out what's the best migration approach; it doesn't block your tests, and still gives us freedom to update it when necessary.

cardil:
One problem with transferring ownership is that to do it in Github way I would need a permission to create repositories in knative organization. At least for migration time.

coryrc:
Could put it in knative-community if another repo is needed

adrcunha:
Creating a repo involves TOC and admins. I'll defer to @chaodaiG and @chizhg what's the best way for handling this tool donation.

adrcunha:
> Could put it in knative-community if another repo is needed

No. knative/community is intended for documentation and samples, not tools.

coryrc:
> > Could put it in knative-community if another repo is needed
> 
> No. knative/community is intended for documentation, not tools.

No, knative-community: https://github.com/knative-community

chizhg:
Hi @cardil , do you think `eventing/test/test_images` would be an appropriate place to hold this tool? Or if you don't want to build these images in e2e tests (as when we run e2e tests, we will run `./test/upload-test-images.sh` to build all images under this folder) , how about a new folder like `eventing/test/test_tool_images`?

cardil:
> Hi @cardil , do you think `eventing/test/test_images` would be an appropriate place to hold this tool? Or if you don't want to build these images in e2e tests (as when we run e2e tests, we will run `./test/upload-test-images.sh` to build all images under this folder) , how about a new folder like `eventing/test/test_tool_images`?

Yes. I don't want to build images each time. `./test/upload-test-images.sh` will publish images to be public on gcr.io?

If so, that's a good proposition. What about putting tool to: `eventing/test/tools/wathola` together with Dockerfiles?

chizhg:
`eventing/test/tools/wathola` sounds good to me :)

BTW you probably won't need Dockerfiles. If you add them as what we have in https://github.com/knative/eventing/tree/master/test/test_images, `./test/upload-test-images.sh` will be able to build them into images and upload to gcr.io.

cardil:
> `eventing/test/tools/wathola` sounds good to me :)
> 
> BTW you probably won't need Dockerfiles. If you add them as what we have in https://github.com/knative/eventing/tree/master/test/test_images, `./test/upload-test-images.sh` will be able to build them into images and upload to gcr.io.

Yeah. But, one problem remains. Doing that I'll add a lot of dependencies to `knative/eventing`. Also, pushing images to gcr.io will be manual, or not?

chizhg:
> > `eventing/test/tools/wathola` sounds good to me :)
> > BTW you probably won't need Dockerfiles. If you add them as what we have in https://github.com/knative/eventing/tree/master/test/test_images, `./test/upload-test-images.sh` will be able to build them into images and upload to gcr.io.
> 
> Yeah. But, one problem remains. Doing that I'll add a lot of dependencies to `knative/eventing`.
It seems most dependencies here already exist in eventing, so there should not be that many new dependencies being introduced. Since you'll need someone to approve your PR, the approver can tell whether it's good or not to keep these dependencies.

>  Also, pushing images to gcr.io will be manual, or not?
Before the e2e tests starts running on Prow, `./test/upload-test-images.sh` will be run to push the test images to gcr.io. You can follow the same pattern to push these images automatically before running your tests.

cardil:
> Before the e2e tests starts running on Prow, `./test/upload-test-images.sh` will be run to push the test images to gcr.io. You can follow the same pattern to push these images automatically before running your tests.

But, those images are not public. I'd like to have them public as then they can be used directly in testing in other repos (`eventing-operator`, downstream tests). 

This was directly requested by @houshengbo here https://github.com/knative/eventing/issues/2271#issuecomment-573773102

Now. It is possible as images are already public on quay.io. So, I wanted to retain that.

chizhg:
> > Before the e2e tests starts running on Prow, `./test/upload-test-images.sh` will be run to push the test images to gcr.io. You can follow the same pattern to push these images automatically before running your tests.
> 
> But, those images are not public. I'd like to have them public as then they can be used directly in testing in other repos (`eventing-operator`, downstream tests).
> 
> This was directly requested by @houshengbo here [knative/eventing#2271 (comment)](https://github.com/knative/eventing/issues/2271#issuecomment-573773102)
> 
> Now. It is possible as images are already public on quay.io. So, I wanted to retain that.

From the `serving-upgrade-test.sh` in serving-operator, you can see it's cloning serving repo to local and generating artifacts from source code. If you move the code of Wathola test tool to eventing, the images can also be built and uploaded from eventing-operator by following the same pattern.

cardil:
> From the `serving-upgrade-test.sh` in serving-operator, you can see it's cloning serving repo to local and generating artifacts from source code. If you move the code of Wathola test tool to eventing, the images can also be built and uploaded from eventing-operator by following the same pattern.

Yes. I know that it is possible to do that, either by clone or by vendoring. So you need to clone or vendor (time), build (time), deploy (time). All that takes time, and makes tests running slower :sob:. Also, test dependencies are introduced :face_with_head_bandage:.

What I was proposing is to have that tool separate, pre build, pre published. That makes everything faster and easier. Also, no new dependencies need to be introduced! :rainbow: 

So, that's my proposition. If you think that's not worth it, I will put this tool to eventing test tools directory. However, I will not be happy with that outcome.

chizhg:
> > From the `serving-upgrade-test.sh` in serving-operator, you can see it's cloning serving repo to local and generating artifacts from source code. If you move the code of Wathola test tool to eventing, the images can also be built and uploaded from eventing-operator by following the same pattern.
> 
> Yes. I know that it is possible to do that, either by clone or by vendoring. So you need to clone or vendor (time), build (time), deploy (time). All that takes time, and makes tests running slower 😭. Also, test dependencies are introduced 🤕.
> 
> What I was proposing is to have that tool separate, pre build, pre published. That makes everything faster and easier. Also, no new dependencies need to be introduced! 🌈

I understand your concern here, and agree that having this tool pre-built will make the process faster. However, I think this approach has some big issues that are more important than the time it can save.

1. You mentioned that we can add a Job to publish the three images, but how long should the time interval be? Actually regardless how short the time will be, we cannot guarantee the images used in eventing tests are the latest. Then if an error happens because the images are not latest, it will be really hard to debug and root cause as code will not be the single source of truth.

2. There will be no versioning for this tool if we put this tool in other repos other than eventing. Suppose some day you need to modify this tool to make it working with a new eventing release, e.g. 0.15, and if it is a breaking change, all tests for previous eventing releases will break because they can only use the latest tool. If you move this tool to eventing, there will not be such a problem as it will be properly versioned along with eventing.

> 
> So, that's my proposition. If you think that's not worth it, I will put this tool to eventing test tools directory. However, I will not be happy with that outcome.

cardil:
I think you're not right on lack of versioning and control as even now images are used with versions like: `quay.io/cardil/wathola-receiver:v0.7.0`. You can extend that by adding sha256 so it will be perfectly repoducable. Also, if source will be under knative organization you will have perfect control of it as well. That's what I was proposing.

But, let it be. I'm not completely happy I wasn't able to persuade you, but let's move forward. I'll put the code to eventing.

Closing this one.

chizhg:
@cardil I see, you've already taken versioning into consideration. However, `test-infra` is not a good place for this kind of version control, you can see unlike other Knative project, `test-infra` only has a `master` branch. And given the affinity between this tool and `eventing`, I think `eventing` is the best place to hold and version this tool.

I'm sorry that you are not happy with this direction, but based on what we currently have, I think moving it to eventing would be the optimal choice.

PR #1636, opened 2020-01-25
Deal with missing tags/manifests in get_latest_knative_yaml_source

### LINKED ISSUE #1629: Presubmit tests on untagged release branch picks up the wrong latest release in installing its dependencies
If the presubmit tests is run on a release branch that has not been tagged (handled by auto-release Prow jobs), `get_latest_knative_yaml_source` will return a wrong release.

The reason is because in https://github.com/knative/test-infra/blob/dfeb254e57938f9e30b179034e9833e196ec899f/scripts/library.sh#L619, it will return a tag that is on the previous release.

To fix it, `major_minor` in that function should be derived from the release branch name, e.g. get `v0.11` from `release-0.11`.

/kind bug

adrcunha:
* When running on a release branch, `get_latest_knative_yaml_source` prefers to get the current version from tags. If no tag exist, extract the current version from the release branch name. 

* If Knative project X creates a release branch *release-A* before serving release A is available, `start_latest_knative_serving` will fail as it cannot find a manifest for serving version A. In this case, `get_latest_knative_yaml_source` should return the latest nightly release.

Fixes #1629.

### DIFF
--- a/scripts/library.sh
+++ b/scripts/library.sh
@@ -597,14 +597,19 @@ function get_canonical_path() {
   echo "$(cd ${path%/*} && echo $PWD/${path##*/})"
 }
 
-# Returns whether the current branch is a release branch.
-function is_release_branch() {
+# Returns the current branch.
+function current_branch() {
   local branch_name=""
   # Get the branch name from Prow's env var, see https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md.
   # Otherwise, try getting the current branch from git.
   (( IS_PROW )) && branch_name="${PULL_BASE_REF:-}"
   [[ -z "${branch_name}" ]] && branch_name="$(git rev-parse --abbrev-ref HEAD)"
-  [[ ${branch_name} =~ ^release-[0-9\.]+$ ]]
+  echo "${branch_name}"
+}
+
+# Returns whether the current branch is a release branch.
+function is_release_branch() {
+  [[ $(current_branch) =~ ^release-[0-9\.]+$ ]]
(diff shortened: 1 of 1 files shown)


scripts/library.sh
```diff
@@ -615,19 +620,29 @@ function get_latest_knative_yaml_source() {
...
   # If it's a release branch, the yaml source URL should point to a specific version.
   if is_release_branch; then
-    # Get the latest tag name for the current branch, which is likely formatted as v0.5.0
-    local tag_name="$(git describe --tags --abbrev=0)"
-    # The given repo might not have this tag, so we need to find its latest release manifest with the same major&minor version.
-    local major_minor="$(echo ${tag_name} | cut -d. -f1-2)"
-    local yaml_source_path="$(gsutil ls gs://knative-releases/${repo_name}/previous/${major_minor}.*/${yaml_name}.yaml \
+    local major_minor="UNKNOWN"
+    # Extract the release major&minor version from either the latest tag or the branch name.
+    local tag_name="$(git describe --tags --abbrev=0 2> /dev/null)"
```
chizhg: `git describe --tags --abbrev=0` will return the tag name whose commit ID is nearest to the current branch (sorry the comment was not accurate). For instance, if `release-0.13` has been cut but not been tagged, this command will return `v0.12.0`. So this approach won't work.

Instead of using tags, how about we only check the branch name by only keeping line 629-630?

adrcunha: Thanks, updated.

PR #1647, opened 2020-01-30
Fix how coverage tool reads git attributes

coryrc:
It doesn't work now.

### DIFF
--- a/tools/coverage/git/git.go
+++ b/tools/coverage/git/git.go
@@ -37,10 +37,9 @@ func hasGitAttr(attr string, fileName string) bool {
 	attrCmd.Wait()
 
 	cleaned := strings.TrimSpace(val.String())
-	n := len(cleaned)
-	// TODO: was only checking "true" before this; I doubt there's ever a
-	// situation where it returns "true" instead of "set"
-	return cleaned[n-4:n] == "true" || cleaned[n-3:n] == "set"
+	// TODO: Prior to this rewrite, the code was checking only for "true"
+	// `git check-attr` returns set, unset, or unspecified. Not sure where "true" comes from, but keeping it just in case.
+	return strings.HasSuffix(cleaned, "true") || strings.HasSuffix(cleaned, "set")
 }
 
 func IsCoverageSkipped(filePath string) bool {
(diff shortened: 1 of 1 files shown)


tools/coverage/git/git.go
```diff
@@ -30,27 +29,18 @@ const (
...
-	valCmd.Wait()
+	attrCmd.Wait()
 
-	//fmt.Println(strings.ToLower(strings.TrimSpace(val.String())))
-	return strings.ToLower(strings.TrimSpace(val.String())) == "true"
+	cleaned := strings.TrimSpace(val.String())
+	n := len(cleaned)
+	// TODO: was only checking "true" before this; I doubt there's ever a
+	// situation where it returns "true" instead of "set"
+	return cleaned[n-4:n] == "true" || cleaned[n-3:n] == "set"
```
adrcunha: Use `strings.HasSuffix`.

coryrc: Sweet! Thanks!

coryrc:
@adrcunha I updated the comment to:

```
	// TODO: Prior to this rewrite, the code was checking only for "true"
	// `git check-attr` returns set, unset, or unspecified. Not sure where "true" comes from, but keeping it just in case.
```

adrcunha:
/approve

I'll let @steuhs LGTM.

steuhs:
Please test the image with the change with the -dev job first

tools/coverage/git/git.go
```diff
@@ -30,27 +29,17 @@ const (
...
-	//fmt.Printf("filename=*%s*\n", fileName)
-	attrCmd := exec.Command("git", "check-attr", attr, "--", fileName)
-	valCmd := exec.Command("cut", "-d:", "-f", "3")
-
-	pr, pw := io.Pipe()
-	attrCmd.Stdout = pw
-	valCmd.Stdin = pr
 	var val bytes.Buffer
-	valCmd.Stdout = &val
+	attrCmd := exec.Command("git", "check-attr", attr, "--", fileName)
```
chizhg: It looks you can save a few lines by using `pkg/test/cmd`, like in https://github.com/knative/test-infra/blob/42c9b74cda686778f2eb215c461325bd84a8551e/tools/cleanup/cleanup.go#L94. But it's up to you.

PR #1651, opened 2020-01-31
Enable welcome plugin

chizhg:
Enable welcome plugin:
```
The welcome plugin posts a welcoming message when it detects a user's first contribution to a repo.
```

### DIFF
--- a/ci/prow/plugins.yaml
+++ b/ci/prow/plugins.yaml
@@ -92,14 +92,15 @@ plugins:
   - lifecycle
   - milestone
   - owners-label
+  - project
   - shrug
   - size
   - skip
   - trigger
   - verify-owners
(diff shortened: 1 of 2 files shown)


adrcunha:
What's the welcome message? What do we want to display so it's actually helpful for new contributors (note that we already have a message for non-members due to /ok-to-test)?

chizhg:
> What's the welcome message? What do we want to display so it's actually helpful for new contributors (note that we already have a message for non-members due to /ok-to-test)?

The welcome message will be like https://github.com/kubernetes/test-infra/pull/10382#issuecomment-445301361. It does not have much practical significance, just meant to make the Prow robot smarter, like the `cat` and `dog` plugin :)

Issue #1658, opened 2020-02-04
Capture errors introduced by Go1.14 


chaodaiG:
Go1.14 is going to be released very soon: https://github.com/golang/go/releases/tag/go1.14beta1

With early manually testing there were already a couple of bugs discovered and fixes are on the way: https://github.com/knative/serving/pull/6725

We could set up Prow jobs running continuously or even in presubmit test to avoid uncaught errors

chizhg:
I think it will be enough to run optional presubmit tests with Go 1.14, we can change it to be required at some time point.
Do we really need to run the Prow job continuously?

PR #1659, opened 2020-02-04
added optional presubmits for go 1.14

yt3liu:

Part of https://github.com/knative/test-infra/issues/1658

### DIFF
--- a/ci/prow/config.yaml
+++ b/ci/prow/config.yaml
@@ -5541,46 +5541,6 @@ periodics:
     - name: test-account
       secret:
         secretName: test-account
-- cron: "16 */2 * * *"
-  name: ci-serving-go114-continuous
-  agent: kubernetes
-  labels:
-    prow.k8s.io/pubsub.project: knative-tests
-    prow.k8s.io/pubsub.topic: knative-monitoring
-    prow.k8s.io/pubsub.runID: ci-serving-go114-continuous
-  decorate: true
-  extra_refs:
-  - org: knative
-    repo: serving
-    base_ref: master
-    path_alias: knative.dev/serving
-  spec:
-    containers:
-    - image: gcr.io/knative-tests/test-infra/prow-tests-go114:stable
-      imagePullPolicy: Always
-      command:
-      - runner.sh
-      args:
-      - "./test/presubmit-tests.sh"
-      - "--all-tests"
-      volumeMounts:
-      - name: test-account
-        mountPath: /etc/test-account
-        readOnly: true
-      env:
-      - name: GOOGLE_APPLICATION_CREDENTIALS
-        value: /etc/test-account/service-account.json
-      - name: E2E_CLUSTER_REGION
-        value: us-central1
-      resources:
-        requests:
-          memory: 12Gi
-        limits:
-          memory: 16Gi
-    volumes:
-    - name: test-account
-      secret:
-        secretName: test-account
 - cron: "50 8 * * *"
   name: ci-knative-serving-0.9-continuous
   agent: kubernetes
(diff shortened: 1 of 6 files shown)


ci/prow/testgrid.yaml
```diff
@@ -56,6 +56,10 @@ default_dashboard_tab:
   alert_options:
     alert_mail_to_addresses: "knative-productivity-dev@googlegroups.com"
 test_groups:
+- name: ci-knative-serving-continuous
+  gcs_prefix: knative-prow/logs/ci-knative-serving-continuous
```
chaodaiG: this doesn't look right, you may want to ask @chizhg

chizhg: The current testgrid config generator does not support such use case.
And it does not sound right to me to add a continuous job that uses Go 1.14, as once the job fails on Go 1.14, there is no point to run it again and again since it will always fail.

yt3liu: I'll create a separate PR for the CI then. 

Why is the case that a continuous job will keep failing on Go 1.14? I think after the serving team make the change to use 1.14, wouldn't the CI using go1.14 be more useful in that case? 

The main reason for adding CI currently is to make sure once they support it, if it ever gets broken, it's easier to track down which change break it since the go1.14 version will be optional until that's released everywhere.

chizhg: I reconsidered about this, and I think you are right.

But instead of making these changes, you can probably make them custom jobs, that should make your life easier:
https://github.com/knative/test-infra/blob/77b975af074880c49957946a6d861e063e11f232/ci/prow/config_knative.yaml#L48
https://github.com/knative/test-infra/blob/77b975af074880c49957946a6d861e063e11f232/ci/prow/config_knative.yaml#L355

yt3liu: Cool testgrid result looks more reasonable after switcching to custom job :D 
PTAL

ci/prow/periodic_config.go
```diff
@@ -83,7 +83,7 @@ func generateCron(jobType, jobName string, timeout int) string {
 
 	var res string
 	switch jobType {
-	case "continuous", "custom-job", "auto-release": // Every hour
+	case "continuous", "custom-job", "auto-release", "continuous-go114": // Every hour
```
chizhg: Also for the changes in this file?

ci/prow/presubmit_config.go
```diff
@@ -51,7 +51,7 @@ func generatePresubmit(title string, repoName string, presubmitConfig yaml.MapSl
 	generateJob := true
 	for i, item := range presubmitConfig {
 		switch item.Key {
-		case "build-tests", "unit-tests", "integration-tests":
+		case "build-tests", "build-tests-go114", "unit-tests", "integration-tests", "integration-tests-go114":
```
chizhg: Also for the changes in this file?

ci/prow/make_config.go
```diff
@@ -966,7 +997,7 @@ func collectMetaData(periodicJob yaml.MapSlice) {
 			releaseVersion := ""
 			for _, item := range jobConfig {
 				switch item.Key {
-				case "continuous", "dot-release", "auto-release", "performance",
+				case "continuous", "continuous-go114", "dot-release", "auto-release", "performance",
```
chizhg: This change can be removed if you use custom jobs?

ci/prow/config_knative.yaml
```diff
@@ -25,13 +25,31 @@ presubmits:
...
+      dot-dev: true
+      optional: true
+      go114: true
+      resources:
+        requests:
+          memory: 12Gi # Real request for this pod is 16 as sidecar requests 4
+        limits:
+          memory: 16Gi
+      args:
+      - "--build-tests"
```
chizhg:
```suggestion
        - "--build-tests"
```

ci/prow/config.yaml
```diff
@@ -216,6 +216,94 @@ presubmits:
...
+    agent: kubernetes
+    context: pull-knative-serving-build-tests-go114
+    always_run: true
+    rerun_command: "/test pull-knative-serving-build-tests-go114"
+    trigger: "(?m)^/test (all|pull-knative-serving-build-tests-go114),?(\\s+|$)"
+    decorate: true
+    optional: true
+    path_alias: knative.dev/serving
+    branches:
+    - "release-0.9"
```
chizhg: This is not correct, and probably not an easy fix..

ci/prow/periodic_config.go
yt3liu: Removed.

ci/prow/make_config.go
yt3liu: Removed.

ci/prow/presubmit_config.go
yt3liu: Removed.

ci/prow/config.yaml
yt3liu: This is an optional job. So extra job should not impact the release if changes are pushed in release-0.9.

ci/prow/config_knative.yaml
yt3liu: Changed it to add the extra space. Note: The indention is actually very inconsistent in this file.

PR #1665, opened 2020-02-05
Fix spelling errors

mattmoor:
Produced via:
  `github.com/client9/misspell`
/assign chaodaig
/cc chaodaig


chaodaiG:
This doesn't seem like a typo, will this PR be produced again if we don't merge it? @mattmoor

PR #1669, opened 2020-02-06
Testgroup Failure Alert Modification (PR#1609):

### LINKED ISSUE #1609: Making testgrid alerts more useful and standardize the alerts
Currently testgrid will send email alerts when Prow jobs fail, as configured in https://github.com/knative/test-infra/blob/master/ci/prow/testgrid.yaml. By default it will send alerts for every job failure, and for some CI jobs it will send alerts if there are more than 3 failures. 

In this way it will send a lot of alerts, most of which are not useful. Further more, we now have FlakyTestReporter and Monitoring for alerting, and many alerts they send are duplicate.

We should distinguish responsibilities for these tools and avoid sending duplicate alerts:
1. Testgrid should only alert for general Knative job failures that we care about, e.g.
- Alert for any release job failure
- Alert for release CI job failures if they fail 3 times in series
2. Monitoring (or any monitoring system we would replace in the future) should only alert for infra related issues, e.g.
- Alert if any maintenance job failures, e.g. prow auto bumper, cleanup job, issue tracker job, etc.
- Alert if any infra related issue happens, e.g. boskos fails to acquire project, github no such host issue.
3. FlakyTestReporter should still alert for flaky test cases, no changes are required.

/kind good-first-issue

peterfeifanchen:

Previously, alerts were generated on every test fail with the exception
of continuous jobs on master of a repo. We are changing this to 3
consecutive failures on tagged continuous jobs of a repo and on any failures
for auto-release, dot-release and nightly-release.

Fixes #1609 

/cc @chizhg @yt3liu @efiturri

### DIFF
--- a/ci/prow/templates/testgrid_config_header.yaml
+++ b/ci/prow/templates/testgrid_config_header.yaml
@@ -31,7 +31,6 @@ default_test_group:
   use_kubernetes_client: true    # ** This field is deprecated and should always be true **
   is_external: true              # ** This field is deprecated and should always be true **
   alert_stale_results_hours: 26  # Alert if tests haven't run for a day (1 day + 2h)
-  num_failures_to_alert: 1       # Alert for every failure
   num_passes_to_disable_alert: 1 # Consider a failing test passing if it has 1 or more consecutive passes
 
 default_dashboard_tab:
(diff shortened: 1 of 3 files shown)


chizhg:
/lgtm
/approve

Thanks!

PR #1680, opened 2020-02-08
Document how/what files are excluded from coverage

steuhs:
Adds documentation on how and what files are excluded by the coverage tool

Fixes #

### DIFF
--- a/tools/coverage/README.md
+++ b/tools/coverage/README.md
@@ -20,7 +20,7 @@ This tool allows users to specify which files to skip, from reporting and aggreg
 User can add the git attribute `coverage-excluded=true` to any directory or file to exclude files.
 [Here](https://github.com/knative/serving/blob/master/.gitattributes) is an example.
 
-### Other automatically excluded files
+### Other Automatically Excluded Files
 - Source file (say `abc.go`) without corresponding `_test.go` file (abc_test.go)
 - Any file that has the gitattribute of `linguist-generated=true`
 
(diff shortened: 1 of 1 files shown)


tools/coverage/README.md
```diff
@@ -14,6 +14,16 @@ The code coverage tool has two major features:
 
 See the [design document](design.md).
 
+## Skipped Files
+
+This tool allows users to specify which files to skip, from reporting and aggregation.
+User can add the git attribute `coverage-excluded=true` to any directory or file to exclude files.
```
coryrc: leave off the =true
(yes, it can be used that way, but git attributes already can be a binary value so why make it a string value of "true"?)

tools/coverage/README.md
```diff
@@ -14,6 +14,16 @@ The code coverage tool has two major features:
...
+[Here](https://github.com/knative/serving/blob/master/.gitattributes) is an example.
+
+### Other Automatically Excluded Files
+- Source file (say `abc.go`) without corresponding `_test.go` file (abc_test.go)
+- Any file that has the gitattribute of `linguist-generated=true`
```
coryrc: Does these mean vendor files? Would be nice to say explicitly

steuhs: No. linguist-generated is an attribute used in other knative repos, which is not created or managed by me. I don't think I am the best person to give out an definition for that. I know some of the files with the attr are machine generated and the dev team request us to exempt linguist-generated files from coverage.

tools/coverage/README.md
steuhs: Good to know.
However, please take a look at the example linked. It's to be consistent with the rest of the file.

PR #1690, opened 2020-02-11
Revert "Update prow to v20200204-7e8cd997a"

chaodaiG:
Reverts knative/test-infra#1682

This change broke Prow and was discovered in < 1 minute, Prow is already reverted to previous working version, reverting this PR so code and Prow are consistent

/assign @chizhg

### DIFF
--- a/ci/prow/config.yaml
+++ b/ci/prow/config.yaml
@@ -29,10 +29,10 @@ plank:
     grace_period: 15000000000 # 15s
     utility_images:
       # Update these versions when updating plank version in cluster.yaml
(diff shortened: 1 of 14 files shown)


chizhg:
/lgtm

That's weird, we may need to reconsider if we still want auto-updater for these PRs.

chaodaiG:
I think there was a not disclosed breaking change, and possibly for something really good:

Screen Shot 2020-02-11 at 10 45 53 AM

chaodaiG:
@chizhg, I think it won't be worse than what we have today, without auto updater we need to do manual inspection, with auto updater I would still recommend doing manual inspection for at least 1 minute

PR #1692, opened 2020-02-11
Also skip cluster teardown if --skip-teardowns

chizhg:
https://github.com/knative/test-infra/pull/1687 forgot to skip cluster teardown, this PR adds it.

/cc @chaodaiG

### DIFF
--- a/scripts/e2e-tests.sh
+++ b/scripts/e2e-tests.sh
@@ -231,9 +231,12 @@ function create_test_cluster() {
   [[ -n "${GCP_PROJECT}" ]] && test_cmd_args+=" --gcp-project ${GCP_PROJECT}"
   [[ -n "${E2E_SCRIPT_CUSTOM_FLAGS[@]}" ]] && test_cmd_args+=" ${E2E_SCRIPT_CUSTOM_FLAGS[@]}"
   local extra_flags=()
-  if (( IS_BOSKOS )); then # Add arbitrary duration, wait for Boskos projects acquisition before error out
+  if (( IS_BOSKOS )); then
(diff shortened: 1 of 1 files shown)


chaodaiG:
Curious if there is any case that we do want teardown, WDYT?

chizhg:
> Curious if there is any case that we do want teardown, WDYT?

- If we do want teardown, we can simply run `./e2e-tests.sh`.
- If we do not want any teardown, we can run the script with `./e2e-tests.sh --skip-tear-downs`.
- If we want other teardowns but not cluster teardown, we can run with `./e2e-tests.sh --run-tests`.

I think these commands cover all use cases, is there any other that you think we are missing?

chaodaiG:
I would suggest not cleaning up by default, WDYT?

chizhg:
I feel the UX is better to clean up the resources unless users explicitly want to keep them, this way we don't leave idle stuff on their local environment by default. But that's just my opinion.

chaodaiG:
Just realized we might be talking with different assumptions. Let me express the UX I would like to see:
BYO cluster -> default NOT teardown anything
No cluster    -> default teardown everything
Prow             -> default NOT teardown anything

Is this what you mean?

chizhg:
> Just realized we might be talking with different assumptions. Let me express the UX I would like to see:
> BYO cluster -> default NOT teardown anything
> No cluster -> default teardown everything

I don't want to add such complex logic in our current shell script. So what I'm proposing and doing is:
BYO cluster -> default teardown anything
No cluster -> default teardown everything

> Prow -> default NOT teardown anything
> 
> Is this what you mean?

Issue #1693, opened 2020-02-11
Updating Prow images to 20200204 stopped Prow from functioning


chaodaiG:
Knative has an automated Prow images updating process, which picked one of Prow images updating PRs in k8s/test-infra repo, based on the fact that the PR was not reverted within 24 hours. It has been doing pretty good since it was introduced, until the one from 20200204 from commit https://github.com/kubernetes/test-infra/commit/5f861f7489e1c1a870caa038b72dc54595a21631#diff-00deeeaf0a7a2aba7d06744db0da6d43

The PR for updating images in knative is #1682 , after merging this PR and applying the images update to Prow cluster, the entire Prow stopped working, even `prow.knative.dev` was not responding

PR #1694, opened 2020-02-11
Add missing Prow configs so that we can update to latest images

chaodaiG:
Part of #1693 

/assign @chizhg 
/cc @katharine

### DIFF
--- a/config/prow/config_start.yaml
+++ b/config/prow/config_start.yaml
@@ -178,6 +178,7 @@ rules:
     verbs:
       - get
       - list
+      - watch
       # Required when deck runs with `--rerun-creates-job=true`
       # **Warning:** Only use this for non-public deck instances, this allows
       # anyone with access to your Deck instance to create new Prowjobs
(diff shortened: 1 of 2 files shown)


chaodaiG:
@Katharine as fyi

ci/prow/deployments/deck.yaml
```diff
@@ -40,6 +40,7 @@ spec:
         - --tide-url=http://tide/
         - --spyglass
         - --oauth-url=/github-login
+        - --cookie-secret=/etc/cookie/secret
```
chizhg: What is this cookie-secret? Is it also required to make this work?

chaodaiG: Sorry for not making it clear, this is a by product

chaodaiG: https://github.com/kubernetes/test-infra/blob/master/prow/ANNOUNCEMENTS.md
It was announced to be needed long time ago

chizhg: I see, nice!

chizhg:
/lgtm
/hold
In case @Katharine has extra comments.

PR #1697, opened 2020-02-12
additional labels for eventing-contrib

lionelvillard:
it has additional labels for categorizing eventing-contrib issues by channel and source type.

### DIFF
--- a/config/prow/plugins.yaml
+++ b/config/prow/plugins.yaml
@@ -74,6 +74,15 @@ label:
   - proposal/0.17
   - proposal/0.18
   - proposal/0.19
+  - channel/kafka
+  - channel/natss
+  - source/awssqs
+  - source/camel
+  - source/ceph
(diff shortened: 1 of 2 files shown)


chizhg:
Hi @lionelvillard , this change needs to be made in `config/prow/templates/prow_plugins.yaml`, and then you'll need to run `make -C config/prow config` to generate the config file.

config/prow/plugins.yaml
```diff
@@ -74,6 +74,15 @@ label:
   - proposal/0.17
   - proposal/0.18
   - proposal/0.19
+  - channel/kafka
```
chizhg: It looks Prow does not support setting additional_labels for a repo... Can you add a comment here to say these labels are specific to eventing?

lionelvillard: done

PR #1698, opened 2020-02-12
chore: Add "client-contrib" as new repo

rhuss:


config/prow/issue_tracker_config.go
```diff
@@ -69,6 +69,15 @@ func generateIssueTrackerPeriodicJobs() {
 		daysToRot:   rotDefault,
 		daysToClose: closeDefault,
 	}.generateJobs()
+
+	// generate for knative/client-contrib
```
chizhg: Do you really need it now for this new repo? It's not enabled for `knative/client` yet 😆

config/prow/config_knative.yaml
```diff
@@ -151,6 +151,18 @@ presubmits:
         - "./test/presubmit-integration-tests-latest-release.sh"
       dot-dev: true
 
+  knative/client-contrib:
+    - repo-settings:
+      go112-branches: []
```
chizhg: For now we'll have to add a dummy branch here to make the config-generator work :( 
See https://github.com/knative/test-infra/blob/810ad47815885a5f2ceee52539d4aef4ec4486d5/config/prow/config_knative.yaml#L287

config/prow/config_knative.yaml
```diff
@@ -151,6 +151,18 @@ presubmits:
         - "./test/presubmit-integration-tests-latest-release.sh"
       dot-dev: true
 
+  knative/client-contrib:
```
chizhg: To be consistent, it's better to also add the simple periodics part for this client-contrib repo, see the setting for client - https://github.com/knative/test-infra/blob/810ad47815885a5f2ceee52539d4aef4ec4486d5/config/prow/config_knative.yaml#L444

config/prow/issue_tracker_config.go
rhuss: ok, we can skip that for now (and I will discuss with the team whether we want to enable the stale bot for all client stuff)

config/prow_setup.md
```diff
@@ -69,7 +69,7 @@
 1. Make sure that _Knative Robots_ is an Admin of the repo.
 
 1. Add the new repo to [config_knative.yaml](./prow/config_knative.yaml),
-   without any presubmits. Run `make config` to regenerate
+   without any presubmits. Check the top-level section `presubmits:` and `periodics:` for blueprints for what to add. Then run `make config` to regenerate
```
rhuss: @chizhg I'm not sure what is meant "without any presubmits". Actually you _have to_ add a new section to `presubmits:`, or ? I was a bit confused what to add initially, so I just tried to copy over the config for `knative/client` (but overlooked the periodics part). A bit clarification (maybe even with an example ?) would be cool here.

Also, a link to reference documentation for the possible fields would be awesome (if such documentation exists).

chizhg: You can ignore that "without any presubmits" part. The doc is assuming you will create two PRs, one is setting up Prow for a new repo, and the next one is adding Prow job for the repo, but one PR that contains both is also fine.

Currently we don't have other reference doc for this, but thanks for helping us identify this gap, we will definitely make some improvements for the docs.

chizhg:
Please run `make config` to regenerate the config files.

config/prow/config_knative.yaml
```diff
@@ -471,6 +485,20 @@ periodics:
       dot-dev: true
       go113: true
 
+  knative/client-contrib:
+    - continuous: true
+      dot-dev: true
+      go113: true
+    - nightly: true
```
chizhg: You haven't added `./hack/release.sh` in that repo, so these jobs (nightly, dot-release, auto-release)  are not needed.

rhuss: good point. I will remove it for now and add later if needed.

PR #1701, opened 2020-02-13
Move config generator code to tools

chizhg:
Move config generator code to tools/config-generator


tools/config-generator/README.md
```diff
@@ -0,0 +1,11 @@
+# config-generator
+
+config-generator is a tool that takes a meta config file (e.g. ../../config/prow/config_knative.yaml)
```
chaodaiG: Thank you for this README, I think we should also mention where templates are. Related, should we move templates here?

chizhg: Done.

PR #1702, opened 2020-02-13
Reorganize a bunch of prow config files

chizhg:
1. There is no reason to use `kubectl replace` rather than `kubectl apply` in creating configmaps
2. After changing to `kubectl apply`, we won't need the empty configmaps in `config_start.yaml`
3. Namespace configs can also be moved out, so `config_start.yaml` can be renamed to `config_rbac.yaml`
4. And I don't see a reason why rbac resources need to be treated specially as they are binded with each Prow service... And since it can also be applied Idempotently, it can also be moved to the cluster config folder.
5. Rename deployments folder to cluster, as it contains not just deployments
6. There should be a sequence in applying all files under `cluster`, so add number prefix for the yaml files under it.

/cc @chaodaiG


chaodaiG:
/lgtm
Please make sure that presubmit scanning still applies on all these yamls
/hold

chaodaiG:
Forgot to ask, what does 400 mean?

chizhg:
> Forgot to ask, what does 400 mean?

That's point 6 in the PR description, basically there should be a sequence in apply all yaml files under `cluster` folder, these numbers are added just for sorting the files, that's how other Knative projects are doing.

PR #1719, opened 2020-02-18
Disable default failure alert

peterfeifanchen:

Disable alerts by default on failures for all test groups. If a test group wants failure alerts, it sets it individually in its test group config instead.

Fixes #

/cc @chizhg


chaodaiG:
/lgtm
/approve

Thanks!

chizhg:
I tried it once but seems it didn't work. We may still need to check with Michelle what's the right fix.

peterfeifanchen:
Sure I'll ping her.

peterfeifanchen:
Michelle suggested to remove the alert email to just the tests we want the alerts. This is a bit of the hack of the generator but let's see if this works first before looking into whether we need to allow the config generator to generate one more layer deep.

peterfeifanchen:
/cc @steuhs 

Stephen since you are on call, can you deploy this when it merges?

PR #1532, opened 2019-11-07
Enable running tests with https for presubmit

taragu:
/lint

**What this PR does, why we need it**:
Run tests with https as a presubmit

**Which issue(s) this PR fixes**:
Part of https://github.com/knative/serving/issues/5148

**Special notes to reviewers**:

**User-visible changes in this PR**: None

PR #1724, opened 2020-02-20
Add https tests to testgrid

taragu:
This PR adds running e2e+conformance tests with https option to testgrid dashboard.

Part of #5148

Not sure if I'm covering all the places I need to change. 

### DIFF
--- a/config/prow/config_knative.yaml
+++ b/config/prow/config_knative.yaml
@@ -421,6 +421,10 @@ periodics:
       full-command: "./test/e2e-tests.sh --ambassador-version latest"
       dot-dev: true
       go113: true
+    - custom-job: https
+      full-command: "./test/e2e-tests.sh --https"
+      dot-dev: true
+      go113: true
     - nightly: true
       dot-dev: true
       go113: true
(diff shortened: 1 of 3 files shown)


chizhg:
You cannot add a tab for presubmit test jobs (the tab `pull-knative-serving-test-coverage` is actually a wrong name)

Please add an entry in config_knative.yaml like https://github.com/knative/test-infra/blob/359e2ba3ab51221e4a91e78f4031361306bbba0d/config/prow/config_knative.yaml#L420, and then run `make config`, the config generator will generate the continuous Prow job and testgrid dashboard tab config for you.

taragu:
@chizhg I see. I think the test is already added in a previous PR, but I'm not able to find the dashboard. Could you point met to where it is?

chizhg:
You only added it as a presubmit job in https://github.com/knative/test-infra/pull/1532. To show the results on testgrid, you need to add a continuous job, which is what I meant in the last comment.
Please let me know if it's still unclear.

taragu:
@chizhg I see. Thanks for the clarification! I've updated the PR.

chizhg:
/lgtm
/approve

Thanks!

PR #1725, opened 2020-02-20
Fix testgrid tab name for coverage and comments for coverage-dev job

chizhg:
1. The test group name for coverage in testgrid config should be `ci-xxx`, instead of `pull-xxx`
2. We want to keep the coverage-dev job for debugging purpose, so the TODO comment is not true, rephrase it a but.

/cc @chaodaiG 
FYI @steuhs

### DIFF
--- a/config/prow/jobs/config.yaml
+++ b/config/prow/jobs/config.yaml
@@ -9543,6 +9543,21 @@ postsubmits:
         args:
         - "--artifacts=$(ARTIFACTS)"
         - "--cov-threshold-percentage=0"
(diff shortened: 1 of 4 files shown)


chaodaiG:
/lgtm

/hold
Please make sure cov-dev is not needed anymore

chizhg:
> /lgtm
> 
> /hold
> Please make sure cov-dev is not needed anymore

Thanks for reminding. Just checked with Stephen, we want to keep the coverage-dev job for debugging purpose, so there is actually no `TODO` to remove it, rephrased the comment a bit.

PR #1727, opened 2020-02-21
Update ko in prow-tests image

### LINKED ISSUE #1557: prow-tests image update fails knative-serving integration tests
After upgrading the tools in prow-tests, knative serving integration tests are not setup correctly and fails due to:
```
Error from server (NotFound): horizontalpodautoscalers.autoscaling "activator" not found
ERROR: Knative setup failed
```

Test PR in OSS: https://github.com/knative/serving/pull/6089
Spyglass link: https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/knative_serving/6089/pull-knative-serving-integration-tests/1197947484156989440

The most likely culprit that breaks the test-infra is ko update.

@yt3liu since she first reported and documented this issue

chizhg:
Since ko is a critical tool we use in the CI/CD flow, I believe it makes sense to pin a version that we have verified, so we don't accidentally introduce breaking changes in doing update. v0.3.0 was released 8 days ago and includes the fix we need. 

Also I run some tests with the new image and it works (the tests fail because they are flaky):
https://prow.knative.dev/view/gcs/knative-prow/logs/ci-knative-serving-continuous/1230632067109228544
https://prow.knative.dev/view/gcs/knative-prow/logs/ci-knative-serving-istio-1.3-no-mesh/1230655850540961792
https://prow.knative.dev/view/gcs/knative-prow/logs/ci-knative-serving-istio-1.3-mesh/1230655724787339264

Fixes #1557 

/cc @chaodaiG 
/cc @yt3liu

### DIFF
--- a/images/prow-tests/Dockerfile
+++ b/images/prow-tests/Dockerfile
@@ -42,7 +42,7 @@ RUN rm -rf /usr/local/go && \
     rm "${GO_TARBALL}"
 
 # Extra tools through go get
-RUN GO111MODULE=on go get github.com/google/ko/cmd/ko
+RUN GO111MODULE=on go get github.com/google/ko/cmd/ko@v0.3.0
 RUN GO111MODULE=on go get github.com/boz/kail/cmd/kail
 RUN go get -u github.com/golang/dep/cmd/dep
 RUN go get -u github.com/jstemmer/go-junit-report


chaodaiG:
/lgtm
Thank you!

/hold
I think v0.3.0 is a moving target, it's still not a pinned revision. I think it's acceptable, but we should bear in mind that the ko we use might not be pointing to v0.3.0 all the time

chizhg:
> /lgtm
> Thank you!
> 
> /hold
> I think v0.3.0 is a moving target, it's still not a pinned revision. I think it's acceptable, but we should bear in mind that the ko we use might not be pointing to v0.3.0 all the time

Do you mean it can have cherrypick fixes on v0.3.0? Or we can possibly install another version of `ko` in different workflows?

chaodaiG:
> > /lgtm
> > Thank you!
> > /hold
> > I think v0.3.0 is a moving target, it's still not a pinned revision. I think it's acceptable, but we should bear in mind that the ko we use might not be pointing to v0.3.0 all the time
> 
> Do you mean it can have cherrypick fixes on v0.3.0? Or we can possibly install another version of `ko` in different workflows?

v0.3.0 is a git tag, which theoretically can be moved to any other commit if someone wants to

chizhg:
> > > /lgtm
> > > Thank you!
> > > /hold
> > > I think v0.3.0 is a moving target, it's still not a pinned revision. I think it's acceptable, but we should bear in mind that the ko we use might not be pointing to v0.3.0 all the time
> > 
> > 
> > Do you mean it can have cherrypick fixes on v0.3.0? Or we can possibly install another version of `ko` in different workflows?
> 
> v0.3.0 is a git tag, which theoretically can be moved to any other commit if someone wants to

I see. But that's not normally people will do :)

/hold cancel

Issue #1638, opened 2020-01-28
Prow config update PRs don't have an approver auto-assigned 


coryrc:
Assignees are set by knative-prow-robot, who selects reviewers and approvers from the OWNERS file.

However all PRs from knative-prow-updater-robot are assigned only to a reviewer. See https://github.com/knative/test-infra/pull/1637 for example.

This it NOT because knative-prow-updater-robot is not a member of Knative, as other PRs created by non-members do get a reviewer and an approver assigned by knative-prow-robot.

adrcunha:
Assignees are set by knative-prow-robot, who selects reviewers and approvers from the OWNERS file. If this is really desired, the config files should be moved to a separate directory and have a custom OWNERS file.

adrcunha:
It would be good to expand on why this is desired, though.

adrcunha:
Hmmm... I think the reason for this issue is because https://github.com/knative/test-infra/pull/1637 was assigned only to coryrc, who's a reviewer but not an approver.

adrcunha:
It seems that the way the update PR is generated is causing this behavior, but I can't see any obvious cause in https://github.com/knative/test-infra/blob/b0c4c4e1691282f24ccb3b5b0399952bf086edfc/tools/prow-auto-bumper/pullrequest.go#L141

PR #1739, opened 2020-02-24
Prow auto bumper won't bother anyone but oncall

coryrc:
No reason to auto-add reviewers; it's the oncall's responsibility to review.

### DIFF
--- a/tools/prow-auto-bumper/pullrequest.go
+++ b/tools/prow-auto-bumper/pullrequest.go
@@ -58,7 +58,7 @@ func generatePRBody(extraMsgs []string) string {
 	var assignment string
 	if err == nil {
 		if oncaller != "" {
-			assignment = "/assign @" + oncaller
+			assignment = fmt.Sprintf("/assign @%s\n/cc @%s\n", oncaller, oncaller)
 		} else {
 			assignment = "Nobody is currently oncall."
 		}


tools/prow-auto-bumper/pullrequest.go
```diff
@@ -58,7 +58,7 @@ func generatePRBody(extraMsgs []string) string {
 	var assignment string
 	if err == nil {
 		if oncaller != "" {
-			assignment = "/assign @" + oncaller
+			assignment = fmt.Sprintf("/assign @%s\n/cc @%s\n", oncaller, oncaller)
```
chaodaiG: are you sure `\n` works?

coryrc: It's used in body just above without problem.

chizhg:
It seems what we stated in https://github.com/knative/test-infra/issues/1638#issue-555923905 and https://github.com/knative/test-infra/pull/1661#issue-371018250 are both wrong, Prow won't automatically cc or assign approvers..
So after this change, the oncall will need to manually assign one approver if they don't have approve power, is it a reasonable flow for you? Or we can consider adding all oncalls to be OWNER of Prow configs?

coryrc:
> So after this change, the oncall will need to manually assign one approver if they don't have approve power, is it a reasonable flow for you?

Yes. Today, #1737 made reviewers of yt3liu and steuhs, neither of whom can approve. So it was just bothering them.

> Or we can consider adding all oncalls to be OWNER of Prow configs?

This is good as well, but not up to me :).

PR #1743, opened 2020-02-25
Explicitly enable pubsub in Prow job generator

chaodaiG:
Prow jobs send pubsub messages based on it's config, which is consumed by a few test-infra tools. One of them is flaky test retryer, only jobs with pubsub enabled are monitored by retryer. 

This PR does:
- Let auto-tls test to send pubsub messages
- Move business logic out of Go code, and let it configurable explicitly
- Remove unnecessary pubsub

/assign @chizhg 
/cc @vagababov

### DIFF
--- a/config/prow/config_knative.yaml
+++ b/config/prow/config_knative.yaml
@@ -31,22 +31,22 @@ presubmits:
           memory: 16Gi
     - unit-tests: true
       dot-dev: true
-      is-monitored: true
+      needs-monitor: true
     - integration-tests: true
       dot-dev: true
-      is-monitored: true
+      needs-monitor: true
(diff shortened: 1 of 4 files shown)


config/prow/config_knative.yaml
```diff
@@ -31,24 +31,27 @@ presubmits:
           memory: 16Gi
     - unit-tests: true
       dot-dev: true
+      is-monitored: true
```
chizhg: Do you think `needs-monitor` would be a better name, as we already have `needs-dind`?

chaodaiG: make sense and done

PR #1744, opened 2020-02-25
All network tests run both e2e and auto tls tests

chaodaiG:
Update all Prow jobs for running network tests, so that they still run auto tls tests

/assign @chizhg 
/assign @ZhiminXiang 

/hold
Let's make sure it's not merged before testing

### DIFF
--- a/config/prow/config_knative.yaml
+++ b/config/prow/config_knative.yaml
@@ -64,73 +64,55 @@ presubmits:
       dot-dev: true
       always_run: false
       optional: true
-      args:
-      - "--run-test"
-      - "./test/e2e-tests.sh --istio-version 1.3-latest --mesh &&
+      full-command: "./test/e2e-tests.sh --istio-version 1.3-latest --mesh &&
       ./test/e2e-tests.sh --run-tests --skip-knative-setup --istio-version 1.3-latest --mesh"
     - custom-test: istio-1.3-no-mesh
       dot-dev: true
       always_run: false
       optional: true
-      args:
-      - "--run-test"
-      - "./test/e2e-tests.sh --istio-version 1.3-latest --no-mesh &&
+      full-command: "./test/e2e-tests.sh --istio-version 1.3-latest --no-mesh &&
       ./test/e2e-tests.sh --run-tests --skip-knative-setup --istio-version 1.3-latest --no-mesh"
     - custom-test: istio-1.4-mesh
       dot-dev: true
       always_run: false
       optional: true
-      args:
-      - "--run-test"
-      - "./test/e2e-tests.sh --istio-version 1.4-latest --mesh &&
+      full-command: "./test/e2e-tests.sh --istio-version 1.4-latest --mesh &&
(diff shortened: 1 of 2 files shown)


chaodaiG:
It doesn't work @chizhg :
https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/knative_serving/7016/pull-knative-serving-istio-1.4-no-mesh/1232454592453677056

chizhg:
😞 `presubmit-tests.sh` does not support this. Can you check if the periodic jobs work? They are run directly by `runner.sh`

chaodaiG:
good catch, would prefer not to alter continuous run status, updated presubmit to use different endpoint

chaodaiG:
not supported either: https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/knative_serving/7016/pull-knative-serving-istio-1.4-no-mesh/1232471958478655489

chizhg:
Hmm, how about `(./test/e2e-tests.sh --istio-version 1.4-latest --no-mesh) && (./test/e2e-tests.sh --run-tests --skip-knative-setup --istio-version 1.4-latest --no-mesh)`?

config/prow/config_knative.yaml
```diff
@@ -64,73 +64,55 @@ presubmits:
       dot-dev: true
       always_run: false
       optional: true
-      args:
-      - "--run-test"
-      - "./test/e2e-tests.sh --istio-version 1.3-latest --mesh &&
+      full-command: "./test/e2e-tests.sh --istio-version 1.3-latest --mesh &&
```
coryrc: This is an ugly hack (and I don't think it will work).

I think adding an additional script is a better solution, it'll basically just look like:

```
./e2e-tests.sh $@
./e2e-tests.sh --run-tests --skip-knative-setup $@
```

(actually, you seem to just be running the same test twice? I can't see what's different except the above options?)

chizhg:
Seems it does not work with the config generator..

coryrc:
> Seems it does not work with the config generator..

I'm guessing you can't use shell in arguments; they are passed directly and not parsed, so you can't use && or () etc.

config/prow/config_knative.yaml
chaodaiG: This is more like a test to see which one works for Prow, and I do agree with you that have a additional script is a better solution in this case, and it could live inside serving repo

chaodaiG:
Added a wrapper in serving, let's discuss on that one: https://github.com/knative/serving/pull/7019

config/prow/config_knative.yaml
adrcunha: Indeed this won't work, as the successful second run will overwrite the results of the first run. Also, presubmit or not, they all work the same way so you can just have a separate job that you can start from command line for testing purposes, no need to merge a test PR.

adrcunha: Apologies, this will actually run, but as presubmit it will miss the presubmit checks. I commented on 1746.

chaodaiG:
/close
in favor of #1783

PR #1745, opened 2020-02-26
Rename auto tls test script

chaodaiG:
The current name of script doesn't follow `e2e-*tests.sh` naming convention and not discoverable by presubmit tests, rename it
/assign @chizhg 

/hold
Will need to have the script before merging

### DIFF
--- a/config/prow/config_knative.yaml
+++ b/config/prow/config_knative.yaml
@@ -49,7 +49,7 @@ presubmits:
       needs-monitor: true
       args:
       - "--run-test"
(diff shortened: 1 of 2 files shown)


chizhg:
/lgtm

Only the continuous test have this issue right? Other tests are all run with `./presubmit-tests.sh --run-tests xxxxx`.

chaodaiG:
that's correct. I'm actually planning to make the other tests use presubmit test as well

PR #1749, opened 2020-02-26
Highlight "code is outdated" error

chizhg:
As reported on Slack https://knative.slack.com/archives/CQBKVH4QY/p1582704778000800, highlight the "code is outdated" error on Spyglass.

Note:
We don't have a central point that defines this error message, but it's the same pattern that every repository follows. I added a TODO to move it to the common test infrastructure.

/cc @chaodaiG

### DIFF
--- a/config/prow/staging/jobs/config.yaml
+++ b/config/prow/staging/jobs/config.yaml
@@ -98,10 +98,6 @@ presubmits:
           secretName: test-account
   - name: pull-knative-prow-robot-serving-unit-tests
     agent: kubernetes
-    labels:
-      prow.k8s.io/pubsub.project: knative-tests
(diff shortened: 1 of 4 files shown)


config/prow/staging/jobs/config.yaml
```diff
@@ -98,10 +98,6 @@ presubmits:
           secretName: test-account
   - name: pull-knative-prow-robot-serving-unit-tests
     agent: kubernetes
-    labels:
-      prow.k8s.io/pubsub.project: knative-tests
-      prow.k8s.io/pubsub.topic: knative-monitoring
```
chaodaiG: where is this from?

chaodaiG: nvm, it's staing

PR #1755, opened 2020-02-27
Report flakes for eventing-contrib in Github and Slack

chaodaiG:
As discussed in eventing channel on Slack, it's expected to also report flakes status for eventing-contrib, which will be reported to both `eventing-sources` and `eventing-channels`

/assign @chizhg 
/cc @vaikas 
/cc @grantr 
/cc @lionelvillard

### DIFF
--- a/tools/flaky-test-reporter/config/config.yaml
+++ b/tools/flaky-test-reporter/config/config.yaml
@@ -75,6 +75,15 @@ jobConfigs:
     slackChannels:
       - name: eventing
         identity: C9JP909F0
+  - name: ci-knative-eventing-contrib-continuous
+    repo: eventing-contrib
+    type: postsubmit
+    issueRepo: eventing-contrib
+    slackChannels:
+      - name: eventing-sources
+        identity: CQBKVH4QY
+      - name: eventing-channels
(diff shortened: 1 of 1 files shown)


tools/flaky-test-reporter/config/config.yaml
```diff
@@ -75,6 +75,15 @@ jobConfigs:
     slackChannels:
       - name: eventing
         identity: C9JP909F0
+  - name: ci-knative-eventing-contrib-continuous
+    repo: eventing-contrib
+    type: postsubmit
+    issueRepo: eventing-contrib
+    slackChannels:
```
chizhg: It will be reported in these two channels **at the same time** right?

chaodaiG: yep, that's expected

chizhg: It seems you do not need this `ch` in https://github.com/knative/test-infra/blob/dfe8d3851bdb8f18e9d83e5ad9df1fbf1ca1d7ee/tools/flaky-test-reporter/slack_notification.go#L83. Guess it was written before we attended the Go concurrency class :)

PR #1766, opened 2020-02-29
Run check-config in a separate presubmit test job

### LINKED ISSUE #1686: Presubmit tests in test-infra should also include testing for staging prow configuration
Since staging prow configuration file is also generated from the config generator tool, if there are some changes to the tool, the auto-generated staging prow configuration will also be changed.

Currently in presubmit tests we are only checking the prod prow configuration file, see https://github.com/knative/test-infra/blob/64615f92c4ebdd4e4423015ffc0db45877660a0d/test/presubmit-tests.sh#L39, we should also check staging prow configuration file here.

/kind good-first-issue

chizhg:
Run check-config in a separate presubmit test job, other than build tests. This allows people to find out the error reason more clearly.
Also fix #1686 to also check prow-staging config files.


config/prow_common.mk
```diff
@@ -153,47 +153,11 @@ update-prow-cluster: update-almost-all-cluster-deployments update-all-boskos-dep
 # In emergency, could easily edit this file, deleting all these lines
 confirm-master:
 	@if git diff-index --quiet HEAD; then true; else echo "Git working space is dirty -- will not update server"; false; fi;
-ifneq ("$(shell git branch --show-current)","master")
```
coryrc: `git rev-parse --abbrev-ref HEAD` can be affected by a dirty working space or something
`git branch --show-current` is exactly designed for this purpose

config/prow/Makefile
```diff
@@ -15,10 +15,8 @@
 PROJECT       := knative-tests
 PROW_GCS      := knative-prow
 PROW_HOST     := https://prow.knative.dev
-TESTGRID_GCS  := knative-testgrid
```
coryrc: Could you instead leave this here and remove the TESTGRID_GCS from prow_common.mk? It's what I meant to do but overlooked.

Removing TESTGRID_CONFIG is good though, you should continue removing it.

config/prow/jobs/config.yaml
```diff
@@ -3428,6 +3428,70 @@ presubmits:
...
+  - name: pull-knative-test-infra-check-config
+    agent: kubernetes
+    context: pull-knative-test-infra-check-config
+    always_run: true
+    rerun_command: "/test pull-knative-test-infra-check-config"
+    trigger: "(?m)^/test (all|pull-knative-test-infra-check-config),?(\\s+|$)"
+    decorate: true
+    path_alias: knative.dev/test-infra
+    branches:
+    - "release-0.10"
```
coryrc: This is weird..... we don't have branches in test-infra

test/presubmit-config-check.sh
```diff
@@ -0,0 +1,66 @@
...
+source $(dirname $0)/../scripts/library.sh
+
+set -e
+
+trap 'echo "Please rerun \`make -C config/prow config\`."' ERR
+header "Checking generated config for production prow and testgrid"
+export PROW_CONFIG=$(mktemp)
+export PROW_JOB_CONFIG=$(mktemp)
+export PROW_PLUGINS=$(mktemp)
+export TESTGRID_CONFIG=$(mktemp)
```
coryrc: You aren't actually writing to these files. This bash script has to invert the logic of the Makefile; first copy each of the files to the temp file, run make config, then compare them.

test/presubmit-config-check.sh
```diff
@@ -0,0 +1,66 @@
...
+diff --ignore-matching-lines="^# Copyright " "${TESTGRID_CONFIG}" "config/prow/testgrid/testgrid.yaml"
+
+trap 'echo "Please rerun \`make -C config/prow-staging config\`."' ERR
+header "Checking generated config for staging prow"
+subheader "Regenerating config for staging prow"
+make -C config/prow-staging config
+subheader "Comparing the generated config files with the existing config files"
+diff --ignore-matching-lines="^# Copyright " "${PROW_CONFIG}" "config/prow-staging/core/config.yaml"
+diff --ignore-matching-lines="^# Copyright " "${PROW_JOB_CONFIG}" "config/prow-staging/jobs/config.yaml"
+diff --ignore-matching-lines="^# Copyright " "${PROW_PLUGINS}" "config/prow-staging/core/plugins.yaml"
```
coryrc: This should be able to be function-ized

test/presubmit-tests.sh
```diff
@@ -35,8 +35,6 @@ function post_build_tests() {
     echo "*** Checking ${makefile}"
     make -n -C $(dirname ${makefile}) || { failed=1; echo "--- FAIL: ${makefile}"; }
   done
-  subheader "Checking config files"
-  make -C config/prow test || { failed=1; echo "--- FAIL"; }
```
coryrc: As inappropriate as it is for this to have been in the Makefile to begin with, wouldn't an easier solution be to just run `make -C config/prow-staging test` as well?

config/prow/jobs/config.yaml
chizhg: I know... But for some reason, we'll have to have at least one branch to get the config generator work. I have changed the branch name to `legacy` to be less confusing.
(Config generator is not related to this PR and should have separate discussion.)

test/presubmit-config-check.sh
chizhg: We are using `?=` in the Makefile, like https://github.com/knative/test-infra/blob/4dac123b9a3d14d2e07b4a59fbac0b0e99a09e92/config/prow_common.mk#L32.

If these env vars are defined as what we are doing here with `export XXX=xx`, the make command will actually use them. So it actually works :)

test/presubmit-tests.sh
chizhg: Staging Prow does not have Testgrid config file. I don't want to add a special logic in the Makefile, as it's already complicated enough :(

config/prow/Makefile
chizhg: I have removed both TESTGRID_GCS and TESTGRID_CONFIG from prow_common.mk, as they are only used in production Prow.

test/presubmit-config-check.sh
chizhg: Done, thanks.

config/prow_common.mk
chizhg: It seems the `git` in the prow-tests image does not support `git branch --show-current`.  I'll take a further look on this tomorrow.

chaodaiG:
This is a nice move, thanks for doing the work!

config/prow/Makefile
coryrc: There's a different TESTGRID_GCS in prow-staging, are you sure?

test/presubmit-config-check.sh
coryrc: I thought that only related to make variables and not environment variables. Or do all environment variables become make variables?
I don't like that the environment could make the command misbehave, but you're not making it worse. So if it works, it works.

config/prow/Makefile
coryrc: Oh, because it has GENERATE_TESTGRID_CONFIG set to false. Okay.

test/presubmit-tests.sh
coryrc: k

config/prow_common.mk
coryrc: Then our git at least a year old... https://stackoverflow.com/a/55088865

test/presubmit-config-check.sh
```diff
@@ -27,19 +34,19 @@ export TESTGRID_CONFIG=$(mktemp)
 subheader "Regenerating config for production prow and testgrid"
 make -C config/prow config
 subheader "Comparing the generated config files with the existing config files"
-diff --ignore-matching-lines="^# Copyright " "${PROW_CONFIG}" "config/prow/core/config.yaml"
-diff --ignore-matching-lines="^# Copyright " "${PROW_JOB_CONFIG}" "config/prow/jobs/config.yaml"
-diff --ignore-matching-lines="^# Copyright " "${PROW_PLUGINS}" "config/prow/core/plugins.yaml"
-diff --ignore-matching-lines="^# Copyright " "${TESTGRID_CONFIG}" "config/prow/testgrid/testgrid.yaml"
+diff_config_files "${PROW_CONFIG}" "config/prow/core/config.yaml"
+diff_config_files "${PROW_JOB_CONFIG}" "config/prow/jobs/config.yaml"
+diff_config_files "${PROW_PLUGINS}" "config/prow/core/plugins.yaml"
```
coryrc: I meant the set of these three should be in a function

config/prow/jobs/config.yaml
```diff
@@ -3428,6 +3428,70 @@ presubmits:
...
+        env:
+        - name: GOOGLE_APPLICATION_CREDENTIALS
+          value: /etc/test-account/service-account.json
+        - name: E2E_CLUSTER_REGION
+          value: us-central1
+      volumes:
+      - name: test-account
+        secret:
+          secretName: test-account
+  - name: pull-knative-test-infra-check-config
```
peterfeifanchen: Are these two of the same jobs? Duplicate?

peterfeifanchen: pull-knative-test-infra-check-config duplicated twice here?

coryrc: Duplicate. Only the image name has changed.
The config generator is currently designed to always create two jobs for every desired one, one with older image and one with latest.

peterfeifanchen: Ignore this, ironically duplicated this.

coryrc: See my other comment: https://github.com/knative/test-infra/pull/1766#discussion_r386577737

config/prow/jobs/config.yaml
```diff
@@ -3428,6 +3428,70 @@ presubmits:
...
+    agent: kubernetes
+    context: pull-knative-test-infra-check-config
+    always_run: true
+    rerun_command: "/test pull-knative-test-infra-check-config"
+    trigger: "(?m)^/test (all|pull-knative-test-infra-check-config),?(\\s+|$)"
+    decorate: true
+    path_alias: knative.dev/test-infra
+    branches:
+    - "legacy"
```
peterfeifanchen: I am a bit confused, if all these jobs have "legacy" as branches, do they get run then. Why are there so many jobs marked as legacy? Is this part of the generator that takes in the prow config for the legacy branches and automatically generates all this? Like I feel there may be two changes here that is not made very clear by the commit message.

1) We seem to have gotten rid of release-0.10 and replaced it with this fake "legacy" branch. 
2) There is an extra periodic job that does the config check
3) How do I know if these changes affect the production cluster vs. staging cluster?

I don't know what the effects of 1) is and I am not sure how 2) was checked previously (manually call make?). For 3) it seems we only have a single prow config file generated right, so what is the part about production vs. staging clusters?

test/presubmit-config-check.sh
chizhg: Done.

config/prow/jobs/config.yaml
chizhg: 1) Either `release-0.10` or `legacy` is added for making it work for config generator, the Prow jobs themselves do not have any effect as there is only one master branch in this repo.
(This is actually a bug in config generator, but we haven't got time to fix it)

2) We don't have an extra periodic job that does the config check. Previously we are doing this check in the presubmit-build-tests, but it was not clear and missing some important checks. This PR is moving the check to a separate Prow job.

3) We have a set of config files for both production and staging Prow, see `config/prow` and `config/prow-staging`. The only configuration files they share are the cluster configs in `config/prow/cluster` (which in the future we will also make a copy in `config/prow-staging` while we implement Prow auto-update)

test/presubmit-config-check.sh
```diff
@@ -0,0 +1,84 @@
...
+  local prow_job_config="$3"
+  local prow_env="$4"
+  diff_config_files "${prow_config}" "config/${prow_env}/core/config.yaml"
+  diff_config_files "${prow_plugin_config}" "config/${prow_env}/core/plugins.yaml"
+  diff_config_files "${prow_job_config}" "config/${prow_env}/jobs/config.yaml"
+}
+
+set -e
+
+trap 'echo "Please rerun \`make -C config/prow config\`."' ERR
```
coryrc: Could you add "--- FAIL: " to the beginning of the output for this and all trap?

test/presubmit-config-check.sh
```diff
@@ -0,0 +1,84 @@
...
+header "Checking generated config for production prow and testgrid"
+export PROW_CONFIG=$(mktemp)
+export PROW_JOB_CONFIG=$(mktemp)
+export PROW_PLUGINS=$(mktemp)
+export TESTGRID_CONFIG=$(mktemp)
+subheader "Regenerating config for production prow and testgrid"
+make -C config/prow config
+subheader "Comparing the generated config files with the existing config files"
+diff_prow_config_files "${PROW_CONFIG}" "${PROW_PLUGINS}" "${PROW_JOB_CONFIG}" "prow"
```
coryrc: You use the same 3 variables for both calls; you could drop all the intermediate variables and just use the globals in the function

config/prow_common.mk
chizhg: It's already failing in presubmit-build-tests for other PRs, see https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/knative_test-infra/1767/pull-knative-test-infra-build-tests/1234494216499892227#1:build-log.txt%3A655. 

I added a TODO here to use `--show-current` after we update Git in the prow-tests image. Let's do it in followup PRs?

test/presubmit-config-check.sh
chizhg: Done.

test/presubmit-config-check.sh
chizhg: Done.

test/presubmit-config-check.sh
```diff
@@ -0,0 +1,78 @@
...
+  --job-config-path="$(realpath "config/prow/jobs/config.yaml")" \
+  --plugin-config="$(realpath "config/prow/core/plugins.yaml")"
+
+header "Validating staging Prow config files"
+bazel run @k8s//prow/cmd/checkconfig -- \
+  --config-path="$(realpath "config/prow-staging/core/config.yaml")" \
+  --job-config-path="$(realpath "config/prow-staging/jobs/config.yaml")" \
+  --plugin-config="$(realpath "config/prow-staging/core/plugins.yaml")"
+
+trap '--- FAIL: echo "Testgrid config file has errors, please check."' ERR
```
coryrc: This and others should be trap 'echo "--- FAIL: Testgrid config file has errors, please check."' ERR

config/prow_common.mk
coryrc: k

test/presubmit-config-check.sh
chizhg: My bad.. Fixed.

PR #1772, opened 2020-03-03
Reduce Boskos Janitor count

coryrc:
Now that janitor properly cleans images, we don't need so many (k8s
has 16 for 256 projects...).

/cc @nbarthwal 
/assign @chizhg

### DIFF
--- a/config/prow/cluster/400-boskos-janitor.yaml
+++ b/config/prow/cluster/400-boskos-janitor.yaml
@@ -21,7 +21,7 @@ metadata:
     app: boskos-janitor
   namespace: test-pods
 spec:
-  replicas: 35  # Number of distributed janitor instances
+  replicas: 8  # Number of distributed janitor instances; k8s has ~1/16 as many projects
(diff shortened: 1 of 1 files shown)


config/prow/cluster/400-boskos-janitor.yaml
```diff
@@ -21,7 +21,7 @@ metadata:
     app: boskos-janitor
   namespace: test-pods
 spec:
-  replicas: 35  # Number of distributed janitor instances
+  replicas: 8  # Number of distributed janitor instances; k8s has ~1/16 as many projects
```
chizhg: Shall we be a bit less aggressive? Maybe change to 15 first?

coryrc: We only have one cleaning

coryrc: Add each supports multiple cleaning threads AFAICT

PR #1774, opened 2020-03-03
Update the prow-tests image

chizhg:
The Git version we use in prow-tests image is pretty old, see https://github.com/knative/test-infra/pull/1766#discussion_r386185336. So update the image to use the latest kubekins-e2e image.


chaodaiG:
/lgtm
/approve

Please work with oncall to validate it before giving it `stable` tag

Issue #1746, opened 2020-02-26
Support running multiple scripts in a single prow job


chaodaiG:
Currently there are 2 ways of defining what runs in a Prow job:
- Use [`presubmit-tests.sh`](https://github.com/knative/test-infra/blob/master/scripts/presubmit-tests.sh) as entrypoint
- Invoke a script directly

As experimented in #1744, none of these currently work with multiple scripts, will need to figure out a way supporting running multiple scripts in a single Prow job

adrcunha:
Easiest way is to expand --run-test to allow multiple parameters, then loop through them: https://github.com/knative/test-infra/blob/01760e2303d6a17936658f155004da3019d0f43c/scripts/presubmit-tests.sh#L362

chaodaiG:
The PR #1780  doesn't work as expected  was reverted in #1790 

/reopen

chaodaiG:
Some context, this is needed after this work was done: https://github.com/knative/serving/issues/6862

It used to be in all of our test flows, there was a single entry point like `test/e2e-tests.sh`, `test/e2e-upgrade-tests.sh`. Now we need to support multiple scripts in a single flow, see this PR for example: https://github.com/knative/test-infra/pull/1783/files

chaodaiG:
Duplicate of #1811 
/close

PR #1780, opened 2020-03-04
Allow presubmit run multiple sctips

chaodaiG:
Fix: #1746 

/assign @chizhg 
/assign @coryrc

### DIFF
--- a/scripts/presubmit-tests.sh
+++ b/scripts/presubmit-tests.sh
@@ -93,19 +93,19 @@ function run_build_tests() {
   local failed=0
   # Run pre-build tests, if any
   if function_exists pre_build_tests; then
-    pre_build_tests || {failed=1; step_failed "pre_build_tests"}
+    pre_build_tests || { failed=1; step_failed "pre_build_tests"; }
   fi
   # Don't run build tests if pre-build tests failed
   if (( ! failed )); then
     if function_exists build_tests; then
-      build_tests || {failed=1; step_failed "build_tests"}
+      build_tests || { failed=1; step_failed "build_tests"; }
     else
-      default_build_test_runner || {failed=1; step_failed "default_build_test_runner"}
+      default_build_test_runner || { failed=1; step_failed "default_build_test_runner"; }
     fi
   fi
   # Don't run post-build tests if pre/build tests failed
   if (( ! failed )) && function_exists post_build_tests; then
-    post_build_tests || {failed=1; step_failed "post_build_tests"}
+    post_build_tests || { failed=1; step_failed "post_build_tests"; }
   fi
   results_banner "Build" ${failed}
   return ${failed}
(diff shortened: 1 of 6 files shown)


scripts/presubmit-tests.sh
```diff
@@ -368,7 +368,9 @@ function main() {
       header "Documentation only PR, skipping running custom test"
       exit 0
     fi
-    ${TEST_TO_RUN} || failed=1
+    for test_to_run in ${TESTS_TO_RUN[@]}; do
+      ${test_to_run} || failed=1
```
coryrc: can you change "failed=1" to something like { failed=1; echo "--- FAIL: Test ${test_to_run}"; }?
Here and all uses of lone failed=1?

chaodaiG: I think it's a good idea. Updated and made a new Spyglass entry so that these messages could be highlighted

chaodaiG:
@coryrc, another lgtm is needed
@chizhg , please take a look at spyglass change

coryrc:
/lgtm

I thought an ending ; was needed in the {} but I guess not

chaodaiG:
> /lgtm
> 
> I thought an ending ; was needed in the {} but I guess not

You're right, it's needed

chaodaiG:
Now it works: https://github.com/knative/serving/pull/7135
@coryrc

scripts/presubmit-tests.sh
```diff
@@ -368,17 +369,19 @@ function main() {
       header "Documentation only PR, skipping running custom test"
       exit 0
     fi
-    ${TEST_TO_RUN} || failed=1
+    for test_to_run in ${TESTS_TO_RUN[@]}; do
```
chizhg: Could you update the document for this change?

scripts/presubmit-tests.sh
```diff
@@ -179,7 +179,7 @@ function default_build_test_runner() {
   for pkg in ${tagged_pkgs}; do
     # `go test -c` lets us compile the tests but do not run them.
     if ! capture_output "${report}" go test -c -tags="${tags}" ${pkg} ; then
-      failed=1
+      failed=1;
```
chizhg: Not sure what's the best practice for shell, but it seems we do not need this `;`?

scripts/presubmit-tests.sh
```diff
@@ -211,19 +211,19 @@ function run_unit_tests() {
   local failed=0
   # Run pre-unit tests, if any
   if function_exists pre_unit_tests; then
-    pre_unit_tests || failed=1
+    pre_unit_tests || { failed=1; step_failed "pre_unit_tests"; }
```
chizhg: It seems we do not need this `;` in the end either?

chizhg: Hmm, I saw your comment above, so this ; in the end is needed?

scripts/presubmit-tests.sh
chaodaiG: added

scripts/presubmit-tests.sh
chaodaiG: yep it's needed

scripts/presubmit-tests.sh
chaodaiG: deleted

chizhg:
/lgtm
Thanks @chaodaiG !

PR #1781, opened 2020-03-04
Add permissions for release service accounts

chizhg:
As release related jobs are running the same set of tests as the integration tests, the service account used in these jobs should be granted the same permissions. Currently we are missing some permissions for the two service accounts on the boskos projects, and it's causing the knative-gcp auto-release and nightly-release jobs to fail, see https://testgrid.knative.dev/knative-gcp#auto-release.

This PR adds the permissions for these service accounts.

/cc @nachocano 
/cc @chaodaiG 

BTW I'm not sure why we chose to use different service accounts for tests and release jobs, @adrcunha could you please provide some background? Thanks.

### DIFF
--- a/config/prow/boskos/set_up_boskos_project.sh
+++ b/config/prow/boskos/set_up_boskos_project.sh
@@ -51,15 +51,23 @@ readonly RESOURCES=(
 
     "roles/storage.admin"
     "prow-job@knative-tests.iam.gserviceaccount.com"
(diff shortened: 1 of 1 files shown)


adrcunha:
It's a security best practice having separate service accounts for test and release jobs.

chizhg:
> It's a security best practice having separate service accounts for test and release jobs.

I see, that makes sense. Thanks!

PR #1786, opened 2020-03-04
Dop the Istio add on

dprotaso:

This was breaking the [old smoke tests in serving](https://github.com/knative/serving/issues/6810#issuecomment-593000261) (since you had two versions of istio on the cluster). 

Generally serving installs Istio via it's own yaml so we shouldn't need to install it via the add on.

Thanks for the review

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -179,9 +179,6 @@ This is a helper script for Knative E2E test scripts. To use it:
 1. By default `knative_teardown()` and `test_teardown()` will be called after
    the tests finish, use `--skip-teardowns` if you don't want them to be called.
 
-1. By default Istio is installed on the cluster via Addon, use
-   `--skip-istio-addon` if you choose not to have it preinstalled.
-
 1. You can force running the tests against a specific GKE cluster version by
    using the `--cluster-version` flag and passing a full version as the flag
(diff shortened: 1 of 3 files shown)


chizhg:
We have removed the smoke tests for serving in https://github.com/knative/serving/pull/7085, do we still need this change?

chizhg:
I think it makes sense to never install Istio add-on since it's a GKE feature, and we should not have any dependencies for GKE in the e2e tests.
I'm not sure about the history when and why we had this, cc @chaodaiG for opinions.
/cc @chaodaiG

chaodaiG:
Istio addon was added as decided by serving at that time. Since knative now has it's own istio I think we can drop it from here. But this also means that we need to change default knative setup in test-infra since the addon path is not tested anymore

chaodaiG:
Never mind, the default installer is already validated in this PR

chizhg:
Yes, that's what Dave asked in the Slack channel - why there are two e2e-tests.sh :)

chaodaiG:
I see, I'm ok with the change
/approve

chizhg:
/lgtm
/hold
In case @chaodaiG has some comments.

PR #1792, opened 2020-03-07
Modify Prow config files based on the deprecation notice

chizhg:
1. According to the deprecation notice in https://github.com/kubernetes/test-infra/blob/master/prow/ANNOUNCEMENTS.md, update Prow config file.
2. A bit improvement on the config-check script.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -18,23 +18,41 @@
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 
+# Download the Go rules
 http_archive(
     name = "io_bazel_rules_go",
-    sha256 = "f04d2373bcaf8aa09bccb08a98a57e721306c8f6043a2a0ee610fd6853dcde3d",
-    urls = ["https://github.com/bazelbuild/rules_go/releases/download/0.18.6/rules_go-0.18.6.tar.gz"],
+    urls = [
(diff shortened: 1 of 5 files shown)


chizhg:
@cjwagner could you please help check why the `checkconfig` fails for this PR? The error message is `no default decoration config provided for plank`, but that part in my PR is the same as it in k8s test-infra - https://github.com/kubernetes/test-infra/blob/master/config/prow/config.yaml#L8.

cjwagner:
> @cjwagner could you please help check why the `checkconfig` fails for this PR? The error message is `no default decoration config provided for plank`, but that part in my PR is the same as it in k8s test-infra - https://github.com/kubernetes/test-infra/blob/master/config/prow/config.yaml#L8.

It looks like you run checkconfig with bazel: https://github.com/knative/test-infra/blob/3f96c9e98f31595fe33eaa2d95f5e2170522acd7/test/presubmit-config-check.sh#L58-L61
and the version of our repo that you have is out of date so checkconfig probably doesn't know about the new config fields yet:
https://github.com/knative/test-infra/blob/3f96c9e98f31595fe33eaa2d95f5e2170522acd7/WORKSPACE#L50-L54

Check config is meant to be run at the same commit as the Prow components it is validating config for. The [gcr.io/k8s-prow/checkconfig](https://gcr.io/k8s-prow/checkconfig) image is published with the Prow component images with the same image tag so you can easily update the checkconfig image tag when you update the Prow component image tags. I'd suggest using that image directly if you can.

chizhg:
Thanks @cjwagner ! I have created https://github.com/knative/test-infra/issues/1810 for tracking and will move this check to a separate job.

chizhg:
Close this issue for now and opened #1810  and #1812 for further tracking.

PR #1793, opened 2020-03-07
Remove one TODO as prow-tests image has been updated

chizhg:
Remove one TODO in the Makefile as prow-tests image has been updated


chizhg:
The image update did not update Git, so we still cannot use the new command in Git...
/close

PR #1802, opened 2020-03-11
Run ./hack/release.sh from the release branch

### LINKED ISSUE #1262: dot-release is running the ./hack/release.sh in master branch
`eventing-contrib` repo is doing a dot-release but the job ended up with an error.

From the job logs https://prow.knative.dev/log?job=ci-knative-eventing-contrib-dot-release&id=1161353664162435072, we can see the error is `2019/08/13 19:14:03 Error enumerating files: lstat github/config/: no such file or directory`. However, `github/config/` is a path written in the `./hack/release.sh` of master branch but not 0.8 branch.

Our dot-release job should use `./hack/release.sh` in branch 0.8, instead of master.

/cc @adrcunha 

FYI @chaodaiG since he found the reason of this bug.

chizhg:
Make sure `./hack/release.sh` is finally run from the release branch.

Fixes #1262 

/cc @chaodaiG 
FYI @grantr @mattmoor

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -509,6 +509,7 @@ function main() {
     # HACK HACK HACK
     # Rerun the release script from the release branch. Fixes https://github.com/knative/test-infra/issues/1262
     ./hack/release.sh "$@"
+    exit "$?"
   fi
 
   function_exists build_release || abort "function 'build_release()' not defined"
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -501,7 +501,8 @@ function main() {
   parse_flags "$@"
 
   # Checkout specific branch, if necessary
-  local current_branch="$(git rev-parse --abbrev-ref HEAD)"
+  local current_branch
+  current_branch="$(git rev-parse --abbrev-ref HEAD)"
   if [[ -n "${RELEASE_BRANCH}" && -z "${FROM_NIGHTLY_RELEASE}" && "${current_branch}" != "${RELEASE_BRANCH}" ]]; then
     setup_upstream
     setup_branch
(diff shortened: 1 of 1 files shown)


scripts/release.sh
```diff
@@ -509,6 +509,7 @@ function main() {
     # HACK HACK HACK
     # Rerun the release script from the release branch. Fixes https://github.com/knative/test-infra/issues/1262
     ./hack/release.sh "$@"
+    exit "$?"
```
chaodaiG: are you sure stop here?

chizhg: Yes, at this point, the logic below would have been executed in the `./hack/release.sh` in the release branch, so we should stop here.

scripts/release.sh
```diff
@@ -498,9 +498,23 @@ function publish_artifacts() {
 
 # Entry point for a release script.
 function main() {
+  parse_flags "$@"
+
+  # Checkout specific branch, if necessary
+  local current_branch="$(git rev-parse --abbrev-ref HEAD)"
```
chaodaiG: I know it's rare, but `local` swallow the error after `=`
```suggestion
  local current_branch
  current_branch="$(git rev-parse --abbrev-ref HEAD)"
```

chaodaiG:
/lgtm

Feel free ignore my suggestion if it's an urgent fix

chizhg:
> /lgtm
> 
> Feel free ignore my suggestion if it's an urgent fix

It's not. I need to do some more test before merging it in.

chizhg:
I've done some tests and it works.
/hold cancel
/assign @chaodaiG

PR #1815, opened 2020-03-14
Move Prow/Testgrid config generation to update-codegen.sh

chaodaiG:
The UX for code generation isn't that friendly, user need to remember run `make config` in 2 different places, and the logic is maintained in 2 different places. Moving them to udpate-codegen.sh so that it's easier to understand

/assign @chizhg @coryrc

### DIFF
--- a/config/prow_common.mk
+++ b/config/prow_common.mk
@@ -22,10 +22,6 @@ ZONE          ?= us-central1-f
 JOB_NAMESPACE ?= test-pods
 
 SKIP_CONFIG_BACKUP        ?=
-GENERATE_MAINTENANCE_JOBS ?= true
-GENERATE_TESTGRID_CONFIG  ?= true
-
-CONFIG_GENERATOR_DIR ?= ../../tools/config-generator
 
 # Any changes to file location must be made to staging directory also
 # or overridden in the Makefile before this file is included.
@@ -43,7 +39,6 @@ BOSKOS_RESOURCES ?= boskos/boskos_resources.yaml
 
 SET_CONTEXT   := gcloud container clusters get-credentials "$(CLUSTER)" --project="$(PROJECT)" --zone="$(ZONE)"
 UNSET_CONTEXT := kubectl config unset current-context
-MAKE_CONFIG   := go run "${CONFIG_GENERATOR_DIR}"
 
 .PHONY: help get-cluster-credentials unset-cluster-credentials config
 help:
@@ -51,7 +46,6 @@ help:
 	@echo "'Update' means updating the servers and can only be run by oncall staff."
 	@echo "Common usage:"
 	@echo " make update-prow-cluster: Update all Prow things on the server to match the current branch. Errors if not master."
-	@echo " make config: Update all generated files"
 	@echo " make update-testgrid-config: Update the Testgrid config"
 	@echo " make get-cluster-credentials: Setup kubectl to point to Prow cluster"
 	@echo " make unset-cluster-credentials: Clear kubectl context"
@@ -62,22 +56,6 @@ get-cluster-credentials:
 
 unset-cluster-credentials:
 	$(UNSET_CONTEXT)
-
-# Generate the Prow config file
-config $(PROW_JOB_CONFIG) $(PROW_CONFIG) $(PROW_PLUGINS) $(TESTGRID_CONFIG): $(KNATIVE_CONFIG) $(wildcard $(CONFIG_GENERATOR_DIR)/templates/*.yaml)
-	$(MAKE_CONFIG) \
-		--env=$(ENVIRONMENT) \
-		--gcs-bucket=$(PROW_GCS) \
-		--generate-testgrid-config=$(GENERATE_TESTGRID_CONFIG) \
-		--generate-maintenance-jobs=$(GENERATE_MAINTENANCE_JOBS) \
-		--image-docker=gcr.io/$(PROJECT)/test-infra \
-		--plugins-config-output=$(PROW_PLUGINS) \
-		--prow-config-output=$(PROW_CONFIG) \
-		--prow-jobs-config-output=$(PROW_JOB_CONFIG) \
-		--prow-host=$(PROW_HOST) \
-		--testgrid-config-output=$(TESTGRID_CONFIG) \
-		--testgrid-gcs-bucket=$(TESTGRID_GCS) \
-		$(KNATIVE_CONFIG)
 
 .PHONY: update-prow-config update-prow-job-config update-prow-plugins update-all-boskos-deployments update-boskos-resource update-almost-all-cluster-deployments update-single-cluster-deployment update-prow test update-testgrid-config confirm-master
 
--- a/config/prow_common.mk
+++ b/config/prow_common.mk
@@ -40,7 +40,7 @@ BOSKOS_RESOURCES ?= boskos/boskos_resources.yaml
 SET_CONTEXT   := gcloud container clusters get-credentials "$(CLUSTER)" --project="$(PROJECT)" --zone="$(ZONE)"
 UNSET_CONTEXT := kubectl config unset current-context
 
-.PHONY: help get-cluster-credentials unset-cluster-credentials config
+.PHONY: help get-cluster-credentials unset-cluster-credentials
 help:
 	@echo "Help"
 	@echo "'Update' means updating the servers and can only be run by oncall staff."
--- a/hack/update-codegen.sh
+++ b/hack/update-codegen.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# Copyright 2019 The Knative Authors
+# Copyright 2020 The Knative Authors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/hack/update-codegen.sh
+++ b/hack/update-codegen.sh
@@ -20,5 +20,48 @@ set -o pipefail
 
 source $(dirname $0)/../scripts/library.sh
 
+CONFIG_GENERATOR_DIR="${REPO_ROOT_DIR}/tools/config-generator"
+CONFIG_DIR="${REPO_ROOT_DIR}/config/prow"
+PROW_GCS="knative-prow"
+GENERATE_TESTGRID_CONFIG="true"
+GENERATE_MAINTENANCE_JOBS="true"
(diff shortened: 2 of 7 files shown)


config/prow_common.mk
```diff
@@ -42,15 +38,13 @@ BOSKOS_RESOURCES ?= boskos/boskos_resources.yaml
 
 SET_CONTEXT   := gcloud container clusters get-credentials "$(CLUSTER)" --project="$(PROJECT)" --zone="$(ZONE)"
 UNSET_CONTEXT := kubectl config unset current-context
-MAKE_CONFIG   := go run "${CONFIG_GENERATOR_DIR}"
 
 .PHONY: help get-cluster-credentials unset-cluster-credentials config
```
coryrc: remove "config" from this list

hack/update-codegen.sh
```diff
@@ -20,5 +20,48 @@ set -o pipefail
...
 
+CONFIG_GENERATOR_DIR="${REPO_ROOT_DIR}/tools/config-generator"
+CONFIG_DIR="${REPO_ROOT_DIR}/config/prow"
+PROW_GCS="knative-prow"
+GENERATE_TESTGRID_CONFIG="true"
+GENERATE_MAINTENANCE_JOBS="true"
+PROJECT="knative-tests"
+PROW_HOST="https://prow.knative.dev"
+TESTGRID_GCS="knative-testgrid"
+KNATIVE_CONFIG="config_knative.yaml"
```
coryrc: I'd rather these be args to generate_config()?

test/presubmit-config-check.sh
```diff
@@ -16,43 +16,8 @@
 
 source $(dirname $0)/../scripts/library.sh
 
-# Run diff on the two configuration files.
-# Parameters: $1 - the regenerated temp file.
-#             $2 - the existing file.
-function diff_config_files() {
-  diff --ignore-matching-lines="^# Copyright " "$1" "$2"
```
coryrc: You lost this feature

hack/verify-codegen.sh
```diff
@@ -30,18 +35,23 @@ trap "cleanup" EXIT SIGINT
...
 
-# Save working tree state
 mkdir -p "${TMP_DIFFROOT}/pkg"
-cp -aR "${REPO_ROOT_DIR}/Gopkg.lock" "${REPO_ROOT_DIR}/vendor" "${TMP_DIFFROOT}"
+# Save working tree state
+for dir in ${DIRS_TOBE_INSPECTED[@]}; do
+  cp -aR "${REPO_ROOT_DIR}/${dir}" "${TMP_DIFFROOT}"
+done
 
 "${REPO_ROOT_DIR}/hack/update-codegen.sh"
```
coryrc: Could we run this in the temporary directory so we don't have to mess up the actual repo dir?

hack/verify-codegen.sh
```diff
@@ -30,18 +35,23 @@ trap "cleanup" EXIT SIGINT
...
 echo "Diffing ${REPO_ROOT_DIR} against freshly generated codegen"
 ret=0
-diff -Nupr --no-dereference "${REPO_ROOT_DIR}/vendor" "${TMP_DIFFROOT}/vendor" || ret=1
-diff -Nup --no-dereference "${REPO_ROOT_DIR}/Gopkg.lock" "${TMP_DIFFROOT}/Gopkg.lock" || ret=1
+for dir in ${DIRS_TOBE_INSPECTED[@]}; do
+  diff -Nupr --no-dereference "${REPO_ROOT_DIR}/${dir}" "${TMP_DIFFROOT}/${dir}" || ret=1
```
coryrc: please change "ret=1" to {ret=1; echo "--- FAIL: file difference ${REPO_ROOT_DIR}/${dir}"; }

config/prow_common.mk
chaodaiG: done

hack/update-codegen.sh
chaodaiG: I'm not sure if it's gonna be any cleaner unless we make them named params

hack/verify-codegen.sh
chaodaiG: done

hack/verify-codegen.sh
chaodaiG: done

test/presubmit-config-check.sh
chaodaiG: Is it because of this line?
```
# Copyright [[.Year]] The Knative Authors
```

If so, I would rather prefer let it fail when this is different, `update-codegen.sh` can fix it right?

chizhg:
I actually feel `make config` is more expressive and easy-to-remember (or maybe I'm just too used to it). One direction I was thinking about is to make one single `make config` command to generate config files for both prow and staging-prow, but probably there is no good way to support it with Makefile.

chizhg:
`make config` is also mentioned in https://github.com/knative/test-infra/blob/master/guides/prow_setup.md, can you update the doc as well?

hack/verify-codegen.sh
```diff
@@ -30,18 +35,24 @@ trap "cleanup" EXIT SIGINT
...
+done
 
 "${REPO_ROOT_DIR}/hack/update-codegen.sh"
 echo "Diffing ${REPO_ROOT_DIR} against freshly generated codegen"
 ret=0
-diff -Nupr --no-dereference "${REPO_ROOT_DIR}/vendor" "${TMP_DIFFROOT}/vendor" || ret=1
-diff -Nup --no-dereference "${REPO_ROOT_DIR}/Gopkg.lock" "${TMP_DIFFROOT}/Gopkg.lock" || ret=1
+for dir in ${DIRS_TOBE_INSPECTED[@]}; do
+  diff -Nupr --no-dereference "${REPO_ROOT_DIR}/${dir}" "${TMP_DIFFROOT}/${dir}" || \
+    { ret=1; echo "--- FAIL: file difference ${REPO_ROOT_DIR}/${dir}"; }
```
chizhg: Nit: would `${REPO_ROOT_DIR}/${dir} is not up-to-date`  be a better error message?

hack/update-codegen.sh
```diff
@@ -20,5 +20,48 @@ set -o pipefail
 
 source $(dirname $0)/../scripts/library.sh
 
+CONFIG_GENERATOR_DIR="${REPO_ROOT_DIR}/tools/config-generator"
```
chizhg: Would it be more clear if we move the code to a separate script called `./hack/update-configs.sh`, and run it here like how we run `./hack/update-deps.sh`?

test/presubmit-config-check.sh
coryrc: Right, this is correct.
We added this because you can't approve a PR (like an automated one) from two days ago if today happens to be Jan 2. It's really annoying.

hack/update-codegen.sh
coryrc: Then why not just keep it in `make` and call it there?

I think it makes sense that `update-codegen.sh` calls it, but I don't care as much where the code lives.

test/presubmit-config-check.sh
chaodaiG: For this once in a year thing which can be fixed by `update-codegen.sh`, my personal preference would be just let it fail instead of this hack. WDYT? @chizhg for opinion as well

hack/update-codegen.sh
coryrc: I don't like that inputs are all global vars, because if someone misses one in a later change, it would be hard to detect.

hack/verify-codegen.sh
coryrc: I meant, can we run `hack/update-codegen.sh` from the temporary directory instead of the repo directory? If not, that's fine, but if we run this on our own machines, it will overwrite local changes to our repo, report things are broken, but it will then be compliant, which is confusing user experience...

test/presubmit-config-check.sh
coryrc: I was wrong, all build tests start failing in the new year, every single year:
https://github.com/knative/test-infra/pull/1606
It needs to be kept.

hack/update-codegen.sh
chizhg: Then developers can run `./hack/update-configs.sh` if they ONLY want to update the config files? Otherwise they will have to run `./hack/update-codegen.sh` and it takes a much longer time.

chaodaiG: agree with Chi, I think it makes sense moving to a separate file

chaodaiG: done

hack/update-codegen.sh
chaodaiG: I agree with you, removed all global vars, but just invoke `go run` twice seems cleaner to me. We can revisit this if it ever grows

hack/verify-codegen.sh
chaodaiG: This is the same across all repos, so far haven't seen any problems yet, so might just keep it

test/presubmit-config-check.sh
coryrc: Thinking about this more, given these files are generated by a computer program, they shouldn't have a copyright at all!
The source files can, but their copyright should be fixed by the author and not dynamically. So this is the appropriate fix: #1833

hack/verify-codegen.sh
coryrc: k. Feel free to resolve

hack/update-codegen.sh
coryrc: It was already in a separate file, just leave it in the Makefile then...

hack/update-codegen.sh
coryrc: Yes, I like this result the best.

hack/update-codegen.sh
chaodaiG: it's a single invocation vs. multiple invocations, I can make the scripts do 2 `make` call just to make the user experience slightly better

test/presubmit-config-check.sh
chaodaiG: Agree

chizhg:
@chaodaiG could you rebase it? It has merge conflicts.

chaodaiG:
> @chaodaiG could you rebase this PR? It has merge conflicts.

done. thanks

chizhg:
/lgtm
/hold
In case @coryrc has other comments.

Issue #459, opened 2019-02-11
More prow jobs doing less stuff


mattmoor:
We've been cramming stuff (like markdown checks) into existing CI legs, which:
1. Makes it less clear from the failing job what exactly failed
2. Makes each leg longer (at a minimum this is relevant to flake reruns)

My recollection is that the k8s.io/test-infra folks have separated out things like `verify-bazel` and whatnot into separate legs, so it's clear even without making the jump to gubernator what's wrong.

adrcunha:
At a minimum, this depends on #382, or our already-hard-to-read Prow config will become as terrible as k8s' Prow config.

It also requires refactoring how we run presubmit tests (and thus the CI tests), which seems to offer a good opportunity to:

1. migrate the (already complex) bash scripts to Go (preferred) or some other language
1. drop kubetest (#424)
1. show better errors in build tests (#428)

Of course, there's the dirty way of doing it, which basically means adding more flags to `presubmit-tests.sh` and replicating the presubmit jobs.

adrcunha:
[WIP] Assorted comments about the new test-infra:
* it should allow running old tests (e.g., running tests from the 0.5 branch, even though HEAD is now 0.6)
* prefer providing test services  as opposed to providing shared code

evankanderson:
One additional observation (which may be obvious) for the new test setup:

Test setup should probably be thought as a nesting set (stack) of operations. At a minimum:

1. Provision resources (i.e. GKE cluster, kind cluster, etc for e2e tests)
2. Setup environment (i.e. load base CRDs, controllers)
3. Create test-specific resources and validate behavior (within go test, typically)

It seems like step 1 and step 2 should be separable -- step 1 would be the responsibility of the test-infra repo, while step 2 would be the responsibility of the repo under test (i.e. via an operator, helm chart, or `ko apply`). For manual testing, users would start at either 3 or 2 in most cases, but could also have a harness to perform 1 (and there could be different harnesses for 1 for kind, GKE, etc).

evankanderson:
Thinking about this a bit more:

It seems like `test-infra` should run a tool which performs step 1 and can bootstrap step 2 via some sort of contract.

It seems like step 2 should live in the component repo, and should be responsible for building the necessary artifacts (docker containers, etc) and then launching them on the cluster.

[...]

PR #1816, opened 2020-03-15
Run verify-codegen for docs only PRs

### LINKED ISSUE #1799: Verify codegen for docs only PR
CI skips e2e tests for docs only PRs, but it should run the complete build tests (verify codegen etc).

In client, we noticed a docs only PR had 'autgenerated code out of date' but build-tests failed to catch that (https://github.com/knative/client/pull/710).

To check, here is a debug PR https://github.com/knative/client/pull/717 with a couple of commits:
1. First; docs only change in commit with out of date autogenerated code, but build tests passed for it. [logs](https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/knative_client/717/pull-knative-client-build-tests/1237121301668171776#1:build-log.txt%3A82)

2. Second; added a subsequent commit with code change and build tests runs properly to catch the codegen errors. [logs](https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/knative_client/717/pull-knative-client-build-tests/1237225238014464000#1:build-log.txt%3A804)

More logs: [PR history](https://prow.knative.dev/pr-history/?org=knative&repo=client&pr=717)

chizhg:
For docs only PRs, still run `verify-codegen` check.

Fixes #1799 

/cc @chaodaiG @coryrc 
FYI @navidshaikh


chaodaiG:
The code change looks good to me. However, my feeling is that it might not be preferable for every repo to verify codegen for md only changes. How about we make this another leg of presubmit, and make each repo override by themselves.

chizhg:
> The code change looks good to me. However, my feeling is that it might not be preferable for every repo to verify codegen for md only changes. How about we make this another leg of presubmit, and make each repo override by themselves.

Not sure if I got your point: do you mean adding a new env var like `SKIP_CODEGEN_CHECK_FOR_DOC_PRS`? Or if it meant something else, but my concern is adding such extra logic is pretty hidden and can take time for another repo that needs this to find out this special logic and use it correctly (like knative/client found this error in a very specific PR). 

Given `verify codegen` usually does not take much time, and this change will only impact doc-only PRs, my personal feeling is this approach is better than adding extra complexity.

chaodaiG:
My point was instead of make it part of "build test", we should have another test just named "verify codegen", and default is skip when it's documentation only. It's really simple when a repo needs to override:
```
function verify_codegen() {
  ./hack/verify-codegen.sh
}
```

chizhg:
It seems what you meant is https://github.com/knative/test-infra/issues/459. I agree with you and the issue that we should move `verify codegen` out of `build tests`, but it also involves `md checks` and `license checks`, which will all require non-trivial changes for the scripts, existing Prow jobs and config-generator . 
I can work on it later when I get extra bandwidth, but for now let's fix the issue first?

chaodaiG:
sgtm

/lgtm

PR #1822, opened 2020-03-17
Minor improvements to the config-generator

chizhg:
- There is no concept of legacy branches now as all repos have been converted to use the knative.dev path alias, so remove the setting
- If there is no go112-branches in the repo settings, default the image to `prow-tests:stable`, this helps get rid of the hacky settings in order to use Go 113 image.

/cc @chaodaiG


chaodaiG:
Awesome! We're one step closer to be clean!

/lgtm

Don't be too excited yet, there is going to be `go113` when go1.14 is fully released

PR #1843, opened 2020-03-23
Add postsubmit job for Prow config updater

chizhg:
Add postsubmit job for Prow config updater

Note:
It's becoming quite annoying to use config generator to add a new maintenance job. I strongly feel this kind of jobs should be added manually as it's normally a one-time config and is tied to a specific tool (regardless whether we want to keep maintaining the current config generator, or seek alternative solution)

/cc @chaodaiG

### DIFF
--- a/config/prow/config_knative.yaml
+++ b/config/prow/config_knative.yaml
@@ -253,6 +253,8 @@ presubmits:
       dot-dev: true
     - integration-tests: true
       dot-dev: true
+    - go-coverage: true
+      dot-dev: true
     - custom-test: check-config
       dot-dev: true
       command:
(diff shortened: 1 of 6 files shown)


chaodaiG:
Strongly agree that adding a maintenance job is cumbersome! For the short term, I have an alternative solution: making a maintenance job template, and overrides are read directly from `knative-config.yaml`. The reason for this is that the top portion of the maintenance jobs are almost identical, this can save some copy/paste.
I'm OK with your approach too, since we are not maintaining that much maintenance jobs right now
/lgtm

/hold
unhold when oncall is ready :)

chaodaiG:
Is the build cluster ready?

chizhg:
> Is the build cluster ready?

Yes, I have created secrets on it.

PR #1844, opened 2020-03-23
Add restrictions for Prow staging process

chizhg:
1. Do not allow developers to directly change production Prow core and cluster configs, as prow-config-updater tool will automatically roll out staging changes to production.
2. prow-auto-bumper tool only updates staging Prow config files.

Plus some fixes to the prow-auto-updater tool:
1. Running the prow-updater-tool requires `bazel`, `gcloud` and `kubectl`, I'm now building the image based on `kubekins-e2e`, will make optimizations by only installing components we need. I have added a TODO.
2. Add the missed `Makefile` in the prow-auto-updater tool
3. In running command to update `testgrid`, use the env vars as it needs the `GOOGLE_APPLICATION_CREDENTIALS` to be set.


tools/prow-config-updater/config/config.go
```diff
@@ -54,39 +55,42 @@ const (
 
 var (
 	// Prow config root paths.
-	prodProwConfigRoot    = filepath.Join(configPath, string(ProdProwEnv))
-	stagingProwConfigRoot = filepath.Join(configPath, string(StagingProwEnv))
+	ProdProwConfigRoot    = filepath.Join(configPath, string(ProdProwEnv))
+	StagingProwConfigRoot = filepath.Join(configPath, string(StagingProwEnv))
+	// Prow config templates paths.
+	ProdProwConfigTemplatesPath = filepath.Join(configTemplatePath, string(ProdProwEnv))
```
mattmoor: Format Go code:
```suggestion
	ProdProwConfigTemplatesPath    = filepath.Join(configTemplatePath, string(ProdProwEnv))
```

test/presubmit-config-check.sh
```diff
@@ -18,6 +18,10 @@ source $(dirname $0)/../scripts/library.sh
 
 set -e
 
+trap 'echo "--- FAIL: Directly changing production Prow config files is not allowed, please revert the change."' ERR
```
chaodaiG: let's use add_trap instead

test/presubmit-config-check.sh
```diff
@@ -18,6 +18,10 @@ source $(dirname $0)/../scripts/library.sh
...
+header "Checking Prow config files are not modified by mistake"
```
chaodaiG:
```suggestion
header "Checking to make sure Prow productions config files are not modified manually"
```

test/presubmit-config-check.sh
chaodaiG: Also, the error message should be more actionable, like move the changes to staging dir

chaodaiG:
/lgtm

/hold
in case you still have last minute change

test/presubmit-config-check.sh
chizhg: We don't have `add_trap` function in `library.sh`. Do you want me to add it there?

And I changed the error message.

chaodaiG: I think it's fine for this script, since these 2 traps don't collide with each other

PR #1848, opened 2020-03-24
Bug fixes for prow-config-updater

chizhg:

1. In the `make` command to update Prow configs, we use `gcloud`. But `gcloud` does not use `GOOGLE_APPLICATION_CREDENTIALS` env var, see the answer [here](https://serverfault.com/questions/848580/how-to-use-google-application-credentials-with-gcloud-on-a-server). So if `GOOGLE_APPLICATION_CREDENTIALS` is provided, run `gcloud auth activate-service-account` to use this key file before running the `make` command.

2. The wildcard `*` needs to be expanded by shell, so the `cp` command has to be run in a shell process. See the answer [here](https://stackoverflow.com/questions/41685393/exec-command-with-cp-exits-with-status-1)

3. A couple of nit fixes and improvements based on the tests result.

4. The `knative-prow-robot` is using a different email address `adrcunha+knative-prow-robot@google.com`, so change the Prow job.

/cc @chaodaiG

### DIFF
--- a/tools/prow-config-updater/config/config.go
+++ b/tools/prow-config-updater/config/config.go
@@ -110,6 +110,8 @@ func UpdateProw(env ProwEnv, dryrun bool) error {
 	return helpers.Run(
 		fmt.Sprintf("Updating Prow configs with command %q", updateCommand),
 		func() error {
+			// Use the default GOOGLE_APPLICATION_CREDENTIALS to authenticate with the gcloud services,
+			// it will fallback to use the local credentials if the env var does not exist.
 			kf := os.Getenv("GOOGLE_APPLICATION_CREDENTIALS")
 			if kf != "" {
 				authCommand := "gcloud auth activate-service-account --key-file=" + kf
--- a/tools/prow-config-updater/config/config.go
+++ b/tools/prow-config-updater/config/config.go
@@ -110,7 +110,14 @@ func UpdateProw(env ProwEnv, dryrun bool) error {
 	return helpers.Run(
 		fmt.Sprintf("Updating Prow configs with command %q", updateCommand),
 		func() error {
-			out, err := cmd.RunCommand(updateCommand)
+			kf := os.Getenv("GOOGLE_APPLICATION_CREDENTIALS")
+			if kf != "" {
+				authCommand := "gcloud auth activate-service-account --key-file=" + kf
+				if _, err := cmd.RunCommand(authCommand); err != nil {
+					return fmt.Errorf("error activating service account with %q", kf)
+				}
+			}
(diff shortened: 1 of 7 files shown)


tools/prow-config-updater/config/config.go
```diff
@@ -110,7 +110,14 @@ func UpdateProw(env ProwEnv, dryrun bool) error {
 	return helpers.Run(
 		fmt.Sprintf("Updating Prow configs with command %q", updateCommand),
 		func() error {
-			out, err := cmd.RunCommand(updateCommand)
+			kf := os.Getenv("GOOGLE_APPLICATION_CREDENTIALS")
+			if kf != "" {
```
chaodaiG: I think this deserves a comment, if this is not set then we assume it's local?

tools/prow-config-updater/config/config.go
```diff
@@ -110,7 +110,14 @@ func UpdateProw(env ProwEnv, dryrun bool) error {
...
+				authCommand := "gcloud auth activate-service-account --key-file=" + kf
+				if _, err := cmd.RunCommand(authCommand); err != nil {
+					return fmt.Errorf("error activating service account with %q", kf)
```
chaodaiG: don't we want to print error?

tools/prow-config-updater/config/config.go
```diff
@@ -110,7 +110,14 @@ func UpdateProw(env ProwEnv, dryrun bool) error {
...
+				}
+			}
+			out, err := cmd.RunCommand(updateCommand, cmd.WithEnvs(os.Environ()))
```
chaodaiG: do we have to pass `cmd.WithEnvs(os.Environ())` every time now? if so I think `cmd` package needs to be fixed

tools/prow-config-updater/config/config.go
chizhg: Added the comment. If the env var is not set, it will fallback to use the local credential.

tools/prow-config-updater/config/config.go
chizhg: We don't need to, if there is an error, it will be shown as a comment on the pull request.

tools/prow-config-updater/config/config.go
chizhg: We don't have to pass it every time, e.g. the `cp` command does not use it (we don't have to add it here either, but just adding it for safety)
And it does not necessarily have to be `os. Environ()`, it can be any env vars that users want to set for running this command.

chaodaiG:
/lgtm

/hold
please check coverage job

chizhg:
> /lgtm
> 
> /hold
> please check coverage job

The error is fixed. Now it fails because the coverage is below the threshold. I will try to bump it up later in followup PRs.

PR #1849, opened 2020-03-24
Update vendor to use pkg test

chizhg:
Some libs have been moved to pkg to avoid circular dependencies, update test-infra to vendor them.


pkg/testgrid/testgrid_test.go
```diff
@@ -56,7 +57,7 @@ func TestXMLOutput(t *testing.T) {
 `
 
 	// Create a test file
-	if err := CreateXMLOutput(tc, name); err != nil {
+	if err := pkgTestgrid.CreateXMLOutput(tc, name); err != nil {
```
chaodaiG, line 60: This test should like in pkg repo

chaodaiG:
I don't understand why testgrid coverage is 0 now

pkg/testgrid/testgrid_test.go
chizhg: It's already in https://github.com/knative/pkg/blob/master/test/testgrid/testgrid_test.go, I just forgot to delete it here.

Deleted.

chizhg:
> I don't understand why testgrid coverage is 0 now

It's because the current test is for `testgrid/yaml.go` but not `testgrid/testgrid.go`

chaodaiG:
> > I don't understand why testgrid coverage is 0 now
> 
> It's because the current test is for `testgrid/yaml.go` but not `testgrid/testgrid.go`

Should it be renamed to `yaml_test.go`?

chizhg:
> > > I don't understand why testgrid coverage is 0 now
> > 
> > 
> > It's because the current test is for `testgrid/yaml.go` but not `testgrid/testgrid.go`
> 
> Should it be renamed to `yaml_test.go`?

Ah yes, done.

PR #1840, opened 2020-03-21
Add net-http01 to meta config

mattmoor:
cc @chaodaiG 

Looks like `make config` isn't the right thing anymore, so I'm opening this hoping CI will tell me if anything is wrong 🤞

PR #1851, opened 2020-03-24
Set up https://github.com/knative-sandbox/operator

evankanderson:
## What this PR does, why we need it:

This sets up initial prow configuration for https://github.com/knative-sandbox/operator, the location of the new unified operator for Knative, sponsored by the operations WG.

## Special notes to reviewers:

I think I passed the typing test, but please double check.

Also, this is set up in a new org, so there may need to be some additional permissions checks and may be even new tokens provisioned (not sure0.

## User-visible changes in this PR:

* Initial automation and testgrid integration for the unified operator.

## Side note

There seems to be a lot of duplication in the Prow configuration. It might be worth suggesting some form of either defaulting or templating support in Prow to the upstream maintainers. I found myself needing to type the same "agent:Kubernetes\ndecorate: true" line many times, as well as repeating the `name` of the configuration stanza at least 3 times within the block.

### DIFF
--- a/config/prow/config_knative.yaml
+++ b/config/prow/config_knative.yaml
@@ -404,13 +404,9 @@ presubmits:
 
   knative-sandbox/operator:
     - build-tests: true
-      dot-dev: true
     - unit-tests: true
-      dot-dev: true
     - integration-tests: true
-      dot-dev: true
     - go-coverage: true
-      dot-dev: true
 
 periodics:
   knative/serving:
@@ -978,14 +974,14 @@ periodics:
 
   knative-sandbox/operator:
     - continuous: true
-      dot-dev: true
       go113: true
     - nightly: true
-      dot-dev: true
       go113: true
     - dot-release: true
-      dot-dev: true
       go113: true
+      env-vars:
+      - ORG_NAME=knative-sandbox
     - auto-release: true
-      dot-dev: true
       go113: true
+      env-vars:
+      - ORG_NAME=knative-sandbox
--- a/config/prow/config_knative.yaml
+++ b/config/prow/config_knative.yaml
@@ -402,6 +402,16 @@ presubmits:
     - integration-tests: false
     - go-coverage: false
 
+  knative-sandbox/operator:
+    - build-tests: true
+      dot-dev: true
+    - unit-tests: true
+      dot-dev: true
+    - integration-tests: true
+      dot-dev: true
+    - go-coverage: true
+      dot-dev: true
+
 periodics:
   knative/serving:
     - continuous: true
@@ -953,6 +963,20 @@ periodics:
       go113: true
 
   knative/net-kourier:
+    - continuous: true
+      dot-dev: true
+      go113: true
+    - nightly: true
+      dot-dev: true
+      go113: true
+    - dot-release: true
+      dot-dev: true
+      go113: true
+    - auto-release: true
+      dot-dev: true
+      go113: true
+
+  knative-sandbox/operator:
     - continuous: true
       dot-dev: true
       go113: true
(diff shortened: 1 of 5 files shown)


config/prow/jobs/config.yaml
```diff
@@ -4671,6 +4671,105 @@ presubmits:
...
+        - "--github-token=/etc/covbot-token/token"
+        volumeMounts:
+        - name: covbot-token
+          mountPath: /etc/covbot-token
+          readOnly: true
+      volumes:
+      - name: covbot-token
+        secret:
+          secretName: covbot-token
+
```
mattmoor: Remove trailing whitespace:
```suggestion

```

chizhg:
It seems you added the config files in a wrong way, the only file you need to make changes is `config/prow/config_knative.yaml`, and then all other config files will be auto-generated after running `./hack/generate_configs.sh`, see more details in https://github.com/knative/test-infra/blob/master/guides/prow_setup.md (the doc is a bit outdated though, I will make the improvements soon)

evankanderson:
From #1840, I realized that I needed to run `./hack/update-codegen.sh`, which made slightly different changes than I'd done by hand.

evankanderson:
Also, `./hack/update-codegen.sh` actually fails for me, so I ran `./hack/generate-configs.sh` instead:

```shell
$ ./hack/update-codegen.sh 
/Users/evana/go/src/knative.dev/test-infra/hack/update-deps.sh: line 44: DEP_FLAGS[@]: unbound variable
```

config/prow/config_knative.yaml
```diff
@@ -965,3 +975,17 @@ periodics:
...
       go113: true
+
+  knative-sandbox/operator:
+    - continuous: true
+      dot-dev: true
+      go113: true
+    - nightly: true
+      dot-dev: true
+      go113: true
+    - dot-release: true
```
chizhg: We need a hack to set the org name as env var for release jobs for non-knative projects, see https://github.com/knative/test-infra/blob/master/config/prow/config_knative.yaml#L906. 
So please add below lines to these two job settings:
```
env-vars:
    - ORG_NAME=knative-sandbox
```

config/prow/config_knative.yaml
```diff
@@ -402,6 +402,16 @@ presubmits:
     - integration-tests: false
     - go-coverage: false
 
+  knative-sandbox/operator:
+    - build-tests: true
+      dot-dev: true
```
chizhg: Do projects in this org also use the `knative.dev` path alias for dep? If not, please remove all `dot-dev: true` settings.

chizhg:
> Also, `./hack/update-codegen.sh` actually fails for me, so I ran `./hack/generate-configs.sh` instead:
> 
> ```shell
> $ ./hack/update-codegen.sh 
> /Users/evana/go/src/knative.dev/test-infra/hack/update-deps.sh: line 44: DEP_FLAGS[@]: unbound variable
> ```

Hmm, I'm running it on MAC but do not see the error. Which system are you using?

evankanderson:
I'm using the following `bash`:

```shell
$ bash --version
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin18)
Copyright (C) 2007 Free Software Foundation, Inc.
```

config/prow/config_knative.yaml
evankanderson: We haven't figured this out yet, so I'll remove it.

config/prow/config_knative.yaml
evankanderson: So this is only for `-release` steps, correct?

Updated.

chizhg:
> I'm using the following `bash`:
> 
> ```shell
> $ bash --version
> GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin18)
> Copyright (C) 2007 Free Software Foundation, Inc.
> ```

Thanks, I have created https://github.com/knative/test-infra/issues/1856 for tracking.

chizhg:
/lgtm
/approve
Thanks!

PR #1861, opened 2020-03-27
Update janitor to make it wokr

chaodaiG:
Only changed staging, assume this is how it's working now?
/assign @chizhg


chaodaiG:
There is a chicken and egg problem, integration test will for sure fail when no boskos is available
/override pull-knative-test-infra-integration-tests

PR #1868, opened 2020-03-30
fix two shell syntax issues

rhuss:
* IS_PROW used in [[ ]] instead of $IS_PROW or within (( ))
* -gt for numerical comparison


chizhg:
Thanks for the fix!
/lgtm
/approve

Issue #1855, opened 2020-03-25
Improve the prow_setup doc


chizhg:
The [prow_setup](https://github.com/knative/test-infra/blob/master/guides/prow_setup.md) doc is now outdated, and some words are not accurate. We need to improve it.

/assign

PR #1869, opened 2020-03-30
chore: Remove gocoverage for client-contrib

rhuss:
Client contrib dispatches the build over various sub plugin directories.
This works nicely with shell scripts called, but not
for coverage where go code is expected in the top-level dir.

### DIFF
--- a/config/prow/config_knative.yaml
+++ b/config/prow/config_knative.yaml
@@ -178,8 +178,6 @@ presubmits:
       dot-dev: true
     - integration-tests: true
       dot-dev: true
-    - go-coverage: true
-      dot-dev: true
 
   knative/eventing:
     - repo-settings:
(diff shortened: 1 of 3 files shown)


rhuss:
One question to a the `custom-test` entry. Is there any documentation how this could be used ? (also, what again is this `dot-dev` field ? and other fields like `alaways_run` ?

chizhg:
`dot-dev` field is needed when you are using `knative.dev` instead of `github.com/knative` in the import path.
`always-run` field is needed if you want the job to be run in every pull request. If it's not set, the job will not be run unless you type `/test [job_name]` in the PR comment.

We are missing document on this. I have created https://github.com/knative/test-infra/issues/1855 for tracking but haven't got time to work on it...

PR #1881, opened 2020-04-01
Enable API for boskos projects to support Workload Identity testing

chizhg:
As requested in https://github.com/google/knative-gcp/pull/709#issuecomment-607025735, enable a new API for boskos projects.

/cc @grac3gao 
FYI @chaodaiG @yt3liu


grac3gao:
Thanks for the quick enabling!

PR #1883, opened 2020-04-01
Add the kntest skeleton code and cluster subcommands

chizhg:
1. Add `kntest` skeleton code
2. Add cluster subcommands

Note:
1.  I'm placing `kntest` under the root folder instead of `tools/` to make it more noticeable to other contributors.
2. The code of cluster subcommands is mostly copied from `knative/pkg/testutils/prow-cluster-operation`, I think we should remove `knative/pkg/testutils/prow-cluster-operation` after this PR is merged.
3. I also want to make `knative/pkg/testutils/junithelper` and `knative/pkg/testutils/metahelper` as subcommands of `kntest tools`, then we can have one central point to maintain these tools. Please let me know how you feel about that.


kntest/README.md
```diff
@@ -0,0 +1,9 @@
+# kntest
+
+kntest is developed to simplify the [shell scripts](../scripts) used by all Knative projects and
+make the test process more robust.
```
mattmoor, line 4: Remove trailing whitespace:
```suggestion
make the test process more robust.
```

mattmoor, line 4: Format markdown:
```suggestion
kntest is developed to simplify the [shell scripts](../scripts) used by all
Knative projects and make the test process more robust.
```

kntest/README.md
```diff
@@ -0,0 +1,9 @@
+# kntest
+
+kntest is developed to simplify the [shell scripts](../scripts) used by all Knative projects and
+make the test process more robust. 
+
+This project is still under active development, and this doc will contain more details
+in the future.
```
mattmoor, line 7: Format markdown:
```suggestion
This project is still under active development, and this doc will contain more
details in the future.
```

kntest/README.md
```diff
@@ -0,0 +1,9 @@
...
+
+See also [the design doc](https://docs.google.com/document/d/1-pGZaITHwZrMxFLIvBoOhgyh-vSb0z4xpG4yCEouIzs/edit) shared with Knative community.
```
mattmoor, line 9: Format markdown:
```suggestion
See also
[the design doc](https://docs.google.com/document/d/1-pGZaITHwZrMxFLIvBoOhgyh-vSb0z4xpG4yCEouIzs/edit)
shared with Knative community.
```

kntest/README.md
```diff
@@ -0,0 +1,11 @@
+# kntest
+
+kntest is developed to simplify the [shell scripts](../scripts) used by all Knative projects and
+make the test process more robust.
```
mattmoor, line 4: Format markdown:
```suggestion
kntest is developed to simplify the [shell scripts](../scripts) used by all
Knative projects and make the test process more robust.
```

kntest/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+This project is still under active development, and this doc will contain more
+details in the future.
+
+See also
```
mattmoor, line 9: Format markdown:
```suggestion
See also
[the design doc](https://docs.google.com/document/d/1-pGZaITHwZrMxFLIvBoOhgyh-vSb0z4xpG4yCEouIzs/edit)
shared with Knative community.
```

kntest/README.md
mattmoor, line 4: Format markdown:
```suggestion
kntest is developed to simplify the [shell scripts](../scripts) used by all
Knative projects and make the test process more robust.
```

kntest/README.md
mattmoor, line 9: Format markdown:
```suggestion
See also
[the design doc](https://docs.google.com/document/d/1-pGZaITHwZrMxFLIvBoOhgyh-vSb0z4xpG4yCEouIzs/edit)
shared with Knative community.
```

kntest/README.md
mattmoor, line 4: Format markdown:
```suggestion
kntest is developed to simplify the [shell scripts](../scripts) used by all
Knative projects and make the test process more robust.
```

kntest/README.md
```diff
@@ -0,0 +1,11 @@
+# kntest
+
+kntest is developed to simplify the [shell scripts](../scripts) used by all
+Knative projects and make the test process more robust.
```
mattmoor, line 4: Format markdown:
```suggestion
kntest is developed to simplify the [shell scripts](../scripts) used by all
Knative projects and make the test process more robust.
```

chizhg:
@googlebot I consent

kntest/pkg/cluster/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Prerequisite
+
+- `GOOGLE_APPLICATION_CREDENTIALS` set
+
+## Usage
+
+This tool can be invoked from command line with following parameters:
+
+- `--min-nodes`: minumum number of nodes, default 1
```
mattmoor, line 13: Found potential misspelling(s):
```suggestion
- `--min-nodes`: minimum number of nodes, default 1
```

chizhg:
/hold
Wait for https://github.com/knative/pkg/pull/1186 to be merged and change cluster deletion to be async.

kntest/pkg/cluster/README.md
```diff
@@ -0,0 +1,58 @@
+## prow-cluster-operation
```
chaodaiG, line 1: Change this

kntest/pkg/cluster/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+`kntest cluster` command is used for creating, deleting or getting a GKE cluster
```
chaodaiG, line 3: let's consider adding another subcommand
```
kntest cluster gke
```

coryrc: It needs to be a single command that chooses based on environment variables.

Otherwise, we have to write a bunch of bash to decide which cluster command to call.... that's the whole thing we're trying to avoid.

kntest/pkg/cluster/ops/create.go
```diff
@@ -40,6 +39,35 @@ const (
...
+		return nil, fmt.Errorf("failed acquiring GKE cluster: '%v'", err)
+	}
+
+	// At this point we should have a cluster ready to run test. Need to save
+	// metadata so that following flow can understand the context of cluster, as
+	// well as for Prow usage later
+	writeMetaData(gkeOps.Cluster, gkeOps.Project)
+
+	// set up kube config points to cluster
+	// TODO(chaodaiG): this probably should also be part of clustermanager lib
```
coryrc, line 57: Or not?

kntest/pkg/cluster/ops/create.go
```diff
@@ -40,6 +39,35 @@ const (
 	projectKey        = "E2E:Project"
 )
 
+// Create creates a GKE cluster and configures gcloud after successful GKE create request
+func Create(o *RequestWrapper) (*clm.GKECluster, error) {
```
coryrc, line 43: These should be methods on (o *RequestWrapper)

kntest/pkg/cluster/ops/delete.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The Knative Authors
+Copyright 2020 The Knative Authors
```
coryrc, line 2: Don't change the copyright date; it should always be the earliest

kntest/pkg/cluster/options.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"github.com/spf13/cobra"
+
+	"knative.dev/test-infra/kntest/pkg/cluster/ops"
+)
+
+func addOptions(clusterCmd *cobra.Command, rw *ops.RequestWrapper) {
+	pf := clusterCmd.PersistentFlags()
+	req := &rw.Request
+	pf.Int64Var(&req.MinNodes, "min-nodes", 0, "minimal number of nodes")
```
coryrc, line 28: min & max nodes should have good defaults

kntest/pkg/cluster/options.go
```diff
@@ -0,0 +1,41 @@
...
+	pf.Int64Var(&req.MaxNodes, "max-nodes", 0, "maximal number of nodes")
+	pf.StringVar(&req.NodeType, "node-type", "", "node type")
```
coryrc, line 30: The rest of these should not be arguments -- they need to be passed in via the environment.

Think about how Redhat folks would update this to work in their environment

test/e2e-tests.sh
```diff
@@ -28,27 +28,29 @@ source $(dirname $0)/../scripts/e2e-tests.sh
 # Read metadata.json and get value for key
 # Parameters: $1 - Key for metadata
 function get_meta_value() {
-  go run ${REPO_ROOT_DIR}/vendor/knative.dev/pkg/testutils/metahelper --get $1
+  go run "${REPO_ROOT_DIR}"/vendor/knative.dev/pkg/testutils/metahelper --get "$1"
```
coryrc, line 31: The second quote should actually be just after "metahelper"

kntest/pkg/cluster/ops/create.go
chizhg: I think this comment can be removed, but cc @chaodaiG for opinions.

kntest/pkg/cluster/options.go
coryrc: Add that as a comment?

kntest/pkg/cluster/README.md
chizhg: Even the cluster provider can be an environment variable, e.g. the command can be called with `kntest cluster ${PROVIDER}`, so it doesn't matter how we provide the commands IMO.

chizhg: I'm going to support both env vars and flags, the kntest command will use the env vars by default, but users can also use flags to overwrite them. Does it sound a reason direction to you guys?

kntest/pkg/cluster/ops/create.go
chaodaiG: It doesn't really matter at this point, kntest will play it out

kntest/pkg/cluster/README.md
chaodaiG: I'm OK kntest supporting some env vars, but hesitating on extensively using env vars. Do we have a boundary here? Also brings in more opinions @mattmoor

chizhg: I'm going to withdraw what I said above - in this PR I'm not going to add any support for env vars in `kntest`, I'm going to support `kntest cluster gke` as a subcommand though.

My point is the current `kntest` design is focusing on simplifying the CI/CD flow we currently have, and it's still not clear how we want to support the downstream processes. So I'd rather not adding it at this point but leave it to the next step of `kntest`.

kntest/pkg/cluster/gke/README.md
```diff
@@ -0,0 +1,58 @@
+## kntest cluster gke
+
+`kntest cluster gke` command is used for creating, deleting or getting a GKE cluster
```
mattmoor, line 3: Format markdown:
```suggestion
`kntest cluster gke` command is used for creating, deleting or getting a GKE
cluster
```

kntest/cmd/kntest/main.go
```diff
@@ -0,0 +1,43 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+    "log"
```
mattmoor, line 20: Format Go code:
```suggestion
	"log"
```

kntest/cmd/kntest/main.go
```diff
@@ -0,0 +1,43 @@
...
+
+    "github.com/spf13/cobra"
```
mattmoor, line 22: Format Go code:
```suggestion
	"github.com/spf13/cobra"
```

kntest/pkg/cluster/commands.go
```diff
@@ -0,0 +1,16 @@
+package cluster
+
+import (
+    "github.com/spf13/cobra"
```
mattmoor, line 4: Format Go code:
```suggestion
	"github.com/spf13/cobra"
```

kntest/pkg/cluster/commands.go
```diff
@@ -0,0 +1,16 @@
...
+
+    "knative.dev/test-infra/kntest/pkg/cluster/gke"
```
mattmoor, line 6: Format Go code:
```suggestion
	"knative.dev/test-infra/kntest/pkg/cluster/gke"
```

kntest/pkg/cluster/commands.go
```diff
@@ -0,0 +1,16 @@
...
+)
+
+func AddCommands(topLevel *cobra.Command) {
+    var clusterCmd = &cobra.Command{
+        Use:   "cluster",
+        Short: "Cluster related commands.",
+    }
```
mattmoor, line 13: Format Go code:
```suggestion
	var clusterCmd = &cobra.Command{
		Use:   "cluster",
		Short: "Cluster related commands.",
	}
```

kntest/pkg/cluster/commands.go
```diff
@@ -0,0 +1,16 @@
...
+
+    gke.AddCommands(clusterCmd)
```
mattmoor, line 15: Format Go code:
```suggestion
	gke.AddCommands(clusterCmd)
```

kntest/pkg/cluster/gke/ops/delete.go
```diff
@@ -14,41 +14,37 @@ See the License for the specific language governing permissions and
...
 	var err error
 	if err = gkeOps.Delete(); err != nil {
-		return fmt.Errorf("failed deleting cluster: '%v'", err)
+		return fmt.Errorf("failed deleting cluster: '%w'", err)
+	}
+	// Unset context with best effort.
+	// The commands will try to unset the current context and delete it from kubeconfig.
+	cc, _ := cmd.RunCommand("kubectl config current-context")
+	if _, err := cmd.RunCommand("kubectl config unset current-context"); err != nil {
+		cmd.RunCommand("kubectl config unset contexts."+cc)
```
mattmoor, line 46: Format Go code:
```suggestion
		cmd.RunCommand("kubectl config unset contexts." + cc)
```

kntest/pkg/cluster/gke/README.md
mattmoor, line 3: Format markdown:
```suggestion
`kntest cluster gke` command is used for creating, deleting or getting a GKE
cluster
```

kntest/pkg/cluster/README.md
coryrc: I don't support this at all. If we configure things via arguments, then we have to use bash scripts to configure and pass things in as arguments -- the only things we can set in the Prow Job config is environment variables or args to the script, which then has to parse.

There must not be a subcommand or we aren't fixing the problem -- the complexity is left in the bash script. If we're only supporting one then we don't need a subcommand.

We either support environment variables or a config file. Nothing else takes the complexity of decision making out of bash.

chizhg: Honestly I don't see why subcommand is a problem in the context you provided. Like I said above we can run the command like `kntest cluster ${PROVIDER}`, and `PROVIDER` can be an env var in both approaches regardless whether we support subcommands or not.

However, the drawback of not having subcommands is clear: they will share all the flags (like 
the command to create `kind` cluster will also have the `gcp-project` flag), we cannot define and set clear ownership on these commands. These are also the issues [kubetest2](https://github.com/kubernetes/test-infra/tree/master/kubetest2) is trying to avoid.

Regarding the env vars, I definitely agree it's an option, but just as you said, we are still not sure whether we should support environment variables or a config file. So at this stage I'd like to first get rid of the huge bash functions in https://github.com/knative/test-infra/blob/master/scripts/e2e-tests.sh#L185-L321, and we can always iterate on it when we get a clearer picture.

coryrc: I'm okay with having flags if we will soon support the exact same options as env var/config file.

Otherwise, we're going to have this in our bash script:

```
if [[${PROVIDER}="gke"]]; then
kntest cluster gke --max-nodes ${GKE_MAX_NODES} --min-nodes ${GKE_MIN_NODES} --project-name=${GKE_PROJECT_ID} etc
elif [[${PROVIDER}="kind"]]; then
kntest cluster kind --other-kind-option ${OTHER_KIND_ENV} --another-kind-option ${ANOTHER_KIND_ENV}
elif [[${PROVIDER=...etc...
fi
```

Let's just solve the problem completely. `kntest cluster create`. That should be the only line and the complete command in the script.

We have/had this same problem with namespace being hardcoded in e2e_flags.go. Don't make me hack shit to get it done -- make the only available option the right one.

chizhg: I agree that `kntest cluster create` makes some sense, but does `kntest cluster ${PROVIDER} create` indeed add too much complexity?

chizhg:
@googlebot rescan

kntest/pkg/cluster/README.md
chizhg: Done.

kntest/pkg/cluster/ops/create.go
chizhg: Removed the comment.

kntest/pkg/cluster/ops/create.go
chizhg: Good point. Done.

kntest/pkg/cluster/ops/delete.go
chizhg: Changed back to 2019

kntest/pkg/cluster/options.go
chizhg: Added the comment

test/e2e-tests.sh
chizhg: Done.

kntest/pkg/cluster/options.go
chizhg: As we discussed in  https://github.com/knative/test-infra/pull/1883#discussion_r402314355, I'm going to defer the changes for env vars to following PRs.

And one more note: I believe env vars can sometimes cause negative effects, as you cannot restrict on where they can be set. That said, they can be set literally anywhere that makes them hard to manage and discover, that's exactly one of the problems we have in the other project. I think another option we can go with would be always using flags, and we can define groups of flags for each subcommand, e.g. for cluster related command, we can pass the flags like `--cluster "--project=xxx --name=xxx --version=xxx"`, in this way all the settings will be managed in one place, mandatorily.

This is of course just my thoughts, and we can certainly have more discussions. Also loop @chaodaiG @yt3liu in for more ideas.

kntest/pkg/cluster/README.md
chaodaiG: The switch statement Cory mentioned above doesn't seem to be avoidable in advance usage of `kntest` no matter which direction we go, as provider specific logics differ so much from each other, for example `--project` flag is a gcloud only logic, it's meaningless in all other platforms. But in default cases that shouldn't be an issue, caller should expect calling `kntest cluster ${PROVIDER}` and get a default cluster

kntest/pkg/cluster/options.go
chaodaiG: Added one more comment above. Agree that command line arguments are more preferable to me.

coryrc: I'd be okay if all cluster creation args were passed as one variable, like you wrote.

coryrc: Except what if one flow needs to say "make a cluster twice as big as normal"... how would we do that with this regime? (maybe we don't need to support that)
It wouldn't be any easier with env vars though.
Config file could be copied and do *2 on an arg.

kntest/pkg/cluster/gke/README.md
```diff
@@ -0,0 +1,59 @@
...
+- `--backup-regions`: backup regions to be used if cluster creation in primary
+  region failed, comma separated list, default "us-west1,us-east1"
+- `--addons`: GKE addons, comma separated list, default empty
+- `--skip-creation`: should skip creation or not
+
+## Flow
+
+### Create
+
+1. Acquiring cluster if kubeconfig already points to it
```
coryrc, line 33: My understanding of what you wrote, is if kubeconfig points to something (is liveness checked?), this does nothing except maybe write to metadata.json.

If that's not the case, you should reword.

kntest/pkg/cluster/gke/README.md
```diff
@@ -0,0 +1,59 @@
...
+   Failed obtaining project name will fail the tool
+
+1. Get default cluster name if not provided as a parameter
+1. Delete cluster if cluster with same name and location already exists in GKE
+1. Create cluster
+1. Write cluster metadata to `${ARTIFACT}/metadata.json`
+
+### Delete
+
+1. Acquiring cluster if kubeconfig already points to it
```
coryrc, line 48: My understanding of what you wrote, is if kubeconfig points to something, even if the name doesn't match what's passed in, that's the cluster that is deleted.

If that's not the case, you should reword.

kntest/pkg/cluster/gke/ops/create.go
```diff
@@ -14,18 +14,17 @@ See the License for the specific language governing permissions and
...
 	"log"
 	"strconv"
 
 	container "google.golang.org/api/container/v1beta1"
+	"knative.dev/pkg/test/cmd"
 	"knative.dev/pkg/test/gke"
 	clm "knative.dev/pkg/testutils/clustermanager/e2e-tests"
-	"knative.dev/pkg/testutils/clustermanager/e2e-tests/common"
-	"knative.dev/pkg/testutils/clustermanager/prow-cluster-operation/options"
 	"knative.dev/pkg/testutils/metahelper/client"
```
coryrc, line 28: Please consider calling this "metaclient" instead of just client.
Today it's clear enough, but "client" alone is super vague.

test/e2e-tests.sh
```diff
@@ -28,27 +28,32 @@ source $(dirname $0)/../scripts/e2e-tests.sh
...
 function knative_setup() {
   start_latest_knative_serving
 }
 
-# Run a prow-cluster-operation tool
+# Run "kntest cluster" tool
 # Parameters: $1..$n - parameters passed to the tool
 function run_prow_cluster_tool() {
-  go run ${REPO_ROOT_DIR}/vendor/knative.dev/pkg/testutils/clustermanager/prow-cluster-operation $@
+  if [[ "${REPO_NAME}" == "test-infra" ]]; then
```
coryrc, line 41: Isn't this file only run in test-infra?
Also, this is exactly the kind of BS that kntest is supposed to get rid of. I think we should just get it into prow-tests immediately.

kntest/pkg/cluster/gke/README.md
```diff
@@ -0,0 +1,59 @@
...
+- `--name`: cluster name, default empty
+- `--release-channel`: GKE release channel, default empty
+- `--resource-type`: Boskos resource type
+- `--version`: GKE version
+- `--backup-regions`: backup regions to be used if cluster creation in primary
+  region failed, comma separated list, default "us-west1,us-east1"
+- `--addons`: GKE addons, comma separated list, default empty
+- `--skip-creation`: should skip creation or not
+
+## Flow
```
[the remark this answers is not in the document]
chizhg: Changed.

kntest/pkg/cluster/gke/README.md
```diff
@@ -0,0 +1,66 @@
...
+- `--addons`: GKE addons, comma separated list, default empty
+- `--skip-creation`: should skip creation or not
+
+## Subcommands
+
+### Create
+
+`kntest cluster gke create` will always create a new cluster.
+
+The flow is:
```
mattmoor, line 35: Format markdown:
```suggestion
The flow is:

```

kntest/pkg/cluster/gke/README.md
```diff
@@ -0,0 +1,66 @@
...
+1. Get default cluster name if not provided as a parameter
+1. Delete cluster if cluster with same name and location already exists in GKE
+1. Create a new cluster with the config being provided
+1. Write cluster metadata to `${ARTIFACT}/metadata.json`
+
+### Delete
+
+`kntest cluster gke delete` will delete the existing cluster.
+
+The flow is:
```
mattmoor, line 52: Format markdown:
```suggestion
The flow is:

```

kntest/pkg/cluster/gke/README.md
```diff
@@ -0,0 +1,59 @@
...
+1. Get default cluster name if not provided as a parameter
+1. Delete cluster if cluster with same name and location already exists in GKE
+1. Create cluster
+1. Write cluster metadata to `${ARTIFACT}/metadata.json`
+
+### Delete
+
+1. Acquiring cluster if kubeconfig already points to it
+1. If cluster name is defined then getting cluster by its name
+1. If no cluster is found from previous step then it fails
```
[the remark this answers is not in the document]
chizhg: Fiixed.

kntest/pkg/cluster/gke/README.md
chizhg: To add some background here, most of the code and docs are copied from https://github.com/knative/pkg/tree/master/testutils/clustermanager/prow-cluster-operation written by Chao a few months ago. I'm going delete them after this PR gets merged.

Now `kntest cluster gke create` will always try to create a new cluster though. I have reworded the doc.

kntest/pkg/cluster/gke/README.md
chizhg: Done.

kntest/pkg/cluster/gke/ops/create.go
chizhg: I think this is the more preferable way to name a Go package, as the upper level package name is `metahelper`, changing the package name to `metaclient` would be duplicate.
One example is https://github.com/knative/eventing/tree/master/pkg/client/injection/client.

kntest/pkg/cluster/gke/README.md
```diff
@@ -0,0 +1,67 @@
...
+1. Get default cluster name if not provided as a parameter
+1. Delete cluster if cluster with same name and location already exists in GKE
+1. Create a new cluster with the config being provided
+1. Write cluster metadata to `${ARTIFACT}/metadata.json`
+
+### Delete
+
+`kntest cluster gke delete` will delete the existing cluster.
+
+The flow is:
```
mattmoor, line 53: Format markdown:
```suggestion
The flow is:

```

mattmoor, line 53: Format markdown:
```suggestion
The flow is:

```

test/e2e-tests.sh
chizhg: `kntest` is still under active development, so at the staging I would prefer to install it as part of the CI flow to always get its latest version, otherwise we would need to update the `prow-tests` image every time we have update `kntest`. It would be super annoying as now we don't have the automated process to update the images.

For why we have that `if` logic, in this repo we always run the local version of `kntest` but for other repos they need to `go get`, if we also use `go get` in this repo, we won't be able to run and test `kntest` changes in this PR.

kntest/pkg/cluster/gke/README.md
```diff
@@ -0,0 +1,59 @@
...
+- `--max-nodes`: maximum number of nodes, default 3
+- `--node-type`: GCE node type, default "e2-standard-4"
+- `--region`: GKE region, default "us-central1"
+- `--zone`: GKE zone, default empty
+- `--project`: GCP project, default empty
+- `--name`: cluster name, default empty
+- `--release-channel`: GKE release channel, default empty
+- `--resource-type`: Boskos resource type
+- `--version`: GKE version
+- `--backup-regions`: backup regions to be used if cluster creation in primary
```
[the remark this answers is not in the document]
chizhg: `--backup-regions` is only used when cluster creation in the main region fails, so in the normal scenario only `--region` will be used. I think `--region=xx,yy,zz` will be confusing as it would be like users want to create 3 clusters in the 3 regions.

kntest/pkg/cluster/gke/README.md
```diff
@@ -0,0 +1,59 @@
...
+### Delete
+
+1. Acquiring cluster if kubeconfig already points to it
+1. If cluster name is defined then getting cluster by its name
+1. If no cluster is found from previous step then it fails
+1. Delete:
+   - [In Prow] Release Boskos project
+   - [Not in Prow] Delete cluster
+
+### Get
```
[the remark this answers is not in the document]
chizhg: This call will be used to get an existing cluster for running tests, and it will do some verification, e.g. make sure the cluster is not a protected cluster. It's not currently used in Knative CI flow but can be used elsewhere.
I have modified the README to be more clear.

kntest/pkg/cluster/options.go
chizhg: Just to be clear, the place I will make this change is at https://github.com/knative/test-infra/blob/master/scripts/e2e-tests.sh#L462-L471, and I want to do it in following PRs as this one is big enough.

For the example you gave, please correct me if I got it wrong, but I think it's not adding much trouble as users can do *2 before passing them as flags?

chizhg:
@googlebot rescan

test/e2e-tests.sh
coryrc: I think one of us has this swapped -- this is test/e2e-tests.sh not scripts/e2e-tests.sh

This file is only used in test-infra AFAICT?

kntest/pkg/cluster/gke/ops/create.go
coryrc: Sorry, I'm fine with the name of the package.
I mean, in this file, to make "client" less ambiguous, to give it an import alias (or whatever they are called):

```
import (
...
    mhclient "knative.dev/pkg/testutils/metahelper/client"
)
```

I don't really care what the name is, as long as it isn't just "client"; "metahelper" "mhclient" "mh" "mhc"...

kntest/pkg/cluster/options.go
coryrc: We have one function to create clusters in the scripts. Currently it only works with GKE; not KIND or anything else.

So each e2e flow must have the exact same cluster creation.

But, for example, the autotls test flow shouldn't need the same size cluster, right? But it does run with the same one.

So if are passing in all arguments as one group of args (like `--cluster "--project=xxx --name=xxx --version=xxx"`) then how would we make the autotls test use half as many nodes?

What if we want a test flow with and without autoscaling?

None of this is being made easier with arguments; if anything, it's now harder than it was before.

kntest/pkg/cluster/gke/README.md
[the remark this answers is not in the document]
coryrc: What does it do to "acquire" it? Does it set kubeconfig? Some env vars?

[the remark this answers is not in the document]
chizhg: It gets the cluster from the kubeconfig. To be more specific, it runs `kubectl config current-context` to get the cluster and makes sure the cluster is the one that user needs.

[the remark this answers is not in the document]
coryrc: So a better description would be "it validates the current kubeconfig context points to a usable cluster"?

test/e2e-tests.sh
chizhg: Correct, this file is now only used in test-infra and is run to make sure it works in presubmit tests.
I will have followup PRs to move the changes to `script/e2e-tests.sh` so that all other Knative projects can get this update.

kntest/pkg/cluster/gke/ops/create.go
chizhg: Changed to `mhclient`

kntest/pkg/cluster/gke/README.md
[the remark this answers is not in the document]
chizhg: Makes sense. Updated.

kntest/pkg/cluster/gke/README.md
```diff
@@ -0,0 +1,69 @@
...
+1. If cluster name is defined then getting cluster by its name
+1. If no cluster is found from previous step then it fails
+1. Delete:
+   - [In Prow] Delete cluster asynchronously and release Boskos project
+   - [Not in Prow] Delete cluster synchronously
+
+### Get
+
+`kntest cluster gke get` will validate the current kubeconfig
+context points to a usable cluster requested by the user.
```
mattmoor, line 65: Format markdown:
```suggestion
`kntest cluster gke get` will validate the current kubeconfig context points to
a usable cluster requested by the user.
```

kntest/pkg/cluster/options.go
chizhg: Even with the current test setup, we do not support disabling autoscaling for the GKE cluster. If we do need it in the future, we can add it as an option in the `create` command.

The `create` command now supports `--min-nodes` and `--max-nodes` option, developers can set them if they want a different size cluster. "half as many nodes" is a pretty vague request and I doubt people would want that...

kntest/pkg/cluster/gke/README.md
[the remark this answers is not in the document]
coryrc: Then lets change it to "--possible-regions" or "--usable-regions" or "--candidate-regions"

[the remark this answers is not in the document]
chizhg: TBH I don't see why these are better than having `--region` and `--backup-regions`.. By separating them into two flags, users can just provide `--region` if they do not care which other regions are considered to be backups. But I assume changing to `--possible-regions` will require users to always provide multiple regions? Or only one region if they do not care about backups? Either way is more confusing I think.
Further more, we have `--zone` option, do we also want to change it to `--possible-zones`?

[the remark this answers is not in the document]
coryrc: I don't see why `--possible-regions` would not accept a single option.
Or `--region` could be a repeated option instead of having --backup-options:

```
--region us-central1
--region us-west1
```
...etc

Also, zone is confusing with multiple regions being possible. It might be best to have a repeated `--location` option that takes either region or zone:

```
--location us-central1-a
--location us-west1
```
etc

kntest/pkg/cluster/options.go
coryrc: If we accept that flows are never going to have any input into the cluster created for them (never asking for addons or anything else) --- to say another way, if the only configuration allowed is in Prow or on the command line before running the scripts, then `--cluster-args="--project=xxx -name=xxx` will suffice.

kntest/pkg/cluster/gke/README.md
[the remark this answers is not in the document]
chizhg: I double checked the code, another reason we are doing like this is we are following how it's done in bash to make sure they are compatible - https://github.com/knative/test-infra/blob/master/scripts/e2e-tests.sh#L42-L48.

There are also other places tied to it that makes it hard to change, e.g. https://github.com/knative/pkg/blob/master/testutils/clustermanager/e2e-tests/setup.go#L91-L108.

Since this needs to change two repos, I would prefer to not make changes in this PR. I can clean them up as you suggested later for sure.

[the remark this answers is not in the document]
coryrc: If we're just recapitulating mess because that's the way it was in bash, then there's little point in doing this at all. 
I grepped through docs, serving, eventing, test-infra, and pkg, and none set zone ever.
Looking at `gcloud container clusters create`, it takes EITHER --zone or --region.
--region is e.g. us-central1
--zone is e.g. us-central1-a
Additionally, it supports `--node-locations` which allows nodes to be spread across zones in one region.
Supporting zones doesn't make any sense when we have backup regions.
If people really want all the knobs, then they can call `gcloud` directly.
For this case, let's streamline and make it more reliable. No zones, only regions, passed like so:
```
--region us-central1
--region us-west1
--region etc
```

kntest/pkg/cluster/options.go
chizhg: This discussion is going too far away from the original topic. I agree that it's something we need to solve but it's out of the scope of this PR.
We can certainly discuss the different options, including env vars, multiple args and one single arg, and choose one to move forward, but it should not be a blocker for this PR.

coryrc: I disagree. If the shell scripts aren't broken right now, there's no rush. We need to get this right. There's no sense in putting bad code in.
If there's no scenario where scripts need to modify args to kntest (other than possibly add some), then this approach is fine.

kntest/pkg/cluster/gke/README.md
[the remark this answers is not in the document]
chizhg: [cobra](https://github.com/spf13/cobra) does not seem to support flag like this. 

Also like I mentioned above, [`BackupRegions`](https://github.com/knative/test-infra/pull/1883/files#diff-9a42c6de599ac1363d68dd7b4e9d4e65R41) is one field of the request struct, it would be pretty dirty if I make changes in this PR without modifying `pkg` first.

I can clean them up and change to `--possible-regions` in followup PRs. But let's not block this PR because of this?

kntest/pkg/cluster/options.go
chizhg: Then what do you want? Env vars?

kntest/pkg/cluster/gke/README.md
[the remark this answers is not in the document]
coryrc: It does and it's easy -- I'll send you a link directly.

They end up in a []string; you can slice so [1:] elements go to BackupRegions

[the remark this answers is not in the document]
chaodaiG: IMO backup regions should be considered a second class citizen, since it basically is a workaround, I would say that it shouldn't be mixed with the mainstream args. Also `--region a --region b` could be very confusing to users

kntest/pkg/cluster/options.go
coryrc: If you can't think of a scenario where we need scripts to modify args to kntest, then lets move forward.

kntest/pkg/cluster/gke/README.md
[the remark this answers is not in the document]
coryrc: It would only be confusing if they don't know how to read.
It's not a workaround, it's a vital piece of functionality we rely on.

PR #1865, opened 2020-03-28
Switch to Go modules

chizhg:
**What this PR does, why we need it**:
Switch test-infra to use Go modules

PR #1887, opened 2020-04-02
Ignore build tags under hack.

mattmoor:
This unblocks our ability to reference `package main` tooling from `tools.go`


chizhg:
Ah, this is the same change that is included in my experimental PR - https://github.com/knative/test-infra/pull/1865.

/lgtm
/approve
Thanks!

PR #1889, opened 2020-04-06
[WIP]TEST DO NOT MERGE

chizhg:
**What this PR does, why we need it**:

**Which issue(s) this PR fixes**:
Fixes #

**Special notes to reviewers**:

**User-visible changes in this PR**:

PR #1892, opened 2020-04-06
dep collector works with Go modules

chizhg:
Modify `dep-collector` to make it working with Go modules. Most of the ideas are borrowed from https://github.com/google/ko/pull/19, by using https://godoc.org/golang.org/x/tools/go/packages package.

In this PR it tests with Go dep and `third_party/VENDOR-LICENSE` does not change.

In https://github.com/knative/test-infra/pull/1889 it tests with Go modules, the diff can be seen in https://github.com/knative/test-infra/pull/1889/files#diff-da080de4f67002eef05fc37666055a19. 
There are two changes to the `third_party/VENDOR-LICENSE` file:
1. The import path changes, now it does not have `/vendor/`
2. A few extra licenses in the subdirectories are added, primarily because Go modules does not support pruning non-go files.

### DIFF
--- a/tools/dep-collector/gobuild.go
+++ b/tools/dep-collector/gobuild.go
@@ -1,0 +1,81 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+    "encoding/json"
+    "fmt"
+    gb "go/build"
+    "path/filepath"
+    "strings"
+
+    "knative.dev/pkg/test/cmd"
+)
+
+// https://golang.org/pkg/cmd/go/internal/modinfo/#ModulePublic
+type modInfo struct {
+    Path string
+    Dir  string
+}
+
+type gobuild struct {
+    mod *modInfo
+}
+
+// moduleInfo returns the module path and module root directory for a project
+// using go modules, otherwise returns nil.
+//
+// Related: https://github.com/golang/go/issues/26504
+func moduleInfo() *modInfo {
+    output, err := cmd.RunCommand("go list -mod=readonly -m -json")
+    if err != nil {
+        return nil
+    }
+    var info modInfo
+    if err := json.Unmarshal([]byte(output), &info); err != nil {
+        return nil
+    }
+    return &info
+}
+
+// importPackage wraps go/build.Import to handle go modules.
+//
+// Note that we will fall back to GOPATH if the project isn't using go modules.
+func (g *gobuild) importPackage(s string) (*gb.Package, error) {
+    if g.mod == nil {
+        return gb.Import(s, gb.Default.GOPATH, gb.ImportComment)
+    }
+
+    // If we're inside a go modules project, try to use the module's directory
+    // as our source root to import:
+    // * paths that match module path prefix (they should be in this project)
+    // * relative paths (they should also be in this project)
+    gp, err := gb.Import(s, g.mod.Dir, gb.ImportComment)
+    return gp, err
+}
+
+func (g *gobuild) qualifyLocalImport(ip string) (string, error) {
+    if g.mod == nil {
+        gopathsrc := filepath.Join(gb.Default.GOPATH, "src")
+        if !strings.HasPrefix(WorkingDir, gopathsrc) {
+            return "", fmt.Errorf("working directory must be on ${GOPATH}/src = %s", gopathsrc)
+        }
+        return filepath.Join(strings.TrimPrefix(WorkingDir, gopathsrc+string(filepath.Separator)), ip), nil
+    } else {
+        return filepath.Join(g.mod.Path, ip), nil
+    }
+}
--- a/tools/dep-collector/gobuild.go
+++ b/tools/dep-collector/gobuild.go
@@ -17,23 +17,23 @@ limitations under the License.
 package main
 
 import (
-    "encoding/json"
-    "fmt"
-    gb "go/build"
-    "path/filepath"
-    "strings"
+	"encoding/json"
+	"fmt"
+	gb "go/build"
+	"path/filepath"
+	"strings"
 
-    "knative.dev/pkg/test/cmd"
+	"knative.dev/pkg/test/cmd"
 )
 
 // https://golang.org/pkg/cmd/go/internal/modinfo/#ModulePublic
 type modInfo struct {
-    Path string
-    Dir  string
+	Path string
+	Dir  string
 }
 
 type gobuild struct {
-    mod *modInfo
+	mod *modInfo
 }
 
 // moduleInfo returns the module path and module root directory for a project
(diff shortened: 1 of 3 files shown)


tools/dep-collector/gobuild.go
```diff
@@ -0,0 +1,81 @@
...
+*/
+
+package main
+
+import (
+    "encoding/json"
+    "fmt"
+    gb "go/build"
+    "path/filepath"
+    "strings"
```
mattmoor, line 24: Format Go code:
```suggestion
	"encoding/json"
	"fmt"
	gb "go/build"
	"path/filepath"
	"strings"
```

tools/dep-collector/gobuild.go
```diff
@@ -0,0 +1,81 @@
...
+
+    "knative.dev/pkg/test/cmd"
```
mattmoor, line 26: Format Go code:
```suggestion
	"knative.dev/pkg/test/cmd"
```

tools/dep-collector/gobuild.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+// https://golang.org/pkg/cmd/go/internal/modinfo/#ModulePublic
+type modInfo struct {
+    Path string
+    Dir  string
```
mattmoor, line 32: Format Go code:
```suggestion
	Path string
	Dir  string
```

tools/dep-collector/gobuild.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+type gobuild struct {
+    mod *modInfo
```
mattmoor, line 36: Format Go code:
```suggestion
	mod *modInfo
```

tools/dep-collector/gobuild.go
```diff
@@ -0,0 +1,81 @@
...
+func moduleInfo() *modInfo {
+    output, err := cmd.RunCommand("go list -mod=readonly -m -json")
+    if err != nil {
+        return nil
+    }
+    var info modInfo
+    if err := json.Unmarshal([]byte(output), &info); err != nil {
+        return nil
+    }
+    return &info
```
mattmoor, line 52: Format Go code:
```suggestion
	output, err := cmd.RunCommand("go list -mod=readonly -m -json")
	if err != nil {
		return nil
	}
	var info modInfo
	if err := json.Unmarshal([]byte(output), &info); err != nil {
		return nil
	}
	return &info
```

tools/dep-collector/gobuild.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// importPackage wraps go/build.Import to handle go modules.
+//
+// Note that we will fall back to GOPATH if the project isn't using go modules.
+func (g *gobuild) importPackage(s string) (*gb.Package, error) {
+    if g.mod == nil {
+        return gb.Import(s, gb.Default.GOPATH, gb.ImportComment)
+    }
```
mattmoor, line 61: Format Go code:
```suggestion
	if g.mod == nil {
		return gb.Import(s, gb.Default.GOPATH, gb.ImportComment)
	}
```

tools/dep-collector/gobuild.go
```diff
@@ -0,0 +1,81 @@
...
+
+    // If we're inside a go modules project, try to use the module's directory
+    // as our source root to import:
+    // * paths that match module path prefix (they should be in this project)
+    // * relative paths (they should also be in this project)
+    gp, err := gb.Import(s, g.mod.Dir, gb.ImportComment)
+    return gp, err
```
mattmoor, line 68: Format Go code:
```suggestion
	// If we're inside a go modules project, try to use the module's directory
	// as our source root to import:
	// * paths that match module path prefix (they should be in this project)
	// * relative paths (they should also be in this project)
	gp, err := gb.Import(s, g.mod.Dir, gb.ImportComment)
	return gp, err
```

tools/dep-collector/gobuild.go
```diff
@@ -0,0 +1,81 @@
...
+func (g *gobuild) qualifyLocalImport(ip string) (string, error) {
+    if g.mod == nil {
+        gopathsrc := filepath.Join(gb.Default.GOPATH, "src")
+        if !strings.HasPrefix(WorkingDir, gopathsrc) {
+            return "", fmt.Errorf("working directory must be on ${GOPATH}/src = %s", gopathsrc)
+        }
+        return filepath.Join(strings.TrimPrefix(WorkingDir, gopathsrc+string(filepath.Separator)), ip), nil
+    } else {
+        return filepath.Join(g.mod.Path, ip), nil
+    }
```
mattmoor, line 80: Format Go code:
```suggestion
	if g.mod == nil {
		gopathsrc := filepath.Join(gb.Default.GOPATH, "src")
		if !strings.HasPrefix(WorkingDir, gopathsrc) {
			return "", fmt.Errorf("working directory must be on ${GOPATH}/src = %s", gopathsrc)
		}
		return filepath.Join(strings.TrimPrefix(WorkingDir, gopathsrc+string(filepath.Separator)), ip), nil
	} else {
		return filepath.Join(g.mod.Path, ip), nil
	}
```

tools/dep-collector/imports.go
```diff
@@ -19,74 +19,76 @@ package main
...
 	gb "go/build"
-	"path/filepath"
 	"sort"
 	"strings"
 )
 
-func CollectTransitiveImports(binaries []string) ([]string, error) {
+type importInfo struct {
+	importPath string
+	dir string
```
mattmoor, line 28: Format Go code:
```suggestion
	dir        string
```

tools/dep-collector/imports.go
```diff
@@ -19,74 +19,76 @@ package main
...
 }
 
-func visit(pkg *gb.Package, visited map[string]struct{}) error {
-	if _, ok := visited[pkg.ImportPath]; ok {
+func visit(g *gobuild, pkg *gb.Package, visited map[string]importInfo) error {
+	if _, ok := visited[pkg.Dir]; ok {
 		return nil
 	}
-	visited[pkg.ImportPath] = struct{}{}
+	visited[pkg.Dir] = importInfo{dir: pkg.Dir, importPath:pkg.ImportPath}
```
mattmoor, line 76: Format Go code:
```suggestion
	visited[pkg.Dir] = importInfo{dir: pkg.Dir, importPath: pkg.ImportPath}
```

tools/dep-collector/gobuild.go
```diff
@@ -0,0 +1,81 @@
...
+
+// moduleInfo returns the module path and module root directory for a project
+// using go modules, otherwise returns nil.
+// If there is something wrong in parsing the json output, it will return an error.
+//
+// Related: https://github.com/golang/go/issues/26504
+func moduleInfo() (*modInfo, error) {
+	output, err := cmd.RunCommand("go list -mod=readonly -m -json")
+	if err != nil {
+		return nil, nil
```
chaodaiG, line 47: what if go doesn't support `-m -json` one day? Is there a way to make this check more strict? i.e. check if there is a command line exception

chizhg: I don't think there is a reliable way to detect if an option is supported or not, as the error code returned by the command is tied to the implementation details.

The `-m` flag is used to list modules instead of packages and I believe there is minor chance for it to be changed. I can run `go list -m` to check if it's using Go modules or not, and then run the `go list -mod=readonly -m -json` to get the json output. If you think it's okay I'll make the change.

chaodaiG: This would be preferable to me. The underline issue of invoking command is like life time binding, and it breaks without us being prepared. I think it's ok to return nil and nil for `go list -m` or make it more explicit by wrapping it inside `function isMod`. Anything failed afterwards should be treated as errors

chizhg: Makes sense. Done.

PR #1897, opened 2020-04-07
fix dep collector by not using pkg

chizhg:
In our CI we are using `go get` to install `dep-collector`. In last change we introduced dependencies on `knative.dev/pkg` for it and it's causing some annoying dependency problems when most of the projects are still using `go dep`, see https://prow.knative.dev/view/gcs/knative-prow/logs/ci-knative-net-kourier-continuous/1247293590225293312.

This PR removes the dependency on `knative.dev/pkg`.

/cc @chaodaiG

### DIFF
--- a/tools/dep-collector/gobuild.go
+++ b/tools/dep-collector/gobuild.go
@@ -20,10 +20,9 @@ import (
 	"encoding/json"
 	"fmt"
 	gb "go/build"
+	"os/exec"
 	"path/filepath"
 	"strings"
-
-	"knative.dev/pkg/test/cmd"
 )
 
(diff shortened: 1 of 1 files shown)


chaodaiG:
That doesn't feel like the most proper fix, why not just `go get -u`?

chizhg:
> That doesn't feel like the most proper fix, why not just `go get -u`?

You meant `go get -u` at https://github.com/knative/test-infra/pull/1897#issuecomment-610548201? I'm not sure how that will help.. Maybe I missed something? And moreover we don't have a proper way to test it before merging it in, and I'm worried it can introduce more issues that are unexpected...

PR #1916, opened 2020-04-18
Enable integration test job for net-certmanager

ZhiminXiang:


chizhg:
You need to change config_knative.yaml and run ./hack/generate-configs.sh to generate the actual Prow job.

ZhiminXiang:
hmm interesting. If I follow the instruction to run `./hack/update-codegen.sh` to fix the test failure, my change will be reverted.

ZhiminXiang:
> You need to change config_knative.yaml and run ./hack/generate-configs.sh to generate the actual Prow job.

Done

PR #1928, opened 2020-04-21
Move another 20 Boskos projects to build cluster

chaodaiG:
/assign chizhg
/cc chizhg

PR #1929, opened 2020-04-21
Move more repos to use build cluster

chaodaiG:
/assign chizhg
/cc chizhg

/hold
Needs to do it after #1928


chaodaiG:
The load on build cluster was very light after #1927 , so migrate these repos as well

PR #1932, opened 2020-04-21
Move all repos other than serving to build cluster

chaodaiG:
/assign chizhg
/cc chizhg

/hold
Build cluster may need more nodes to handle these requests


chaodaiG:
/unhold

build cluster can still hold

PR #1934, opened 2020-04-21
Move everything to build cluster

chaodaiG:
Finally moving all serving jobs to use build cluster

/assign chizhg
/cc chizhg

/hold
Will do this when Prow is quieter


chaodaiG:
Rebased, need another lgtm @chizhg

PR #1937, opened 2020-04-22
Do not generate most of the maintenance jobs with config-generator

### LINKED ISSUE #1936: Do not generate most of the maintenance jobs with config-generator
Currently, we are generating all Prow jobs with the config-generator tool, including the maintenance jobs. This makes it pretty tedious to add new maintenance jobs, and also, there is no easy way to bump up the images versions used in the jobs with the autobumper tool.

Since most of the maintenance jobs are pretty stable, there is little value to generate them with code. Now that we have supported having multiple Prow job config files, these maintenance jobs need to be moved to a separate config file.

/assign

chizhg:
Do not generate most of the maintenance jobs with config-generator. Instead, move them to a separate Prow job config file.

Fixes #1936

### DIFF
--- a/config/prow/jobs/config.yaml
+++ b/config/prow/jobs/config.yaml
@@ -10757,190 +10757,242 @@ periodics:
       args:
       - "--artifacts=$(ARTIFACTS)"
       - "--cov-threshold-percentage=50"
-- cron: "0 19 * * 1,3,5"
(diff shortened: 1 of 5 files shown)


coryrc:
I hate the way you have with pulling a script with curl in order to update the Prow cluster, but I like this change.

/lgtm

PR #1942, opened 2020-04-22
Improve presubmit config check

chizhg:
Currently we are using config-generator to generate testgrid config, so there is little chance for it to be broken. And we are going to move the config to Prow job annotation, this will be deprecated soon anyway. 
Also change the last checker to run as a custom Prow job that will only be run if specific files change.

/cc @chaodaiG


tools/prow-config-updater/presubmit-checker/README.md
```diff
@@ -1,5 +1,5 @@
 # presubmit-checker
 
-presubmit-checker is a checker that will be run as a presubmit test to disallow
+presubmit-checker is a checker that will be run as a unit test in presubmit to disallow
 developers to directly modify the core and cluster configuration files for
 production Prow.
```
mattmoor, line 5: Format markdown:
```suggestion
presubmit-checker is a checker that will be run as a unit test in presubmit to
disallow developers to directly modify the core and cluster configuration files
for production Prow.
```

chizhg:
/hold cancel
/assign @chaodaiG 

Please note I have to manually update the Prow job config to unblock this PR as I deleted the shell script in this PR.

chaodaiG:
/lgtm

You could have done this in 2 separate steps

chizhg:
> /lgtm
> 
> You could have done this in 2 separate steps

I know, but I'm too lazy to do that 😄

Issue #1734, opened 2020-02-23
Document default resource requests for Prow jobs


coryrc:
It seems we have default resource requests of 8.6GiB memory and 4 vCPU, but I can't find where this is configured. We should document this.

chaodaiG:
They are configured here:
https://github.com/knative/test-infra/blob/master/config/prow/cluster/200-cpu-limitrange.yaml
https://github.com/knative/test-infra/blob/master/config/prow/cluster/200-memory-limitrange.yaml

The reason for 8.6 GiB instead of 8GiB I believe was because of each sidecar takes 200M

coryrc:
Ah, thanks. I will find an appropriate place to document.
/assign

coryrc:
oops, wrong number
still open

coryrc:
/open
?

coryrc:
we have deleted default resource... but we should document what the config generator does somewhere

PR #1944, opened 2020-04-22
Prow jobs explicitly set it's own resources requests/limit

chaodaiG:
This removes the unnecessary resource allocation of:
- janitor pods
- maintenance jobs

Other than these, some of the jobs don't really need to use that much cpu and memory, this enables us tweaking them on a per job basis

/assign chizhg

Related to: #1734 
/cc coryrc


config/build-cluster/cluster/200-cpu-limitrange.yaml
```diff
@@ -1,26 +0,0 @@
...
-# limitations under the License.
-
-apiVersion: v1
-kind: LimitRange
-metadata:
-  name: cpu-limit-range
-  namespace: test-pods
-spec:
-  limits:
-    - default:
```
coryrc, line 22: I'd rather keep these sensible defaults

tools/config-generator/main.go
```diff
@@ -440,12 +462,12 @@ func newbaseProwJobTemplateData(repo string) baseProwJobTemplateData {
 	data.Args = make([]string, 0)
 	data.Volumes = make([]string, 0)
 	data.VolumeMounts = make([]string, 0)
-	data.Resources = make([]string, 0)
 	data.Env = make([]string, 0)
 	data.ExtraRefs = []string{"- org: " + data.OrgName, "  repo: " + data.RepoName}
 	data.Labels = make([]string, 0)
 	data.Optional = ""
 
+	setReourcesReqForJob(defaultResource, &data)
```
coryrc, line 470: function name is misspelled

tools/config-generator/main.go
```diff
@@ -177,6 +177,28 @@ var (
 	emittedOutput bool
 
 	projNameRegex = regexp.MustCompile(`.+-[0-9\.]+$`)
+
+	// defaultResource setting for each job
```
coryrc, line 181: Lacking CPU requests

config/build-cluster/cluster/200-cpu-limitrange.yaml
chaodaiG: Do you mean cpu or both?

tools/config-generator/main.go
chaodaiG: done

tools/config-generator/main.go
chaodaiG: fixed

tools/config-generator/main.go
chaodaiG: added

config/build-cluster/cluster/200-cpu-limitrange.yaml
coryrc: Both. It causes very weird intermittent problems when the default requests (.1cpu) are used by some jobs. It's just nice prevention-in-depth to keep.

chaodaiG: For lightweight jobs I actually would prefer it uses the default .1cpu as request, as well as any other pods such as janitor

chaodaiG: The other problem is that sidecar in each Prow job pod also requests these resources, which is pretty confusing

chaodaiG: So when we say each container requests 2 CPUs, a Prow job actually requests 4 CPUs

coryrc: Why don't we have resource requests on those?
I also think our janitors need a little more than .1 cpu, but we can reduce their quantity now that they've been fixed.

chaodaiG: Prow job sidecar is really lightweight, it shouldn't need 2 CPUs. If janitors need a little more than .1 cpu we can grant it there

coryrc: I asked someone to double-check, but looks like the sidecar is a problem. Could you boost the janitor CPU allocation slightly and reduce their number at the same time? The latter could be delayed if you want.

chaodaiG: done

coryrc: Ah, we had a misunderstanding. The sidecar shares the resources dedicated to the pod, so it doesn't hurt to keep the defaults.

coryrc:
/lgtm

I guess all the jobs have resources now so that should be good enough.

chaodaiG:
> /lgtm
> 
> I guess all the jobs have resources now so that should be good enough.

Per the offline discussion, we can explicitly set the sidecar requests, will skip setting it on every single job then

/hold

config/build-cluster/cluster/200-memory-limitrange.yaml
```diff
@@ -22,5 +22,5 @@ spec:
     - default:
         memory: 8Gi
       defaultRequest:
-        memory: 4Gi
+        memory: 8Gi
```
chizhg, line 25: What's the reason for changing this?

chaodaiG: It used to be sidecar 4Gb, test container 4Gb, so actually a pod requests 8Gb. Now that we set sidecar to be real low, so we can make the real request of 8Gb

chizhg: I see, great improvements on resource utilization!

PR #1946, opened 2020-04-22
Clean up tool supports multiple yaml locations

chaodaiG:
Expand it's capacity so it's easier to support prow, staging prow, and build cluster boskos resource defs

/assign chizhg
/cc chizhg


chizhg:
> Expand it's capacity so it's easier to support prow, staging prow, and build cluster boskos resource defs
> 
> /assign chizhg
> /cc chizhg

We won't need boskos on the Prow cluster after the migration is done, I think. 

/lgtm

PR #1956, opened 2020-04-23
Fixing staging Prow: don't use build cluster

chaodaiG:
Auto config updater failed in the past few days, this caused by all staging Prow jobs were pointing to build cluster and they don't have access to the build cluster.

/assign chizhg
/cc chizhg

/hold
There is a potential race condition between the current auto bumping PR https://github.com/knative-prow-robot/test-infra/pull/35 and the PR kicked off by this PR


chaodaiG:
Closed https://github.com/knative-prow-robot/test-infra/pull/35 manually, so there is no problem of race condition

/unhold

PR #1976, opened 2020-04-27
run_go_tool always install/update the go tool

chizhg:
1. When developers run the script on local, they may get different result with the CI run as `run_go_tool` will not `get` or `install` the tool if it already exists, thus can possibly have different version of tools. This PR changes to always `get` or `install` the tool. The side effect should be minor as `go get` will not do anything if it's already the latest version.

2. A few other cleanups, including removing `githubhelper` tool and modifying the `Dockerfile` headers.

/cc @chaodaiG


images/backups/Dockerfile
```diff
@@ -15,7 +15,6 @@
 FROM gcr.io/go-containerregistry/gcrane as gcrane
 
 FROM gcr.io/cloud-builders/gcloud-slim
-LABEL maintainer "Jon Johnson <jonjohnson@google.com>"
```
chaodaiG, line 18: should there be an alternative maintainer?

images/prow-tests/Dockerfile
```diff
@@ -59,13 +58,3 @@ RUN ko version > /ko_version
 # Extra tools through gem
 RUN gem install mixlib-config -v 2.2.4  # required because ruby is 2.1
 RUN gem install mdl -v 0.5  # required because later version of mdl requires ruby >= 2.4
-
```
chaodaiG, line 62: This can only be done when all repos relying on githubhelper are using latest test-infra. Do we know whether this is done already? How about release branches?

chaodaiG:
Generally ok with me, would like to point out the implication here: this change means all the tools consumed by `go install` need to be backward compatible

chizhg:
> implication

That's already the case without this change, as the when the job is run by Prow, the tools will always be installed as the latest version as they do not exist in the prow-tests image.  But yeah, I agree that we need a better versioning story for this tool.

images/prow-tests/Dockerfile
chizhg: You are right, it cannot be deleted as release branches are still having it. Reverted the change.
Thanks for pointing it out!

images/backups/Dockerfile
chizhg: `maintainer` is not required in the Dockerfile.  Or we can use the team email if you think we really need one?

chaodaiG:
/lgtm

Since Jon is no longer part of Knative, it doesn't make much sense still list him as maintainer. We can decide what to do with backup image later

coryrc:
This change makes it nearly impossible to test a custom version of the tool and will screw up the workspace of anyone who runs it and builds their own version

chizhg:
> This change makes it nearly impossible to test a custom version of the tool and will screw up the workspace of anyone who runs it and builds their own version

This will be using the same version as run by Prow. At the first place why would developers want to use a different version of tools as CI to run their tests?

coryrc:
Because I'm developing a patch for `ko` and have my own version installed; if I'm also debugging a problem in Knative my version gets blown away.

A test flow should not be arbitrarily modifying my workstation. But that's why we discussed running tests with docker.

PR #1979, opened 2020-04-27
Add shellcheck and others to prow-tests image

coryrc:
The former for adding a presubmit
The others for being prepared for future use from the image (currently
they are used directly via "go run" in the scripts).


chaodaiG:
/assign chizhg

I think Chi has different opinions over how tools should be installed in Knative

Issue #1970, opened 2020-04-24
Unified story of using k8s/test-infra tools with bazel


chaodaiG:
There are currently a few places where bazel are used in this repo, based on the usage this can be divided into these categories:
- Used only by test-infra:
  - [Update testgrid config](https://github.com/knative/test-infra/blob/f645de8d9a500a3fd00149f1b4e693029d678132/config/prow_common.mk#L120)
  - [run_job.sh](/config/prow/run_job.sh)
- Used by other repos 
  - Installed in [Prow test image](https://github.com/knative/test-infra/blob/f645de8d9a500a3fd00149f1b4e693029d678132/images/prow-tests/Dockerfile#L53)

For the usage only by test-infra, the bazel runtime is declared by an out-dated [WORKSPACE](/WORKSPACE) file, which has been so old that it's almost not possible to keep it up-to-date. We need a better story with using bazel

/cc @chizhg

chizhg:
@cjwagner Do you have a suggestion on how to solve this issue? How are other repos running commands from k8s/test-infra repo such as `mkpj`? Thanks!

cjwagner:
All Prow binaries are published as container images and all images are published together whenever one changes so that a consistent image version tag can be used across Prow components. Using components with different versions together or building with bazel from an arbitrary commit can cause in problems due to version skew between components. e.g. using a newer version of `mkpj` than `plank` could create PJs that use fields that are not yet considered valid by plank.
All the available images can be found under [gcr.io/k8s-prow/](https://gcr.io/k8s-prow/)
Here is an example use of the `mkpj` and `mkpod` containers: https://github.com/kubernetes/test-infra/blob/c5d8bee32b32fba5d2fa278acc0950ab88715d7f/prow/pj-on-kind.sh#L29-L30

coryrc:
So we'd `docker run` them somehow?

cjwagner:
> So we'd `docker run` them somehow?

Yes, see my link above for some example `docker run` commands for `mkpj` and `mkpod`.

chizhg:
> > So we'd `docker run` them somehow?
> 
> Yes, see my link above for some example `docker run` commands for `mkpj` and `mkpod`.

Thanks Cole!!

It seems `//testgrid/cmd/configurator` is not built as an image, or maybe it was in a different registry? If indeed not, can we built it as an image as well?

cjwagner:
[...]

PR #1981, opened 2020-04-27
Run Prow job ad hoc using docker command instead of bazel

chaodaiG:
Keeps us using updated mkpj while not relying on bazel

Part of: #1970

/assign chizhg
/cc cjwagner

### DIFF
--- a/config/prow/run_job.sh
+++ b/config/prow/run_job.sh
@@ -33,7 +33,7 @@ JOB_CONFIG_YAML=${REPO_ROOT_DIR}/config/prow/jobs
 docker run -i --rm \
     -v "${PWD}:${PWD}" -v "${CONFIG_YAML}:${CONFIG_YAML}" -v "${JOB_CONFIG_YAML}:${JOB_CONFIG_YAML}" \
     -w "${PWD}" \
-    gcr.io/k8s-prow/mkpj \
+    gcr.io/k8s-prow/mkpj:v20200427-84e5e2b2c \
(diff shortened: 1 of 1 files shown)


config/prow/run_job.sh
```diff
@@ -26,13 +26,17 @@ cd ${REPO_ROOT_DIR}
...
-
 JOB_YAML=$(mktemp)
 CONFIG_YAML=${REPO_ROOT_DIR}/config/prow/core/config.yaml
 JOB_CONFIG_YAML=${REPO_ROOT_DIR}/config/prow/jobs
-${run_mkpj} --job=$1 --config-path=${CONFIG_YAML} --job-config-path=${JOB_CONFIG_YAML} > ${JOB_YAML}
+
+docker run -i --rm \
+    -v "${PWD}:${PWD}" -v "${CONFIG_YAML}:${CONFIG_YAML}" -v "${JOB_CONFIG_YAML}:${JOB_CONFIG_YAML}" \
+    -w "${PWD}" \
+    gcr.io/k8s-prow/mkpj \
```
chizhg, line 36: How about we pin the version here? Otherwise you'll need to download a new image almost every time. `mkpj` does not change much often AFAICT.

chizhg: Though the best approach would be `prow-auto-bumper` also update this image.

chaodaiG: agreed, done.

PR #1997, opened 2020-04-29
Mirror copy from staging to prod configs while auto update

chaodaiG:
/assign chizhg
/cc chizhg

### DIFF
--- a/tools/prow-config-updater/updater.go
+++ b/tools/prow-config-updater/updater.go
@@ -17,8 +17,10 @@ limitations under the License.
 package main
 
 import (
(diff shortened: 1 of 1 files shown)


chizhg:
@googlebot rescan

chizhg:
Hmm, does the current logic have issues? What's the reason for this change?

chaodaiG:
Sorry for not making it clear. The current copying logic doesn't take care of configs that are deleted, so deleting before copying

chizhg:
I see. Thanks.
/lgtm

PR #1998, opened 2020-04-30
revert the changes in function run_go_tool

chizhg:

/cc @chaodaiG


chaodaiG:
This might happen to other tools that we rely on, may be we should reconsider doing this for other tools as well?

PR #1999, opened 2020-05-01
Switch the master branches for all repos to run with Go 1.14

chizhg:
We don't see issues for serving and knative-gcp when they run the tests with Go 1.14, so aggressively switch Prow jobs on master branches for all repos to run with Go 1.14. Also fixes some previously misconfigured Prow jobs.

/cc @chaodaiG 
/hold


chaodaiG:
let it rip

PR #2000, opened 2020-05-01
Configurator use docker run

chaodaiG:
/assign chaodaiG
/cc chaodaiG

/hold
Configurator is not fully published yet, current latest has a dirty tag


chaodaiG:
Updated to use configurator image in prod
/unhold

Fixes: #1970

PR #2003, opened 2020-05-03
Now it's safe to switch to go-mod

chizhg:

/cc @chaodaiG


third_party/VENDOR-LICENSE/github.com/go-sql-driver/mysql/utils.go
```diff
@@ -0,0 +1,701 @@
...
+func stringToInt(b []byte) int {
+	val := 0
+	for i := range b {
+		val *= 10
+		val += int(b[i] - 0x30)
+	}
+	return val
+}
+
+// returns the string read as a bytes slice, wheter the value is NULL,
```
mattmoor, line 378: Found potential misspelling(s):
```suggestion
// returns the string read as a bytes slice, whether the value is NULL,
```

PR #2010, opened 2020-05-05
Add a new wait function that also confirms HTTP connection.

tcnghia:

Some times static IP is available, but it is not yet accessible due to requests not going all the way to the proxy Pod.

This new wait function will confirms that we can receive some HTTP status (like 404, 503, etc) from the proxy before proceeding.

Fixes #

### DIFF
--- a/scripts/library.sh
+++ b/scripts/library.sh
@@ -225,14 +225,14 @@ function wait_until_service_has_external_ip() {
 function wait_until_service_has_external_http_address() {
   local ns=$1
   local svc=$2
-  local sleep_seconds=20
+  local sleep_seconds=6
   local attempts=150
 
   echo -n "Waiting until service $ns/$svc has an external address (IP/hostname)"
   for attempt in $(seq 1 $attempts); do  # timeout after 15 minutes
     local address=$(kubectl get svc $svc -n $ns -o jsonpath="{.status.loadBalancer.ingress[0].ip}")
     if [[ -n "${address}" ]]; then
-      echo -e "Service $2.$1 has IP $address"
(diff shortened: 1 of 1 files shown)


scripts/library.sh
```diff
@@ -219,6 +219,42 @@ function wait_until_service_has_external_ip() {
...
+  local attempts=150
+
+  echo -n "Waiting until service $ns/$svc has an external address (IP/hostname)"
+  for attempt in $(seq 1 $attempts); do  # timeout after 15 minutes
+    local address=$(kubectl get svc $svc -n $ns -o jsonpath="{.status.loadBalancer.ingress[0].ip}")
+    if [[ -n "${address}" ]]; then
+      echo -e "Service $ns/$svc has IP $address"
+    else
+      address=$(kubectl get svc $svc -n $ns -o jsonpath="{.status.loadBalancer.ingress[0].hostname}")
+      if [[ -n "${hostname}" ]]; then
```
chizhg, line 238: `hostname` is an `undefined var`?

scripts/library.sh
```diff
@@ -219,6 +219,42 @@ function wait_until_service_has_external_ip() {
...
+        echo -e "Service $ns/$svc has hostname $address"
+      fi
+    fi
+    if [[ -n "${address}" ]]; then
+      local status=$(curl -s -o /dev/null -w "%{http_code}" http://"${address}")
+      if [[ $status != "" && $status != "000" ]]; then
```
chizhg, line 244: Hmm what does "000" mean?

vagababov: probably a special curl return code :)

PR #2011, opened 2020-05-05
Followups on go-mod migration

chizhg:
1. Reenable `update_licenses`
2. Add release versions for the floating deps, similar as other Knative repos

Note:
Have to skip build tests and markdown links check for `./third_party` as files are generated by the `go-licenses` tool and we don't have control over them.

/cc @chaodaiG


third_party/VENDOR-LICENSE/github.com/go-sql-driver/mysql/utils.go
```diff
@@ -0,0 +1,701 @@
...
+func stringToInt(b []byte) int {
+	val := 0
+	for i := range b {
+		val *= 10
+		val += int(b[i] - 0x30)
+	}
+	return val
+}
+
+// returns the string read as a bytes slice, wheter the value is NULL,
```
mattmoor, line 378: Found potential misspelling(s):
```suggestion
// returns the string read as a bytes slice, whether the value is NULL,
```

mattmoor, line 378: Found potential misspelling(s):
```suggestion
// returns the string read as a bytes slice, whether the value is NULL,
```

PR #2020, opened 2020-05-05
Full support build cluster: auto config update also covers build cluster

chaodaiG:
Together with various fixes. Note that the postsubmit job hasn't ran for several days due to the filtering path being incorrect

/assign chizhg
/cc chizhg


chizhg:
/lgtm
Thanks, hopefully the issues are fixed.

PR #2030, opened 2020-05-06
Add shellcheck presubmit files

coryrc:

Add the files needed to run a presubmit job with shellcheck to this repo and to the prow-tests image.

Also fix `list_changed_files` function to use the correct variable, though it seems to have been working anyway

/cc @chizhg 
/assign @chizhg

### DIFF
--- a/images/prow-tests/Dockerfile
+++ b/images/prow-tests/Dockerfile
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 # This is the base Dockerfile for knative-tests images, please remember to also
-# do `make push` in `../prow-tests-go112` directory to make sure these 2 images are in
+# do `make push` in `../prow-tests-*` directories to make sure these images are in
 # sync
 
 # TODO(chizhg): probably build the image from scratch?
@@ -32,8 +32,8 @@ RUN docker-credential-gcr configure-docker
 # Replace kubectl with a working version
 # TODO(chizhg): remove once https://github.com/knative/test-infra/issues/1858 is fixed
 ARG KUBECTL_VERSION=v1.17.4
-RUN rm -f $(which kubectl) && \
-    wget https://storage.googleapis.com/kubernetes-release/release/${KUBECTL_VERSION}/bin/linux/amd64/kubectl -O /usr/local/bin/kubectl && \
+RUN rm -f "$(which kubectl)" && \
+    wget "https://storage.googleapis.com/kubernetes-release/release/${KUBECTL_VERSION}/bin/linux/amd64/kubectl" -O /usr/local/bin/kubectl && \
     chmod +x /usr/local/bin/kubectl
 
 # Extra tools through apt-get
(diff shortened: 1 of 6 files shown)


images/prow-tests/Dockerfile
```diff
@@ -71,8 +71,12 @@ RUN gem install mdl -v 0.5  # required because later version of mdl requires rub
 # Install our own tools
 RUN go get knative.dev/pkg/testutils/junithelper
 
+# And our scripts
+ENV SCRIPTS /scripts
+COPY scripts /scripts
+
 # Temporarily add test-infra to the image to build custom tools
-ADD . /go/src/knative.dev/test-infra
+COPY . /go/src/knative.dev/test-infra
```
coryrc, line 79: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy

"Although ADD and COPY are functionally similar, generally speaking, COPY is preferred. That’s because it’s more transparent than ADD."

coryrc: to preempt any questions :-D

images/prow-tests/Dockerfile
```diff
@@ -71,8 +71,12 @@ RUN gem install mdl -v 0.5  # required because later version of mdl requires rub
 # Install our own tools
 RUN go get knative.dev/pkg/testutils/junithelper
 
+# And our scripts
+ENV SCRIPTS /scripts
```
chizhg, line 75: Is this line needed?

scripts/library.sh
```diff
@@ -652,6 +652,21 @@ function get_canonical_path() {
   echo "$(cd ${path%/*} && echo $PWD/${path##*/})"
 }
 
+# List changed files in the current PR.
+# This is implemented as a function so it can be mocked in unit tests.
+# It will fail if a file ever contained a newline character (which is bad practice anyway)
```
chizhg, line 657: This command will fail if the file name contains a newline character, or the file content? It would be surprising if it's the latter..

scripts/shellcheck-presubmit.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+set -e
+
+source "$(dirname "${BASH_SOURCE[0]}")/library.sh"
+
+if [[ IS_PROW -eq 0 ]]; then
```
chizhg, line 7:
```suggestion
if (( ! IS_PROW )); then
```

images/prow-tests/Dockerfile
coryrc: If you don't have it and something in the repo refers to `/scripts` directly, then you couldn't run those scripts locally without adding a symlink in your root directory.

scripts/library.sh
coryrc: fileNAME yes, I can update the comment to include the word `name`

images/prow-tests/Dockerfile
chizhg: Sorry it's still confusing to me. By `run those scripts locally` do you mean running them in the Docker container?

coryrc: No, run them locally NOT in a docker container.
Today there's no reason, but I'm trying to preempt a future issue.

chizhg: Got it.

/lgtm

PR #2032, opened 2020-05-07
Fix config updater

chizhg:
Not entirely sure if this can fix the issue, but changing to `go run` makes it almost impossible to test a postsubmit job.

/cc @chaodaiG


chaodaiG:
Not sure which failure you are trying to fix, but these are certainly good changes, so
/lgtm

/hold
As I think we can switch back to use custom image if you prefer

chizhg:
As per the offline discussion, we'll revert the `go run` change for the Prow job config.

PR #2033, opened 2020-05-07
Revert some changes to kntest

chizhg:
Go modules has a caveat that makes it hard to install kntest via `go get`:
https://github.com/golang/go/issues/30354
https://github.com/kubernetes/test-infra/issues/14712

So revert the current use of kntest, until we find a more reliable way for users to get it.

/cc @chaodaiG


chaodaiG:
This is sad

/lgtm

PR #2040, opened 2020-05-08
run shellcheck

chizhg:
Run shellcheck on the scripts:
`shellcheck -f diff scripts/*.sh hack/*.sh test/*.sh | git apply`


chizhg:
/hold
Need to make sure it does not break other tests.

scripts/library.sh
```diff
@@ -696,7 +696,7 @@ function get_latest_knative_yaml_source() {
     local major_minor="${branch_name##release-}"
     # Find the latest release manifest with the same major&minor version.
     local yaml_source_path="$(
-      gsutil ls gs://knative-releases/${repo_name}/previous/v${major_minor}.*/${yaml_name}.yaml 2> /dev/null \
+      gsutil ls gs://knative-releases/"${repo_name}"/previous/v"${major_minor}".*/"${yaml_name}".yaml 2> /dev/null \
```
coryrc, line 699: This would be better with double quotes across most of the line, i.e. before `gs:` and after `.yaml` rather than all the groupings inside

scripts/presubmit-tests.sh
```diff
@@ -181,7 +181,7 @@ function default_build_test_runner() {
     | sort | uniq | tr '\n' ' ')"
   for pkg in ${tagged_pkgs}; do
     # `go test -c` lets us compile the tests but do not run them.
-    if ! capture_output "${report}" go test -c -tags="${tags}" ${pkg} ; then
+    if ! capture_output "${report}" go test -c -tags="${tags}" "${pkg}" ; then
```
coryrc, line 184: I wasn't sure $pkg is always a single package, so I was worried about this one

scripts/presubmit-tests.sh
```diff
@@ -270,7 +270,7 @@ function default_integration_test_runner() {
   local failed=0
   for e2e_test in $(find test/ -name e2e-*tests.sh); do
     echo "Running integration test ${e2e_test}"
-    if ! ${e2e_test} ${options}; then
+    if ! ${e2e_test} "${options}"; then
```
coryrc, line 273: I think this will break something; I think $options is intended to be split

scripts/release.sh
```diff
@@ -50,9 +50,9 @@ function tag_images_in_yamls() {
   for file in $@; do
     [[ "${file##*.}" != "yaml" ]] && continue
     echo "Inspecting ${file}"
-    for image in $(grep -o "${DOCKER_BASE}/[a-z\./-]\+@sha256:[0-9a-f]\+" ${file}); do
+    for image in $(grep -o "${DOCKER_BASE}/[a-z\./-]\+@sha256:[0-9a-f]\+" "${file}"); do
       for region in "" ${GEO_REGIONS// /. }; do
-        gcloud -q container images add-tag ${image} ${region}${image%%@*}:${TAG}
+        gcloud -q container images add-tag "${image}" "${region}""${image%%@*}":"${TAG}"
```
coryrc, line 55: The last part would be better as `"${region}${image%%@*}:${TAG}"` without all the extra double-quotes inside

scripts/release.sh
```diff
@@ -148,14 +148,14 @@ function setup_upstream() {
...
-    git remote add upstream ${REPO_UPSTREAM}
+    git remote add upstream "${REPO_UPSTREAM}"
   fi
 }
 
 # Fetch the release branch, so we can check it out.
 function setup_branch() {
   [[ -z "${RELEASE_BRANCH}" ]] && return
-  git fetch ${REPO_UPSTREAM} ${RELEASE_BRANCH}:upstream/${RELEASE_BRANCH}
+  git fetch "${REPO_UPSTREAM}" "${RELEASE_BRANCH}":upstream/"${RELEASE_BRANCH}"
```
coryrc, line 158: same double-quote-around-the-group-comment here

test/e2e-tests.sh
```diff
@@ -23,7 +23,7 @@
 # Calling this script without arguments will create a new cluster in
 # project $PROJECT_ID, run the tests and delete the cluster.
 
-source $(dirname $0)/../scripts/e2e-tests.sh
+source $(dirname "$0")/../scripts/e2e-tests.sh
```
coryrc, line 27: while you are here...... could you replace all the `$0` with `${BASH_SOURCE[0]}` ?

scripts/e2e-tests.sh
```diff
@@ -171,7 +171,7 @@ function resolve_k8s_version() {
     E2E_CLUSTER_VERSION="${gke_versions[0]}"
     echo "Using latest version, ${E2E_CLUSTER_VERSION}"
   else
-    local latest="$(echo "${gke_versions[@]}" | tr ' ' '\n' | grep -E ^${target_version} | sort -V | tail -1)"
+    local latest="$(echo "${gke_versions[@]}" | tr ' ' '\n' | grep -E ^"${target_version}" | sort -V | tail -1)"
```
coryrc, line 174: this would be better with the `"` before the `^` instead of just after it

coryrc:
/lgtm
(because you have the hold)

scripts/library.sh
chizhg: done

scripts/release.sh
chizhg: done

scripts/release.sh
chizhg: done

test/e2e-tests.sh
chizhg: done

scripts/e2e-tests.sh
chizhg: done

scripts/presubmit-tests.sh
chizhg: `tagged_pkgs` is a list of packages so `$pkg` is always a single one.

scripts/presubmit-tests.sh
chizhg: I have removed the quotes. 
However, `$options` here is always empty as we have `local options=""` at the beginning... I added a comment there and maybe we can remove it in the future.

PR #2041, opened 2020-05-08
Update prow-tests to use gvm

### LINKED ISSUE #1823: Use gvm to install and switch different Go versions (?)
For Knative projects, we need to have Prow jobs running with different Go versions, e.g. 1.12, 1.13, 1.14. The current way we handle them is to build [multiple test images](https://github.com/knative/test-infra/tree/master/images) with different Go versions installed, and we need to rebuild all the images if we want to make any changes to the base image.

There is a tool called [gvm](https://github.com/moovweb/gvm), with which we can install and switch multiple different Go versions in one single image. With it, we won't need to build and maintain multiple images for multiple Go versions.

Note:
Having multiple images may not be a problem when we have image auto-update process, so whether we need gvm depends on which way we want to move forward with.

coryrc:
Fixes #1823

The image defaults to Go 1.13 per usual.

Set the environment variable GO_VERSION to something acceptable to
`gvm use` to get a different version of Go instead of using separate images.

Also fixes the way githubhelper is installed, adding comment that it is deprecated.

/cc @chizhg 
/cc @chaodaiG


coryrc:
@chizhg can I install kntest now to fix #2033 ? :-D

images/prow-tests/my-gvm.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
```
chizhg, line 1: Please add copyright header.

images/prow-tests/runner.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
```
chizhg, line 1: Ditto.

images/prow-tests/Dockerfile
```diff
@@ -41,32 +41,47 @@ RUN apt-get install -y uuid-runtime  # for uuidgen
...
+# Install the tool:
+RUN curl -s -S -L https://raw.githubusercontent.com/moovweb/gvm/master/binscripts/gvm-installer | bash
+# gvm requires one to "source /root/.gvm/scripts/gvm" after installing
+#  but in Dockerfile, each RUN is its own shell, so source'd in-RUN env vars are not propagated
+# So have created "my-gvm.sh" to source the above then run gvm
+COPY images/prow-tests/my-gvm.sh /usr/local/bin
+# Install our versions of Go
+RUN my-gvm.sh install go1.12 --prefer-binary
+RUN my-gvm.sh install go1.13 --prefer-binary
+RUN my-gvm.sh install go1.14 --prefer-binary
```
chizhg, line 62: I cannot remember clearly, but does `gvm` support installing minor go versions? Is there a way to update to a patch release like `go1.14.1`?

coryrc: yes:
```
root@fe6a5d65896e:/workspace# gvm install go1.13.1 --prefer-binary
Installing go1.13.1 from binary source
```

chizhg: Then let's install the minor version `go1.14.1` so the environment is more deterministic and more clear when we really need an update?

coryrc: It does... but we've never needed it before and then you also have to update all the Prow jobs to use the new GO_VERSION... but I'm fine with that if you are

coryrc: There's also a Go 1.14.2 so we should probably jump straight to it

chizhg: Hmm I tried 1.14.2 on Mac but it told me the binary was not found. But if works in the container I agree we can directly jump to it.

chizhg: And it seems we don't need to specify the minor version when using it, `gvm use go1.14` will just work.

coryrc: Interactively I installed go1.11.2:
`gvm install go1.11.2 --prefer-binary`
then
`gvm use go1.11` and it said:
`Now using version go1.11.2`

so you are right!

Installing go1.14.2 worked:
```
root@435c7b9d5d89:/workspace# gvm install go1.14.2 --prefer-binary
Installing go1.14.2 from binary source
```
So I'll install 1.14.2 in the image and check the use command works, then upload.

coryrc:
```
root@f9b34d9b819c:/workspace# gvm use go1.14
Now using version go1.14.2
root@f9b34d9b819c:/workspace# go version
go version go1.14.2 linux/amd64
root@f9b34d9b819c:/workspace# gvm use go1.13
Now using version go1.13.10
root@f9b34d9b819c:/workspace# go version
go version go1.13.10 linux/amd64
root@f9b34d9b819c:/workspace# 
```

chizhg:
/lgtm
/approve

Thanks Cory!

Issue #1364, opened 2019-09-18
Staging mechanism for new prow-tests images


coryrc:
Problem: No good or documented way to test new prow-tests images.

Solution:
Building on https://github.com/knative/test-infra/issues/329, add an optional presubmit using an alpha tagged prow-tests image, change a CI flow to use a beta-tagged image, and document the procedure.

adrcunha:
Don't change a CI flow. Instead, create a new one that exercises the components/tools. Add tests as necessary.

coryrc:
We discussed in person; plan is new presubmit same as CI master branch, then CI master branch using the image.

adrcunha:
So I don't get confused again about this:

* new presubmit job uses "alpha" image;
* test-infra CI flow uses "beta" image;
* everyone else uses "prod" image;

coryrc:
Oh, as another part of modifying the config generator, may want the ability to leave old releases on an old image (f.ex. updating to go 1.13 would break releases pre-0.9), but I won't actually set an old image at this time.

coryrc:
Don't forget to update the playbook as well

coryrc:
And, not test-infra CI flow using beta, knative/serving master CI flow uses beta image. It's the same test that is run in the new presubmit.

test-infra doesn't do enough interesting work to serve as a valid test case.

adrcunha:
We don't want to use knative/serving as a guinea pig. If test-infra doesn't have good enough test cases for the image, add new ones, even if they're used exclusively by the beta flow.

coryrc:
The presubmit runs in knative/serving. It runs the exact same job as CI master branch. There's no reason for it to fail, and it will be easy to revert.

We currently have no testing. This is better.

adrcunha:
> The presubmit runs in knative/serving. It runs the exact same job as CI master branch.

Almost. The CI job runs everything in sequence.

> There's no reason for it to fail, and it will be easy to revert.

Famous last words. 🤣 It's even easier and more contained if we use test-infra.

> We currently have no testing. This is better.

We do have testing, and they're meant to test the infrastructure, not Knative. If they're not enough for the prow-tests image, this is yet another reason to add more tests.

coryrc:
Creating a new CI flow as copy of the integration-tests CI flow in each repo.

coryrc:
2044 should have fixed this

coryrc:
[...]

PR #2044, opened 2020-05-09
Add beta prow-tests CI flows and cleanup code

### LINKED ISSUE #1734: Document default resource requests for Prow jobs

coryrc:
Add a copy of all continuous flows using the prow-tests:beta label and
running rarely but enough to get testing in a day across multiple
repos, also run mostly at low-use times to spread the load out.

Add comments explaining the generator better.

Use existing hash algorithm instead of the code that was here. This
will result in a one-time change to the minute that all periodic jobs
run at.

We currently have no way to test new prow-tests images.

Fixes #1734


tools/config-generator/periodic_config.go
```diff
@@ -226,7 +257,35 @@ func generatePeriodic(title string, repoName string, periodicConfig yaml.MapSlic
...
+	// If job is a continuous run, add a duplicate for pre-release testing of new prow-tests image
+	// It will run less often than source job
+	if isContinuousJob {
+		betaData := data.Clone()
+
+		// Change the name and image
+		betaData.PeriodicJobName += "-beta-prow-tests"
+		betaData.Base.Image = strings.ReplaceAll(betaData.Base.Image, ":stable", ":beta")
+
+		// Run 2 or 4 times a day
```
chaodaiG, line 273: The frequency of continuous job was computed based on timeout, if you don't like to change the timeout of these beta test jobs, at least there should be comment explaining the reason why not use generateCron function

coryrc: Sure, it's because their purpose is to test the beta prow-tests and has nothing to do with whatever the underlying job is. I will make it clearer.

tools/config-generator/types_test.go
```diff
@@ -0,0 +1,21 @@
...
+func TestJobDetailMap(t *testing.T) {
+	j := NewJobDetailMap()
+
+	local := []string{"continuous", "nightly"}
+
+	for _, t := range local {
+		j.Add("serving", t)
+	}
+
+	for i, _ := range local {
```
mattmoor, line 14: Format Go code:
```suggestion
	for i := range local {
```

mattmoor, line 14: Format Go code:
```suggestion
	for i := range local {
```

chizhg:
/lgtm
/hold
Please unhold when you are ready to get it merged.

coryrc:
Fixes #1364

PR #2057, opened 2020-05-12
Attempt fixing dep-collector

chizhg:
**What this PR does, why we need it**:
This is an attempt to fix dep-collector to get it working for release branches that were not switched to Go modules, see https://github.com/knative/pkg/pull/1325

/cc @chaodaiG

PR #2058, opened 2020-05-12
Fix CI/CD flows on release branches

chizhg:
https://github.com/knative/test-infra/pull/2057 didn't fix the issue. 

It's the way we installed `dep-collector` that caused this issue. After we switched test-infra to Go modules, the way we install `dep-collector` should be changed to `GO111MODULE=on go get knative.dev/test-infra/tools/dep-collector`, otherwise it will look for the license db file in the non Go modules way, which is from the vendor path where the tool resides, that's why we are seeing it looking for the db in `/home/prow/go/src/knative.dev/test-infra/vendor/github.com/google/licenseclassifier/licenses/licenses.db`.

However, the way we install `dep-collector` is hardcoded in the test script, so it's unpractical to make changes to every release branch to fix the issue. Since the tool will not be installed again if it already exists, we can install the tool from the prow-tests image. But also since `gvm` will change the GOPATH when it installs a new Go version, the current `prow-tests` image will not work well with non Go modules projects.

So this PR is adding a temporary `prow-tests-go113` image to fix the issue on all existing release branches (<= 0.14). After we only support release branches that is >= 0.15, this image can be safely deleted.

/cc @chaodaiG

### DIFF
--- a/images/prow-tests-go113/README.md
+++ b/images/prow-tests-go113/README.md
@@ -3,4 +3,4 @@ projects switched to use Go modules. Unless urgent fixes are needed, this image
 should never be touched.
 
 This image should be deleted when we stop supporting non Go modules branches,
-and the latest one is `release-0.14`. 
+and the latest one is `release-0.14`.
--- a/images/prow-tests-go113/README.md
+++ b/images/prow-tests-go113/README.md
@@ -1,0 +1,6 @@
+This is an image used in the CI/CD flows for release branches before Knative
+projects switched to use Go modules. Unless urgent fixes are needed, this image
+should never be touched.
+
+This image should be deleted when we stop supporting non Go modules branches,
+and the latest one is `release-0.14`. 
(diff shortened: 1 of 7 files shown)


images/prow-tests-go113/README.md
```diff
@@ -0,0 +1,6 @@
+This is an image used in the CI/CD flows for release branches before Knative
+projects switched to use Go modules. Unless urgent fixes are needed, this image
+should never be touched.
+
+This image should be deleted when we stop supporting non Go modules branches,
+and the latest one is `release-0.14`.
```
mattmoor, line 6: Remove trailing whitespace:
```suggestion
and the latest one is `release-0.14`.
```

mattmoor, line 6: Format markdown:
```suggestion
and the latest one is `release-0.14`.
```

chaodaiG:
as long as it works

/lgtm

PR #2067, opened 2020-05-14
Fix missing copyright and warn on silent go-license failure

coryrc:
/assign @chizhg 
/assign @chaodaiG 
/cc @chizhg

### DIFF
--- a/scripts/library.sh
+++ b/scripts/library.sh
@@ -544,7 +544,7 @@ function update_licenses() {
   local dst=$1
   local dir=$2
   shift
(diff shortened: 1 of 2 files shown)


scripts/library.sh
```diff
@@ -544,7 +544,7 @@ function update_licenses() {
   local dst=$1
   local dir=$2
   shift
-  run_go_tool github.com/google/go-licenses go-licenses save "${dir}" --save_path="${dst}" --force || return 1
+  run_go_tool github.com/google/go-licenses go-licenses save "${dir}" --save_path="${dst}" --force || { echo "--- FAIL: go-licenses failed to update licenses"; return 1; }
```
chizhg, line 547: Can we add line break here?

coryrc: done

PR #2068, opened 2020-05-15
Create script to "vendor" test-infra/scripts

coryrc:
Tracks the contents of test-infra/scripts into each repo at
scripts/test-infra directory.

The script provides both installation assistance (with --first-time
command-line argument) as well as its usual function of updating
scripts/test-infra directory and is intended to be called by
hack/update-deps.sh in each repo. The --branch option lets you choose
a particular branch (instead of master) if desired.

Helps with #1823 (because knative/client is preventing rollout of new image through every release and needs script updates)

### DIFF
--- a/scripts/update-test-infra.sh
+++ b/scripts/update-test-infra.sh
@@ -39,7 +39,7 @@ declare SCRIPTS_REF=master
 while [[ $# -ne 0 ]]; do
   parameter="$1"
   case ${parameter} in
-    --branch)
+    --ref)
       shift
       SCRIPTS_REF="$1"
       ;;
@@ -49,7 +49,10 @@ while [[ $# -ne 0 ]]; do
     --update)
       DO_UPDATE=1
       ;;
-    *) abort "unknown option ${parameter}" ;;
+    *)
+      echo "unknown option ${parameter}"
+      exit 1
+      ;;
   esac
   shift
 done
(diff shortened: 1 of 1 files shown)


mattmoor:
I'm not comfortable making a change of this magnitude less than a week before we are supposed to cut the test-infra/pkg release and pinning them everywhere.  This should wait until the first week of 0.16.

TBH, I haven't heard a particularly clear articulation of what the problem is, so can we start with an issue that writes up the issue in detail before we prescribe solutions?

scripts/update-test-infra.sh
```diff
@@ -0,0 +1,91 @@
...
+    echo "test-infra scripts installed to scripts/test-infra from branch ${SCRIPTS_REF}"
+}
+
+function run() {
+  if (( FIRST_TIME_SETUP )); then
+    if [[ ! -d .git ]]; then
+      echo "I don't believe you are in a repo root; exiting"
+      exit 5
+    fi
+    git remote add test-infra https://github.com/knative/test-infra.git || true
```
chizhg, line 74: Why do we ignore the error?

coryrc: It errors if the remote is already set, so running the script twice would otherwise fail.
I don't want to add and remove the remote, because you override the upstream by changing the remote.
I could instead make it an option?

chizhg: I see. I think adding a comment, or printing out a message like "the test-infra upstream has been set, skipping" would be helpful here. But it's also fine with the current way as people normally won't have `test-infra` as upstream.

chizhg:
/lgtm
/hold
Feel free to unhold if you don't want to make changes for my comment.

scripts/update-test-infra.sh
coryrc: Message added!

PR #2077, opened 2020-05-15
Unify images to reduce burden and add testing

coryrc:
With the beta prow-tests flows, we can now test things before
releasing.

Unify prow-tests and coverage; coverage-dev is replaced in
functionality with prow-tests:beta (or another :tag if desired).

Unify the support programs into one new image. None rely on ENTRYPOINT
being set, so there's no problem grouping them together.

Additionally, the individual utility images are already decaying and differing from one another. We need to keep them up-to-date or tech debt will build.

/cc @chizhg 
/cc @chaodaiG


chizhg:
/lgtm for me
Defer to @chaodaiG for the `support-utilties`

chaodaiG:
I like this kind of incremental changes. It's a good idea to put coverage tool in prow-tests image. 

However, I'm not sure it's best practice bundle all other tools into a single image. Not that they interfere with each other, I'm more concerned about the different support levels they are on, i.e. if one of the tool is broken, how can we still ship the image. One of the other thing I have been wanting, is automatically pushing images when changes occur, and in my opinion having them in separate images would make the automation easier to manage

Also, for any tools that don't need source code from repo other than test-infra, I think we should consider getting away from using images

coryrc:
> However, I'm not sure it's best practice bundle all other tools into a single image. Not that they interfere with each other, I'm more concerned about the different support levels they are on, i.e. if one of the tool is broken, how can we still ship the image. One of the other thing I have been wanting, is automatically pushing images when changes occur, and in my opinion having them in separate images would make the automation easier to manage

Yeah I wasn't sure. If you've got a plan to automate creating an image and uploading, that sounds fine to me. Each image is templated; how about a Makefile that takes in the tool name and creates the image (so we only need one Dockerfile)?

> Also, for any tools that don't need source code from repo other than test-infra, I think we should consider getting away from using images

How else would they run?

chaodaiG:
> Yeah I wasn't sure. If you've got a plan to automate creating an image and uploading, that sounds fine to me. Each image is templated; how about a Makefile that takes in the tool name and creates the image (so we only need one Dockerfile)?

That's totally fine with me
 
> How else would they run?

We can use `go run` instead, like this example: https://github.com/knative/test-infra/blob/7a28f47cdbcb11b92359528689b57030fafc127d/config/prod/prow/jobs/custom-jobs.yaml#L306

coryrc:
> We can use `go run` instead, like this example:
> 
> https://github.com/knative/test-infra/blob/7a28f47cdbcb11b92359528689b57030fafc127d/config/prod/prow/jobs/custom-jobs.yaml#L306

Oh, clone the repo and always run from head. They are non-critical so that could be okay.

images/tool/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+
+# Temporarily add test-infra to the image to build custom tools
+COPY . /go/src/knative.dev/test-infra
+
+# TODO: jobs should not run `/command`, just `command`
+
+RUN go install "/go/src/knative.dev/test-infra/tools/${TOOLS_SUBDIR}${TOOL_NAME}"
+RUN cp "$(which ${TOOL_NAME})" /
+
+RUN [ "${TOOL_NAME}" = flaky-test-reporter ] && mkdir -p /config || true
```
chaodaiG, line 31: Why ignore this error?

coryrc: Because it's not an error if TOOL_NAME != flaky-test-reporter

chaodaiG: if `mkdir -p /config` failed it'll also be ignored right?

coryrc: Dockerfile doesn't support conditionals, so not sure what else to do

coryrc: unless I can write an `if`?

PR #2096, opened 2020-05-19
Consolidate prow-tests images

### LINKED ISSUE #1823: Use gvm to install and switch different Go versions (?)

coryrc:
Now that the prow-tests image with gvm has been tested on the
beta-prow-tests flows and is working, we can consolidate down to a
single image in the jobs (which also simplifies the config generator a
little).

A few -go114 jobs were missing GO_VERSION=go1.14 and this fixes that. It was not trivial to set GO_VERSION=go1.13 in the config so this was not used. Likely we should make the go version mandatory in the input config.

Coverage is not yet figured out, so don't modify its output yet.

Fixes #1823

### DIFF
--- a/tools/config-generator/main.go
+++ b/tools/config-generator/main.go
@@ -305,16 +305,6 @@ func exclusiveSlices(a1, a2 []string) []string {
 	return res
 }
 
-// getGo114ID returns image identifier for go114 images
-func getGo114ID() string {
-	return "-go114"
-}
-
-// getGo113ID returns image identifier for go113 images
-func getGo113ID() string {
-	return "-go113"
-}
-
 // strip out all suffixes from the image name
 func stripSuffixFromImageName(name string, suffixes []string) string {
 	parts := strings.SplitN(name, ":", 2)
@@ -340,10 +330,6 @@ func addSuffixToImageName(name string, suffix string) string {
 		parts[0] = fmt.Sprintf("%s%s", parts[0], suffix)
 	}
 	return strings.Join(parts, ":")
-}
-
-func getGo114ImageName(name string) string {
-	return addSuffixToImageName(stripSuffixFromImageName(name, []string{getGo113ID()}), getGo114ID())
 }
 
 // Consolidate whitelisted and skipped branches with newly added
@@ -455,6 +441,10 @@ func createCommand(data baseProwJobTemplateData) []string {
 	return append(c, data.Args...)
 }
 
+func envNameToKey(key string) string {
+	return "- name: " + key
+}
+
 // addEnvToJob adds the given key/pair environment variable to the job.
 func (data *baseProwJobTemplateData) addEnvToJob(key, value string) {
 	// Value should always be string. Add quotes if we get a number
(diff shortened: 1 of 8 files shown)


coryrc:
Wait for #2077 then unhold
/hold

tools/config-generator/main.go
```diff
@@ -455,14 +441,38 @@ func createCommand(data baseProwJobTemplateData) []string {
...
 	}
 
-	data.Env = append(data.Env, "- name: "+key, "  value: "+value)
+	data.Env = append(data.Env, envNameToKey(key), "  value: "+value)
+}
+
+func (data *baseProwJobTemplateData) SetGoVersion(version string) {
+	match, _ := regexp.MatchString(`go\d+[.]\d+`, version)
+	if !match {
+		panic(fmt.Errorf("Bad version string to SetGoVersion: %q", version))
```
chizhg, line 461: Lowercase for error message :)

tools/config-generator/main.go
```diff
@@ -612,8 +622,7 @@ func parseBasicJobConfigOverrides(data *baseProwJobTemplateData, config yaml.Map
 		(*data).ExtraRefs = append((*data).ExtraRefs, "  "+(*data).PathAlias)
 	}
 	if needGo114 {
-		data.addEnvToJob("GO_VERSION", "go1.14")
-		(*data).Image = getGo114ImageName((*data).Image)
+		data.SetGoVersion("go1.14")
```
chizhg, line 625: Probably better to define `go1.14` as a constant, as it's used multiple times.

coryrc: also done, more-or-less

tools/config-generator/types.go
```diff
@@ -53,3 +72,21 @@ func NewTestGridMetaData() TestGridMetaData {
 		nonAligned: make([]NonAlignedTestGroup, 0),
 	}
 }
+
+func NewGoVersion(version string) GoVersion {
```
chizhg, line 76: Do we need this function? Seems it's unused.

tools/config-generator/types.go
```diff
@@ -1,5 +1,19 @@
 package main
```
chizhg, line 1: Copyright header is missed for this file.

tools/config-generator/types.go
coryrc: It is currently unused, but we should transition the config input template from hard-coded "Go114" and "Go113" keys to `GoVersion: "go1.14"` or similar, and wanted to have a validator

tools/config-generator/types.go
coryrc: done

tools/config-generator/types.go
chizhg: Got it.

PR #2098, opened 2020-05-20
Prepare for coverage in prow-tests

coryrc:
All coverage-go114 jobs now have GO_VERSION env variable.

prow-tests now has coverage installed and a utility script to
translate `/coverage` to `runner.sh coverage` meaning it will use the
GO_VERSION of Go to run coverage.

Add beta-prow-tests jobs for coverage too.


images/prow-tests/in-gvm-env.sh
```diff
@@ -0,0 +1,40 @@
...
+set -e
+set -x
+
+if [[ ! -v GO_VERSION ]]; then
+  echo "--- FAIL: need GO_VERSION defined for in-gvm-env.sh"
+  exit 1
+fi
+gvm use "${GO_VERSION}"
+# Get our original Go directory back into GOPATH
+pushd /go || exit 5
```
chizhg, line 28: Why using 5 - `I/O error`?

images/prow-tests/in-gvm-env.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
```
chizhg, line 1: What's the difference between this script and `my-gvm.sh`, can we give them better names, or add a README to explain what are the scripts supposed to do? And also it seems the code in the two scripts are somewhat duplicated.

images/prow-tests/in-gvm-env.sh
coryrc: Who said that's what it means?

images/prow-tests/in-gvm-env.sh
coryrc: Added comments to each.

Renaming `my-gvm.sh` => `source-gvm-and-run.sh`

tools/config-generator/periodic_config.go
```diff
@@ -321,16 +328,44 @@ func generateGoCoveragePeriodic(title string, repoName string, _ yaml.MapSlice)
...
+		if repo.Go114 {
 			data.Base.SetGoVersion(GoVersion{1, 14})
 		}
 		addExtraEnvVarsToJob(extraEnvVars, &data.Base)
 		addMonitoringPubsubLabelsToJob(&data.Base, data.PeriodicJobName)
 		configureServiceAccountForJob(&data.Base)
 		executeJobTemplate("periodic go coverage", readTemplate(periodicCustomJob), title, repoName, data.PeriodicJobName, false, data)
-		return
+
+		betaData := data.Clone()
```
chaodaiG, line 342: A periodic job doesn't exercise the core logic of coverage job, the coverage diffing part, can we make this job somehow go through that code path to make sure?

images/prow-tests/in-gvm-env.sh
chizhg: I checked from https://www.cyberciti.biz/faq/linux-bash-exit-status-set-exit-statusin-bash/.
Actually it does not matter to me as long as it's a non-zero issue, just being curious.

tools/config-generator/periodic_config.go
coryrc: Not in any way we can display in TestGrid that I can think of

coryrc: The :coverage-dev job presubmit still exists

images/prow-tests/in-gvm-env.sh
coryrc: https://en.wikipedia.org/wiki/Exit_status says there are various "conventions" and no certain meanings.

I just picked an arbitrary non-1 number, but 2 is also an arbitrary non-1 number :)

tools/config-generator/periodic_config.go
chaodaiG: Maybe we don't need it? this much custom logic in config generator makes me worry about maintenance

coryrc: The time for that ended a long time ago.

I want to have some sign coverage works with `prow-tests:beta`. This verifies at least half its functionality.

coryrc: Also, it's two function calls and a nice, clear, documented struct.

chaodaiG: I see line 342 to 369, like 28 lines of code rather than two function calls though

coryrc: Sure, but it's mostly naming things and static assignment of values. The only functions of consequence are a call to `executeJobTemplate` and `metaData.AddNonAlignedTest`.

We need some testing before releasing. This is decent and encompasses a lot of the functionality of the program, even if it isn't all of it.

chaodaiG: Well, I hate to block you because of this. Let's clean this mess up later

chaodaiG:
/approve
Defer to @chizhg for lgtm

PR #2112, opened 2020-05-22
Cleanup images

coryrc:
coverage-go114:latest is already pointing to the prow-tests image, see
https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/knative_serving/8034/pull-knative-serving-go-coverage-dev/1263625182585556992
and https://github.com/knative/serving/pull/8034

We have unified all prow-tests* and coverage* images into the single
prow-tests image. Now delete the old.

/cc @chizhg
/cc @chaodaig
/assign @chizhg

PR #2113, opened 2020-05-22
Fixing coverage jobs tool path in prow jobs

chaodaiG:
Context: https://github.com/knative/test-infra/pull/2112 switched all coverage jobs to use `prow-tests` image instead of `coverage` image, but forgot to update the invocation path from `coverage` to `/coverage`, see where coverage tool is located here:
https://github.com/knative/test-infra/blob/f44a701116b0c4013c464db9cc9e295407f384e5/images/prow-tests/Dockerfile#L111

/assign coryrc chizhg

### DIFF
--- a/config/prod/prow/jobs/config.yaml
+++ b/config/prod/prow/jobs/config.yaml
@@ -495,7 +495,7 @@ presubmits:
         command:
         - runner.sh
         args:
-        - "coverage"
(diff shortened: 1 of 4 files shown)


chaodaiG:
/hold
This doesn't fully solve the problem

Issue #2104, opened 2020-05-20
Separate prow job update job from others


chaodaiG:
Currently, once a PR containing prow jobs only changes, the config is updated right away by https://prow.knative.dev/?job=post-knative-prow-config-updater. This could be separated out of the process by simply updating prow job configs if the PR contains prow job changes

chizhg:
FYI https://github.com/kubernetes/test-infra/tree/master/prow/plugins/updateconfig could be helpful here. It can even be configured on different repos.

chaodaiG:
> FYI https://github.com/kubernetes/test-infra/tree/master/prow/plugins/updateconfig could be helpful here. It can even be configured on different repos.

Sweet!

PR #2127, opened 2020-05-27
Use updateconfig plugin for updating prow job configs

chaodaiG:
Fixes: #2104 

See  https://github.com/kubernetes/test-infra/tree/master/prow/plugins/updateconfig

/assign chizhg albertomilan

### DIFF
--- a/config/staging/prow/core/plugins.yaml
+++ b/config/staging/prow/core/plugins.yaml
@@ -73,6 +73,7 @@ config_updater:
   maps:
     config/prod/staging/jobs/*.yaml:
       name: job-config
+      gzip: true
 # Plugins enabled for each repo.
 plugins:
   knative-prow-robot/serving:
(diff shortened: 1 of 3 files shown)


config/staging/prow/core/plugins.yaml
```diff
@@ -121,6 +126,32 @@ plugins:
   - welcome
   - wip
   - yuks
+  knative-prow-robot/test-infra:
```
chizhg, line 129: Will it work to have two `knative-prow-robot/test-infra`s in this file?

chaodaiG: Actually, I think at least it won't fail prow, see https://github.com/knative/test-infra/blob/350178ab2a0e4780bc9d039cf7a72b3d4ce183b9/config/staging/prow/core/config.yaml#L136. 

It might override the wrong way, but since this only affects staging prow, and staging prow is already covered by auto config updater, we should be fine

PR #2131, opened 2020-05-29
Update go.mod to specify the module is go1.14

dprotaso:
Thus the go commands will default to -mod=vendor

See: https://golang.org/doc/go1.14#go-command


chizhg:
/lgtm
/approve
Thanks!

PR #2141, opened 2020-06-02
Define scripts/ as a submodule

### LINKED ISSUE #2078: Investigate using multiple Go modules to solve the scripts vendor issue
At this point (May 15th, 2020), we have switched all Knative projects to use Go modules, but it brings in a regression that makes it impossible to only update vendored scripts from `test-infra`, when vendored `pkg` is not at HEAD. More discussions can be found from https://github.com/knative/test-infra/issues/2069.

Using multiple Go modules in the test-infra repo can potentially solve this issue, but we need to evaluate carefully to make sure it does not introduce other side effects.

The doc for multiple modules repository is in https://github.com/golang/go/wiki/Modules#faqs--multi-module-repositories.

/assign
FYI @coryrc @chaodaiG @albertomilan

chizhg:
I did some experiment with Go multiple modules to vendor `scripts/`, and did not see any issues.

The only changes needed in each repo will be changing the floating deps from `knative.dev/test-infra` to `knative.dev/test-infra/scripts` in for example https://github.com/knative/serving/blob/master/hack/update-deps.sh#L35. And even if it's not changed, it can still work in the old way.

Fixes #2078 

/cc @chaodaiG @albertomilan


chaodaiG:
Looks promising, can you fix the build error first?

chizhg:
> Looks promising, can you fix the build error first?

I didn't find an elegant way to fix the build errors. It seems `go build -v ./...` cannot work with a repo that has multiple Go modules and will get an error `go: directory scripts is outside main module`, I'm not getting the error when I run the command locally though...
I think we can just skip building the submodule it contains mainly scripts, and since we are already skipping a few directories, the approach is not that dirty.

chaodaiG:
> > Looks promising, can you fix the build error first?
> 
> I didn't find an elegant way to fix the build error. It seems `go build -v ./...` cannot work with a repo that has multiple Go modules and will get an error `go: directory scripts is outside main module`, I'm not getting the error when I run the command locally though...
> 
> I think we can just skip building the submodule it contains mainly scripts, and since we are already skipping a few directories, the approach is not that dirty.

I couldn't repro the error locally either, maybe we should figure out why this happened instead of going straight to the workaround?

chizhg:
> > > Looks promising, can you fix the build error first?
> > 
> > 
> > I didn't find an elegant way to fix the build error. It seems `go build -v ./...` cannot work with a repo that has multiple Go modules and will get an error `go: directory scripts is outside main module`, I'm not getting the error when I run the command locally though...
> > I think we can just skip building the submodule it contains mainly scripts, and since we are already skipping a few directories, the approach is not that dirty.
> 
> I couldn't repro the error locally either, maybe we should figure out why this happened instead of going straight to the workaround?

I actually can reproduce the error by running `./test/presubmit-tests.sh --build-tests`. The error happens in command `go test -c -tags="e2e library local tools" ./scripts`.

There are two issues with this command:
1. Currently we are assuming all packages that have build tags are test packages, but it's not necessarily true.
2. As we are now defining `./scripts` as a go submodule, even `go build -tags="tools" ./scripts` will get the `go: directory scripts is outside main module` error, the right way to build it will be `cd scripts/ && go build -tags="tools" .`.

Given the first issue and the fact the submodule only contains a `dummy.go` file, I think it does make sense for us to explicitly skip building it. I have added a TODO to support building normal go packages with build tags.

PR #2144, opened 2020-06-02
Also update prow core/plugin using configupdater

chaodaiG:
This is in preparing for stopping using in house auto config updater. After this change, the postsubmit jobs just need to do `make -C config/prod update-prow-cluster` etc.

/assign chizhg

PR #2145, opened 2020-06-02
Updating prow configs straight in postsubmit jobs instead of auto con…

chaodaiG:
This change moves from updating prow cluster with auto-config-updater to directly using make rules in postsubmit jobs. Key updates:

- Stop auto-config-updater
- Stop checking for changes against prod prow cluster config
- Use direct make rules in postsubmit jobs, separated as:
  - prow cluster deployments
  - build cluster deployments
  - boskos resources
  - testgrid update

The implicit impact is knative staging prow will not be tracked any more

/assign chizhg

/hold
Better be merged after #2144

### DIFF
--- a/config/prod/prow/jobs/custom-jobs.yaml
+++ b/config/prod/prow/jobs/custom-jobs.yaml
@@ -274,7 +274,7 @@ postsubmits:
     path_alias: knative.dev/test-infra
     max_concurrency: 1
     cluster: "prow-trusted"
-    run_if_changed: "^config/prod/prow/cluster/.*.yaml$"
+    run_if_changed: "^config/prod/(prow/cluster/.*.yaml|build-cluster/(cluster|boskos)/.*.yaml|prow/testgrid/testgrid.yaml)$"
     branches:
     - "master"
     spec:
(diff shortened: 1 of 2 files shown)


config/prod/prow/jobs/custom-jobs.yaml
```diff
@@ -291,44 +268,31 @@ periodics:
 
 postsubmits:
   knative/test-infra:
-  - name: post-knative-prow-config-updater
+  - name: post-knative-prow-cluster-config-updater
```
chizhg, line 271: Do we need 4 jobs? Can we just have one job updating all?

config/prow_common.mk
```diff
@@ -100,10 +100,14 @@ update-almost-all-cluster-deployments:: confirm-master
 	$(SET_CONTEXT)
 	@for f in $(filter-out $(wildcard $(PROW_DEPLOYS)/*boskos*.yaml),$(wildcard $(PROW_DEPLOYS)/*.yaml)); do kubectl apply -f $${f}; done
 	$(UNSET_CONTEXT)
+
+update-almost-all-build-cluster-deployments:: confirm-master
```
chizhg, line 104: We are adding too many commands in this Makefile I think... Most of the commands are actually not needed, as all the deployments are idempotent.

config/prod/prow/jobs/custom-jobs.yaml
chaodaiG: I'd like to divide it at least to 2:
- prow cluster deployment (This will be deprecated pretty soon)
- prow build cluster deployment

config/prow_common.mk
chaodaiG: I'll take a look

config/prod/prow/jobs/custom-jobs.yaml
chaodaiG: Actually I think you're right, combined to be a single job

config/prow_common.mk
chaodaiG: Removed some of the targets

chizhg:
/lgtm
Feel free to unhold.

chaodaiG:
/unhold
Let's see how it goes

PR #2147, opened 2020-06-03
Use go version defined in go.mod to run Prow jobs

### LINKED ISSUE #2118: Go version should be defined in-repo
Each repo should be able define what GO_VERSION becomes when running jobs. The configuration generator can provide a default value, but each repo should have a way to define what version of Go it desires.

Advantages:
1. Each repo can decide when to switch Go versions w/o waiting on test-infra to update jobs
2. Running tests in docker allows the GO_VERSION to be determined accurately without having to keep scripts updated or intertwined with the config generator.
3. Simplifies the config generator
4. Historical versions will stay accurate

I recommend this be done an extensible way by having [runner.sh](https://github.com/knative/test-infra/blob/master/images/prow-tests/runner.sh) source a shell script in each repo root (or some other location?), if it exists.

/cc @peterfeifanchen @chizhg

chizhg:
Now that all Knative projects have been switched to go mod, we can parse `go.mod` file to dynamically determine the go version to use in the CI/CD flows. In this way, whenever a project wants to a new Go version, they can just update the `go.mod` file, no changes will be needed for Prow jobs as long as that version of Go has been installed in the `prow-tests` image.

Fixes #2118 

/cc @chaodaiG @albertomilan @peterfeifanchen

### DIFF
--- a/images/prow-tests/Dockerfile
+++ b/images/prow-tests/Dockerfile
@@ -75,8 +75,6 @@ RUN GO111MODULE=on go get -u github.com/google/go-licenses
 RUN bazel version > /bazel_version
 RUN ko version > /ko_version
 
-# Set default Go version
-ENV GO_VERSION go1.13
 # Note, it's pointless to run `source-gvm-and-run.sh use go1.13` in this Dockerfile because the PATH changes it makes won't stay in effect
 # We wrap the runner with our own which will run `gvm use $GO_VERSION` for us.
 RUN mv "$(which runner.sh)" "$(dirname "$(which runner.sh)")/kubekins-runner.sh"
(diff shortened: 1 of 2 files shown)


chaodaiG:
Does submodule also define go version?

chizhg:
> > > Does submodule also define go version?
> > 
> > 
> > It does, but no other repos (other than test-infra) are currently using Go submodules, and they don't have plan to AFAIK.
> > In the future if some repos are indeed going to use Go submodules extensively (not just a `dummy.go` file as we have in `test-infra/scripts`), we'll have to make more changes to test-infra accordingly, at least `build-tests` and `unit-tests`. Thus, I think we don't need to consider Go submodules for now and it's not a blocker for this PR.
> 
> I think that makes sense to me. How about older release branches that don't have go.mod?

According to the current logic, it will use the default version `go1.13` if `go.mod` file is not found and `GO_VERSION` env var is not defined.

chaodaiG:
> According to the current logic, it will use the default version `go1.13` if `go.mod` file is not found and `GO_VERSION` env var is not defined.

I see, so it still keeps backward compatible
Thanks!
/lgtm

PR #2148, opened 2020-06-03
Remove the logic to set go version via Prow jobs

chizhg:
After https://github.com/knative/test-infra/pull/2147 is merged and we have a new image, we can simplify the Prow config generator to remove the logic that sets go version via Prow jobs. This change also significantly simplifies the generated Prow job config file, as we don't need to differentiate the release branches for presubmit/postsubmit jobs. 

/cc @chaodaiG @albertomilan @peterfeifanchen 
/hold

### DIFF
--- a/tools/config-generator/main.go
+++ b/tools/config-generator/main.go
@@ -308,69 +308,6 @@ func exclusiveSlices(a1, a2 []string) []string {
 	return res
 }
 
-// Consolidate whitelisted and skipped branches with newly added
-// whitelisted/skipped. To make the logic easier to maintain, this function
-// makes the assumption that the outcome follows these rules:
-//   - Special branch logics always apply on master and future branches
-// Based on the previous rule, if there is a special branch logic, the 2 Prow
-// jobs that serves different branches become:
-//   - Standard job definition:
-//		- whitelisted: [release-0.1]
-//		- skipped: []
-//   - Standard job definition + branch special logic #1:
-//		- whitelisted: []
-//		- skipped: [release-0.1]
-// And when there is a new special logic comes up with different list of release
-// branches to exclude, for example [release-0.1, release-0.2], then the desired
-// outcome becomes:
-//   - Standard job definition:
-//		- whitelisted: [release-0.1]
-//		- skipped: []
-//   - Standard job definition + branch special logic #1: (This will never run)
-//		- whitelisted: []
-//		- skipped: []
-//   - Standard job definition + branch special logic #2:
-//		- whitelisted: [release-0.2]
-//		- skipped: []
-//   - Standard job definition + branch special logic #1 + branch special logic #2:
-//		- whitelisted: []
-//		- skipped: [release-0.1, release-0.2]
-// Noted that only jobs with all special branch logics have something in
-// skipped, while all other jobs only have whitelisted. This rule also applies
-// when there is a third branch specific logic and so on.
-// This function takes the logic above, and determines whether generate
-// whitelisted or skipped as output.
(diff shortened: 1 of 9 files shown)


tools/config-generator/main.go
```diff
@@ -853,41 +796,14 @@ func recursiveSBL(repoName string, data interface{}, generateOneJob func(data in
 // executeJobTemplateWrapper takes in consideration of repo settings, decides how many variants of the
 // same job needs to be generated and generates them.
 func executeJobTemplateWrapper(repoName string, data interface{}, generateOneJob func(data interface{})) {
-	var sbs []specialBranchLogic
-
 	switch data.(type) {
```
albertomilan, line 799: Is this switch necessary with the new logic? It seems for all cases we are invoking `generateOneJob(data)`

tools/config-generator/main.go
```diff
@@ -551,7 +511,7 @@ func setResourcesReqForJob(res yaml.MapSlice, data *baseProwJobTemplateData) {
 // parseBasicJobConfigOverrides updates the given baseProwJobTemplateData with any base option present in the given config.
 func parseBasicJobConfigOverrides(data *baseProwJobTemplateData, config yaml.MapSlice) {
 	(*data).ExtraRefs = append((*data).ExtraRefs, "  base_ref: "+(*data).RepoBranch)
-	var needDotdev, needGo114 bool
+	var needDotdev bool
 	for i, item := range config {
 		switch item.Key {
```
albertomilan, line 516: I would suggest including a `default` case in this switch which logs the unexpected config, in case any client is still providing the deprecated "go*" values.

tools/config-generator/main.go
chizhg: You are right, it's not needed anymore. But I'm a bit worried that we'll need the `specialBranchLogic` again in the future, so I feel we should not clean up too much of the code.

cc @chaodaiG for opinions since he was the initial author for the logic here.

tools/config-generator/main.go
chizhg: This tool is actually only used by us, and I have removed all the `go114` field from our meta config files, so it's safe to deprecate it here.

tools/config-generator/main.go
chaodaiG: I have the same feeling with Chi, would suggest we keep the logic for a little longer.

I have another question though: it's nice to converge master and release branch jobs, but would this mean that previous release branch with no go.mod will use go1.13?

chizhg: That is true. But nothing regresses as the infra is only supposed to support the last 4 releases, and all of them are already using Go 1.13.

chaodaiG: Ack

chaodaiG:
/approve
From my side

tools/config-generator/main.go
albertomilan: I'm fine keeping the code, especially if it's going to be restored in the near future. But I would want to hear we will need it in the short-term (during the next few weeks). Otherwise, I will consider removing it for health purposes. We will always have the history of the file to get it back if we need it.

tools/config-generator/main.go
albertomilan: It's definitely not a requirement, so I'm ok with it. This was mostly a suggestion not only for this case but to cover all unexpected cases and root cause the issue quickly. If you feel confident with the current state, merge it.

tools/config-generator/main.go
chizhg: Makes sense. I have cleaned up more of the code.

PR #2167, opened 2020-06-04
Taking a hook bug fix

chaodaiG:
Take in a bug fix from https://github.com/kubernetes/test-infra/pull/17846, so that updateconfig plugin error can be surfaced

/assign chizhg albertomilan


chaodaiG:
/hold
Depends on https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/post-test-infra-push-prow/1268669829938679812

PR #2173, opened 2020-06-05
Use runner.sh for the config-updater job

chizhg:
The logic to enable DIND is in `runner.sh`, so we need to use it as the test driver for the config-updater job.

/cc @chaodaiG


chaodaiG:
Good catch, thanks!
/lgtm

Issue #1871, opened 2020-03-30
Text under DataRaceAnalysis tab of Spyglass log page does not have correct line feeds


chizhg:
See https://prow.knative.dev/view/gcs/knative-prow/logs/ci-knative-serving-continuous/1232488190519021569

/bug
/kind good-first-issue

chizhg:
[...]

PR #2179, opened 2020-06-08
Install kntest in Dockerfile and use it to create junit xml file

chizhg:
In https://github.com/knative/test-infra/pull/2033 we reverted a change that started using `kntest`, as `go get` does not work properly. Now that we have decided to provide utilities for developers to run test flows from the `prow-tests` image, and for now developers normally do not run the test scripts directly so they won't need `kntest` on their local, let's start installing `kntest` in the Dockerfile and use it to create junit xml file to fix https://github.com/knative/test-infra/issues/1871

/cc @chaodaiG

### DIFF
--- a/images/prow-tests/Dockerfile
+++ b/images/prow-tests/Dockerfile
@@ -109,7 +109,8 @@ RUN go install knative.dev/test-infra/tools/coverage
 # TODO: Prow jobs run `runner.sh coverage` instead `/coverage`
 COPY images/prow-tests/coverage.sh /coverage
 
-# RUN go install knative.dev/test-infra/kntest/cmd/kntest # Uncomment when you wish to have kntest installed
+# Install kntest
+RUN go install knative.dev/test-infra/kntest/cmd/kntest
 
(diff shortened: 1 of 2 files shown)


chaodaiG:
This is cool. I think we still should support local development mode though, and understand that `go get` doesn't work. I would still expect a warning message like `kntest not installed, please install via ./install_kntest.sh` script etc. 

Note: the `./install_kntest.sh` is just an idea, you can also point to `git clone && go install .`

chizhg:
> This is cool. I think we still should support local development mode though, and understand that `go get` doesn't work. I would still expect a warning message like `kntest not installed, please install via ./install_kntest.sh` script etc.
> 
> Note: the `./install_kntest.sh` is just an idea, you can also point to `git clone && go install .`

Added the `not installed` error as suggested when running kntest commands.

chizhg:
It's fine, it will only impact test-infra repo without the scripts being synced to other repos. I will roll out the image to production as soon as the postsubmit job finishes building it.

/hold cancel

Issue #2177, opened 2020-06-05
Auto add/delete release branch jobs


chaodaiG:
Branch specific jobs, including continuous, release jobs are set up for current release and previous release branches, currently they are manually added and deleted. 

Since all these are defined in https://github.com/knative/test-infra/blob/master/config/prod/prow/config_knative.yaml, we can automate the process by monitoring the branches changes in each repo and send out PR updating them

PR #2188, opened 2020-06-10
Update release jobs

chaodaiG:
This is part 1 of #2177, this PR adds a function to config generator, to let it modify `config_knative.yaml`, aka config generator input. The modification is based on active release branches, which creates/deletes release branch jobs

Note for reviewers: all changes in `.yaml` files were generated, please focus on the `.go` file change to save some time

/assign chizhg albertomilan

### DIFF
--- a/tools/config-generator/udpaterelease.go
+++ b/tools/config-generator/udpaterelease.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package main
 
 import (
@@ -14,7 +30,9 @@ import (
 
 const (
 	maxReleaseBranches = 4
-	latest             = "0.15"
+	// Assuming all repo with 0.14 branch also has 0.15 branch, this will be
+	// derived from the branch in following PR
+	latest = "0.15"
 )
 
 func upgradeReleaseBranchesTemplate(name string) error {
@@ -76,16 +94,16 @@ func getReposMap(val interface{}, latest string) interface{} {
 		if !skipCiUpdate && len(ciBranches) > 0 {
 			repoConfigs = updateConfigForJob(repoConfigs, ciBranches, latest,
 				func(jobConfig yaml.MapSlice) string {
-					ciBranch, _ := getBranch(jobConfig)
-					return ciBranch
+					branch, _ := getBranch(jobConfig)
+					return branch
 				})
 		}
 
 		if !skipReleaseUpdate && len(releaseBranches) > 0 {
 			repoConfigs = updateConfigForJob(repoConfigs, releaseBranches, latest,
 				func(jobConfig yaml.MapSlice) string {
-					ciBranch, _ := getBranch(jobConfig)
-					return ciBranch
+					_, branch := getBranch(jobConfig)
+					return branch
 				})
 		}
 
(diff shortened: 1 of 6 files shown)


tools/config-generator/udpaterelease.go
```diff
@@ -0,0 +1,220 @@
...
+func sortFunc(strSlice []string) {
+	sort.Slice(strSlice, func(i, j int) bool {
+		return versionComp(strSlice[i], strSlice[j]) > 0
+	})
+}
+
+func versionComp(v1, v2 string) int {
+	leftMajor, leftMinor := majorMinor(v1)
+	rightMajor, rightMinor := majorMinor(v2)
+	if leftMajor == rightMajor && leftMinor == rightMinor {
```
albertomilan, line 190: Being sort.Slice based on a boolean function execution, I'd simplify this to:
```golang
if leftMajor != rightMajor {
    return leftMajor - rightMajor
}
if leftMinor != rightMinor {
    return leftMinor - rightMinor
}
return 0
```

tools/config-generator/udpaterelease.go
```diff
@@ -0,0 +1,220 @@
...
+)
+
+const (
+	maxReleaseBranches = 4
+	// Assuming all repo with 0.14 branch also has 0.15 branch, this will be
+	// derived from the branch in following PR
+	latest = "0.15"
+)
+
+func upgradeReleaseBranchesTemplate(name string) error {
```
albertomilan, line 38: Nit: I'd be more specific with the `name` argument name. `configFileName` might be more concrete on the content and purpose

tools/config-generator/udpaterelease.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+	content, err := ioutil.ReadFile(name)
+	if err != nil {
+		return fmt.Errorf("Cannot read file %q: %w", name, err)
+	}
+	if err = yaml.Unmarshal(content, &config); err != nil {
+		return fmt.Errorf("Cannot parse config %q: %w", name, err)
+	}
+	for i, repos := range config {
+		if repos.Key == "presubmits" {
```
albertomilan, line 52: Is this a common Go pattern? Why not:
```golang
if repos.Key != "presubmits" {
    config[i].Value = getReposMap(repos.Value, latest)
}
```

tools/config-generator/udpaterelease.go
```diff
@@ -0,0 +1,220 @@
...
+
+		config[i].Value = getReposMap(repos.Value, latest)
+	}
+
+	updated, err := yaml.Marshal(&config)
+	// This shouldn't happen, just catch it in case
+	if err != nil {
+		return fmt.Errorf("failed marshal modified content: %w", err)
+	}
+	return ioutil.WriteFile(name, updated, info.Mode())
```
albertomilan, line 64: This might be correct in the short-term and the WriteFile contract might not be subject to change, but we're effectively binding our contract to it.
Consider decoupling the invocation to the return value.

tools/config-generator/udpaterelease.go
```diff
@@ -0,0 +1,220 @@
...
+}
+
+func getReposMap(val interface{}, latest string) interface{} {
```
albertomilan, line 65: Why are you passing latest (a global visible constant) as an argument? Is this a common convention?

chaodaiG: This is not, as I commented above, this logic will very soon be replaced with real branch names

tools/config-generator/udpaterelease.go
chaodaiG: Sorry but not sure what the recommendation is, did you mean make this function return string instead of make this function in place update the file?

tools/config-generator/udpaterelease.go
albertomilan: Thanks for resolving this

tools/config-generator/udpaterelease.go
albertomilan: The recommendation is to decouple the return value from WriteFile from our function return value. Right now, both contracts are bound together and that might cause issues on the future.
I'm suggesting to reconsider that and make sure we want to *always* return whatever `WriteFile` function returns or invoke it independently and `return nil` afterwards.

chaodaiG: I think that's expected to let it return from WriteFile, and I think it's pretty common pattern in go

PR #2189, opened 2020-06-11
Add fetch github branches

chaodaiG:
Fetch github branches for generating jobs, depends on https://github.com/knative/pkg/pull/1400

/assign chizhg albertomilan

### DIFF
--- a/tools/config-generator/getlivebranch.go
+++ b/tools/config-generator/getlivebranch.go
@@ -39,17 +39,17 @@ func latestReleaseBranch(gc ghutil.GithubOperations, repo string) (string, error
 
 // filterLatest returns latest release branch in the form of
 // [MAJOR].[MINOR], if there is no valid release branch exist in the form of
-// `release-[MAJOR]-[MINOR]`, then it returns "0.0"
+// `release-[MAJOR]-[MINOR]`, then it returns ""
 func filterLatest(branches []*github.Branch) string {
 	var (
 		reReleaseBranch = regexp.MustCompile(`^release\-(\d+\.\d+)$`)
-		latest          = "0.0"
+		latest          = ""
 	)
 
 	for _, branch := range branches {
 		if matches := reReleaseBranch.FindStringSubmatch(*branch.Name); len(matches) > 1 {
 			release := matches[1]
-			if versionComp(release, latest) > 0 {
+			if latest == "" || versionComp(release, latest) > 0 {
 				latest = release
 			}
 		}
--- a/tools/config-generator/getlivebranch.go
+++ b/tools/config-generator/getlivebranch.go
@@ -34,10 +34,13 @@ func latestReleaseBranch(gc *ghutil.GithubClient, repo string) (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("failed listing branches for repo %q: %w", repo, err)
 	}
-	return latestReleaseBranches(branches), nil
+	return filterLatest(branches), nil
 }
 
-func latestReleaseBranches(branches []*github.Branch) string {
+// filterLatest returns latest release branch in the form of
+// [MAJOR].[MINOR], if there is no valid release branch exist in the form of
+// `release-[MAJOR]-[MINOR]`, then it returns "0.0"
+func filterLatest(branches []*github.Branch) string {
 	var (
 		reReleaseBranch = regexp.MustCompile(`^release\-(\d+\.\d+)$`)
 		latest          = "0.0"
--- a/tools/config-generator/getlivebranch.go
+++ b/tools/config-generator/getlivebranch.go
@@ -1,0 +1,56 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/google/go-github/v27/github"
+	"knative.dev/pkg/test/ghutil"
+)
+
+func latestReleaseBranch(gc *ghutil.GithubClient, repo string) (string, error) {
+	parts := strings.Split(repo, "/")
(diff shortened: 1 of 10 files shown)


tools/config-generator/main.go
```diff
@@ -945,6 +946,7 @@ func main() {
 	flag.StringVar(&jobNameFilter, "job-filter", "", "Generate only this job, instead of all jobs")
 	flag.StringVar(&preCommand, "pre-command", "", "Executable for running instead of the real command of a job")
 	flag.BoolVar(&upgradeReleaseBranches, "upgrade-release-branches", false, "Update release branches jobs based on active branches")
+	flag.StringVar(&githubTokenPath, "github-token-path", "", "Token path for authenticating with github, used only when --upgrade-release-branches is on")
```
albertomilan, line 949: I tend to prefer interfaces and arguments which do not allow inconsistencies. A user could request `upgrade-release-branches` and not provide the token (or viceversa). Does it make sense to have a single program argument which encapsulates both options? E.g: `upgrade-release-branches-with-github-token` string

tools/config-generator/getlivebranch.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		return "", fmt.Errorf("failed listing branches for repo %q: %w", repo, err)
+	}
+	return latestReleaseBranches(branches), nil
+}
+
+func latestReleaseBranches(branches []*github.Branch) string {
+	var (
+		reReleaseBranch = regexp.MustCompile(`^release\-(\d+\.\d+)$`)
+		latest          = "0.0"
```
albertomilan, line 43: Is "0.0" a valid return value for this function or should we capture a failure?

chaodaiG: `release-0.0` is not a vaild release, and shouldn't appear at all. So `0.0` just means there is no release branch, I think I can add a comment to make it clearer

chaodaiG: Updated

tools/config-generator/main.go
chaodaiG: To be honest, it makes sense to me combining these two params, I'm just not very comfortable with such a long param name

albertomilan: We are fully aligned on that.
I'm open to both options, but lean a little bit more towards the single-parameter option.

Feel free to choose the one you prefer.

tools/config-generator/getlivebranch.go
albertomilan: I'm more concerned about a scenario where a `release-*` branch is not localized and "0.0" is returned. This will hide effectively the error when it seems this would be the best place to manage it.

chaodaiG: I agree that this is the best place finding errors. But it's not super clear to me what scenario should trigger an error. i.e. when there is a branch named `release-something` it might be intended, and shouldn't fail this tool

tools/config-generator/updaterelease_test.go
```diff
@@ -85,6 +88,11 @@ func TestUpgradeReleaseBranchesTemplate(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			fgc := fakeghutil.NewFakeGithubClient()
+			fgc.Branches = make(map[string][]*github.Branch)
+			fgc.Branches["org1/repo1"] = []*github.Branch{
+				&github.Branch{Name: &latest},
```
mattmoor, line 94: Format Go code:
```suggestion
				{Name: &latest},
```

tools/config-generator/updaterelease_test.go
```diff
@@ -246,11 +251,19 @@ func TestGetReposMap(t *testing.T) {
...
```
mattmoor, line 257: Format Go code:
```suggestion
				{Name: &latest},
```

tools/config-generator/udpaterelease.go
```diff
@@ -71,6 +72,13 @@ func getReposMap(val interface{}, latest string) interface{} {
 			skipCiUpdate      bool
 			skipReleaseUpdate bool
 		)
+		repoName := getString(repo.Key)
+		latest, err := latestReleaseBranch(gc, repoName)
+		if err != nil {
+			return nil, fmt.Errorf("failed getting latest release branches: %w", err)
+		}
+
+		log.Printf("latest branch for repo %q is %q", repoName, latest)
```
chizhg, line 81: Nit: capitalize the first word in the logging message

tools/config-generator/main.go
```diff
@@ -959,7 +962,11 @@ func main() {
 	// Read input config.
 	name := flag.Arg(0)
 	if upgradeReleaseBranches {
-		if err := upgradeReleaseBranchesTemplate(name); err != nil {
+		gc, err := ghutil.NewGithubClient(githubTokenPath)
+		if err != nil {
+			log.Fatalf("failed creating github client from %q: %v", githubTokenPath, err)
```
chizhg, line 967: Ditto

tools/config-generator/getlivebranch.go
chizhg: What will happen if `latestReleaseBranch` returns `release-0.0` as the latest release branch? If the callsite will consider it as an error and skip it, I think it makes more sense to return an error here rather than returning a relatively weird non-existent branch name.

tools/config-generator/getlivebranch.go
```diff
@@ -0,0 +1,59 @@
...
+func filterLatest(branches []*github.Branch) string {
+	var (
+		reReleaseBranch = regexp.MustCompile(`^release\-(\d+\.\d+)$`)
+		latest          = "0.0"
+	)
+
+	for _, branch := range branches {
+		if matches := reReleaseBranch.FindStringSubmatch(*branch.Name); len(matches) > 1 {
+			release := matches[1]
+			if versionComp(release, latest) > 0 {
```
chizhg, line 52: Does it make more sense to use an existing library to do semver version comparison (e.g. https://github.com/coreos/go-semver) rather than making our own?

chaodaiG: semver enforces `x.y.x` style, which doesn't work for us. I have looked at semver under golang, which didn't work either

tools/config-generator/main.go
chaodaiG: done

tools/config-generator/udpaterelease.go
chaodaiG: done

tools/config-generator/getlivebranch.go
chaodaiG: Makes sense. Changed to make the function return "" instead of "0.0", and add a condition at caller for this check

PR #2194, opened 2020-06-12
Use the latest kubectl in prow-tests image

### LINKED ISSUE #1858: Test-infra integration tests are failing after new prow-tests image
See here:
https://testgrid.knative.dev/test-infra#continuous

Integration tests failed to install serving in test-infra, but interestingly not in other repos that also consumes serving nightly builds, an example: https://testgrid.knative.dev/docs

The main difference is test-infra is using cluster management lib while doc is not

chizhg:
We have changed the logic to install serving after https://github.com/knative/serving/issues/7755 is fixed, which works for the latest kubectl.

Remove the logic to pin the kubectl version in Dockerfile to use the `kubectl` installed with `gcloud`.

Fixes #1858 

/cc @chaodaiG

### DIFF
--- a/images/prow-tests/Dockerfile
+++ b/images/prow-tests/Dockerfile
@@ -29,13 +29,6 @@ RUN gcloud components update
 RUN gcloud components install docker-credential-gcr
 RUN docker-credential-gcr configure-docker
 
-# Replace kubectl with a working version
-# TODO(chizhg): remove once https://github.com/knative/test-infra/issues/1858 is fixed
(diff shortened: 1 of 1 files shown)


chaodaiG:
/lgtm

/hold
Would be nice back port the change to old test-infra release branches in case they are needed

chizhg:
Nice catch! 
There are a lot of old branches that need this, and updating test-infra in the release branches will trigger unnecessary dot releases. Probably not worth doing it for now. When we have `kntest`, we need to keep its backward compatibility in such case.

PR #2202, opened 2020-06-15
Add testgrid tabs for all in house tools

chaodaiG:
Add testgrid tabs, as well as alerts for in house tools

/assign chizhg albertomilan


tools/config-generator/customjobs.go
```diff
@@ -0,0 +1,48 @@
...
+			"ci-knative-cleanup-performance-tests",
+			"ci-knative-flakes-reporter",
+			"ci-knative-flakes-resultsrecorder",
+			"ci-knative-prow-auto-bumper",
+			"post-knative-prow-cluster-config-updater",
+			"post-knative-test-infra-image-push",
+		}
+		extras = map[string]string{
+			"num_failures_to_alert": "3",
+			"alert_options":         "\n      alert_mail_to_addresses: \"prime-engprod-sea@google.com\"",
```
chizhg, line 36: Can you change the email address to `serverless-engprod-sea@google.com`, plus all the existing ones?

tools/config-generator/customjobs.go
```diff
@@ -0,0 +1,48 @@
...
+
+// Although custom jobs are not generated by this generator, certain testgrid
+// configs are needed for certain custom jobs
+
+package main
+
+func addCustomJobsTestgrid() {
+	var (
+		jobnames = []string{
+			"ci-knative-backup-artifacts",
```
chizhg, line 25: I'm wondering if we can add a unit test to make sure these Prow jobs exist.
We don't necessarily need to do it in this PR, but adding it will definitely make this part less brittle.

tools/config-generator/customjobs.go
chaodaiG: done

tools/config-generator/customjobs.go
chaodaiG: Added one

chizhg:
This looks cool, thanks!

/lgtm

PR #2208, opened 2020-06-16
Fixing testgrid config, by the way add verify testgrid config make ta…

chaodaiG:
Testgrid doesn't allow dashboard group and dashboard using the same name, fixing it.

By the way add a verify testgrid config make target

/assign chizhg

### DIFF
--- a/config/prow_common.mk
+++ b/config/prow_common.mk
@@ -121,3 +121,25 @@ update-testgrid-config: confirm-master
 		"--output=gs://$(TESTGRID_GCS)/config" \
 		"--yaml=$(realpath $(TESTGRID_CONFIG))"
 
+.PHONY: verify-testgrid-config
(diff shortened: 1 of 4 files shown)


config/prow_common.mk
```diff
@@ -121,3 +121,25 @@ update-testgrid-config: confirm-master
 		"--output=gs://$(TESTGRID_GCS)/config" \
 		"--yaml=$(realpath $(TESTGRID_CONFIG))"
 
+.PHONY: verify-testgrid-config
```
chizhg, line 124: I was considering adding this check as a presubmit test job, similar as how we added the `checkconfig` job.
Or do you prefer adding it as a make target?

chaodaiG: Ah. I see, how do you run 2 commands in a presubmit?

PR #2213, opened 2020-06-17
Increase frequency and concurrency count for the cleanup job

chizhg:
1. Increase frequency to `once per day` and concurrency count to `50` for the cleanup job. I'm running one job in https://prow.knative.dev/view/gcs/knative-prow/logs/ci-knative-cleanup/1273298252417994752 and don't see any quota issues.
2. Use `runner.sh` as the test driver instead of `go run` to automatically set Go version for the jobs.

/cc @chaodaiG @peterfeifanchen

### DIFF
--- a/config/prod/prow/jobs/custom-jobs.yaml
+++ b/config/prod/prow/jobs/custom-jobs.yaml
@@ -100,7 +100,7 @@ periodics:
     - name: backup-account
       secret:
         secretName: backup-account
-- cron: "0 19 * * 1,3,5"
+- cron: "0 19 * * *"
   name: ci-knative-cleanup
   labels:
     prow.k8s.io/pubsub.project: knative-tests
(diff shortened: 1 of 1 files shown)


chaodaiG:
/hold
Can you check the cpu and memory usage to make sure it's not starving other prow jobs?

chizhg:
> /hold
> Can you check the cpu and memory usage to make sure it's not starving other prow jobs?

We have resource limits set on the `test-pods` namespace so it shouldn't matter?
https://github.com/knative/test-infra/blob/master/config/prod/build-cluster/cluster/200-cpu-limitrange.yaml
https://github.com/knative/test-infra/blob/master/config/prod/build-cluster/cluster/200-memory-limitrange.yaml

chaodaiG:
You're right, totally forgot we have limits

/unhold
/lgtm

PR #2060, opened 2020-05-13
Add peribolos job for updating knative-sandbox

chizhg:
**What this PR does, why we need it**:
In preparation for https://github.com/knative/community/pull/135. The jobs here should be added after that PR is merged.
@evankanderson please let me know if you want to enable more `fix`es.

/cc @evankanderson @coryrc @chaodaiG

PR #2220, opened 2020-06-18
Fix job config path for updateconfig plugin

chaodaiG:
Fixes regression caused by #2060 

/assign chizhg


chaodaiG:
/hold
let me test the blob matching to make sure

chizhg:
I was about to ask you why it was not triggered on my PR..
Thanks for the fix!

/lgtm

chaodaiG:
I think it works
/unhold

PR #2223, opened 2020-06-18
Exclude long-job-names warning for checkconfig and clean up code

chizhg:
Exclude long-job-names warning for checkconfig

Kubernetes test-infra does not have this check - https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml#L89-L95

Also clean up the code a bit as a few functions are not being used anymore.

/cc @chaodaiG @evankanderson

### DIFF
--- a/tools/config-generator/main.go
+++ b/tools/config-generator/main.go
@@ -32,13 +32,11 @@ import (
 	"regexp"
 	"runtime"
 	"sort"
-	"strconv"
 	"strings"
 	"text/template"
 	"time"
 
 	"gopkg.in/yaml.v2"
-	"k8s.io/apimachinery/pkg/util/sets"
 	"knative.dev/pkg/test/ghutil"
 )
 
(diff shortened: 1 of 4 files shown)


chaodaiG:
I'm not sure this is the best way of fixing the problem, as blindly pruning can cause other problems, i.e. two jobs sharing same prefix would collide

chaodaiG:
/hold

How about shorten the job name from the very beginning?

chizhg:
> /hold
> 
> How about shorten the job name from the very beginning?

Actually I checked kubernetes test-infra, they don't have this check - https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml#L89-L95. So I think we can simply disable the warning.

I'm keeping other changes in this PR as they are mostly code cleanups.

evankanderson:
Why are all the `tools/` files in this PR?

tools/config-generator/main.go
```diff
@@ -682,37 +495,6 @@ func outputConfig(line string) {
...
-func strExists(arr []string, str string) bool {
-	for _, s := range arr {
-		if str == s {
-			return true
-		}
-	}
-	return false
-}
-
-type specialBranchLogic struct {
```
evankanderson, line 695: It looks like this went away entirely?

chizhg: Yeah, these functions are not used anymore.

chizhg:
> Why are all the `tools/` files in this PR?

They are mostly cleanups and refactor I did in my last commit. I switched to a different approach (simply excluding the warning in the Prow job setting) to fix the issue, but I think these changes are still useful so I kept as is.

evankanderson:
Can you update the PR title and/or first description to indicate that there's also cleanup in this PR?

/lgtm
/hold

Cancel the hold when the title/description is updated.

chizhg:
> Can you update the PR title and/or first description to indicate that there's also cleanup in this PR?
> 
> /lgtm
> /hold
> 
> Cancel the hold when the title/description is updated.

Thanks! Updated the title and description. 
/unhold

PR #2230, opened 2020-06-23
Add periodic peribolos job

chizhg:
Add a periodic peribolos job for `knative-sandbox` that will be run at 10AM PST. everyday.

/cc @evankanderson

### DIFF
--- a/config/prod/prow/jobs/custom/peribolos.yaml
+++ b/config/prod/prow/jobs/custom/peribolos.yaml
@@ -52,3 +52,41 @@ postsubmits:
       - name: oauth
         secret:
           secretName: oauth-token
+
+periodics:
+# Run at 10AM PST.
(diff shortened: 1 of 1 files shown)


config/prod/prow/jobs/custom/peribolos.yaml
```diff
@@ -52,3 +52,41 @@ postsubmits:
       - name: oauth
         secret:
           secretName: oauth-token
+
+periodics:
+# Run at 10AM PST.
+- cron: "0 17 * * *"
+  name: post-knative-sandbox-peribolos
```
evankanderson, line 59: Should this be named "post-"?

chizhg: It should be named as `ci-`. Fixed.

PR #2233, opened 2020-06-23
Add tooling to generate protobufs to test Dockerfile.

markusthoemmes:

We're in the process of generating protobuf APIs in https://github.com/knative/serving/pull/8396 for example, thus need a bit more tooling available to run update-codegen.


chaodaiG:
/lgtm
/approve

Note that the image pushing job will push a `beta` image first, we'll manually update the label once it looks good

Issue #2241, opened 2020-06-24
We need a Prow job to automatically bump up images versions in the Prow job config files


chizhg:
At this moment we have moved all the cluster and core config files for Knative Prow to [oss-test-infra](https://github.com/GoogleCloudPlatform/oss-test-infra) repo, but still having all the Prow job config files in knative test-infra.

A Prow job in oss-test-infra will automatically bump up the image versions for cluster and core config files, but not the Prow job config files. We also need a Prow job to bump up the image versions for the Prow job config files that live in knative test-infra. Ideally we should use https://github.com/kubernetes/test-infra/tree/master/prow/cmd/autobump, but will need some tweaks to allow it only handle Prow job config files.

/assign @chaodaiG

chizhg:
This Prow job can possibly live in the oss-test-infra repo, so we have one single place to maintain all the autobump Prow jobs.

PR #2247, opened 2020-06-26
Bump prow versions using k8s auto bumper

chaodaiG:
Fixes: https://github.com/knative/test-infra/issues/2241

/cc chizhg


chaodaiG:
It'll not run over the weekend, so it's safe to merge it

PR #2248, opened 2020-06-28
Remove dep on knative pkg

### LINKED ISSUE #2249: Remove the dependencies on knative.dev/pkg
Currently `knative/test-infra` depends on `knative/pkg`, and `knative/pkg` also depends on `knative/test-infra`, this results in circular dependency and causes a few problems, see more details in https://github.com/knative/test-infra/issues/2069.

We tried to use Go multiple modules to solve this issue, but it didn't work, see the comments in https://github.com/knative/test-infra/issues/2078.

Another proposed solution is having a shell script (https://github.com/knative/test-infra/blob/master/scripts/update-test-infra.sh) to only update the `scripts/` directory with git commands, but it does not solve the problem from the root, and introduces another dependency management mechanism which diverges from Go modules.

The long term solution for these issues will be getting rid of the circular dependencies between `knative/pkg` and `knative/test-infra` - `knative/pkg` can depend on `knative/test-infra`, but not vice versa, and every other knative repo can depend on both `pkg` and `test-infra`. In this way, we also separate the responsibilities for the two repos - `knative/test-infra` will only host infra related tools and libraries, and `knative/pkg/test` package will host reusable libraries that other projects can use to implement actual test cases.

/assign

chizhg:
As previously discussed, move all the dependent packages in `pkg` repo to `test-infra`, to get rid of the dependency on `knative.dev/pkg`.

Fixes #2249


scripts/library.sh
```diff
@@ -528,10 +528,19 @@ function run_go_tool() {
 # Run kntest tool, error out and ask users to install it if it's not currently installed.
 # Parameters: $1..$n - parameters passed to the tool.
 function run_kntest() {
-  if [[ -z "$(which kntest)" ]]; then
```
chaodaiG, line 531: Can we do this in a separate PR? As well as changes in other shell scripts

chizhg: Sure, I have reverted the script changes in this PR, and will create another PR for it.

chizhg: The e2e tests cannot pass without the changes in `test/e2e-tests.sh`, so I'm keeping that file and will create another PR to only change `scripts/library.sh`

chaodaiG:
Please fix tests

chizhg:
> Please fix tests

The changes in `library.sh` is also needed.. The error here is due to the difference between `$@` and `"$@"`.

chaodaiG:
/lgtm

Interesting, you deleted 433K lines of code by adding something 👍

PR #2250, opened 2020-06-29
Fix autobump

chaodaiG:
Use build cluster as autobump doesn't really need so much permission

Also add a plank image place holder as autobump expects it

/cc chizhg

### DIFF
--- a/config/prod/prow/jobs/custom/test-infra.yaml
+++ b/config/prod/prow/jobs/custom/test-infra.yaml
@@ -270,7 +270,7 @@ periodics:
         secretName: flaky-test-reporter-slack-token
 - cron: "15 15-23 * * 1-5"  # Run at 7:15-15:15 PST (15:15 UTC) Mon-Fri
   name: ci-knative-prow-auto-bumper
-  cluster: prow-trusted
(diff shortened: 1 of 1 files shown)


config/prod/prow/jobs/custom/test-infra.yaml
```diff
@@ -12,6 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Add a fake plank image here so that autobump.sh can bump only job configs
```
chizhg, line 15: It looks this comment is not at the right place. Also it looks you forgot to add the "fake plank image" in the Prow job config?

chaodaiG: Based on the regex, as long as it's in this file it should be fine: https://github.com/knative/test-infra/blob/4440f3f3adad78518dc68411069b0e495af538f3/config/prod/prow/jobs/custom/test-infra.yaml#L302

chizhg: I see, this is hacky...

PR #2253, opened 2020-06-29
Do not generate issue tracker Prow jobs

### LINKED ISSUE #2252: Do not generate issue tracker Prow jobs
Currently the issue tracker Prow jobs are auto-generated by the config generator tool, see https://github.com/knative/test-infra/blob/master/tools/config-generator/issue_tracker_config.go, which makes it hard to bump up the image versions used in the Prow job, https://github.com/knative/test-infra/pull/2251 failed because of this.

Given the fact we have all the other maintenance jobs defined in customized yaml files, we should also move the issue tracker Prow jobs there and stop generating them.

/assign

chizhg:
Do not generate issue tracker Prow jobs to get https://github.com/knative/test-infra/pull/2251 pass.

Fixes #2252 

/cc @chaodaiG


chaodaiG:
I feel that this job will eventually be generated. But since job generator isn't currently robust enough addressing the problem

/lgtm

PR #2254, opened 2020-06-29
Add stdout option to exec.Command

peterfeifanchen:
This redirects the output of the command running to stdout, bypassing
the buffer it is returned to (commands requiring output of a command
should not set this)

/cc @chizhg

### DIFF
--- a/pkg/cmd/command.go
+++ b/pkg/cmd/command.go
@@ -69,6 +69,7 @@ func WithStdout() Option {
 
 // RunCommand will run the command and return the standard output, plus error if there is one.
 func runCommand(cmdLine string, options ...Option) (string, error) {
+	var err error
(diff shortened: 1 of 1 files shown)


pkg/cmd/command.go
```diff
@@ -57,6 +58,15 @@ func WithDir(dir string) Option {
 	}
 }
 
+// WithStdout returns an option that sets the cmd output to stdout.
+// This redirects output to stdout, should not use this option if you
+// need the return value of the command.
+func WithStdout() Option {
```
chizhg, line 64: The `runCommand` function will always return an error if this option is used. See https://github.com/golang/go/blob/master/src/os/exec/exec.go#L535-L537.

PR #2259, opened 2020-06-30
Stop pushing prow-test image from prow so that other images can be pushed

chaodaiG:
/cc @chizhg

### DIFF
--- a/images/push-all.sh
+++ b/images/push-all.sh
@@ -16,14 +16,17 @@
 
 # This script pushes all images under current directory
 
+# prow-tests image crashes prow job, don't build it here
+EXCLUDED="prow-tests"
+
 set -e
 
 CUR_DIR="$(realpath $(dirname $0))"
 
 for dir in ${CUR_DIR}/*; do
-    if [[ -d "${dir}" && -f "${dir}/Makefile" ]]; then
+    if [[ -d "${dir}" && -f "${dir}/Makefile" && "${dir}" != "${CUR_DIR}/${EXCLUDED}" ]]; then
         echo "RUNNING: make -C '${dir}' push"
-        make -C "${dir}" push
+        # make -C "${dir}" push
     else
         echo "Skipping '${dir}'"
     fi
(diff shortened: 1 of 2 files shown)


chizhg:
@googlebot

images/push-all.sh
```diff
@@ -16,14 +16,17 @@
...
 set -e
 
 CUR_DIR="$(realpath $(dirname $0))"
 
 for dir in ${CUR_DIR}/*; do
-    if [[ -d "${dir}" && -f "${dir}/Makefile" ]]; then
+    if [[ -d "${dir}" && -f "${dir}/Makefile" && "${dir}" != "${CUR_DIR}/${EXCLUDED}" ]]; then
         echo "RUNNING: make -C '${dir}' push"
-        make -C "${dir}" push
+        # make -C "${dir}" push
```
chizhg, line 29: Uncomment?

chizhg: Also removing `--no-cache` seems to be in your another branch.

chaodaiG: This is embarrassing, will quickly update it

chaodaiG: Done

images/simple-image.mk
```diff
@@ -37,7 +37,7 @@ IMG = $(REGISTRY)/$(PROJECT)/test-infra/$(IMAGE_NAME)
 TAG := $(shell date +v%Y%m%d)-$(shell git describe --always --dirty --match '^$$')
 
 build:
-	docker build --no-cache --pull -t $(IMG):$(TAG) -f $(DOCKERFILE) $(DOCKERBUILDARGS) ../..
```
chizhg, line 40: IIRC, `--no-cache` was added as we want to always update `gcloud` when we build the `prow-tests` image. Can we make an exception to keep this option when building the `prow-tests` image?

chaodaiG: I see. --no-cache doesn't hurt images other than `prow-tests` as far as I can tell. Let me add it back

chizhg:
@googlebot

PR #2266, opened 2020-07-05
More moving from knative.dev/pkg

chizhg:
Also move `perf-tests` tool and `interactive` package from knative/pkg to knative/test-infra, plus some small refactor and cleanups.


pkg/clustermanager/e2e-tests/get.go
```diff
@@ -14,16 +14,16 @@ See the License for the specific language governing permissions and
...
 )
 
 // Get gets a GKE cluster
-func (rw *RequestWrapper) Get() (*clm.GKECluster, error) {
+func Get(rw *RequestWrapper) (*clm.GKECluster, error) {
 	rw.Request.SkipCreation = true
 	// Reuse `Create` for getting operation, so that we can reuse the same logic
 	// such as protected project/cluster etc.
-	return rw.Create()
+	return Create(rw)
```
albertomilan, line 28: Nit: I'm not sure why you're moving from a method to a function-based approach. They are, in essence, identical but this might introduce aliasing issues in the future with the function names. Could you provide some insight on your decision?

pkg/gcs/gcs.go
```diff
@@ -31,21 +31,21 @@ import (
 	"google.golang.org/api/option"
 )
 
-type Client struct {
+type storageClient struct {
 	*storage.Client
 }
 
 // NewClient creates new GCS client with given service account
-func NewClient(ctx context.Context, serviceAccount string) (*Client, error) {
```
albertomilan, line 39: Could you confirm the returned value is properly changed to be a value and not a reference?

albertomilan:
Changes look good overall. I will release the hold after a couple of minor comments are answered.

/hold
/lgtm

pkg/gcs/gcs.go
chizhg: `Client` is now an interface, and `*storageClient` has implemented the interface, so here it's returning a `*storageClient` as the `Client` type.

pkg/clustermanager/e2e-tests/get.go
chizhg: The reason I'm making this change is one of our other repos are importing these functions and having some unit tests around them, having them as methods won't be unit-testable.
And different packages can have functions with the same name, so function names shouldn't be a problem.

PR #2279, opened 2020-07-08
Add sugar-controller extension setup

navidshaikh:
 - Add setup utils for nightly and release version of eventing sugar controller
   extension
 - The wait for pod running check is deferred to specific extension function so
   they could watch in respective namespace

- Provide `start_release_eventing_sugar_controller` and `start_latest_eventing_sugar_controller` utils for installing sugar controller extension.

### DIFF
--- a/scripts/library.sh
+++ b/scripts/library.sh
@@ -504,20 +504,18 @@ function start_knative_eventing_extension() {
   header "Starting Knative Eventing Extension"
   echo "Installing Extension CRDs from $1"
   kubectl apply -f "$1"
-  # do the wait for running pods in specific extension function as the namespace could differ
+  wait_until_pods_running "$2" || return 1
 }
 
 # Install the stable release of eventing extension sugar controller in the current cluster.
 # Parameters: $1 - Knative Eventing release version, e.g. 0.16.0
 function start_release_eventing_sugar_controller() {
-    start_knative_eventing_extension "https://storage.googleapis.com/knative-releases/eventing/previous/v$1/eventing-sugar-controller.yaml"
-  wait_until_pods_running knative-eventing || return 1
+  start_knative_eventing_extension "https://storage.googleapis.com/knative-releases/eventing/previous/v$1/eventing-sugar-controller.yaml" "knative-eventing"
 }
 
 # Install the sugar cotroller eventing extension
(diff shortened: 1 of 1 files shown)


scripts/library.sh
```diff
@@ -498,6 +498,28 @@ function start_latest_knative_eventing() {
   start_knative_eventing "${KNATIVE_EVENTING_RELEASE}"
 }
 
+# Install Knative Eventing extension in the current cluster.
+# Parameters: $1 - Knative Eventing extension manifest.
+function start_knative_eventing_extension() {
+  header "Starting Knative Eventing Extension"
+  echo "Installing Extension CRDs from $1"
+  kubectl apply -f "$1"
+  # do the wait for running pods in specific extension function as the namespace could differ
```
chizhg, line 507: This is not the pattern used by the `start_knative_eventing` function. Are there scenarios that require eventing to be installed in a different namespace? If yes we may need to change them all.

navidshaikh: This is for the eventing extensions, the resources of some extensions are installed in other namespace, for example kafka-source (https://storage.googleapis.com/knative-nightly/eventing-contrib/latest/kafka-source.yaml), installs in `knative-sources` namespace. So if we'd add kafka setup util here later, it could check for resources in respective namespace. We could revert this and make `start_knative_eventing_extension` to look for running pods in $2. WDYT?

chizhg: Yeah I think that is better, you can pass `knative-eventing` in the function call.

scripts/library.sh
```diff
@@ -498,6 +498,28 @@ function start_latest_knative_eventing() {
...
+}
+
+# Install the stable release of eventing extension sugar controller in the current cluster.
+# Parameters: $1 - Knative Eventing release version, e.g. 0.16.0
+function start_release_eventing_sugar_controller() {
+    start_knative_eventing_extension "https://storage.googleapis.com/knative-releases/eventing/previous/v$1/eventing-sugar-controller.yaml"
```
chizhg, line 513: The indentation for this line is inconsistent with others.

scripts/library.sh
navidshaikh: fixed. PTAL

PR #2291, opened 2020-07-13
Improvements to the peribolos and label_sync Prow jobs

chizhg:
1. Run presubmit Prow jobs for peribolos and label_sync with `--confirm=false`, to prevent the configuration files to be broken like https://github.com/knative/community/pull/181
2. Move some labels in the `label_sync` config file to be common triaging labels

/cc @chaodaiG @albertomilan

### DIFF
--- a/label_sync/labels.yaml
+++ b/label_sync/labels.yaml
@@ -22,19 +22,56 @@
 ---
 default:
   labels:
-    # Signals to other contributors
+    # Common labels for triaging
+    - color: c5def5
+      description: Parked issue that required triaging/revisit in a near future
+      name: kind/TBD
+      target: issues
+      addedBy: humans
     - color: 7057ff
       description: Denotes an issue ready for a new contributor, according to the "help wanted" guidelines.
-      name: 'good first issue'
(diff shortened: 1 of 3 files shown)


label_sync/labels.yaml
```diff
@@ -22,19 +22,56 @@
 ---
 default:
   labels:
-    # Signals to other contributors
+    # Common labels for triaging
+    - color: c5def5
```
albertomilan, line 26: Just curiosity, how are the colors being chosen?

chizhg: Most of the colors are chosen randomly :) while there is a general principle that all "positive" labels should be green but "negative" labels be red, like `lgtm` and `approve` are green or light green, but `hold` is red. And all "neutral" labels are random, ideally neither green nor red.

albertomilan: Thanks for the explanation. Color codes tend to have cultural interpretations and therefore my interest.

Given the impossibility of having different concurrent codes right now and the importance of making it consistent, it sounds great from my side.

PR #2292, opened 2020-07-13
Productivity oncalls owns custom jobs configs

chaodaiG:
So that oncall can approve auto bump PRs like https://github.com/knative/test-infra/pull/2287

/cc chizhg albertomilan

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -11,4 +11,3 @@ aliases:
   - steuhs
   - peterfeifanchen
   - efiturri
-  - nbarthwal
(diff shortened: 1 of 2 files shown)


OWNERS_ALIASES
```diff
@@ -11,4 +11,3 @@ aliases:
   - steuhs
   - peterfeifanchen
   - efiturri
-  - nbarthwal
```
chizhg, line 14: This PR just removes Nikhil from the reviewers, but still not all oncalls will have the `approve` power, so it's different from the PR description?

chaodaiG: `git commit -am` :(

Thanks for noting

PR #2294, opened 2020-07-13
Make Serving e2e resources uniform

mattmoor:


chaodaiG:
/lgtm
/approve

This is pretty cool trick to learn

mattmoor:
image

config/prod/prow/jobs/config.yaml
```diff
@@ -217,6 +222,11 @@ presubmits:
           value: /etc/test-account/service-account.json
         - name: E2E_CLUSTER_REGION
           value: us-central1
+        resources:
```
vaikas, line 225: Just curious why not use anchor here as well?

chizhg: This file is generated, so it does not matter that much to use anchor or not here. More importantly, we are not entirely sure if Prow supports anchors in the Prow job configuration file..

vaikas: ah, cool cool :) 
Note that we use these kind of anchors in our crd definitions as well, so depending on how they get applied, I'd expect things to just magically to work. But if it's generated, doesn't matter :)  Thanks for the explanation.

Issue #414, opened 2019-01-23
Investigate using Failure Aggregator for knative


srinivashegde86:
k8s has a failure aggregator that provides trending view of all the failures. This will be great for admins and WG's to check failures and identify flakes.

k8s version - https://storage.googleapis.com/k8s-gubernator/triage/index.html

chaodaiG:
this looks pretty good. In terms of implementation, not sure how much we could leverage from k8s, if not there is going to be a major investment on re-implementing `test-infra/kettle` for data collection, and `test-infra/triage` for UI. Also we need to have a discussion of whether we want to store data in bigquery or not

srinivashegde86:
idk much about kettle. I initially thought it might be simple to bring up. If this needs a lot of work, we can probably hold on to this for later.

chaodaiG:
sounds good. I think bigquery is certainly worthy looking into when the needs come, and we could iterate on this then.

adrcunha:
The dashboard code lives in https://github.com/kubernetes/test-infra/tree/master/triage

PR #2299, opened 2020-07-15
Replace issue tracker job for test-infra with github action

chaodaiG:
Github actions support stale https://github.com/actions/stale, which uses default github token for management, using it can avoid us managing another github token

Use label `lifecycle/stale` so that it can works with prow lifecycle plugin.

/cc @chizhg @mattmoor


chizhg:
https://github.com/knative/test-infra/issues/414#issuecomment-660328890
It works 🎉 , but the line break is not correct.

PR #2303, opened 2020-07-17
Stale action ignores lifecycle/frozen

chaodaiG:
/cc @chizhg


chaodaiG:
> /lgtm
> /approve
> 
> Do we still need prow for adding the labels?

We don't need prow for adding the labels any more, but prow `lifecycle` plugin works nicely side by side with this. i.e. one can recover an issue from stale state using `/lifecycle` style comment

PR #2308, opened 2020-07-20
Add --help option for presubmit-tests.sh and fix more shellcheck warnings

chizhg:
1. Add --help option for presubmit-tests.sh
2. Fix more shellcheck warnings
3. Use `JOB_TYPE` to determine if `IS_PRESUBMIT=1`

/cc @chaodaiG @albertomilan

### DIFF
--- a/scripts/presubmit-tests.sh
+++ b/scripts/presubmit-tests.sh
@@ -28,7 +28,7 @@ readonly PRESUBMIT_TEST_FAIL_FAST=${PRESUBMIT_TEST_FAIL_FAST:-0}
 readonly NO_PRESUBMIT_FILES=(\.png \.gitignore \.gitattributes ^OWNERS ^OWNERS_ALIASES ^AUTHORS)
 
 # Flag if this is a presubmit run or not.
-(( IS_PROW )) && [[ -n "${PULL_PULL_SHA}" ]] && IS_PRESUBMIT=1 || IS_PRESUBMIT=0
+(( IS_PROW )) && [[ ${JOB_TYPE} == "presubmit" ]] && IS_PRESUBMIT=1 || IS_PRESUBMIT=0
 readonly IS_PRESUBMIT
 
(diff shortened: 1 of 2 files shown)


scripts/presubmit-tests.sh
```diff
@@ -28,7 +28,7 @@ readonly PRESUBMIT_TEST_FAIL_FAST=${PRESUBMIT_TEST_FAIL_FAST:-0}
 readonly NO_PRESUBMIT_FILES=(\.png \.gitignore \.gitattributes ^OWNERS ^OWNERS_ALIASES ^AUTHORS)
 
 # Flag if this is a presubmit run or not.
-(( IS_PROW )) && [[ -n "${PULL_PULL_SHA}" ]] && IS_PRESUBMIT=1 || IS_PRESUBMIT=0
+(( IS_PROW )) && [[ ${JOB_TYPE} == "presubmit" ]] && IS_PRESUBMIT=1 || IS_PRESUBMIT=0
```
albertomilan, line 31: Aren't the two last clauses in the and statement equivalent? When would the JOB_TYPE be presubmit and IS_PRESUBMIT be 0?

chizhg: `JOB_TYPE` is an env var that will be exposed by Prow in a Prow job, see https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#job-environment-variables.

So when the test is run on Prow and it's a presubmit test job, `JOB_TYPE` will be presubmit and `IS_PRESUBMIT` will be 1. In any other situations, `IS_PRESUBMIT` will be 0.

albertomilan: I don't see this as a blocker, but it doesn't seem to be good in the long term to have two environment variables for the same functionality which split use cases. I would push to have a single aggregated one used by all the use cases, otherwise there could be inconsistencies.
In any case, I believe this falls outside of the scope of this change.

PR #2309, opened 2020-07-20
[WIP]Start using kntest to create clusters

chizhg:

Fixes #

/hold


chizhg:
Some notes:
1. `add_trap "run_kntest cluster gke delete > /dev/null &" EXIT SIGINT` should be run before cluster creation
2. `add_trap` should catch all "interrupting` signals
3. `kntest cluster gke delete` should be best effort, it should release the GCP project even if the cluster has not been created yet

PR #2325, opened 2020-07-24
This fixes the return value of the default integration test runner

mattmoor:


mattmoor:
This line was dropped in #2308

PR #2335, opened 2020-08-02
Replace go-junit-report with gotestsum

### LINKED ISSUE #2319: Use gotestsum to generate XML formatted test results
[gotestsum](https://github.com/gotestyourself/gotestsum) is a tool that can be used to print friendly test output and a summary of the `go test` run. It provides some nice features e.g. hiding the test logs for passed logs, and the community is much more active than the [go-junit-report](https://github.com/jstemmer/go-junit-report) tool we are currently using.

We should evaluate if `gotestsum` is a better option and potentially use it to replace `go-junit-report`.

/kind enhancement

chizhg:
1. Use `gotestsum` to generate XML result, in place of `go-junit-report`
2. Fix a bug that `create_junit_xml` cannot generate the XML files correctly if the error message contains space. The reason is we are using `$@` which will automatically split the string with spaces which will convert the error message into multiple arguments. This is a serious issue as most of the error messages are long.

I have tested the changes in https://github.com/knative/serving/pull/8884 (the failing jobs are due to flaky tests)

Fixes #2319

### DIFF
--- a/images/prow-tests/Dockerfile
+++ b/images/prow-tests/Dockerfile
@@ -67,7 +67,9 @@ RUN source-gvm-and-run.sh install go1.14.6 --prefer-binary
 RUN GO111MODULE=on go get github.com/google/ko/cmd/ko@v0.5.1
 RUN GO111MODULE=on go get github.com/boz/kail/cmd/kail
 RUN go get -u github.com/golang/dep/cmd/dep
+# TODO(chizhg): remove once all repos have been updated to use gotestsum.
 RUN go get -u github.com/jstemmer/go-junit-report
+RUN GO111MODULE=on go get -u gotest.tools/gotestsum
 RUN GO111MODULE=on go get -u github.com/raviqqe/liche@v0.0.0-20200229003944-f57a5d1c5be4  # stable liche version for checking md links
 RUN GO111MODULE=on go get -u github.com/google/go-licenses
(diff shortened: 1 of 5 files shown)


chaodaiG:
Does this affect which go version to be used?

chizhg:
> Does this affect which go version to be used?

All Knative repos have been switched to use Go 1.14, and this change is on master branch so it won't be picked up by other release branches which are using Go 1.13, so Go version shouldn't be a problem.

chaodaiG:
> > Does this affect which go version to be used?
> 
> All Knative repos have been switched to use Go 1.14, and this change is on master branch so it won't be picked up by other release branches which are using Go 1.13, so Go version shouldn't be a problem.

Sorry I meant to ask how does it figure out which `go` it uses? Does it work with gvm?

chizhg:
> > > Does this affect which go version to be used?
> > 
> > 
> > All Knative repos have been switched to use Go 1.14, and this change is on master branch so it won't be picked up by other release branches which are using Go 1.13, so Go version shouldn't be a problem.
> 
> Sorry I meant to ask how does it figure out which `go` it uses? Does it work with gvm?

`gotestsum` just runs `go test` under the hood, which uses the current Go version.
And the current Go version is set by `gvm` before actually running the test job.
So yes, it works with `gvm`.

PR #2339, opened 2020-08-06
Omit property in testcase if empty

peterfeifanchen:
To make it more compatible with gotestsum. Omit properties if it is empty.

### DIFF
--- a/pkg/junit/junit.go
+++ b/pkg/junit/junit.go
@@ -56,14 +56,14 @@ type TestSuite struct {
 
 // TestCase holds <testcase/> results
 type TestCase struct {
-	Name       string         `xml:"name,attr"`
-	Time       float64        `xml:"time,attr"` // Seconds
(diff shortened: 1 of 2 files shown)


chizhg:
Please fix the unit tests.

pkg/junit/junit.go
```diff
@@ -91,6 +91,9 @@ func (testCase *TestCase) GetTestStatus() TestStatusEnum {
 
 // AddProperty adds property to testcase
 func (testCase *TestCase) AddProperty(name, val string) {
+	if testCase.Properties == nil {
+		testCase.Properties = &TestProperties{}
+	}
 	property := TestProperty{Name: name, Value: val}
 	testCase.Properties.Properties = append(testCase.Properties.Properties, property)
```
chizhg, line 98: `.Properties.Properties` looks weird... But since it's unrelated to this PR, let's check it later.

PR #2340, opened 2020-08-06
Bump the resources for all serving e2e test legs

mattmoor:


coryrc:
I hope you checked the resource usage to ensure we won't run out

PR #2342, opened 2020-08-07
e2e-tests.sh switches to kubetest2

### LINKED ISSUE #2345: e2e test infra switches to kubetest2
The `kubetest` tool we use in our e2e test infra for creating clusters is being deprecated in favor of [`kubetest2`](https://github.com/kubernetes-sigs/kubetest2). 

It's better in the following ways:
1. `kubetest2` is pretty well-designed. It defines clear boundaries and interfaces between test orchestrator, infra framework and testing framework, and each part can be updated separately. This give us much flexibility to make switches in each phases, like change the infra framework from `GKE` to `KIND`.
2. It provides opportunities to get alignment on tooling side for all the projects we support, as it has already been adopted by a few internal and external projects.

As the first step, we need to switch to `kubetest2` for Knative test infra.

/assign

chizhg:
1. Switch from `kubetest` to `kubetest2`, as we discussed offline
2. Also remove `resolve_k8s_version` as I don't think it's useful, we can just pass `--cluster-version=latest` and the GKE server will automatically choose the version for us
3. Remove `get_test_return_code` and `set_test_return_code` as they are not useful either, since we do not tear down the clusters on Prow

Fixes #2345 

1. I have tested the changes in this PR and in https://github.com/knative/serving/pull/8701 and https://github.com/knative/eventing/pull/3646, everything still works
2. I didn't include `kubetest2` installation steps in this PR as we didn't see interest from the community to run the test scripts on their local, as we discussed offline we should provide `docker run` mechanism to support that, but IMO that should not be a blocker of this switch
3. After this switch is done, I will make some further cleanups to the GCP specific flags, and move the cluster provision related functions to `infra-library.sh`, so that these functions could be imported by internal repos as well, and it'll be much easier to switch to other providers like `KIND`


coryrc:
/lgtm

Some good cleanup in there; thanks!

PR #2344, opened 2020-08-08
Config generator supports run_if_changed config

### LINKED ISSUE #2343: Prow config generator supports override `run_if_changed`
As discussed on Slack: https://knative.slack.com/archives/CCSNR4FCH/p1596842636408900, it's desired to allow certain presubmit job have `run_if_changed` field to be kicked off only when certain files changed, for example `third_party/foo_ingress` under serving repo, when there are changes it's required to pass `foo_ingress` related tests

chizhg:
1. Config generator supports run_if_changed config
2. Also by the way change `always_run` to `always-run` in the meta config file since it's the only exception (other fields are all using `-` as the name separator.

Fixes #2343 

/cc @chaodaiG @mattmoor

### DIFF
--- a/config/prod/prow/config_knative.yaml
+++ b/config/prod/prow/config_knative.yaml
@@ -50,7 +50,7 @@ presubmits:
     dot-dev: true
   - custom-test: istio-1.5-mesh
     dot-dev: true
-    always_run: false
+    always-run: false
     optional: true
     args:
     - --run-test
@@ -64,7 +64,7 @@ presubmits:
         memory: 16Gi
   - custom-test: istio-1.5-no-mesh
     dot-dev: true
-    always_run: false
+    always-run: false
     optional: true
     args:
     - --run-test
@@ -78,7 +78,7 @@ presubmits:
         memory: 16Gi
   - custom-test: istio-1.4-mesh
     dot-dev: true
-    always_run: false
+    always-run: false
     optional: true
     args:
     - --run-test
@@ -92,7 +92,7 @@ presubmits:
         memory: 16Gi
   - custom-test: istio-1.4-no-mesh
     dot-dev: true
-    always_run: false
+    always-run: false
     optional: true
     args:
     - --run-test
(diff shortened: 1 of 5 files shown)


config/prod/prow/config_knative.yaml
```diff
@@ -134,7 +134,8 @@ presubmits:
         memory: 16Gi
   - custom-test: contour-latest
     dot-dev: true
-    always_run: false
+    always-run: false
+    run-if-changed: "^third_party/contour-latest/*"
```
mattmoor, line 138: Cool!

@tcnghia @nak3 we should consider something similar for the istio directories and the mesh legs?

@markusthoemmes @davidor @jmprusi same for kourier?

nak3: Yes, definitely we would like to have it.

cc @JRBANCEL as trying to change the directory structure.

tools/config-generator/templates/prow_presubmit_job.yaml
```diff
@@ -3,6 +3,7 @@
     [[indent_section 6 "labels" .Base.Labels]]
     context: [[.PresubmitPullJobName]]
     always_run: [[.Base.AlwaysRun]]
```
chaodaiG, line 5: Is there a conflict if `always_run: true` and `run_if_changed` is specified? If not, which one overrides the other?

chizhg: This is no conflict, but `always_run: true` has a higher priority, that means:
1. if `always_run: true`, the Prow job will always be run regardless the value of `run_if_changed`
2. if no `always_run:true`, the Prow job will be run if the changed files match with the value of `run_if_changed`.

Details can be found in https://github.com/kubernetes/test-infra/blob/master/prow/config/jobs.go#L347-L363

chaodaiG: kk, thanks for digging it out

PR #2356, opened 2020-08-12
Keep in-gvm-env.sh to use different Go version to install tools

chizhg:
We need it to install some internal tools implemented in different Go versions.

/cc @peterfeifanchen @coryrc

### DIFF
--- a/images/prow-tests/Dockerfile
+++ b/images/prow-tests/Dockerfile
@@ -137,4 +137,3 @@ RUN cd /go/src/knative.dev/test-infra && git reset --hard upstream/release-0.14 
 # Cleanup
 RUN rm -fr /go/src/knative.dev/test-infra
 RUN rm -f /usr/local/bin/source-gvm-and-run.sh
-RUN rm -f /usr/local/bin/in-gvm-env.sh


coryrc:
This is a bad idea. The script isn't designed for that (it doesn't restore state and running it more than once will probably not work) and installing tools during a run is a BAD IDEA. Tools should be installed in the image so our test runs are repeatable.

When installing `go` tools, they only get installed for that version, so the other version of Go wouldn't have the tool available.

chizhg:
> This is a bad idea. The script isn't designed for that (it doesn't restore state and running it more than once will probably not work) and installing tools during a run is a BAD IDEA. Tools should be installed in the image so our test runs are repeatable.
> 
> When installing `go` tools, they only get installed for that version, so the other version of Go wouldn't have the tool available.

The change is made in order to install some tools in the internal `prow-tests` image, not "installing tools during a run".
The default Go version for this image is Go 1.13, but we need to install some internal tools built with Go 1.14 that's where this script is needed.

peterfeifanchen:
our kubekins image is still gcr.io/k8s-testimages/kubekins-e2e:v20200326-e946722-master, could we not just upgrade to the latest kubekins image so our default go is 1.15?

chizhg:
> our kubekins image is still gcr.io/k8s-testimages/kubekins-e2e:v20200326-e946722-master, could we not just upgrade to the latest kubekins image so our default go is 1.15?

The latest kubekins image is using Go 1.13.5 - https://github.com/kubernetes/test-infra/blob/master/images/kubekins-e2e/cloudbuild.yaml#L39. We should not rely on that since we may want to stop using kubekins-e2e image since it installs a lot of stuff we do not need.

peterfeifanchen:
> > our kubekins image is still gcr.io/k8s-testimages/kubekins-e2e:v20200326-e946722-master, could we not just upgrade to the latest kubekins image so our default go is 1.15?
> 
> The latest kubekins image is using Go 1.13.5 - https://github.com/kubernetes/test-infra/blob/master/images/kubekins-e2e/cloudbuild.yaml#L39. We should not rely on that since we may want to stop using kubekins-e2e image since it installs a lot of stuff we do not need.

what's the variants.yaml for then?

chizhg:
> > > our kubekins image is still gcr.io/k8s-testimages/kubekins-e2e:v20200326-e946722-master, could we not just upgrade to the latest kubekins image so our default go is 1.15?
> > 
> > 
> > The latest kubekins image is using Go 1.13.5 - https://github.com/kubernetes/test-infra/blob/master/images/kubekins-e2e/cloudbuild.yaml#L39. We should not rely on that since we may want to stop using kubekins-e2e image since it installs a lot of stuff we do not need.
> 
> what's the variants.yaml for then?

Okay this file is for building images for different k8s versions - https://console.cloud.google.com/gcr/images/k8s-testimages/GLOBAL/kubekins-e2e?gcrImageListsize=30, and it looks `cloudbuild.yaml` is not used anymore. But I still don't think we should rely on that, as the reason I mentioned in https://github.com/knative/test-infra/pull/2356#issuecomment-672636258, and we do have the potential need to install Go tools with different Go versions.

coryrc:
> The change is made in order to install some tools in the internal `prow-tests` image, not "installing tools during a run".
> The default Go version for this image is Go 1.13, but we need to install some internal tools built with Go 1.14 that's where this script is needed.

Oh, oops! Okay then.

PR #2364, opened 2020-08-14
Abstract the infra related code out to a separate script and some cleanups

chizhg:
1. Move the infra related code out of the `e2e-tests.sh` to `infra-library.sh` 
2. Some cleanups to the legacy code in `e2e-tests.sh`

Notes:
1. I have tested the changes for serving and eventing, and neither of them is broken.
2. The goal is to completely decouple the infra setup and test phases. Eventually `e2e-tests.sh` will only take care of SUT setup and test execution, and `infra-library.sh` will include the helper functions for starting up clusters for different cloud providers. This separation is also useful for the future switch to use KinD as the cloud provider and Tekton.
3. There could be some further cleanups which I have added `TODO` comments.


chaodaiG:
Most of the changes are just moving pieces around, looks good to me. Will wait for a couple more pair of eyes

PR #2365, opened 2020-08-14
Add sandbox .github repo

mattmoor:
cc @chaodaiG @chizhg 

Should this have created stuff when I ran codegen?


chaodaiG:
> cc @chaodaiG @chizhg
> 
> Should this have created stuff when I ran codegen?

Probably not. We have switched to org level defaults, so I don't even think this PR is necessary

mattmoor:
I was just following community :)

PR #2378, opened 2020-08-20
Project structure

Yizong98:
This PR creates the skeleton for DKCM.

Fixes #


chizhg:
Please squash your commits and remove the extra files (which you can see from https://github.com/knative/test-infra/pull/2378/files).

You can follow https://github.com/kubernetes/community/blob/master/contributors/guide/github-workflow.md to do these. This doc is for kubernetes but is pretty general, so it can also be applied here.

PR #2382, opened 2020-08-22
Remove dot-dev and ORG_NAME env var from meta job config

### LINKED ISSUE #2373: Small but useful improvements to the config generator tool
The [meta config file](https://github.com/knative/test-infra/blob/master/config/prod/prow/config_knative.yaml) we use to generate Prow jobs is growing pretty large, which makes it quite hard to read and easy to make mistakes. There have been discussions on completely rewriting it, but is not prioritized yet.

Some small and useful improvements can help reduce the meta config size and make it more readable:
1. Make `dot-dev: true` a setting in the repo level, rather than in each job
2. Get rid of the [env vars set in the release jobs](https://github.com/knative/test-infra/blob/master/config/prod/prow/config_knative.yaml#L1083-L1084) as the org name should be able to be got from the running job, if now, it should be handled within the config generator.

FYI @chaodaiG

chizhg:
Remove dot-dev and ORG_NAME env var from meta job config. This change deletes 241 lines from the meta config file.

Fixes #2373

/cc @chaodaiG

### DIFF
--- a/tools/config-generator/main.go
+++ b/tools/config-generator/main.go
@@ -35,6 +35,8 @@ import (
 	"time"
 
 	"gopkg.in/yaml.v2"
+	"k8s.io/apimachinery/pkg/util/sets"
+
 	"knative.dev/test-infra/pkg/ghutil"
 )
 
@@ -48,9 +50,13 @@ const (
 	// ##########################################################
 	// commonHeaderConfig contains common header definitions.
 	commonHeaderConfig = "common_header.yaml"
+)
 
-	// generalProwConfig contains config-wide definitions.
-	generalProwConfig = "prow_config.yaml"
+var (
+	// GitHub orgs that are using knative.dev path alias.
+	pathAliasOrgs = sets.NewString("knative", "knative-sandbox")
+	// GitHub repos that are not using knative.dev path alias.
+	nonPathAliasRepos = sets.NewString("knative/docs")
 )
 
 // repositoryData contains basic data about each Knative repository.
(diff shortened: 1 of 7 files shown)


tools/config-generator/main.go
```diff
@@ -205,6 +204,11 @@ func newbaseProwJobTemplateData(repo string) baseProwJobTemplateData {
 	data.Timeout = 50
 	data.OrgName = strings.Split(repo, "/")[0]
 	data.RepoName = strings.Replace(repo, data.OrgName+"/", "", 1)
+	data.ExtraRefs = []string{"- org: " + data.OrgName, "  repo: " + data.RepoName}
+	if data.OrgName == "knative" || data.OrgName == "knative-sandbox" {
+		data.PathAlias = "path_alias: knative.dev/" + data.RepoName
```
chaodaiG, line 209: Not quite sure, but I'm afraid that not every repo is on knative.dev yet. At least docs is not yet
https://github.com/knative/website/blob/master/static/_redirects

chizhg: Thanks for noticing this. It looks docs is the only repo that is still in `dep` and not using path alias.. 
I have added some extra logic in the config generator tool, which makes the code dirty but also makes the meta config file much cleaner.

chaodaiG: How about move docs over to knative.dev? It's not too much code

chizhg: That will break the Prow jobs for the release branches, as they are not in Go modules and have to be under `GOPATH`. To be backward compatible, we'll have to get back the per branch config generation logic that I deleted in https://github.com/knative/test-infra/pull/2148, which I don't want to..

I'm trying to switch docs to Go mod - https://github.com/knative/docs/pull/2718, if it can be merged, we will be able to delete this special case in the config generator after 4 release cycles.

chaodaiG: ack

PR #2383, opened 2020-08-22
Fix Peribolos Prow jobs by using a different GitHub token and running it in the build cluster

### LINKED ISSUE #2289: Peribolos Prow jobs fail when --fix-team-repos is set to true
When `--fix-team-repos` is set to true, Peribolos Prow jobs will fail, see https://prow.knative.dev/view/gcs/knative-prow/logs/ci-knative-peribolos/1282517750937292801#1:build-log.txt%3A307, and the error message `Entrypoint received interrupt: terminated` is not quite actionable.

We should figure out why it fails and fix it, it can probably be fixed by https://github.com/kubernetes/test-infra/pull/17895

chizhg:
1. Run Peribolos Prow jobs with `--fix-team-repos=true` option on the `prow-trusted` build cluster, which seemingly to be able to fix the `Entrypoint received interrupt: terminated` error, see the discussions in https://github.com/kubernetes/test-infra/issues/18930
2. Use a dedicated GitHub oauth token for running Peribolos jobs, which I think is the reason for the `sleep time for token reset exceeds max sleep time` error we received on Slack - https://knative.slack.com/archives/CCSNR4FCH/p1597949495009800. I also updated our oncall playbook.

Fixes #2289 

/cc @chaodaiG @mattmoor @albertomilan

### DIFF
--- a/config/prod/prow/jobs/custom/peribolos.yaml
+++ b/config/prod/prow/jobs/custom/peribolos.yaml
@@ -23,6 +23,7 @@ presubmits:
     decorate: true
     path_alias: knative.dev/community
     run_if_changed: "^peribolos/knative.yaml$"
+    cluster: "prow-trusted"
     branches:
     - "master"
     spec:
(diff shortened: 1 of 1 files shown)


chaodaiG:
Prow job change lgtm
/approve

Defer lgtm to Evan for peribolos side of changes

PR #2385, opened 2020-08-25
Make E2E_SCRIPT value global

nak3:
Since https://github.com/knative/test-infra/commit/67da4f8de8cd43ec361b0c43950b2ab1a160304d, `E2E_SCRIPT` became local value as `e2e_script`.
But E2E_SCRIPT is used for log file's suffix like
https://github.com/knative/test-infra/blob/master/scripts/infra-library.sh#L72.

Hence this patch makes E2E_SCRIPT value global.

/cc @chizhg @chaodaiG


chizhg:
/lgtm
/approve
Thanks! Sorry for the breakage!

PR #2388, opened 2020-08-26
Install Go 1.15 in the prow-tests image

chizhg:
Install Go 1.15 in the prow-tests image

/cc @chaodaiG @coryrc


coryrc:
Can you pick a specific patch version so it's consistent if the image is rebuilt?

chizhg:
> Can you pick a specific patch version so it's consistent if the image is rebuilt?

Good point. 
However, since Go 1.15 does not have any minor releases yet - https://golang.org/doc/devel/release.html#go1.15, I tried to use `go1.15.0` but it gives an error `ERROR: Failed to download binary go`, so we'll have to use `go1.15` for now.

coryrc:
Ha, they don't tag it right: https://github.com/golang/go/tags

sigh

/lgtm

PR #2394, opened 2020-08-27
Grant container admin role to the Prow tests service account

chizhg:
I was surprised that Prow tests service account does not have `container.admin` role, that's why we have a so complicated `acquire_cluster_admin_role` function in https://github.com/knative/test-infra/blob/master/scripts/library.sh#L330-L366.

Grant this role then we can change the function to one line.

/cc @chaodaiG


chaodaiG:
It doesn't hurt, let's do it
/lgtm

PR #2396, opened 2020-08-28
update config_knative with new async repo

beemarie:
This is a PR for updating the config_knative.yaml file with a new async repo to set up Prow.

Fixes #

This should set up prow for the async-component repo.

### DIFF
--- a/config/prod/prow/config_knative.yaml
+++ b/config/prod/prow/config_knative.yaml
@@ -833,7 +833,7 @@ periodics:
   - nightly: true
   - dot-release: true
   - auto-release: true
-  knative-sandbox/discovery:
+  knative-sandbox/async-component:
   - continuous: true
   - nightly: true
   - dot-release: true
--- a/config/prod/prow/config_knative.yaml
+++ b/config/prod/prow/config_knative.yaml
@@ -334,6 +334,11 @@ presubmits:
     args:
     - --run-test
     - ./test/e2e-upgrade-tests.sh
+  knative-sandbox/async-component:
+  - build-tests: false
+  - unit-tests: true
+  - integration-tests: false
+  - go-coverage: true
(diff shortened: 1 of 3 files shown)


config/prod/prow/config_knative.yaml
```diff
@@ -833,6 +838,11 @@ periodics:
   - nightly: true
   - dot-release: true
   - auto-release: true
+  knative-sandbox/discovery:
```
mattmoor, line 841: I think you forgot to rename?

beemarie: ugh, sorry. Thanks!

config/prod/prow/config_knative.yaml
```diff
@@ -334,6 +334,11 @@ presubmits:
     args:
     - --run-test
     - ./test/e2e-upgrade-tests.sh
+  knative-sandbox/async-component:
+  - build-tests: false
```
chizhg, line 338: Could you explain why we don't need build-tests for this repo?

beemarie: I think we will need build tests, but I left this as false since they don't exist yet. Same for integration tests.

chizhg: build-tests will just run `go build -v ./...` by default, I think it shouldn't hurt to add now as long as the repo contains Go code?

beemarie: okay, thanks. That makes sense.

chizhg: Hmm, you probably forgot to run `./hack/generate-configs.sh`

beemarie: thanks! just ran & pushed the generated file.

PR #2400, opened 2020-08-28
Cleanup prow-tests Dockerfile

coryrc:
We were on really old kubekins-e2e image.

I think we can safely remove a bunch of obsolete things:
* coverage is never called as /coverage
* Using newer kubekins should give newer kubectl + the referenced
issue is allegedly fixed anyway

/assign @chizhg

### DIFF
--- a/images/prow-tests/Dockerfile
+++ b/images/prow-tests/Dockerfile
@@ -12,12 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# This is the base Dockerfile for knative-tests images, please remember to also
-# do `make push` in `../prow-tests-*` directories to make sure these images are in
-# sync
-
-# TODO(chizhg): probably build the image from scratch?
-FROM gcr.io/k8s-testimages/kubekins-e2e:v20200326-e946722-master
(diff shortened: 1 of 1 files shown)


images/prow-tests/Dockerfile
```diff
@@ -29,13 +24,6 @@ RUN gcloud components update
 RUN gcloud components install docker-credential-gcr
 RUN docker-credential-gcr configure-docker
 
-# Replace kubectl with a working version
-# TODO(chizhg): remove once https://github.com/knative/test-infra/issues/1858 is fixed
-ARG KUBECTL_VERSION=v1.17.4
```
chizhg, line 34: This cannot be removed yet as some old release versions are still using the old way of installing Knative serving - https://github.com/knative/serving/blob/release-0.15/vendor/knative.dev/test-infra/scripts/library.sh#L489-L499.

Could you update the comment to say `remove once test infra stops supporting releases <=0.15`?

coryrc: done

chizhg:
Thanks!
/lgtm
/approve

PR #2404, opened 2020-09-01
functions implementation

Yizong98:
Develop functions for clerk
/assign chizhg chaodaiG

### DIFF
--- a/tools/dkcm/clerk/clerk.go
+++ b/tools/dkcm/clerk/clerk.go
@@ -18,35 +18,39 @@ package clerk
 import (
 	"database/sql"
 	"fmt"
+	"strings"
 	"time"
 
 	"k8s.io/apimachinery/pkg/util/uuid"
 	"knative.dev/test-infra/pkg/mysql"
 )
 
+// ClusterParams is a struct for common cluster parameters shared by both Cluster and Request
+type ClusterParams struct {
+	Zone     string
+	Nodes    int64
+	NodeType string
+}
+
 // Cluster stores a row in the "Cluster" db table
 type Cluster struct {
-	ClusterID string
+	*ClusterParams
+	ClusterID int64
 	ProjectID string
 	Status    string
-	Zone      string
-	Nodes     int
-	NodeType  string
 }
 
 // Request stores a request for the "Request" table
 type Request struct {
+	*ClusterParams
 	accessToken string
 	requestTime time.Time
-	Zone        string
-	Nodes       int
-	NodeType    string
 	ProwJobID   string
-	ClusterID   string
+	ClusterID   int64
 }
 
-// ClusterConfig is the struct to return to Prow once the cluster is available
-type ClusterConfig struct {
+// Response is the struct to return(end product) to Prow once the cluster is available
+type Response struct {
 	ClusterName string
 	ProjectID   string
 	Zone        string
@@ -54,25 +58,30 @@ type ClusterConfig struct {
 
 type ClerkOperations interface {
 	//check cluster available, if available, return cluster id and access token
-	CheckAvail(nodes int, nodeType string, zone string) (bool, string, string)
-	// get with token
-	GetCluster(accessToken string) (ClusterConfig, error)
+	CheckAvail(cp ClusterParams) (bool, int64)
+	// get with cluster id stored in the Cluster database
+	GetCluster(clusterID int64) (Response, error)
 	// delete a cluster entry
-	DeleteCluster(accessToken string) error
+	DeleteCluster(clusterID int64) error
 	// Insert a cluster entry
-	InsertCluster(nodes int, nodeType string, projectID string, status string, zone string) error
+	InsertCluster(c *Cluster) error
 	// Update a cluster entry
-	UpdateCluster(clusterID string, fields []string, values []string) error
-	// List clutsers within a time interval (use for checking after downtime to see stale clusters)
-	ListClusters(window time.Duration) ([]Cluster, error)
+	UpdateCluster(clusterID int64, opts ...UpdateOption) error
+	// List clutsers (use for checking after downtime to see stale clusters)
+	ListClusters() ([]Cluster, error)
 	// Generate Unique access token for Prow
 	generateToken() string
 	// Insert a request entry
-	InsertRequest(Zone string, Nodes int, NodeType string, ProwJobID string) error
+	InsertRequest(r *Request) error
 	// Update a request entry
-	UpdateRequest(requestID string, fields []string, values []string) error
+	UpdateRequest(requestID int64, opts ...UpdateOption) error
 	// List requests within a time interval (use for checking after downtime to see stale requests)
 	ListRequests(window time.Duration) ([]Request, error)
+}
+
+// compatible with both Row and Rows and unit test friendly
+type scannable interface {
+	Scan(dest ...interface{}) error
 }
 
 // DB holds an active database connection. This implements all the fuunctions of ClerkOperations.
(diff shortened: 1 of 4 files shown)


tools/dkcm/clerk/clerk.go
```diff
@@ -75,77 +76,199 @@ type ClerkOperations interface {
...
-	return true, "", ""
+// check available clusters in Cluster db and return zone and projectID that would be used by ProwJob
+func (db *DB) CheckAvail(nodes int, nodeType string, zone string) (bool, int64) {
+	// check whether available cluster exists
+	row := db.QueryRow("SELECT * FROM Cluster WHERE Status = 'Ready' AND Nodes = ? AND NodeType = ? AND Zone = ?",
+		nodes, nodeType, zone)
+	c, err := populateCluster(row)
+	// no available cluster is found
+	if err != nil {
+		return false, 0
```
chaodaiG, line 127: To be safe, return -1 instead of 0

tools/dkcm/clerk/clerk.go
```diff
@@ -75,77 +76,199 @@ type ClerkOperations interface {
...
 
 // insert a cluster entry into db
-func (db *DB) InsertCluster(nodes int, nodeType string, accessToken string, projectID string, prowJobID string, status string, zone string) error {
-	return nil
+func (db *DB) InsertCluster(nodes int, nodeType string, projectID string, status string, zone string) error {
+	stmt, err := db.Prepare(`INSERT INTO Cluster(Nodes, NodeType, ProjectID, Status, Zone)
+							VALUES (?,?,?,?,?,?)`)
+	defer stmt.Close()
+
+	if err != nil {
```
chaodaiG, line 139: This should be above `defer`, otherwise the process might crash due to nil pointer

tools/dkcm/clerk/clerk.go
```diff
@@ -75,77 +76,199 @@ type ClerkOperations interface {
...
+		return err
+	}
+
+	_, err = stmt.Exec(nodes, nodeType, projectID, status, zone)
+	return err
+}
+
+func updateQueryString(dbName string, id int, fields []string, values []interface{}) string {
+	updateStatement := fmt.Sprintf(`UPDATE %s SET `, dbName)
+	idStatement := fmt.Sprintf(` WHERE ID = % d`, id)
```
chaodaiG, line 149: should this be `%d` no space?

tools/dkcm/clerk/clerk.go
```diff
@@ -75,77 +76,199 @@ type ClerkOperations interface {
...
+	var fieldStatements []string
+	for idx, val := range fields {
+		fieldStatements = append(fieldStatements, fmt.Sprintf("%s = %v", val, values[idx]))
+	}
+	return updateStatement + strings.Join(fieldStatements, `, `) + idStatement
```
chaodaiG, line 154: Using `fmt.Sprintf` can avoid some of the leading and trailing space in previous strings, which could be easier to maintain in the future

tools/dkcm/clerk/clerk.go
```diff
@@ -75,77 +76,199 @@ type ClerkOperations interface {
...
 
 }
 
 // DeleteCluster deletes a row from Cluster db
-func (db *DB) DeleteCluster(accessToken string) error {
-	return nil
+func (db *DB) DeleteCluster(clusterID int) error {
+	delString := "DELETE FROM Cluster WHERE ID = ?"
+	stmt, err := db.Prepare(delString)
+	defer stmt.Close()
```
chaodaiG, line 188: ditto move `defer` lower

chaodaiG: In this case actually need to only defer when err == nil

tools/dkcm/clerk/clerk.go
```diff
@@ -75,77 +76,199 @@ type ClerkOperations interface {
...
+	if err == nil {
+		err = checkAffected(stmt, 1, clusterID)
```
chaodaiG, line 190: If you worry about wrong number of rows deleted, might want to do `GetCluster` before delete and make sure the row to be deleted exist, so that the problem can be better isolated

tools/dkcm/clerk/schema.sql
```diff
@@ -0,0 +1,36 @@
...
+  Zone varchar(1023) NOT NULL,
+  Nodes varchar(1023) NOT NULL,
+  NodeType varchar(1023) NOT NULL,
+  ProwJobID varchar(1023) NOT NULL,
+  ClusterID int,
+  CONSTRAINT FkCluster 
+  FOREIGN KEY (ClusterID) 
+      REFERENCES Clusters(ID),
+  PRIMARY KEY (ID)
+);
```
chaodaiG, line 36: Add a new line at EOF

tools/dkcm/clerk/clerk.go
```diff
@@ -26,7 +27,7 @@ import (
 
 // Cluster stores a row in the "Cluster" db table
 type Cluster struct {
```
chizhg, line 29: The `Cluster` and `Request` struct have a few fields in common. It'll be better to extract these fields to a common struct, something like below:
```
type ClusterConfig struct {
  Zone      string
  Nodes     int
  NodeType  string
}

type Cluster struct {
  ClusterConfig
  ...
}

type Request struct {
  ClusterConfig
  ...
}
```

Then you can use `ClusterConfig` as the parameter of the `CheckAvail` function.

tools/dkcm/clerk/clerk.go
```diff
@@ -54,17 +55,17 @@ type ClusterConfig struct {
 
 type ClerkOperations interface {
 	//check cluster available, if available, return cluster id and access token
-	CheckAvail(nodes int, nodeType string, zone string) (bool, string, string)
-	// get with token
-	GetCluster(accessToken string) (ClusterConfig, error)
+	CheckAvail(nodes int, nodeType string, zone string) (bool, string)
```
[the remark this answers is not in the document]
chizhg: I think it's better to define a separate `ClusterConfig` struct and use it as the parameter here, as I commented above. WDYT @chaodaiG ?

tools/dkcm/clerk/clerk.go
chaodaiG: That makes sense to me

tools/dkcm/clerk/clerk.go
```diff
@@ -54,17 +55,17 @@ type ClusterConfig struct {
...
+	// get with id
+	GetCluster(clusterID int) (ClusterConfig, error)
 	// delete a cluster entry
-	DeleteCluster(accessToken string) error
+	DeleteCluster(clusterID int) error
```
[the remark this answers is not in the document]
Yizong98: Would be solved by ON DELETE CASCADE

tools/dkcm/clerk/clerk.go
```diff
@@ -54,17 +55,17 @@ type ClusterConfig struct {
 
 type ClerkOperations interface {
 	//check cluster available, if available, return cluster id and access token
-	CheckAvail(nodes int, nodeType string, zone string) (bool, string, string)
-	// get with token
-	GetCluster(accessToken string) (ClusterConfig, error)
+	CheckAvail(nodes int, nodeType string, zone string) (bool, string)
+	// get with id
```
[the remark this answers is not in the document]
Yizong98: solved.

tools/dkcm/clerk/clerk.go
```diff
@@ -54,17 +55,17 @@ type ClusterConfig struct {
...
+	DeleteCluster(clusterID int) error
 	// Insert a cluster entry
 	InsertCluster(nodes int, nodeType string, projectID string, status string, zone string) error
 	// Update a cluster entry
-	UpdateCluster(clusterID string, fields []string, values []string) error
-	// List clutsers within a time interval (use for checking after downtime to see stale clusters)
-	ListClusters(window time.Duration) ([]Cluster, error)
+	UpdateCluster(clusterID int, fields []string, values []string) error
+	// List clutsers (use for checking after downtime to see stale clusters)
+	ListClusters() ([]Cluster, error)
```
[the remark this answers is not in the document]
Yizong98: ClusterConfig is client facing which means we provide it to Prow. Cluster is used for internal debugging purpose(e.g. checking running clusters after downtime). Therefore I believe Cluster makes sense

tools/dkcm/clerk/clerk.go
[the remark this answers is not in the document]
Yizong98: agreed

tools/dkcm/clerk/clerk.go
```diff
@@ -18,34 +18,38 @@ package clerk
...
-	Nodes       int
-	NodeType    string
 	ProwJobID   string
-	ClusterID   string
+	ClusterID   int64
 }
 
-// ClusterConfig is the struct to return to Prow once the cluster is available
+// ClusterConfig is the struct to return(end product) to Prow once the cluster is available
 type ClusterConfig struct {
```
chizhg, line 53: As discussed offline, it's better to rename `ClusterConfig` to `Response`.

tools/dkcm/clerk/schema.sql
```diff
@@ -0,0 +1,40 @@
+/*
```
chizhg, line 1: Please format the copyright header.

PR #2405, opened 2020-09-03
adding ci-knative-serving-contour-latest to beta-prow-tests

peterfeifanchen:
we want to test Go compiler versions without breaking prod

/cc @chizhg @mattmoor

### DIFF
--- a/config/prod/prow/jobs/custom/experimental.yaml
+++ b/config/prod/prow/jobs/custom/experimental.yaml
@@ -47,6 +47,8 @@ periodics:
         value: /etc/test-account/service-account.json
       - name: E2E_CLUSTER_REGION
         value: us-central1
+      - name: GO_VERSION
+        value: devel
   volumes:
   - name: test-account
     secret:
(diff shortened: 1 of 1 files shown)


config/prod/prow/jobs/custom/experimental.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - image: gcr.io/knative-tests/test-infra/prow-tests:beta
+      imagePullPolicy: Always
+      command:
+      - runner.sh
+      args:
+      - "./test/presubmit-tests.sh"
+      - "--run-test"
+      - "./test/e2e-tests.sh --contour-version latest"
+      - "--run-test"
+      - "./test/e2e-auto-tls-tests.sh --contour-version latest"
```
mattmoor, line 40: We don't need these

mattmoor (commented):
/lgtm
/hold

In case you want to remove the excess (it provisions another cluster, and isn't needed to repro the bug)

config/prod/prow/jobs/custom/experimental.yaml
peterfeifanchen: any of the args or just the second set of --run-test? I just copied the original contour ci.

mattmoor (commented):
/lgtm

PR #2407, opened 2020-09-03
generate-config.sh don't import library.sh

chaodaiG:
It's not really needed


coryrc:
/lgtm

Seems unnecessary though

PR #2412, opened 2020-09-08
Add kubetest2 wrapper command for kntest

chizhg:
1. Implement a `kntest kubetest2` subcommand for ease of running integration tests for Knative.
2. Remove all the cluster creation related env vars from the base bash script, and instead using flags to change the default configuration.
3. Other nit fixes and cleanups.

### DIFF
--- a/kntest/pkg/kubetest2/gke/options.go
+++ b/kntest/pkg/kubetest2/gke/options.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package gke
 
 import (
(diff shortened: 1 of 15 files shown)


kntest/pkg/kubetest2/gke/options.go
```diff
@@ -0,0 +1,42 @@
...
+
+package gke
+
+import (
+	"github.com/spf13/cobra"
+
+	"knative.dev/test-infra/pkg/clustermanager/kubetest2"
+)
+
+func addOptions(gkeCmd *cobra.Command, cfg *kubetest2.GKEClusterConfig) {
```
albertomilan, line 25: Is this essentially duplicating kubetest2 parameters? How decoupled can they be?

 I'm concerned about future changes and how that might affect this tool.

kntest/pkg/kubetest2/gke/options.go
```diff
@@ -0,0 +1,42 @@
...
+	f.StringVar(&cfg.Machine, "machine", "e2-standard-4", "The machine type for the GKE cluster.")
+	f.IntVar(&cfg.MinNodes, "min-nodes", 1, "The minimum number of nodes.")
+	f.IntVar(&cfg.MaxNodes, "max-nodes", 3, "The maximum number of nodes.")
+	f.StringVar(&cfg.Network, "network", "e2e-network", "The network name for the GKE cluster.")
+	f.StringVar(&cfg.Version, "version", "latest", "The version of the GKE cluster.")
+	f.StringVar(&cfg.Scopes, "scopes", "cloud-platform", "Scopes for the GKE cluster, should be comma-separated.")
+	f.StringVar(&cfg.Addons, "addons", "", "Addons for the GKE cluster, should be comma-separated.")
+	f.StringVar(&cfg.PrivateClusterAccessLevel, "private-cluster-access-level", "", "Private cluster access level, if not empty, must be one of 'no', 'limited' or 'unrestricted'")
+	f.StringVar(&cfg.PrivateClusterMasterIPRange, "private-cluster-master-ip-range", "172.16.0.%d/28", "The master IP range for the private cluster. The last digit must be left unformatted to allow retrying.")
```
albertomilan, line 41: This can have some security concerns. Could you split this argument as two: subnet range (172.16.0) and mask (28) or some other combination?
Assuming invokers will provide a %d literal is risky, at minimum.

pkg/clustermanager/kubetest2/gke.go
```diff
@@ -0,0 +1,173 @@
...
+		log.Println("Will use boskos to provision the GCP project")
+		timeout := cc.BoskosAcquireTimeoutSeconds
+		if timeout == 0 {
+			timeout = boskosAcquireDefaultTimeoutSeconds
+		}
+		kubetest2Flags = append(kubetest2Flags, "--boskos-acquire-timeout-seconds="+strconv.Itoa(timeout))
+	} else {
+		if cc.GCPProjectID == "" {
+			var err error
+			cc.GCPProjectID, err = cmd.RunCommand("gcloud config get-value project")
```
albertomilan, line 92: This should be a documented behavior. Also, do we want to have it?

albertomilan: I've seen some brief explanation below, but this can have undesired consequences if not known. I would remove this backup case unless there is a strong requirement including it.

kntest/pkg/kubetest2/gke/options.go
chizhg: Could you explain a bit why this could cause security issues? I can split the argument as two, but I would like to understand the underlying reason for this.

pkg/clustermanager/kubetest2/gke.go
chizhg: This is the default behavior for `gcloud` command - if the command does not provide a project, e.g. `gcloud container clusters create test`, it will use the current GCP project, I think it's better to follow the same pattern? (This is mostly for local usage, so there is not a strong requirement to include it but I feel it does not hurt much)

kntest/pkg/kubetest2/gke/options.go
chizhg: These flags are useful for Knative and KubeRun, and here we give the default values that are generally used in the CI flows. If users want to pass extra kubetest2 flags, they could use `--extra-kubetest2-flags` that is supported in the parent `kntest kubetest2` command.

So normally the flags here won't need to be changed unless we want to change the supported flags used for Knative and KubeRun tests, which does not happen quite often.

kntest/pkg/kubetest2/gke/options.go
albertomilan: From my understanding on the change, we are requesting to the invoker to provide an IP range value which includes a %d literal in order to fan out the last octet of the address. This is forcing invokers to provide this value in every case or our logic will fail.
This behavior is undesirable, as making this mandatory does not introduce any particular benefit to the customer and it forces every use case to follow that rule.

In terms of security, letting customers provide formatting string literals (and specially one that is expecting particular data types) is not a recommended approach. It is the basis of the string format vulnerability/attack and, from what I've seen, we're not validating the input in any way (_fmt.Sprintf(cc.PrivateClusterMasterIPRange, i))_). I'm unfamiliar with how Go deals with these cases, but it is not a recommended practice overall.

chizhg: Ack, these are very valid reasons. I have changed the flags as suggested.

kntest/pkg/kubetest2/gke/options.go
albertomilan: By doing so, we'll effectively allow users to know an internal dependency of our tool (kubetest2) and that will complicate migration to other tools in the future. Could we find a set or parameters that could hide the dependency and introduce kntest as a complete intermediate layer?

pkg/clustermanager/kubetest2/gke.go
albertomilan: This follows the same guidance as the comment above. We're using gcloud, but ideally our customers should not care or know how we're fulfilling their requests. I'd rather implement a behavior that is holistically consistent and independent from gcloud in its definition.
What do you think?

kntest/pkg/kubetest2/gke/options.go
chizhg: I got your point. 
The major concern for me on considering kntest as a complete intermediate layer is, every time the test flows need a new flag that is supported by `kubetest2`, like in https://github.com/google/knative-gcp/blob/master/test/e2e-wi-tests.sh#L158, we will need to add it here first, this process will increase the maintenance cost a bit.

pkg/clustermanager/kubetest2/gke.go
chizhg: Makes perfect sense. I have updated the logic to throw out an error if users do not provide the GCP project ID in the non-CI environment.

kntest/pkg/kubetest2/gke/options.go
albertomilan: Ideally, every time kubetest2 accepts a new flag/parameter is because cluster creation/deployment/test execution allows a new functionality and we want to provide it through our own interface. If we don't do that, both tools can become incompatible with each other and it will be extremely hard for us to fix it without affecting customers.
What we want to do is introduce a facade pattern which maps almost 1-to-1 to what kubetest2 does. This map is not exactly direct, otherwise there's no point for the existence of kntest in between, but if we decide to set it we should introduce as much encapsulation as possible. Unless it's strictly necessary to surface kubetest2 parameters in our own interface, I'd push to hide it (even though it might complicate further expansions).

chizhg: My another concern for this is it would be a breaking change to remove the support for passing kubetest2 flags transparently, but what you said it completely reasonable. I will manually update the dependencies of the repos that are impacted by this change.

I have made the corresponding change to remove the `--extra-kubetest2-flags`, please refer to the commit history for better diff check.

PR #2415, opened 2020-09-09
Remove boskos deployment files

chizhg:
As they have been moved to oss-test-infra repo.

/cc


coryrc:
For future reference they are now here: https://github.com/GoogleCloudPlatform/oss-test-infra/tree/master/prow/knative/cluster/build

PR #2420, opened 2020-09-10
Add unit test for outputConfig

joshua-bone:
/lint

Adds a simple unit test to check that outputConfig prints lines properly. The test utils for output redirection will be reused for more complex unit tests in future CLs.


tools/config-generator/...
```diff
@@ -0,0 +1,511 @@
+mode: set
```
chaodaiG, line 1: This file doesn't seem to be necessary

tools/config-generator/testutils_test.go
```diff
@@ -0,0 +1,37 @@
...
+	output = &bytes.Buffer{}
+}
+
+func GetOutput() string {
+	return output.(*bytes.Buffer).String()
+}
+
+func AssertOutput(t *testing.T, expected string) {
+	s := GetOutput()
+	if s != expected {
```
chaodaiG, line 34: The output of cmp is more user friendly, see https://chromium.googlesource.com/external/github.com/google/go-cmp/+/refs/heads/priority/cmp/example_test.go#94

chaodaiG (commented):
Generally lgtm

tools/config-generator/main_test.go
```diff
@@ -0,0 +1,52 @@
...
+	outputConfig("")
+	if diff := cmp.Diff(GetOutput(), ""); diff != "" {
+		t.Errorf("Incorrect output for empty string: (-got +want)\n%s", diff)
+	}
+
+	outputConfig(" \t\n")
+	if diff := cmp.Diff(GetOutput(), ""); diff != "" {
+		t.Errorf("Incorrect output for whitespace string: (-got +want)\n%s", diff)
+	}
+	if emittedOutput {
```
chaodaiG, line 40: where is this variable defined?

joshua-bone: It's a global variable defined in main.go which gets reset in a very non-intuitive fashion.

https://github.com/knative/test-infra/blob/4c79c7ed276bd3c25eec73c8a1dbde7fd5c2e603/tools/config-generator/main.go#L184

chaodaiG: Thank you!

PR #2421, opened 2020-09-11
kntest kubetest2 supports more flags and bug fix

chizhg:
1. Support `--release-channel` and `--enable-stackdriver-kubernetes` flags as they are needed for internal testing
2. Support use `--gcp-service-account` to specify a different service account other than `GOOGLE_APPLICATOIN_CREDENTIALS`
3. Change `--version` to `--cluster-version` to be consistent with the `gcloud` command
4. Fix a bug in the cluster creation retry logic - create a new `flags` array each time it retries, otherwise every time it retries, duplicated flags will be appended to the `kubetest2Flags` array.

/cc

### DIFF
--- a/kntest/pkg/kubetest2/gke/options.go
+++ b/kntest/pkg/kubetest2/gke/options.go
@@ -35,8 +35,9 @@ func addOptions(gkeCmd *cobra.Command, cfg *kubetest2.GKEClusterConfig) {
 	f.IntVar(&cfg.MinNodes, "min-nodes", 1, "The minimum number of nodes.")
 	f.IntVar(&cfg.MaxNodes, "max-nodes", 3, "The maximum number of nodes.")
 	f.StringVar(&cfg.Network, "network", "e2e-network", "The network name for the GKE cluster.")
-	f.StringVar(&cfg.ReleaseChannel, "release-channel", "", "The release channel of the GKE cluster.")
-	f.StringVar(&cfg.Version, "cluster-version", "latest", "The version of the GKE cluster.")
+	f.StringVar(&cfg.ReleaseChannel, "release-channel", "", "The release channel of the GKE cluster, can be one of (empty, rapid, regular, stable). When it's used, --cluster-version must be empty or a valid version in the channel."+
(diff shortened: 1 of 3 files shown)


kntest/pkg/kubetest2/gke/options.go
```diff
@@ -34,10 +35,12 @@ func addOptions(gkeCmd *cobra.Command, cfg *kubetest2.GKEClusterConfig) {
 	f.IntVar(&cfg.MinNodes, "min-nodes", 1, "The minimum number of nodes.")
 	f.IntVar(&cfg.MaxNodes, "max-nodes", 3, "The maximum number of nodes.")
 	f.StringVar(&cfg.Network, "network", "e2e-network", "The network name for the GKE cluster.")
-	f.StringVar(&cfg.Version, "version", "latest", "The version of the GKE cluster.")
+	f.StringVar(&cfg.ReleaseChannel, "release-channel", "", "The release channel of the GKE cluster.")
```
albertomilan, line 38: Include in the comment a reference on how this argument overrides, and has preference, over the cluster-version one.

kntest/pkg/kubetest2/gke/options.go
```diff
@@ -34,10 +35,12 @@ func addOptions(gkeCmd *cobra.Command, cfg *kubetest2.GKEClusterConfig) {
...
+	f.StringVar(&cfg.Version, "cluster-version", "latest", "The version of the GKE cluster.")
```
albertomilan, line 39: Notify this argument is mandatory if no release-channel is provided in the comment

kntest/pkg/kubetest2/gke/options.go
chizhg: Done.

kntest/pkg/kubetest2/gke/options.go
chizhg: Actually this argument can be empty if no release-channel is provided, then it will use the default version, currently it's using `1.15.12-gke.2`, but I have added some comments to explain how it works with release-channel.

PR #2426, opened 2020-09-15
configure knative-sandbox/eventing-github

lionelvillard:

 knative-sandbox/eventing-github is a new repo

Fixes #

### DIFF
--- a/config/prod/prow/config_knative.yaml
+++ b/config/prod/prow/config_knative.yaml
@@ -663,6 +663,10 @@ periodics:
         memory: 12Gi
       limits:
         memory: 16Gi
+  knative-sandbox/eventing-github:
+  - continuous: true
+  - nightly: true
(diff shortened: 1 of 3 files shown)


config/prod/prow/config_knative.yaml
```diff
@@ -663,6 +663,10 @@ periodics:
         memory: 12Gi
       limits:
         memory: 16Gi
+  knative-sandbox/eventing-github:
```
chizhg, line 666: Do you also need presubmit tests for this repo? If yes, you'll also need to configure under the `presubmits` block.

lionelvillard: I think we don't need to run the presubmit tests yet since this repo is configured to run GitHub actions and there is no e2e tests yet. 

Let me know if I'm missing something, thanks!

PR #2437, opened 2020-09-17
Mock log.Fatalf calls and add test suite

joshua-bone:
/lint

- Since log.Fatalf calls are untestable, this PR adds a layer of abstraction and a simple test mock. 
- This PR also adds a complete test suite for the functions in utils.go.
- Increases config-generator test coverage from ~13% to 19.6%.


tools/config-generator/main_test.go
```diff
@@ -14,17 +14,13 @@ limitations under the License.
 package main
 
 import (
-	"os"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
 )
 
-func TestMain(m *testing.M) {
```
coryrc, line 23: why?

coryrc:
"Since log.Fatalf calls are untestable, this PR adds a layer of abstraction and a simple test mock."

Won't they still cause the test to fail?

tools/config-generator/main_test.go
joshua-bone: TestMain does setup only once for the package, not for each individual test as I believed. See this discussion: https://github.com/golang/go/issues/27927

coryrc: oh, ha forgot about that, I thought it was a test called Main... dur

joshua-bone:
> "Since log.Fatalf calls are untestable, this PR adds a layer of abstraction and a simple test mock."
> Won't they still cause the test to fail?

The SetupForTesting() function replaces the logFatalf function with a mock function that simply increments a global counter variable without causing the program to exit.

coryrc:
Correct, but why is this an improvement? In either case, the test correctly
fails.

joshua-bone:
> Correct, but why is this an improvement? In either case, the test correctly
> fails.

There are many cases where I want to verify that an input will cause a failure condition. In those cases I want the code to fail but the test to succeed.

coryrc:
Ah, thanks!

coryrc:
/approve
(I can do that now)

PR #2440, opened 2020-09-17
Simplify the wait_for_xx bash functions

### LINKED ISSUE #1611: wait_until_pods_running should be able to rule out error pods that are not related to the final state of the owners
In https://github.com/knative/test-infra/blob/97901db33f62f8e1abe9c8221670a09c5a1cba8a/scripts/library.sh#L133, `wait_until_pods_running` will only succeed if all pods in the given namespace are in `Running` or `Completed` state.

But since k8s has some retry logic, e.g. K8s Job can create a new pod if there is an error, one error pod does not necessarily mean the Job fails - https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/knative_serving/6440/pull-knative-serving-integration-tests/1214219978266382337 is an example. In such scenario `wait_until_pods_running` will return an error that is not necessarily true.

This function should be general enough to consider and rule out error pods that are not related to the final state of the owners, e.g.
1. For Deployments, all pods should be X/X Running
2. For Jobs, it should depend on the success criteria
...

FYI @mattmoor

chizhg:
- Use `kubectl wait pod --for=condition=Ready -n "$1" -l '!job-name' --timeout=5m` for `wait_for_pods_running`
- Use `kubectl wait job --for=condition=Complete --all -n "$1"` for `wait_until_batch_job_complete`

Fixes #1611

### DIFF
--- a/scripts/library.sh
+++ b/scripts/library.sh
@@ -129,13 +129,18 @@ function wait_until_object_does_not_exist() {
 function wait_until_pods_running() {
   echo "Waiting until all pods in namespace $1 are up"
   kubectl wait pod --for=condition=Ready -n "$1" -l '!job-name' --timeout=5m || return 1
+  # Also wait for all the job pods to be completed.
+  # This is mainly for maintaining backward compatibility.
+  if [[ $(kubectl get jobs --ignore-not-found=true -n "$1") ]]; then
+    kubectl wait job --for=condition=Complete --all -n "$1" --timeout=5m || return 1
+  fi
 }
 
 # Waits until all batch jobs complete in the given namespace.
 # Parameters: $1 - namespace.
 function wait_until_batch_job_complete() {
   echo "Waiting until all batch jobs in namespace $1 run to completion."
-  kubectl wait job --for=condition=Complete --all -n "$1" || return 1
+  kubectl wait job --for=condition=Complete --all -n "$1" --timeout=5m || return 1
 }
 
 # Waits until the given service has an external address (IP/hostname).
--- a/scripts/library.sh
+++ b/scripts/library.sh
@@ -134,21 +134,8 @@ function wait_until_pods_running() {
 # Waits until all batch jobs complete in the given namespace.
 # Parameters: $1 - namespace.
 function wait_until_batch_job_complete() {
-  echo -n "Waiting until all batch jobs in namespace $1 run to completion."
-  for i in {1..150}; do  # timeout after 5 minutes
-    local jobs=$(kubectl get jobs -n $1 --no-headers \
-                 -ocustom-columns='n:{.metadata.name},c:{.spec.completions},s:{.status.succeeded}')
-    # All jobs must be complete
-    local not_complete=$(echo "${jobs}" | awk '{if ($2!=$3) print $0}' | wc -l)
-    if [[ ${not_complete} -eq 0 ]]; then
-      echo -e "\nAll jobs are complete:\n${jobs}"
-      return 0
-    fi
-    echo -n "."
-    sleep 2
-  done
-  echo -e "\n\nERROR: timeout waiting for jobs to complete\n${jobs}"
-  return 1
+  echo "Waiting until all batch jobs in namespace $1 run to completion."
+  kubectl wait job --for=condition=Complete --all -n "$1" || return 1
 }
 
 # Waits until the given service has an external address (IP/hostname).
--- a/scripts/library.sh
+++ b/scripts/library.sh
@@ -69,12 +69,12 @@ function abort() {
 # Parameters: $1 - character to use for the box.
 #             $2 - banner message.
 function make_banner() {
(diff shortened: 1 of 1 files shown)


scripts/library.sh
```diff
@@ -127,70 +127,15 @@ function wait_until_object_does_not_exist() {
...
-  echo -e "\n\nERROR: timeout waiting for pods to come up\n${pods}"
-  if [[ -n "${failed_pod}" ]]; then
-    echo -e "\n\nFailed Pod (data in YAML format) - ${failed_pod}\n"
-    kubectl -n $1 get pods "${failed_pod}" -oyaml
-    echo -e "\n\nPod Logs\n"
-    kubectl -n $1 logs "${failed_pod}" --all-containers
-  fi
-  return 1
+  echo "Waiting until all pods in namespace $1 are up"
+  kubectl wait pod --for=condition=Ready -n "$1" -l '!job-name' --timeout=5m || return 1
```
chaodaiG, line 131: Just to make sure that Ready covers `Running`, `Completed`, etc right?

chaodaiG:
I believe you have tested it already, so
/lgtm
/hold
In case my comment makes sense to you

scripts/library.sh
chizhg: Thanks for pointing this out.

Actually the function name is not correct - other than `wait_until_pods_running` it also `wait_until_job_pods_completed`. But since the function is being used everywhere, let's also wait for the job pods to be completed to avoid regressions.

scripts/library.sh
```diff
@@ -127,70 +127,18 @@ function wait_until_object_does_not_exist() {
...
+  # Also wait for all the job pods to be completed.
+  # This is mainly for backward compatibility and should not throw out an error is there is no job in the ns.
+  kubectl wait job --for=condition=Complete --all -n "$1" --timeout=5m || true
```
mattmoor, line 134: This should be Completed (with a `d`), I believe.

We definitely don't want `--all` with Completed, maybe `-ljobname` (no `!`).

scripts/library.sh
```diff
@@ -127,70 +127,18 @@ function wait_until_object_does_not_exist() {
...
-      echo -e "\nAll jobs are complete:\n${jobs}"
-      return 0
-    fi
-    echo -n "."
-    sleep 2
-  done
-  echo -e "\n\nERROR: timeout waiting for jobs to complete\n${jobs}"
-  return 1
+  echo "Waiting until all batch jobs in namespace $1 run to completion."
+  kubectl wait job --for=condition=Complete --all -n "$1" --timeout=5m || return 1
```
mattmoor, line 141: Completed?

mattmoor: `-ljobname` as well

scripts/library.sh
mattmoor: So my $0.02, we should just use this one line: Ready and `-l '!jobname'`.

This doesn't properly handle Job pods today, which are what result in `Completed` because we need one pod to be Completed, not all (this waits for all).

scripts/library.sh
chizhg: This kubectl command I added is waiting on `job` instead of `pod`

The status of a completed job pod is like this:
```
conditions:
  - lastProbeTime: null
    lastTransitionTime: "2020-09-17T19:41:07Z"
    reason: PodCompleted
    status: "True"
    type: Initialized
  - lastProbeTime: null
    lastTransitionTime: "2020-09-17T19:41:21Z"
    reason: PodCompleted
    status: "False"
    type: Ready
  - lastProbeTime: null
    lastTransitionTime: "2020-09-17T19:41:21Z"
    reason: PodCompleted
    status: "False"
    type: ContainersReady
  - lastProbeTime: null
    lastTransitionTime: "2020-09-17T19:41:07Z"
    status: "True"
    type: PodScheduled
```

It does not have a `Completed` condition, so the command will simply fail:
```
chizhg@chizhg-macbookpro:~$ kubectl wait pod --for=Completed -n istio-system -l 'job-name' --timeout=5m
error: unrecognized condition: "Completed"
```

scripts/library.sh
chizhg: This command is waiting on `job`, I have tested the command on my cluster and it works:
```
chizhg@chizhg-macbookpro:~$ kubectl wait job --for=condition=Complete --all -n istio-system --timeout=5m
job.batch/istio-init-crd-10-1.4.6 condition met
job.batch/istio-init-crd-11-1.4.6 condition met
job.batch/istio-init-crd-14-1.4.6 condition met
job.batch/istio-security-post-install-1.4.6 condition met
```

mattmoor: Ah ok, not pods, I misread

scripts/library.sh
mattmoor: misread here as well, thanks for clarifying 🤦

mattmoor (commented):
/lgtm

PR #2443, opened 2020-09-20
Partially revert https://github.com/knative/test-infra/pull/2440

chizhg:
The reason for https://github.com/knative-sandbox/net-contour/pull/265 is some pods were being terminated, and `kubectl wait` command will get stuck on waiting on these pods. I have tried to rule out the terminating pods with `kubectl wait pod --for=condition=Ready -n knative-serving -l '!job-name' --field-selector status.phase!=Terminating --timeout=5m` command but it did not work either. So let's partially revert https://github.com/knative/test-infra/pull/2440 (the `wait_until_batch_job_complete` function should not have such issue).

/cc @mattmoor 
FYI @chaodaiG

### DIFF
--- a/scripts/library.sh
+++ b/scripts/library.sh
@@ -127,13 +127,50 @@ function wait_until_object_does_not_exist() {
 # Waits until all pods are running in the given namespace.
 # Parameters: $1 - namespace.
 function wait_until_pods_running() {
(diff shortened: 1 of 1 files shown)


mattmoor (commented):
/lgtm

PR #2444, opened 2020-09-22
kntest kubetest2 supports --extra-gcloud-flags

chizhg:

Fixes #

/cc
/hold


albertomilan:
We have discussed offline the reasoning behind the change and decided the flexibility obtained from it covers the break of the encapsulation principle, especially given the scope and usage of the tool.

/lgtm

PR #2459, opened 2020-09-23
Add unit tests for main.go. Increases coverage from 20% to 46%.

joshua-bone:
/lint
/ok-to-test

What this PR does, why we need it:
Adds unit tests for nearly all functions in main.go.

### DIFF
--- a/tools/config-generator/main_test.go
+++ b/tools/config-generator/main_test.go
@@ -541,7 +541,7 @@ func TestParseSection(t *testing.T) {
 	title := "pet-store"
 	dogs := []interface{}{
 		yaml.MapSlice{
-			yaml.MapItem{Key: "Spot", Value: "Dalmation"},
+			yaml.MapItem{Key: "Spot", Value: "Dalmatian"},
 			yaml.MapItem{Key: "Fido", Value: "Terrier"},
 		},
 		yaml.MapSlice{
@@ -564,7 +564,7 @@ func TestParseSection(t *testing.T) {
 	parseSection(config, title, generate, finalize)
 
 	expected := []string{
-		"pet-store, dogs, Spot, Dalmation",
+		"pet-store, dogs, Spot, Dalmatian",
 		"pet-store, dogs, Fido, Terrier",
 		"pet-store, dogs, Remy, Retriever",
 		"pet-store, cats, Whiskers, Calico",
@@ -726,7 +726,7 @@ func TestStringArrayFlagSet(t *testing.T) {
 func TestParseJob(t *testing.T) {
 	SetupForTesting()
 	dogs := yaml.MapSlice{
-		yaml.MapItem{Key: "Spot", Value: "Dalmation"},
+		yaml.MapItem{Key: "Spot", Value: "Dalmatian"},
 		yaml.MapItem{Key: "Fido", Value: "Terrier"},
 	}
 	cats := yaml.MapSlice{
@@ -739,7 +739,7 @@ func TestParseJob(t *testing.T) {
 	}
 
 	out := parseJob(pets, "dogs")
-	expected := "[{Spot Dalmation} {Fido Terrier}]"
+	expected := "[{Spot Dalmatian} {Fido Terrier}]"
 	if diff := cmp.Diff(fmt.Sprintf("%v", out), expected); diff != "" {
 		t.Errorf("ParseJob did not return expected slice. (-got +want)\n%s", diff)
 	}
@@ -754,7 +754,7 @@ func TestParseGoCoverageMap(t *testing.T) {
 	SetupForTesting()
 	dogs := []interface{}{
 		yaml.MapSlice{
-			yaml.MapItem{Key: "Spot", Value: "Dalmation"},
+			yaml.MapItem{Key: "Spot", Value: "Dalmatian"},
 			yaml.MapItem{Key: "Fido", Value: "Terrier"},
 		},
 		yaml.MapSlice{
--- a/tools/config-generator/main_test.go
+++ b/tools/config-generator/main_test.go
@@ -397,10 +397,10 @@ func TestParseBasicJobConfigOverrides(t *testing.T) {
 
 	repoName := "foo_repo"
 	repositories = []repositoryData{
-		{Name: repo_name, EnablePerformanceTests: false},
+		{Name: repoName, EnablePerformanceTests: false},
 	}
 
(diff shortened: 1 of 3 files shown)


chizhg:
Hmm you are still not a member of Knative? Could you check https://github.com/orgs/knative/invitation to see if there is an unaccepted invitation?

joshua-bone:
> Hmm you are still not a member of Knative? Could you check https://github.com/orgs/knative/invitation to see if there is an unaccepted invitation?

It doesn't show anything :(

chizhg:
> > Hmm you are still not a member of Knative? Could you check https://github.com/orgs/knative/invitation to see if there is an unaccepted invitation?
> 
> It doesn't show anything :(

Check it again? :)

joshua-bone:
> Check it again? :)
Accepted!

tools/config-generator/main_test.go
```diff
@@ -43,3 +54,869 @@ func TestOutputConfig(t *testing.T) {
...
+		t.Errorf("Unexpected path alias: (-got +want)\n%s", diff)
+	}
+
+	nonPathAliasRepos.Insert("foo/subrepo")
+	out = newbaseProwJobTemplateData("foo/subrepo")
+	if diff := cmp.Diff(out.PathAlias, ""); diff != "" {
+		t.Errorf("Unexpected path alias: (-got +want)\n%s", diff)
+	}
+
+	// don't pollute the global setup
```
chaodaiG, line 93: Wish the global things are gone :)

tools/config-generator/main_test.go
```diff
@@ -43,3 +54,869 @@ func TestOutputConfig(t *testing.T) {
...
+	nonPathAliasRepos.Delete("foo/subrepo")
+}
+
+func TestCreateCommand(t *testing.T) {
+	SetupForTesting()
+	preCommand = "" // global
+	in := baseProwJobTemplateData{Command: "foo", Args: []string{"bar", "baz"}}
+	out := createCommand(in)
+	expected := "[foo bar baz]"
+	if diff := cmp.Diff(fmt.Sprintf("%v", out), expected); diff != "" {
```
chaodaiG, line 104: `cmp.Diff` can compare slice too, can do you
```
expected := []string{"foo", "bar", "baz"}
if diff := cmp.Diff(expected, out)
```
instead?

chaodaiG: There are a few other cases like this down below as well

tools/config-generator/main_test.go
```diff
@@ -43,3 +54,869 @@ func TestOutputConfig(t *testing.T) {
...
+	}
+	expectedVolumes = []string{
+		"- name: foo",
+		"  secret:",
+		"    secretName: foo",
+		"  bar",
+		"  baz",
+	}
+	for i := range expectedVolumes {
+		if diff := cmp.Diff(job.Volumes[i], expectedVolumes[i]); diff != "" {
```
chaodaiG, line 230: I think you can directly compare slices as well

tools/config-generator/main_test.go
```diff
@@ -43,3 +54,869 @@ func TestOutputConfig(t *testing.T) {
...
+			t.Errorf("Unexpected volume: (-got +want)\n%s", diff)
+		}
+	}
+}
+
+func TestConfigureServiceAccountForJob(t *testing.T) {
+	SetupForTesting()
+	job := baseProwJobTemplateData{ServiceAccount: ""}
+	configureServiceAccountForJob(&job)
+	if logFatalCalls != 0 || len(job.Volumes) != 0 {
```
chaodaiG, line 240: I'm a little concerned about this global variable used for testing, if all these tests are running in parallel, would this still be valid?

joshua-bone: Great observation. I would definitely say these tests are vulnerable to race conditions if run in parallel. It's not just this global variable, it's all of them. That might be another good argument for rewriting the entire tool.

tools/config-generator/main_test.go
joshua-bone: Done

tools/config-generator/main_test.go
joshua-bone: Done

tools/config-generator/main_test.go
```diff
@@ -43,3 +54,849 @@ func TestOutputConfig(t *testing.T) {
...
+func TestAddMonitoringPubsubLabelsToJob(t *testing.T) {
+	SetupForTesting()
+	job := baseProwJobTemplateData{}
+	addMonitoringPubsubLabelsToJob(&job, "foobar")
+	expected := []string{
+		"prow.k8s.io/pubsub.project: knative-tests",
+		"prow.k8s.io/pubsub.topic: knative-monitoring",
+		"prow.k8s.io/pubsub.runID: foobar",
+	}
+	for i := range expected {
```
chaodaiG, line 173: Can this be replaced by a single cmp.Diff instead of for loop?

tools/config-generator/main_test.go
```diff
@@ -43,3 +54,849 @@ func TestOutputConfig(t *testing.T) {
...
+	content := []string{"bar", "baz"}
+
+	job := baseProwJobTemplateData{}
+	isSecret := false
+	addVolumeToJob(&job, mountPath, name, isSecret, content)
+	expectedVolumeMounts := []string{
+		"- name: foo",
+		"  mountPath: somePath",
+	}
+	for i := range expectedVolumeMounts {
```
chaodaiG, line 193: ditto

tools/config-generator/main_test.go
```diff
@@ -43,3 +54,849 @@ func TestOutputConfig(t *testing.T) {
...
+		if diff := cmp.Diff(job.VolumeMounts[i], expectedVolumeMounts[i]); diff != "" {
+			t.Errorf("Unexpected volume mount: (-got +want)\n%s", diff)
+		}
+	}
+	expectedVolumes := []string{
+		"- name: foo",
+		"  bar",
+		"  baz",
+	}
+	for i := range expectedVolumes {
```
chaodaiG, line 203: ditto

tools/config-generator/main_test.go
joshua-bone: Done

tools/config-generator/main_test.go
joshua-bone: Done

tools/config-generator/main_test.go
joshua-bone: Done

PR #2467, opened 2020-09-28
Add eventing-ceph nightlies and auto-releases

lberk:
An oversight from earlier, lets get this landed for the release


n3wscott (commented):
/lgtm

PR #2470, opened 2020-09-29
Add cloudrun-run-events@knative-releases service account to boskos

peterfeifanchen:
As per eventing needs, we are adding another cloud-run-events service account to all the boskos project for knative-releases project.


grac3gao (commented):
/lgtm

PR #2477, opened 2020-10-02
Do not upgrade all dependencies with --upgrade

chizhg:
1. Upgrading all dependencies to the latest brings in side effects, as other repos that depend on `test-infra` could rely on dependencies in older versions, and none of other Knative repos is having the same behavior. Remove the logic when `./hack/update-deps.sh --upgrade` is run.
2. Pin `github.com/docker/docker` to `v1.4.2`.

/cc @rhuss


rhuss:
@chizhg Sorry, I totally overlooked that PR. Thanks a lot for taking care ! Just curious: Why have you closed it and what would be the alternative solution ?

chizhg:
@rhuss no worries, the reason I closed this PR is because Scott made another PR to not update the dependencies outside of Knative. But `github.com/docker/docker` is still pinning on an incompatible version, I will create another PR to pin it to `v1.4.2`.

PR #2486, opened 2020-10-08
Add --platform=all to our default KO_FLAGS.

mattmoor:
Serving has been canarying this for a bit, this rolls it out everywhere.


n3wscott:
ARM64 is going to be really important very very soon. People are already asking for rPI builds of Knative.

/lgtm
/approve

PR #2487, opened 2020-10-08
Adding go module parsing to understand which deps are needed.

n3wscott:

This adds go module parsing to understand which deps are needed.

/lint


pkg/gomod/modules_test.go
```diff
@@ -0,0 +1,192 @@
...
+*/
+
+package gomod
+
+import (
+	"testing"
+)
+
+func TestModule(t *testing.T) {
+	tests := map[string]struct {
```
vagababov, line 24: please don't do `map` just fold the name into the struct
```
tests := []struct {
 name string..
...
}
```

You'll also get better formatting of the cases.

pkg/gomod/modules_test.go
```diff
@@ -0,0 +1,192 @@
...
+				if want := tt.wantDeps[i]; got != want {
+					t.Errorf("unexpected dep[%d]; got %q, want: %q", i, got, want)
+				}
+			}
+		})
+	}
+}
+
+func TestModules(t *testing.T) {
+	tests := map[string]struct {
```
vagababov, line 105: Ditto.

pkg/gomod/modules_test.go
```diff
@@ -0,0 +1,192 @@
...
+				if want, found := tt.wantPkgs[name]; !found {
+				} else if len(gots) != len(want) {
+					t.Errorf("unexpected pkgs[%q].dep length mismatch; got %v, want: %v", name, gots, want)
+					continue
+				} else {
+					for i, got := range gots {
+						if want := tt.wantPkgs[name][i]; got != want {
+							t.Errorf("unexpected pkgs[%q].dep[%d]; got %q, want: %q", name, i, got, want)
+						}
+					}
```
vagababov, line 178: do you want to just sort and cmp.Equal? There's also a way where you can write a custom comparator to sort on the fly in the cmp.

pkg/gomod/modules.go
```diff
@@ -0,0 +1,82 @@
...
+
+import (
+	"errors"
+	"io/ioutil"
+	"strings"
+
+	"golang.org/x/mod/modfile"
+)
+
+// Modules returns a map of given given modules to their dependencies, and a
```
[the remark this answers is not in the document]
n3wscott: hmmm I did not look at this.

looks like I can tell if it is a Direct/Indirect dep:

```go
// A Require is a single require statement.
type Require struct {
	Mod      module.Version
	Indirect bool // has "// indirect" comment
	Syntax   *Line
}
```

Likely we should ignore indirect deps. What do you think?

pkg/gomod/modules.go
```diff
@@ -0,0 +1,82 @@
...
+			if _, seen := cache[pkg]; seen {
+				continue
+			}
+			cache[pkg] = true
+			dependencies = append(dependencies, pkg)
+		}
+	}
+
+	return packages, dependencies, nil
+}
```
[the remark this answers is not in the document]
n3wscott: Sets are not part of the standardlib, I was writing this originally as an independent cli with no real dep on k8s. But not if I am in testinfra, it is no big deal.

pkg/gomod/modules.go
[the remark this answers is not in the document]
vagababov: The only issue if we depend on a version and dep depends on a different version but I am not sure how we deal with it now. So 🤷

pkg/gomod/modules_test.go
n3wscott: strong disagree

pkg/gomod/modules_test.go
n3wscott: ditto disagree

pkg/gomod/modules_test.go
vagababov: Tell me how i this better?

pkg/gomod/modules_test.go
n3wscott: oh! yes. thanks.

pkg/gomod/modules.go
```diff
@@ -0,0 +1,82 @@
...
+	if len(domain) == 0 {
+		return "", nil, errors.New("no domain provided")
+	}
+
+	b, err := ioutil.ReadFile(gomod)
+	if err != nil {
+		return "", nil, err
+	}
+
+	file, err := modfile.Parse(gomod, b, nil)
```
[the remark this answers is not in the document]
n3wscott: added a comment

pkg/gomod/modules.go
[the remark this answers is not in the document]
n3wscott: I am going to exclude indirect deps for now, likely if you update an indirect dep, you will break something unexpected.

vagababov (commented):
/assign

pkg/gomod/modules_test.go
n3wscott: this requires that each name of each test is unique, giving you a signal you did not just copy/paste a row in the table.

vagababov: Literally never happened to me :-)

vagababov (commented):
For posterity strongly disagree with map for tests.
Duplicate names is a weak argument against much harder typing and significantly worse formatting and (it's test, but still) worse memory usage 👅 
📦 🐬 
/lgtm

PR #2488, opened 2020-10-08
This adds functions that check and calculate refs for best deps based on a release.

n3wscott:

This adds functions that check and calculate refs for best deps based on a release.

/lint

### DIFF
--- a/pkg/gomod/check.go
+++ b/pkg/gomod/check.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package gomod
 
 import (
--- a/pkg/gomod/check.go
+++ b/pkg/gomod/check.go
@@ -65,11 +65,11 @@ func check(module string, packages []string, release string, ruleset git.Ruleset
 		case git.NoRef:
 			nonReady = append(nonReady, ref)
 			if out != nil {
-				_, _ = fmt.Fprintf(out, "✘ %s\n", ref)
+				_, _ = fmt.Fprintln(out, "✘ ", ref)
 			}
 		default:
 			if out != nil {
-				_, _ = fmt.Fprintf(out, "✘ %s\n", ref)
+				_, _ = fmt.Fprintln(out, "✔ ", ref)
 			}
 		}
 	}
--- a/pkg/gomod/check.go
+++ b/pkg/gomod/check.go
@@ -27,11 +27,7 @@ import (
 )
 
 // Check examines a go mod file for dependencies and  determines if each have a release artifact
-// based on the ruleset provided.
-//
-// See Also
-//
-// Check leverages the same rules used by
+// based on the ruleset provided. Check leverages the same rules used by
 // knative.dev/test-infra/pkg/git.Repo().BestRefFor
 func Check(gomod, release, domain string, ruleset git.RulesetType, out io.Writer) error {
 	modulePkgs, _, err := Modules([]string{gomod}, domain)
--- a/pkg/gomod/check.go
+++ b/pkg/gomod/check.go
@@ -18,8 +18,9 @@ package gomod
 
 import (
 	"fmt"
-	"github.com/blang/semver/v4"
 	"io"
+
+	"github.com/blang/semver/v4"
 
 	"knative.dev/test-infra/pkg/git"
 )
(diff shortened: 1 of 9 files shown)


pkg/gomod/check.go
```diff
@@ -0,0 +1,112 @@
...
+
+	"github.com/blang/semver/v4"
+
+	"knative.dev/test-infra/pkg/git"
+)
+
+// Check examines a go mod file for dependencies and  determines if each have a release artifact
+// based on the ruleset provided.
+//
+// See Also
```
vagababov, line 31: see also what? :)

pkg/gomod/check.go
```diff
@@ -0,0 +1,112 @@
...
+
+func check(module string, packages []string, release string, ruleset git.RulesetType, verbose bool) error {
+	this, err := semver.ParseTolerant(release)
+	if err != nil {
+		return err
+	}
+
+	if verbose {
+		fmt.Printf("%s\n", module)
+	}
```
vagababov, line 57: if debug remove
if logging, let's use `log.Println`
By default it goes to stderr, so you'll see on screen

pkg/gomod/check.go
```diff
@@ -0,0 +1,112 @@
...
+func (e *Error) Is(target error) bool {
+	_, is := target.(*Error)
+	return is
+}
+
+// Error implements error.Error()
+func (e *Error) Error() string {
+	return fmt.Sprintf("%s failed because of the following dependencies [%s]",
+		e.Module,
+		strings.Join(e.Dependencies, ", "))
```
vagababov, line 110: just %v will probably print [a, b, c]

pkg/gomod/modules.go
```diff
@@ -84,3 +87,17 @@ func Module(gomod string, domain string) (string, []string, error) {
 
 	return file.Module.Mod.Path, packages.List(), nil
 }
+
+func moduleToRepo(module string) (*git.Repo, error) {
+	url := fmt.Sprintf("https://%s?go-get=1", module)
+	meta, err := golang.GetMetaImport(url)
+	if err != nil {
+		return nil, fmt.Errorf("unable to fetch go import %s: %v", url, err)
```
vagababov, line 95:
```suggestion
		return nil, fmt.Errorf("unable to fetch go import %s: %w", url, err)
```

pkg/gomod/modules.go
```diff
@@ -84,3 +87,17 @@ func Module(gomod string, domain string) (string, []string, error) {
...
+	}
+
+	if meta.VCS != "git" {
+		return nil, fmt.Errorf("unknown VCS: %s", meta.VCS)
```
vagababov, line 99:
```suggestion
		return nil, errros.New("unknown VCS: "+meta.VCS)
```

pkg/gomod/check.go
n3wscott: it makes a godoc header: https://medium.com/@elliotchance/godoc-tips-tricks-cda6571549b

pkg/gomod/check.go
n3wscott: I don't want it to go to stderr

pkg/gomod/check.go
n3wscott: true, I can update.

pkg/gomod/modules.go
n3wscott: I don't understand why you like these changes so much. I think this is worse to read and implement

pkg/gomod/check.go
n3wscott: I don't want `log`, I want `fmt`. Log adds extra stuff to the prefix

pkg/gomod/modules.go
vagababov: less state. straight left to right )

pkg/gomod/check.go
vagababov: Does it work with fucntion level headers or only on package level?

pkg/gomod/check.go
```diff
@@ -0,0 +1,113 @@
...
+		ref, refType := repo.BestRefFor(this, ruleset)
+		switch refType {
+		case git.NoRef:
+			nonReady = append(nonReady, ref)
+			if out != nil {
+				_, _ = fmt.Fprintf(out, "✘ %s\n", ref)
+			}
+		default:
+			if out != nil {
+				_, _ = fmt.Fprintf(out, "✘ %s\n", ref)
```
vagababov, line 75: check here, not x.

pkg/gomod/check.go
n3wscott: it turns out it prints `[a b c]` which is close but not what I want.

pkg/gomod/check.go
n3wscott: good call, moved to a `io.Writer` and the caller can choose.

pkg/gomod/check.go
n3wscott: in my ide, the preview looks correct with the header :idk:

n3wscott: maybe not. I will drop it.

pkg/gomod/check.go
n3wscott: woops!!! thanks

vagababov (commented):
/lgtm

PR #2489, opened 2020-10-08
adding buoy cli

n3wscott:

This adds the buoy cli. See the readme for details.

/lint

### DIFF
--- a/buoy/buoy.go
+++ b/buoy/buoy.go
@@ -24,6 +24,6 @@ import (
 
 func main() {
 	if err := commands.New().Execute(); err != nil {
-		log.Fatal("error during command execution: ", err)
+		log.Fatal("Error during command execution: ", err)
 	}
 }
--- a/buoy/buoy.go
+++ b/buoy/buoy.go
@@ -1,0 +1,29 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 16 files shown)


n3wscott:
/hold

Gonna add some tests.

buoy/commands/check.go
```diff
@@ -0,0 +1,91 @@
...
+			gomodFile := args[0]
+
+			var out io.Writer
+			if verbose {
+				out = os.Stdout
+			}
+
+			err := gomod.Check(gomodFile, release, domain, ruleset, out)
+			if errors.Is(err, gomod.DependencyErr) {
+				_, _ = fmt.Fprintln(os.Stderr, err.Error())
```
vagababov, line 74:
```suggestion
				fmt.Fprintln(os.Stderr, err.Error())
```
do you really need those?

buoy/commands/check.go
```diff
@@ -0,0 +1,91 @@
...
+				os.Exit(1)
+				return nil
```
vagababov, line 76: do you need the `return nil`. One would think go would catch that.

buoy/commands/check.go
n3wscott: if you don't you get warnings

n3wscott:
/unhold

I tested check and float, but not via the `cli`. I think it is good enough. If we really wanted to we could run the same tests and e2e and install the cli locally....  maybe a followup

buoy/buoy.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"log"
+
+	"knative.dev/test-infra/buoy/commands"
+)
+
+func main() {
+	if err := commands.New().Execute(); err != nil {
+		log.Fatal("error during command execution: ", err)
```
vagababov, line 27:
```suggestion
		log.Fatal("Error during command execution: ", err)
```
this is end user message so a complete sentence.

vagababov (commented):
Just a nit

vagababov (commented):
/lgtm

PR #2492, opened 2020-10-09
introduce a bash entry point for update deps

n3wscott:
This will standardize how knative upses update deps from test-infra.

### DIFF
--- a/scripts/library.sh
+++ b/scripts/library.sh
@@ -519,6 +519,8 @@ function add_trap {
 #   "--release <version>" used with upgrade. The release version to upgrade
 #                         Knative components. ex: --release v0.18. Defaults to
 #                         "master".
+# Additional dependencies can be included in the upgrade by providing them in a
+# global env var: FLOATING_DEPS
 function go_update_deps() {
   cd "${REPO_ROOT_DIR}" || return 1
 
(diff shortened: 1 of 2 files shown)


scripts/library.sh
```diff
@@ -513,6 +513,63 @@ function add_trap {
...
+      --upgrade) UPGRADE=1 ;;
+      --release) shift; VERSION="$1" ;;
+      *) abort "unknown option ${parameter}" ;;
+    esac
+    shift
+  done
+
+  if (( UPGRADE )); then
+    echo "--- Upgrading to ${VERSION}"
+    FLOATING_DEPS=( $(run_go_tool knative.dev/test-infra/buoy buoy float ${REPO_ROOT_DIR}/go.mod --release ${VERSION} --domain knative.dev) )
```
mattmoor, line 544: we need the ability to add non-knative things here.

n3wscott:
/hold

We need to also support additional floating deps.

scripts/library.sh
mattmoor: I don't mean to buoy, but to FLOATING_DEPS

n3wscott: ok contour will have to do this:

(update_deps.sh):

```
source $(dirname "$0")/../scripts/library.sh

CONTOUR_VERSION="v1.9.0" # This is for controlling which version of contour we want to use.

FLOATING_DEPS=(
  "github.com/projectcontour/contour@${CONTOUR_VERSION}"
)

go_update_deps "$@"
```

n3wscott:
/unhold

Now we `+=` the floating deps, so you can pass a base set in to start with for non-knative floating deps.

mattmoor (commented):
/lgtm

PR #2494, opened 2020-10-09
delete anything we know about that should not be vendored

n3wscott:
This is to support eventing-rabbitmq to move to `go_update_deps`.


n3wscott:
> /lgtm

ooops, I changed what was deleted but needed to also drop them in the vendored files here :D

PR #2502, opened 2020-10-13
Forcing go-containerregistry

### LINKED ISSUE #2497: Bump go-containerregistry to v0.1.3
I cleaned up my local mod cache and now i cannot update **any** knative project anymore because there is `gitea` transitive dependency on test-infra that is broken.

Because of the circular dependency on pkg, even if I try to use the replace directives pointing to my local code, It still tries to download gitea:

```
=== Update Deps for Golang
--- Go mod tidy and vendor
go: k8s.io/test-infra@v0.0.0-20200803112140-d8aa4e063646 requires
	knative.dev/pkg@v0.0.0-20200711004937-22502028e31a requires
	knative.dev/test-infra@v0.0.0-20201009204121-322fb08edae7 requires
	github.com/google/go-containerregistry@v0.1.1 requires
	github.com/goreleaser/goreleaser@v0.136.0 requires
	code.gitea.io/sdk/gitea@v0.12.0: reading code.gitea.io/sdk/gitea/gitea/go.mod at revision gitea/v0.12.0: unknown revision gitea/v0.12.0
```

With a bump of `go-containerregistry` to latest version this should be solved because it doesn't bring in anymore gitea (via `goreleaser`)

```shell
% go get -v -d github.com/google/go-containerregistry@v0.1.3

% ./hack/update-deps.sh
```

slinkydeveloper:
Signed-off-by: Francesco Guardiani <francescoguard@gmail.com>

Fix #2497, from https://github.com/knative/pkg/pull/1802#discussion_r503782059

This is required because the dep cycle is still importing the old version of this package.


slinkydeveloper:
cc @markusthoemmes

markusthoemmes:
I don't think this is honored by downstream releases depending on this. You'll have to set the replace where it's actually used IIRC (we do the same for K8s deps everywhere too).

slinkydeveloper:
@markusthoemmes i think what we need to do is just break the cycle: we update it here, then we bump test-infra in pkg and this dep should disappear completely (and we can then remove this replace too)

markusthoemmes:
Ah, got you now! Makes sense.

slinkydeveloper:
@markusthoemmes can you approve this?

markusthoemmes (approved):
/lgtm
/approve

PR #2504, opened 2020-10-13
Remove the replace

matzew:
Signed-off-by: Matthias Wessendorf <mwessend@redhat.com>

remove the unneeded replace for `go-containerregistry`


matzew:
/close

nope, not really what we thought

PR #2508, opened 2020-10-13
Clean up prow-tests image a bit

chizhg:
1. Do not add `scripts` in the image
2. Remove `githubhelper` tool as scheduled since release-0.18 has been cut

### DIFF
--- a/images/prow-tests/Dockerfile
+++ b/images/prow-tests/Dockerfile
@@ -96,20 +96,12 @@ COPY images/prow-tests/runner.sh /usr/local/bin
 # Install our own tools
 RUN GO111MODULE=on go get knative.dev/pkg/testutils/junithelper@release-0.17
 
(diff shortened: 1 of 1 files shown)


images/prow-tests/Dockerfile
```diff
@@ -96,20 +96,12 @@ COPY images/prow-tests/runner.sh /usr/local/bin
 # Install our own tools
 RUN GO111MODULE=on go get knative.dev/pkg/testutils/junithelper@release-0.17
 
-# And our scripts
-ENV SCRIPTS /scripts
-COPY scripts /scripts
```
coryrc, line 101: I want to keep this, but it just reminded me to bring it up as another option

coryrc: Ah, nevermind, they aren't going to want it.

PR #2509, opened 2020-10-13
Add Sample Source to release/nightlies

lberk:
/lint

After trying to use sample-source as an example in environments where the `ko` tool wasn't available, it'd be nice to have upstream images of the controller and receive adapter.  This is also true of wanting to see the state of sample-source over the course of different knative versions.


n3wscott (commented):
/lgtm
/approve

PR #2511, opened 2020-10-14
Add ProwJobs annotations field

joshua-bone:
/lint

Add ProwJobs annotations field.


chaodaiG:
Generally lgtm. Don't have a clue why presubmit check failed

/cc @chizhg 
Please also take a look

tools/config-generator/testgrid_config.go
```diff
@@ -168,37 +170,75 @@ func getGcsLogDir(testGroupName string) string {
...
 }
 
+func getTestgroupExtras(projName, jobName string) map[string]string {
+	extras := make(map[string]string)
+	switch jobName {
+	case "continuous":
+		// projName has the release encoded into it, so the main page at http://testgrid.knative.dev
+		// does not mix releases with the master branch
+		if releaseRegex.FindString(projName) != "" {
+			extras["num_failures_to_alert"] = "3"
```
chizhg, line 180: I think you need to quote these numbers like `"\"3\""`, this seems to be the reason that the Prow job config check job is failing.

tools/config-generator/testgrid_config.go
```diff
@@ -168,37 +170,75 @@ func getGcsLogDir(testGroupName string) string {
...
+	}
+	v, ok = tgExtras["alert_options"]
+	if ok {
+		email := quotedEmailPattern.FindStringSubmatch(v)[1] //index 1 is first capture group
+		res := fmt.Sprintf("  testgrid-alert-email: \"%s\"", email)
+		annotations = append(annotations, res)
+	}
+	v, ok = tgExtras["num_failures_to_alert"]
+	if ok {
+		res := "  testgrid-num-failures-to-alert: " + v
```
chizhg, line 225: Also quote the number here.

tools/config-generator/testgrid_config.go
joshua-bone: Unquoted is correct within this method since it writes the Testgrid YAML ([example](https://github.com/knative/test-infra/blob/8f5dc4814ac37f9d504bd6872673a7d1bdbb0f5d/config/prod/prow/testgrid/testgrid.yaml#L55), but quoted numbers are required below in the generateProwJobAnnotations method and have been updated.

tools/config-generator/testgrid_config.go
joshua-bone: Done, thanks!

chaodaiG:
Cool, let's get it merged

/lgtm
/approve

PR #2517, opened 2020-10-19
Bump golang default to 1.14

markusthoemmes:

Nothing in Knative should be using Golang 1.13 explicitly at this point. Having 1.14 as the default seems reasonable (and will fix my problem with client-contrib still running 1.13 and breaking on a dependency upgrade).

### DIFF
--- a/images/prow-tests/runner.sh
+++ b/images/prow-tests/runner.sh
@@ -18,8 +18,8 @@ ORIGINAL_GOPATH="${GOPATH}"
 
 source "${HOME}/.gvm/scripts/gvm"
 
-# By default using Go version 1.13.
-version="go1.13"
+# By default using Go version 1.14.
+version="go1.14"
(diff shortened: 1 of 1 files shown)


coryrc:
Per discussion in slack, changing the default might break the oldest releases of client, or at best, be using a different version than intended.

Trying a fake go.mod which fits [this regex](https://github.com/knative/test-infra/pull/2517/files#diff-b7c2ca8471f7f246757a0922310fd618ac61fd616e3e04ab2b7506c5ce72bd8eR25) might make Go too angry or it will work.

coryrc:
Otherwise adding GO_VERSION env var to the Prow job will do it (can't be in-repo, must be in the prow job). You will have to touch [this thing](https://github.com/knative/test-infra/tree/master/tools/config-generator) if so, I'm so sorry.

markusthoemmes:
Fixed as described.

PR #2519, opened 2020-10-19
Clean up prow-tests image

coryrc:
Using builder pattern reduced size by 2.6 GB

Cleaned up unused markdown link linter and deprecated it.

I don't think updating gcloud is necessary; lets see what the beta tests say.

/assign @chizhg 
/cc @chizhg

### DIFF
--- a/images/prow-tests/Dockerfile
+++ b/images/prow-tests/Dockerfile
@@ -18,11 +18,8 @@ ENV DEBIAN_FRONTEND noninteractive
 # If we don't run update, the apt-get install below don't work
 RUN apt-get update
 
-# TODO: do we have to do this?
-# https://github.com/kubernetes/test-infra/blob/master/images/bootstrap/Dockerfile#L67
-# Is where it is installed, using rapid channel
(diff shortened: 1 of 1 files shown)


images/prow-tests/Dockerfile
```diff
@@ -12,13 +12,17 @@
...
 
 ENV DEBIAN_FRONTEND noninteractive
+# If we don't run update, the apt-get install below don't work
 RUN apt-get update
-RUN gcloud components update
+
+# TODO: do we have to do this?
+# https://github.com/kubernetes/test-infra/blob/master/images/bootstrap/Dockerfile#L67
+# Is where it is installed, using rapid channel
+# RUN gcloud components update
```
chizhg, line 24: We have it because sometimes we might want to update the gcloud version without updating the Dockerfile.

coryrc: So add your response as a comment and keep the line commented out?

chizhg: SGTM

PR #2521, opened 2020-10-20
bypass the golang proxy for upgrades if GOPROXY is not set

n3wscott:

Automation that runs just after a merge to an upstream repo misses updates because golang by default uses the global proxy.

This will bypass the default `GOPROXY` and use `direct` if `GOPROXY` is not set, allowing automation to quickly update downstream dependencies.

### DIFF
--- a/scripts/library.sh
+++ b/scripts/library.sh
@@ -555,7 +555,7 @@ function go_update_deps() {
       echo "Respecting 'GOPROXY=${GOPROXY}'."
     fi
     FLOATING_DEPS+=( $(run_go_tool knative.dev/test-infra/buoy buoy float ${REPO_ROOT_DIR}/go.mod --release ${VERSION} --domain knative.dev) )
-    if [[ ${FLOATING_DEPS[@]} ]]; then
+    if [[ ${#FLOATING_DEPS[@]} > 0 ]]; then
       echo "Floating deps to ${FLOATING_DEPS[@]}"
       go get -d ${FLOATING_DEPS[@]}
     else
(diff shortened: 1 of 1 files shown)


scripts/library.sh
```diff
@@ -543,6 +544,12 @@ function go_update_deps() {
 
   if (( UPGRADE )); then
     echo "--- Upgrading to ${VERSION}"
+    if [ -z ${GOPROXY+x} ]; then
```
mattmoor, line 547: What is `+x`?

n3wscott: ${var+x} is a parameter expansion which evaluates to nothing if var is unset, and substitutes the string x otherwise.

mattmoor: This could absolutely use a comment 🤣

n3wscott: Added.

n3wscott:
/hold

Let me check a downstream.

n3wscott:
/unhold

Works on test infra with no deps, and eventing with deps.

scripts/library.sh
```diff
@@ -541,10 +542,20 @@ function go_update_deps() {
     shift
   done
 
-  if (( UPGRADE )); then
+  if [[ $UPGRADE == 1 ]]; then
     echo "--- Upgrading to ${VERSION}"
+    # From shell parameter expeansion:
```
mattmoor, line 547:
```suggestion
    # From shell parameter expansion:
```

mattmoor (approved):
/lgtm

PR #2526, opened 2020-10-23
pin github.com/docker/docker to an v1.4.2

chizhg:
Though we are planning to move the bash scripts to a separate repo, let's pin `github.com/docker/docker` to a version that is compatible with `knative/client` to get a quick fix, as discussed in https://github.com/knative/test-infra/pull/2477#issuecomment-712719799

/cc @rhuss


rhuss (commented):
Thanks, that looks good to me ! But to be clear, it's not really urgent as the client itself does not have a Docker dependency (but some non-official side projects do have this).

Please unhold if you still want to merge.

/lgtm
/hold
/approve

chizhg:
We don't need it anymore since we have moved the test scripts to knative.dev/hack

PR #2528, opened 2020-10-23
pin to code-generator 18.8

n3wscott:
Current ci is failing:

```
go get: knative.dev/test-infra@v0.0.0-20201022003632-7d3ea6019158 requires
	github.com/google/go-containerregistry@v0.1.3 requires
	k8s.io/code-generator@v0.17.2 requires
	gonum.org/v1/netlib@v0.0.0-20190331212654-76723241ea4e requires
	modernc.org/strutil@v1.0.0: reading modernc.org/strutil/go.mod at revision v1.0.0: unknown revision v1.0.0
```

Because `github.com/google/go-containerregistry` deps on `k8s.io/code-generator` but also uses a replace to lock the dep to 0.18.8. Go mod does not bring those locks along, so we need to also pin our `k8s.io/code-generator` to avoid  the above error.


antoineco (approved):
/lgtm

Issue #2063, opened 2020-05-13
Helper utility for running testing from docker


coryrc:
Any developer should be able to easily run tests on their local machine as if they were running in Prow (same tool versions, etc). We can do this by making it easy to run the test flows in docker.

/assign @coryrc

chizhg:
Implemented in https://github.com/knative/test-infra/pull/2531

PR #2531, opened 2020-10-26
Add a rundk tool to help reproduce the same test flows as run in CI

chizhg:
1. Add a `rundk` tool to help reproduce the same test flows as run in CI
2. Update it to use `CopyAndAddMount` instead of `mount` to make the tool also being able to run on MacOS

#2063


chizhg:
/cc @coryrc @albertomilan 
FYI @zhongduo

zhongduo:
IIUC, this is trying to add a wrapper of the `docker run` command. But from my experience, the main complexity seems in the project setup and service account creation and role binding, as well as the `IS_PROW` check. Is there any plan to make that easier too?

chizhg:
> IIUC, this is trying to add a wrapper of the `docker run` command. But from my experience, the main complexity seems in the project setup and service account creation and role binding, as well as the `IS_PROW` check. Is there any plan to make that easier too?

For the project setup and service account creation and role binding, https://github.com/knative/test-infra/blob/master/config/prod/build-cluster/boskos/set_up_boskos_project.sh is the script we run for our CI and it incorporates everything for all the Knative projects. For each project, they should maintain their own docs for the related setups, e.g. https://github.com/google/knative-gcp/blob/master/test/e2e/README.md. The doc generally looks good to me but some improvements can be made there, e.g. you can update this doc to add instructions on how to run the test flow with this `rundk` tool after this PR is merged.

Also test-infra only exposes `IS_PROW` as an env vars, and it's the responsibility of each Knative project to decide how to use it. So IMO if you find `IS_PROW` check as a problem, the issue should be fixed by the project owner, in this case is `knative-gcp`.

sethvargo:
I've been able to get away with a simple shell script most of the time with our use of prow. Example:

```sh
$ docker run 

# interactive mode (so we get a tty)
-it 

# remove it when were done
--rm

# mount the source code in /src (that's where prow wants them)
-v ~/Development/knative:/src

# mount go module cache - it'll work without this, but it will download all
# dependencies each time, which is annoying
-v ~/Development/go/pkg:/go/pkg

# mount docker on the host to docker in the container - required for DinD
-v $(which docker):/usr/bin/docker
-v /var/run/docker.sock:/var/run/docker.sock

# on a mac, set to host networking to make DinD work. not required on linux
--network host

# mount gcloud credentials to get default app creds, readonly to be safe
-v ~/.config/gcloud:/root/.config/gcloud:ro

# set workdir (this might be the default, didn't test)
--workdir /src

# use our runner
my.image.url

# execute tests
runtests.sh
```

chizhg:
> I've been able to get away with a simple shell script most of the time with our use of prow. Example:
> 
> ```shell
> $ docker run 
> 
> # interactive mode (so we get a tty)
> -it 
> 
> # remove it when were done
> --rm
> 
> # mount the source code in /src (that's where prow wants them)
> -v ~/Development/knative:/src
> 
> # mount go module cache - it'll work without this, but it will download all
> # dependencies each time, which is annoying
> -v ~/Development/go/pkg:/go/pkg
> 
> # mount docker on the host to docker in the container - required for DinD
> -v $(which docker):/usr/bin/docker
> -v /var/run/docker.sock:/var/run/docker.sock
> 
> # on a mac, set to host networking to make DinD work. not required on linux
> --network host
> 
> # mount gcloud credentials to get default app creds, readonly to be safe
> -v ~/.config/gcloud:/root/.config/gcloud:ro
> 
> # set workdir (this might be the default, didn't test)
> --workdir /src
> 
> # use our runner
> my.image.url
> 
> # execute tests
> runtests.sh
> ```

Thanks @sethvargo ! Yes, the underlying implementation of this tool is basically a wrapper of a `docker` command, but it's more generic, more similar as run on Prow and easier to use (e.g. it can be run under any subdirectory of a Knative repo and it will run the test flow from the root folder of that repo).

pkg/interactive/docker.go
```diff
@@ -76,39 +74,36 @@ func (d *Docker) AddMount(typeStr, source, target string, optAdditionalArgs ...s
...
-		err = os.Mkdir(path.Join(tmpDir, d), 0777)
-		if err != nil {
-			log.Fatal(err)
-		}
+
+	tempDir, err := ioutil.TempDir(parentDir, fi.Name())
+	if err != nil {
+		return nil, fmt.Errorf("error creating a temporary directory: %w", err)
+	}
+	if err := os.Chmod(tempDir, 0777); err != nil {
```
albertomilan, line 90: Why not `0770`, given it's a directory?

pkg/interactive/docker.go
```diff
@@ -76,39 +74,36 @@ func (d *Docker) AddMount(typeStr, source, target string, optAdditionalArgs ...s
...
-	// You need empty directories for upper and work (and overlay, though you can mount over directories that have files in them if you *want* to...)
-	mount := NewCommand("sudo", "mount", "-t", "overlay", "-o",
-		fmt.Sprintf("lowerdir=%s,upperdir=%s/upper,workdir=%s/work", externalDirectory, tmpDir, tmpDir),
-		"none", overlayDir)
-	// Print command to run so user knows why it is asking for sudo password (if it does)
-	log.Println(mount)
-	if err = mount.Run(); err != nil {
-		log.Fatalf("Unable to create overlay mount, so giving up: %v", err)
+	cleanup := func() {
+		os.RemoveAll(tempDir)
```
albertomilan, line 94: Can we have some error checking control done here with a log statement?

pkg/interactive/docker.go
albertomilan: I meant `0666`

pkg/interactive/command.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Helper functions for running interactive CLI sessions from Go
+// Helper functions for running interactive docker CLI sessions from Go
```
coryrc, line 17: No, half the library is for running any commands from Go

rundk/main.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+}
+
+func run(cmd interactive.Docker, image string, commandAndArgsOpt ...string) error {
+	// Finally add the image then command to run (if any)
+	cmd.AddArgs(image)
+	cmd.AddArgs("runner.sh")
+	cmd.AddArgs(commandAndArgsOpt...)
+	log.Println(cmd)
+	log.Println("👆 Starting in 3 seconds, ^C to abort!")
```
albertomilan, line 129: If this is an important step, I would suggest to expand this to 5+ seconds. Reading the instruction will take ~1 sec already and we need to give room for the reaction from the user

pkg/interactive/docker.go
```diff
@@ -76,39 +74,36 @@ func (d *Docker) AddMount(typeStr, source, target string, optAdditionalArgs ...s
 	d.AddArgs("--mount", fmt.Sprintf("type=%s,source=%s,target=%s%s", typeStr, source, target, addl))
 }
 
-// AddRWOverlay mounts a directory into the image at the desired location, but with an overlay
-//  so internal changes do not modify the external directory.
-// externalDirectory probably needs to be an absolute path
-// Returns a function to clean up the mount (but does not delete the directory).
-// Uses sudo and probably only works on Linux
-func (d *Docker) AddRWOverlay(externalDirectory, internalDirectory string) func() {
```
coryrc, line 84: I need this function, don't remove it

coryrc: You can add a new one if you want

pkg/interactive/docker.go
```diff
@@ -76,39 +74,36 @@ func (d *Docker) AddMount(typeStr, source, target string, optAdditionalArgs ...s
...
-	tmpDir, err := ioutil.TempDir("", "overlay")
+// CopyAndAddMount copies the source files into a temp directory, and then
+// mounts them as the target. It also returns a function to remove the temp
+// directory for cleaning up.
+func (d *Docker) CopyAndAddMount(typeStr, parentDir, source, target string, optAdditionalArgs ...string) (func(), error) {
+	fi, err := os.Stat(source)
 	if err != nil {
-		log.Fatal(err)
+		return nil, fmt.Errorf("error getting the FileInfo for %q: %w", source, err)
```
coryrc, line 83: The purpose of this library is to write interactive scripts, not programs which can recover from errors and do smart things. Adding error checking makes the scripts much less focused on the task at hand and hard to parse.

If you want different functionality, put it in a different library

coryrc (changes requested):

coryrc:
/hold

Can't delete AddRWOverlay

pkg/interactive/docker.go
chizhg: Changed to `0666`.

pkg/interactive/docker.go
chizhg: Added some checking control here, but since it'll be a best-effort cleanup so its failure should not impact the final results, I'm using `warning` instead of `fatalf` log here.

rundk/main.go
chizhg: It's not a very important step, and also since users could be running this tool again and again, I feel 5+ seconds is a bit too long, so I'm inclined to keep it as 3 seconds.

pkg/interactive/command.go
chizhg: I've reverted the change here, but it's probably better to put these functions under different packages..

pkg/interactive/docker.go
chizhg: I have got the function back.

pkg/interactive/docker.go
chizhg: Hmm I'm not quite sure what makes the difference between them, and we already have another function that returns error instead of fatal - https://github.com/knative/test-infra/blob/master/pkg/interactive/docker.go#L43-L56

pkg/interactive/docker.go
coryrc: um, directories need to be executable and we're writing to it, so why isn't it `0777`?

pkg/interactive/docker.go
coryrc: Because that error is optional -- sometimes it's okay to fail and sometimes it is not.

If the overlay fails, it's unlikely you want to continue

rundk/main.go
coryrc: The purpose is: what if you just want the env setup and to use it yourself; then 3 seconds is fast enough because you'll be watching it.
Then why not wait on user input? I hate when you want to run something which takes a while, you launch it, walk away, then it doesn't start because of unnecessary user input.

pkg/interactive/docker.go
albertomilan: Cory is correct. Executable flag for directories grant permission on its content. Discard this comment.

chizhg: Yeah I just tried `0666` and it will actually throw out an error. I've reverted the change.

pkg/interactive/docker.go
chizhg: I have change it to fatal instead of returning an error.

rundk/README.md
```diff
@@ -0,0 +1,46 @@
...
+        A list of env vars separated by comma that optionally need to be set on local.
+```
+
+### Example
+
+Run E2E tests for a Knative repository:
+
+```shell
+export GOOGLE_APPLICATION_CREDENTIALS=/temp/gcloud-secret-key.json
+rundk ./test/e2e-tests.sh
```
coryrc, line 41: I don't think you should take the script name like this, instead of an argument, because if you run it from a subdirectory and use a relative path, it won't work. It also won't work with relative paths.

rundk/README.md
```diff
@@ -0,0 +1,46 @@
...
+go get knative.dev/test-infra/rundk
+```
+
+## Usage
+
+```shell
+Usage of rundk:
+  -test-image string
+        The image we use to run the test flow. (default "gcr.io/knative-tests/test-infra/prow-tests:stable")
+  -mounts string
```
coryrc, line 27: The likelihood the local directory has the same path as would be useful inside the image is unlikely, so I don't think this feature adds anything

rundk/README.md
```diff
@@ -0,0 +1,46 @@
...
+        A list of extra folders or files separated by comma that need to be mounted to run the test flow.
+  -mandatory-env-vars string
```
coryrc, line 29: These two options don't mention that the env vars get promoted to the image

rundk/README.md
coryrc: So maybe take path -> convert to realpath -> confirm under the repo root -> convert to path relative to repo root?

coryrc (changes requested):

rundk/README.md
chizhg: It will work, since in this tool we set the working directory as the root directory of the repo, as long as you run this tool in one of the repo's subdirectories.

rundk/README.md
chizhg: Users would probably want to mount other files when they run the tests? The flag is empty by default, just an extension in case users may want extra mounts.

rundk/README.md
chizhg: Updated the README here.

rundk/README.md
coryrc: if I'm in the test directory already, and run:
`rundk ./e2e-tests.sh`
it won't work despite auto-completing that way

rundk/README.md
coryrc: What files? The whole git directory gets mounted.
What good does mounting `/home/<myusername>/.kube/config` do, it's not put in a usable place?
You need to be able to define the external and internal paths separately to be useful.

rundk/README.md
chizhg: Good point. I have changed the working directory to be the directory where the tool is run.

rundk/README.md
chizhg: Changed the flag value to be in the format of ``source1:target1,source2:target2`.
Thanks.

rundk/main.go
albertomilan: I do agree 3 seconds might be enough for a user who is used to invoking the tool, but it might leave ~1-2 seconds of reaction to a developer who does read the message for the first time.
In any case, being the outcome controlled I'm fine with either option.

+1 to @coryrc 's comment. Waiting on user's input does not seem to be the best option at all.

rundk/main.go
```diff
@@ -0,0 +1,192 @@
...
+		}
+	}
+
+	// Setup gcloud credentials
+	if useLocalGcloudCredentials {
+		gcloudKey := os.Getenv("GOOGLE_APPLICATION_CREDENTIALS")
+		// If GOOGLE_APPLICATION_CREDENTIALS is not empty, also mount the key file
+		// to the container
+		if gcloudKey != "" {
+			envs["GOOGLE_APPLICATION_CREDENTIALS"] = gcloudKey
```
coryrc, line 87: You should use a constant filename for the file inside the image

rundk/main.go
```diff
@@ -0,0 +1,192 @@
...
+		}
+	}
+
+	// Setup kubeconfig
+	if useLocalKubeconfig {
+		kubeconfig := os.Getenv("KUBECONFIG")
+		// If KUBECONFIG is not empty, also mount the kubeconfig files to the
+		// container
+		if kubeconfig != "" {
+			envs["KUBECONFIG"] = kubeconfig
```
coryrc, line 107: ditto filename

rundk/main.go
```diff
@@ -0,0 +1,192 @@
...
+		}
+		cmd.AddMount("bind", sourceAndTarget[0], sourceAndTarget[1])
+	}
+
+	// Copy and mount source code root dir
+	repoRoot, err := helpers.GetRootDir()
+	if err != nil {
+		log.Fatal("Error getting the repo's root directory: ", err)
+	}
+	cleanup := cmd.CopyAndAddMount("bind", tmpDir, repoRoot, repoRoot)
```
coryrc, line 144: you don't want to mount the repoRoot to `/go/src/...`?

pkg/interactive/docker.go
```diff
@@ -76,6 +78,52 @@ func (d *Docker) AddMount(typeStr, source, target string, optAdditionalArgs ...s
...
+	if err := os.Chmod(tempDir, 0777); err != nil {
+		cleanup()
+		log.Fatalf("error changing file mode for %q: %v", tempDir, err)
+	}
+
+	// Copy all the content if it's a directory
+	if fi.IsDir() {
+		source += "/."
+	}
+	if _, err := cmd.RunCommand(fmt.Sprintf("/bin/bash -c 'cp -r %s %s'", source, tempDir), cmd.WithStdout()); err != nil {
```
coryrc, line 113: why bash? you should be able to run `cp` directly

rundk/README.md
```diff
@@ -0,0 +1,60 @@
...
+      The image we use to run the test flow. (default "gcr.io/knative-tests/test-infra/prow-tests:stable")
+  -entrypoint string
+      The entrypoint executable that runs the test commands. (default "runner.sh")
+  -enable-docker-in-docker
+      Enable running docker commands in the test fllow.
+      By enabling this the container will share the same docker daemon in the host machine, so be careful when using it.
+  --use-local-gcloud-credentials
+      Use the same gcloud credentials as local.
+      It can be set either by setting env var GOOGLE_CLOUD_APPLICATION_CREDENTIALS or from ~/.config/gcloud
+  --use-local-kubeconfig
```
coryrc, line 35: why is this double `--` but the others are single `-` prefix?

rundk/README.md
coryrc: I don't see this change?

rundk/README.md
```diff
@@ -0,0 +1,60 @@
...
+      Use the same kubeconfig as local.
+      It can be set either by setting env var KUBECONFIG or from ~/.kube/config
+  -mounts source1:target1,source2:target2,source3:target3
+      A list of extra folders or files separated by comma that need to be mounted to run the test flow.
+      It must be in the format of source1:target1,source2:target2,source3:target3.
+  -mandatory-env-vars string
+      A list of env vars separated by comma that must be set on local. (default "GOOGLE_APPLICATION_CREDENTIALS")
```
coryrc, line 42: Bad example because there's special handling, `KO_DOCKER_REPO` would be a better example

coryrc (changes requested):

rundk/main.go
chizhg: With Go mod, it does not really matter where we put the repo, it'll work under any folder.

pkg/interactive/docker.go
chizhg: The reason was I used to run as `cp -r xxx/* yyy`, and `*` can only be used with bash. But now since I'm not using `*` anymore, I'll remove the bash part then.

rundk/README.md
chizhg: Sorry, copy&paste error. Fixed.

rundk/README.md
chizhg: It's now empty by default, so I've removed the whole `(default "GOOGLE_APPLICATION_CREDENTIALS")` part.

rundk/main.go
chizhg: Hmm do you mean always mounting the key file to the same target path? If yes are there any reasons or advantages?

rundk/main.go
chizhg: Similar as the above comment related to `GOOGLE_APPLICATION_CREDENTIALS`. Also the `KUBECONFIG` can be pointed to multiple kubeconfig files (for multiple clusters), so it's not quite easy to use constant filenames here.

rundk/main.go
chizhg: And this tool will need to work with other test images, which do not necessarily have their `GOPATH` set to `/go`.

rundk/main.go
coryrc: Yes:
Privacy: It leaks information about the caller. (it's mounting their home directory setup, not the biggest deal in the world)
Debugging: Easier to know the file will be in a fixed position (Chi: running this doesn't work for me; Cory: can you `cat /gac.json`?)
Defense-in-depth: Weird file paths could break things in scripts, setting to a fixed, simple name avoids that

rundk/README.md
```diff
@@ -35,12 +35,12 @@ Usage of rundk:
...
       It can be set either by setting env var KUBECONFIG or from ~/.kube/config
-  -mounts source1:target1,source2:target2,source3:target3
+  --mounts source1:target1,source2:target2,source3:target3
       A list of extra folders or files separated by comma that need to be mounted to run the test flow.
       It must be in the format of source1:target1,source2:target2,source3:target3.
-  -mandatory-env-vars string
-      A list of env vars separated by comma that must be set on local. (default "GOOGLE_APPLICATION_CREDENTIALS")
-  -optional-env-vars string
+  --mandatory-env-vars string
+      A list of env vars separated by comma that must be set on local.
```
coryrc, line 42: You are missing the important text that said environment variables will be promoted to the container

test/e2e-tests.sh
```diff
@@ -0,0 +1,37 @@
...
+# to run a specific set of tests.
+
+export GO111MODULE=on
+
+source $(dirname "${BASH_SOURCE[0]}")/../scripts/e2e-tests.sh
+
+gcloud auth list
+
+# Script entry point.
+initialize "$@" --gcp-project-id=test-project-chizhg
```
coryrc, line 31: I don't think we should hardcode your personal project

rundk/main.go
chizhg: 1. I don't think it will leak information about the caller, they will run the docker command and start the container on their local anyway..
2. If `GOOGLE_APPLICATION_CREDENTIALS` is set on the local, it must be explicitly set by the user, so they must know where the file is, and I don't think it's harder to find than asking us the magic value we hardcoded.
3. I don't seem to realize there could be weird file paths that could break things, especially since we only mount it as a single file. Do you have a concrete example or have you met issues related to that?

test/e2e-tests.sh
chizhg: Sorry this file was for my local testing and I committed it by mistake. Just deleted it.

rundk/README.md
chizhg: Added.

pkg/interactive/docker.go
```diff
@@ -76,6 +78,52 @@ func (d *Docker) AddMount(typeStr, source, target string, optAdditionalArgs ...s
...
+		log.Fatal("error creating a temporary directory: ", err)
+		return nil
+	}
+
+	cleanup := func() {
+		log.Printf("🧹 Removing the temp directory %q", tempDir)
+		if err := os.RemoveAll(tempDir); err != nil {
+			log.Printf("Warning: error removing the temp directory %q: %v", tempDir, err)
+		}
+	}
```
albertomilan, line 102: Being this a directory removal, why not making it an independent function and simply return the directory path as end result?
I'm not sure if this is a common practice in Go, but I'm not able to see the direct benefits of dealing with the function as a return object (it mainly complicates its interface)

pkg/interactive/docker.go
```diff
@@ -76,6 +78,52 @@ func (d *Docker) AddMount(typeStr, source, target string, optAdditionalArgs ...s
...
+
+	if err := os.Chmod(tempDir, 0777); err != nil {
+		cleanup()
+		log.Fatalf("error changing file mode for %q: %v", tempDir, err)
+	}
+
+	// Copy all the content if it's a directory
+	if fi.IsDir() {
+		source += "/."
```
albertomilan, line 111: `source += string(os.PathSeparator)+"."` ?

rundk/README.md
```diff
@@ -0,0 +1,60 @@
...
+## Usage
+
+```shell
+Usage of rundk:
+  --test-image string
+      The image we use to run the test flow. (default "gcr.io/knative-tests/test-infra/prow-tests:stable")
+  --entrypoint string
+      The entrypoint executable that runs the test commands. (default "runner.sh")
+  --enable-docker-in-docker
+      Enable running docker commands in the test fllow.
```
albertomilan, line 30: Nit: Typo in `fllow`

rundk/README.md
```diff
@@ -0,0 +1,60 @@
...
+  --use-local-kubeconfig
+      Use the same kubeconfig as local.
+      It can be set either by setting env var KUBECONFIG or from ~/.kube/config
+  --mounts source1:target1,source2:target2,source3:target3
+      A list of extra folders or files separated by comma that need to be mounted to run the test flow.
+      It must be in the format of source1:target1,source2:target2,source3:target3.
+  --mandatory-env-vars string
+      A list of env vars separated by comma that must be set on local, which will then be promoted to the container.
+  --optional-env-vars string
+      A list of env vars separated by comma that optionally need to be set on local, which will then be promoted to the container.
```
albertomilan, line 44: From a use perspective, it's not clear what might make an env var optional or not, given in both cases they will be promoted to the container.

rundk/main.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	builtUpDefers := make([]func(), 0)
+
+	// Setup command
+	cmd := interactive.NewDocker()
+	envs := interactive.Env{}
+
+	// Setup temporary directory to save the artifacts
+	tmpDir, err := ioutil.TempDir("/tmp", "prow-docker.")
```
albertomilan, line 66: Is the template `prow-docker.` working as expected? From the documentation [1], it seems `*` should be used to avoid overwriting.

[1] https://golang.org/pkg/io/ioutil/#TempDir

rundk/main.go
```diff
@@ -0,0 +1,192 @@
...
+	// Finally add the image then command to run (if any)
+	cmd.AddArgs(image)
+	log.Println(cmd)
+	if len(commandAndArgsOpt) != 0 {
+		cmd.AddArgs(entrypoint)
+		cmd.AddArgs(commandAndArgsOpt...)
+		log.Println("👆 Starting in 3 seconds, ^C to abort!")
+	}
+
+	time.Sleep(time.Second * 3)
```
albertomilan, line 190: Nit: Use a unique reference for the wait time. It will simplify any potential future change which can diverge these values

chizhg: Done.

rundk/main.go
chizhg: The doc also mentioned `The directory name is generated by taking pattern and applying a random string to the end.`, so if it does not have `*`, it will simply do appending instead of replacing so we will get `prow-docker.xyz`. I have verified it on my local.

rundk/README.md
chizhg: This is what we have for the internal version of this tool that was implemented by Cory, and we do have reasons to keep it I think.

rundk/README.md
chizhg: Fixed, thanks.

pkg/interactive/docker.go
chizhg: Fixed.

pkg/interactive/docker.go
chizhg: It's a common pattern in Go: a function returns another function that does the cleanup, and the callsite call `defer cleanup()` to make sure it's always being called, https://github.com/knative/serving/blob/master/cmd/queue/execprobe.go#L109 is one example and there are a lot more examples in other places.

rundk/README.md
albertomilan: Could you briefly describe the difference in the documentation and here so developers can make correct use of it?
I was not worried about the change, but they potential interpretation given the lack of explanation.

chizhg: Done. Please let me know if it looks reasonable to you now.

PR #2537, opened 2020-10-27
Build stripped binaries to reduce sizes

tcnghia:
Build stripped binaries to reduce sizes. See https://github.com/knative/serving/issues/9957#issuecomment-717491411.

Part of https://github.com/knative/serving/issues/9957

**User-visible changes in this PR**: reduced Knative binaries sizes

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -100,6 +100,8 @@ ARTIFACTS_TO_PUBLISH=""
 FROM_NIGHTLY_RELEASE=""
 FROM_NIGHTLY_RELEASE_GCS=""
 export KO_DOCKER_REPO="gcr.io/knative-nightly"
+# Build stripped binary to reduce size
+export GOFLAGS="-ldflags=-s -ldflags=-w"
 export GITHUB_TOKEN=""
 
 # Convenience function to run the hub tool.


mattmoor (approved):
/lgtm

🎉

dprotaso:
I worry about doing this across the board because we'll lose stack trace information from users and don't have a way to hydrate them with a symbol file.

Secondly - can we double check this doesn't break runtime.Caller invocations? ie. zap log stack trace output

Can we roll this back and do verify a few things

dprotaso:
cc @n3wscott

mattmoor:
```
# GOFLAGS='-ldflags=-s -ldflags=-w' CGO_ENABLED=0 go build ./cmd/foo/ && ./foo
2020/10/27 18:32:46 stack trace: goroutine 1 [running]:
main.bar()
        /Users/mattmoor/go/src/knative.dev/serving/cmd/foo/main.go:10 +0x6d
main.foo(...)
        /Users/mattmoor/go/src/knative.dev/serving/cmd/foo/main.go:15
main.main()
        /Users/mattmoor/go/src/knative.dev/serving/cmd/foo/main.go:19 +0x20
```

This information doesn't key off of the symbol table, but something in `runtime` called `pclntab`.

mattmoor:
... which is like another 10MB 🙄

mattmoor:
```go
package main

import (
	"log"
	"runtime"
)

func bar() {
	buf := make([]byte, 1000)
	i := runtime.Stack(buf, true)
	log.Printf("stack trace: %s", string(buf[:i]))
}

func foo() {
	bar()
}

func main() {
	foo()
}
```

For posterity...

PR #2541, opened 2020-10-28
Set up Prow jobs for hack repo

chizhg:
Set up Prow jobs for hack repo

/cc @n3wscott

### DIFF
--- a/config/prod/prow/config_knative.yaml
+++ b/config/prod/prow/config_knative.yaml
@@ -319,13 +319,16 @@ presubmits:
   knative/test-infra:
   - build-tests: true
   - unit-tests: true
+  - go-coverage: true
+  knative/hack:
+  - build-tests: true
+  - unit-tests: true
   - integration-tests: true
     args:
     - --run-test
     - ./test/e2e-tests.sh
-  - go-coverage: true
   - custom-test: kind-tests
-    always-run: false
+    always-run: true
     needs-dind: true
     args:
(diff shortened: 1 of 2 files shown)


config/prod/prow/config_knative.yaml
```diff
@@ -319,13 +319,16 @@ presubmits:
   knative/test-infra:
   - build-tests: true
   - unit-tests: true
+  - go-coverage: true
+  knative/hack:
+  - build-tests: true
```
n3wscott, line 324: I was thinking not to use prow for hack repo, just GitHub actions.

chizhg: We are running some smoke integration tests with GKE there to make sure the basic e2e test flow is not broken, and also there are some bash unit tests, since we are still running Prow jobs for all Knative repos, I think Prow job is better than GitHub Actions for these cases?

n3wscott: I do not think we should put those tests inside of the hack repo. Keep them in test-infra for a smoke test and test-infra can pull in hack

chizhg: Then we are not running presubmit tests for the scripts in the `hack` repo, which meant if the base e2e test flow is broken, we will only find it out after the scripts are updated in `test-infra` repo.

n3wscott (commented):
/lgtm
/approve

PR #2543, opened 2020-10-28
vendor in hack repo

n3wscott:
This change makes the hack repo the new source of truth for the scripts directory.


mattmoor (commented):
/lgtm

PR #2545, opened 2020-10-29
Add a heartbeat job

chaodaiG:
This is for monitoring prow purpose, so that there is always a job to check for whether prow had been functioning. There will be a prometheus alert rule checking the liveness of this job and sends out alert(s) if this job hadn't been succeeded for some time

### DIFF
--- a/config/prod/prow/jobs/custom/test-infra.yaml
+++ b/config/prod/prow/jobs/custom/test-infra.yaml
@@ -81,11 +81,18 @@ presubmits:
           secretName: test-account
 
 periodics:
+# ci-knative-heartbeat is used for prometheus, alert(s) will be sent
+# if this job hadn't been succeeded for some time
 - cron: "*/3 * * * *" # Every 3 minutes
   name: ci-knative-heartbeat
   agent: kubernetes
   decorate: true
   cluster: "default"
+  extra_refs:
+  - org: knative
+    repo: test-infra
+    base_ref: master
+    path_alias: knative.dev/test-infra
   spec:
     containers:
     - image: gcr.io/knative-tests/test-infra/prow-tests:stable
@@ -93,10 +100,11 @@ periodics:
       command:
       - "runner.sh"
       args:
-      - "go"
-      - "test"
-      - "-v"
-      - "./..."
+      - "echo"
+      - "Everything is fine!"
+      resources:
+        requests:
+          memory: 1Gi
 - cron: "15 9 * * *"
   name: ci-knative-backup-artifacts
   agent: kubernetes
--- a/config/prod/prow/jobs/custom/test-infra.yaml
+++ b/config/prod/prow/jobs/custom/test-infra.yaml
@@ -81,6 +81,22 @@ presubmits:
           secretName: test-account
 
 periodics:
+- cron: "*/3 * * * *" # Every 3 minutes
+  name: ci-knative-heartbeat
+  agent: kubernetes
+  decorate: true
+  cluster: "default"
+  spec:
+    containers:
+    - image: gcr.io/knative-tests/test-infra/prow-tests:stable
+      imagePullPolicy: Always
+      command:
+      - "runner.sh"
+      args:
+      - "go"
+      - "test"
+      - "-v"
+      - "./..."
 - cron: "15 9 * * *"
   name: ci-knative-backup-artifacts
   agent: kubernetes


chizhg:
Can we have some PR description here? 😅 Is it for monitoring to collect some metrics?

chaodaiG:
> Can we have some PR description here? 😅 Is it for monitoring to collect some metrics?

Updated, it's for monitoring

config/prod/prow/jobs/custom/test-infra.yaml
```diff
@@ -81,6 +81,22 @@ presubmits:
...
+  decorate: true
+  cluster: "default"
+  spec:
+    containers:
+    - image: gcr.io/knative-tests/test-infra/prow-tests:stable
+      imagePullPolicy: Always
+      command:
+      - "runner.sh"
+      args:
+      - "go"
```
chizhg, line 96: Will it be failing since I assume there are no Go code without checking out a repo? Can we replace it with a simple `echo` command?

config/prod/prow/jobs/custom/test-infra.yaml
```diff
@@ -81,6 +81,22 @@ presubmits:
           secretName: test-account
 
 periodics:
+- cron: "*/3 * * * *" # Every 3 minutes
```
chizhg, line 84: Can we also have a comment for this Prow job here?

config/prod/prow/jobs/custom/test-infra.yaml
chaodaiG: Good catch! Would prefer it goes through cloneref path for better sanity checks, updated with cloneref

config/prod/prow/jobs/custom/test-infra.yaml
chaodaiG: Added

coryrc:
monitoring what?

chaodaiG:
> monitoring what?

Updated PR description

config/prod/prow/jobs/custom/test-infra.yaml
chizhg: Then my concern is your are running this job very frequently (every 3 minutes), and I don't think this `go test` job can be finished within 3 minutes..

chaodaiG: ack, updated

chizhg: Sorry for the back-and-forth here,  the main point of my last comment was actually `3 minutes` - since every time we update the `prow-tests` image, just the image download time will be more than 3 minutes so I'm worried it will be causing issues. 
Can we
1. Change it to be running less frequently (like every 10 minutes) and/or use a more lightweight test image?
2. Set its own resource requests to be much smaller, so that it can be scheduled more easily?

coryrc: How about:
1. change to per 5 minutes
2. just `ls /go/src/knative.dev/test-infra/go.mod` <== will fail if the file is not there

config/prod/prow/jobs/custom/test-infra.yaml
```diff
@@ -81,6 +81,27 @@ presubmits:
           secretName: test-account
 
 periodics:
+# ci-knative-heartbeat job is for monitoring purpose,
```
coryrc, line 84: Should be:
`# Used for prometheus, alert(s) will be sent if this job hadn't been succeeded for some time

config/prod/prow/jobs/custom/test-infra.yaml
chaodaiG: I see what you meant, if there an image of recommendation?

And thanks for reminding, the resources requests will make this job not able to schedule, will update

config/prod/prow/jobs/custom/test-infra.yaml
chaodaiG: done

config/prod/prow/jobs/custom/test-infra.yaml
coryrc: It should not be taking 3 minutes to download the image; it's cached

coryrc: I thought you meant downloading the repo

coryrc:
/lgtm
/hold

I think it's fine just letting you get moving in case you are ready

chaodaiG:
Thank you Cory!
/unhold

config/prod/prow/jobs/custom/test-infra.yaml
chizhg: I meant for downloading a *new* updated image :-)

PR #2546, opened 2020-10-30
Heartbeat job requests only 0.1CPU

chaodaiG:
It's just an echo command, doesn't need that much CPU(Default is 2CPU)

/cc @chizhg @coryrc


chizhg:
Hmm, GKE is down again?

/retest

PR #2548, opened 2020-11-02
Also set eventing continuous tests timeout to be 180m

chizhg:
Only eventing has this `90m` timeout, not sure why and when it was added.
Removing it to also have the default `180m` timeout.

/cc @grantr @vaikas


grantr (approved):
/lgtm

Fixes https://github.com/knative/eventing/issues/4449

PR #2552, opened 2020-11-03
Added integration-test-channel-consolidated and distributed

eric-sap:

The eventing-kafka project needs two separate integration tests for the distributed and consolidated channel types.  I am hoping that this small change will do it, but I could really use some expert eyes as I have never changed anything in this repository before.

**References**
eventing-kafka e2e test PR (should be merged first):  https://github.com/knative-sandbox/eventing-kafka/pull/146

### DIFF
--- a/config/prod/prow/config_knative.yaml
+++ b/config/prod/prow/config_knative.yaml
@@ -465,10 +465,12 @@ presubmits:
   - integration-tests: false
   knative-sandbox/eventing-kafka:
   - custom-test: integration-test-channel-consolidated
+    needs-dind: true
     args:
     - --run-test
     - ./test/e2e-tests.sh --consolidated
   - custom-test: integration-test-channel-distributed
+    needs-dind: true
     args:
     - --run-test
     - ./test/e2e-tests.sh --distributed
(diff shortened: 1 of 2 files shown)


config/prod/prow/config_knative.yaml
```diff
@@ -464,12 +464,19 @@ presubmits:
   - unit-tests: false
   - integration-tests: false
   knative-sandbox/eventing-kafka:
+  - custom-test: integration-test-channel-consolidated
```
chizhg, line 467: Do you still need Docker-in-Docker to be enabled for these two new jobs? If yes, you'll also need to add `needs-dind: true` there.

eric-sap: Yes; the script still needs that (if I'm not mistaken); I have added it back and regenerated the config script

travis-minke-sap: I don't think we need the docker-in-docker anymore eric - that was primarily for building the confluent/librdkafka from inside docker to avoid cross compiling with CGO, etc...

matzew:
/ok-to-test

LGTM

PR #2554, opened 2020-11-04
Remove dep-collector since we now use go-licenses

chizhg:
Remove dep-collector since we now use go-licenses


mattmoor (commented):
/lgtm

PR #2565, opened 2020-11-11
Ensure generated k8s-testgrid config have deterministic order

chaodaiG:
So that verify-codegen isn't flaky any more

/cc @joshua-bone @n3wscott


chizhg:
/lgtm

Thanks Chao!

PR #2568, opened 2020-11-11
adding labels to allow more kind labels for PRs.

n3wscott:
/lint

Adding more labels to support:

```
/kind [api-change, bug, cleanup, deprecation, removal, documentation, enhancement, performance]
```

These kinds are used for the release note generation process. See an example: https://github.com/knative/eventing/actions?query=workflow%3A%22Release+Notes%22

Release notes will be categorized better

### DIFF
--- a/config/label_sync/labels.yaml
+++ b/config/label_sync/labels.yaml
@@ -64,6 +64,7 @@ default:
       target: both
       addedBy: humans
     - color: 6103a0
+      description: Categorizes issue or PR as related to removing dead code, or a feature that had been marked deprecation.
       name: kind/removal
       target: both
       prowPlugin: label
--- a/config/label_sync/labels.yaml
+++ b/config/label_sync/labels.yaml
@@ -27,39 +27,66 @@ default:
       description: Parked issue that required triaging/revisit in a near future.
       name: kind/TBD
       target: issues
+      prowPlugin: label
       addedBy: humans
     - color: 7057ff
       description: Denotes an issue ready for a new contributor.
       name: kind/good-first-issue
       target: issues
+      prowPlugin: label
+      addedBy: anyone
+    - color: e11d21
+      description: Categorizes issue or PR as related to adding, removing, or otherwise changing an API
+      name: kind/api-change
+      target: both
+      prowPlugin: label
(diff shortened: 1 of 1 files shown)


config/label_sync/labels.yaml
```diff
@@ -27,39 +27,66 @@ default:
...
       addedBy: humans
     - color: 7057ff
       description: Denotes an issue ready for a new contributor.
       name: kind/good-first-issue
       target: issues
+      prowPlugin: label
+      addedBy: anyone
+    - color: e11d21
+      description: Categorizes issue or PR as related to adding, removing, or otherwise changing an API
+      name: kind/api-change
```
lionelvillard, line 40: how does this label relate to `area/api`?

config/label_sync/labels.yaml
```diff
@@ -27,39 +27,66 @@ default:
...
-      target: issues
+      target: both
+      addedBy: humans
+    - color: e11d21
+      description: Categorizes issue or PR as related to a feature/enhancement marked for deprecation.
+      name: kind/deprecation
+      target: both
+      addedBy: humans
+    - color: 6103a0
+      name: kind/removal
```
lionelvillard, line 67: can you add a description?

config/label_sync/labels.yaml
n3wscott: area/api is for issues related to api, kind/api-change means the api is changing?

config/label_sync/labels.yaml
n3wscott: sure, added. I am trying to match what we use to use.

grantr:
Are these label names predetermined by the tool, or do we decide them?

n3wscott:
> Are these label names predetermined by the tool, or do we decide them?

@grantr take a look at https://github.com/knative-sandbox/.github/pull/75

you will select them. `/kind foo` is how the release-notes tool organizes the notes I found.

grantr (approved):
Thanks, that helps clarify. I was more asking if the set of possible labels is predetermined by the tool. But the answer to that doesn't need to block this PR.

/lgtm

PR #2590, opened 2020-11-26
A few improvements and cleanups to the image auto-push process

chizhg:
1. Unify all the images to use `make push` command to build and publish new images
2. Pin `gcloud` and `kubetest2` to a fixed version since they are critical in our test flow, and we don't want to update them unintentionally
3. Remove the installation for `kubectl` v1.17.4 since release-0.19 has been out
4. Remove some outdated comments for `gvm`

/cc @coryrc @albertomilan


n3wscott (commented):
/lgtm
/approve

PR #2608, opened 2020-12-10
Fixes and improvements to the rundk tool

chizhg:
- `--image` is a better flag name than `--test-image`, since this tool can be used to run all the flows other than just tests
- `pflag` lib requires all the extra args to be delimited with other flags with `--`, so add it in the examples
- We should only expect users to provide one way to authenticating to `gcloud` and `kubectl`, so change the logics there to be `if ... else ...`

### DIFF
--- a/rundk/main.go
+++ b/rundk/main.go
@@ -10,13 +10,21 @@ import (
 	"time"
 
 	"github.com/spf13/pflag"
+
 	"knative.dev/test-infra/pkg/helpers"
 	"knative.dev/test-infra/pkg/interactive"
 )
 
-// The warmup time before starting the test flow to give users some flexibility
-// to react.
-const warmupTime = 3
+const (
+	// The warmup time before starting the test flow to give users some flexibility
+	// to react.
+	warmupTime = 3
+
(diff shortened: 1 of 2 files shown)


rundk/main.go
```diff
@@ -93,34 +101,42 @@ func setup() (interactive.Docker, func()) {
 				cleanup := cmd.CopyAndAddMount("bind", tmpDir, gcloudKey, gcloudKey)
 				builtUpDefers = append(builtUpDefers, cleanup)
```
albertomilan, line 102: Don't you need to set `envs["CLOUDSDK_CONFIG"]` in this branch of the if-else statement?

chizhg: To add a bit more contexts here:
- `GOOGLE_APPLICATION_CREDENTIALS` is for authenticating with a Google service account
- The credential files under the gcloud config folder are for authenticating with a user account

So we don't need to set the `CLOUDSDK_CONFIG` env var if we are using the `GOOGLE_APPLICATION_CREDENTIALS` env var, since it's only used for overwriting the default value for gcloud config path, which is for a different authentication mechanism.

PR #2622, opened 2021-01-08
Add hermetic go mod aware handling to run_go_tool

Cynocracy:
Allows tools to be pulled in via go modules, which allows us to ensure the version of the tool used by various devs against the same commit matches exactly. When the tool is not found in a module, it will automatically fall back to the old logic.

The only instance where this will fail is if the module includes a different version of the tool than that installed in GOPATH.

Warns when non-hermetic mode is used, as it can be very difficult to main consistency across individual devs machines without
it.

Fixes #

None


chizhg:
@Cynocracy we have moved the scripts to github.com/knative/hack repo, and the scripts changed there will be synced here for backward compatibility. Could you create this PR against that repo instead? Thanks!

Cynocracy:
Ah ok, sure, can do!

Cynocracy:
https://github.com/knative/hack/pull/38

PR #2626, opened 2021-01-12
Drop in replacement using uber/sync

vagababov:
Much easier to read, no pointer thingies, fewer casts,
better all around

/assign @n3wscott @chizhg


vagababov:
Pushed the new dep code

PR #2629, opened 2021-01-14
bump autobumper to working image

mpherman2:
autobumper.sh was broken and will not be able to bump itself out of being broken, so I am manually bumping the autobumper to the current k8s/test-infra image that has the fix.

Fixes #


chizhg:
/ok-to-test
/lgtm
/approve

Thanks!

PR #2638, opened 2021-01-22
change default gke region to us-east1

dprotaso:
this regions picks up k8s versions in the regular channel faster than central

### DIFF
--- a/kntest/pkg/kubetest2/gke/options.go
+++ b/kntest/pkg/kubetest2/gke/options.go
@@ -30,7 +30,7 @@ func addOptions(gkeCmd *cobra.Command, cfg *kubetest2.GKEClusterConfig) {
 	f.StringVar(&cfg.GCPProjectID, "gcp-project-id", "", "GCP project ID for creating the cluster")
 
 	f.StringVar(&cfg.Name, "name", "", "The GKE cluster name.")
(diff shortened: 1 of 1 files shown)


kntest/pkg/kubetest2/gke/options.go
```diff
@@ -30,7 +30,7 @@ func addOptions(gkeCmd *cobra.Command, cfg *kubetest2.GKEClusterConfig) {
 	f.StringVar(&cfg.GCPProjectID, "gcp-project-id", "", "GCP project ID for creating the cluster")
 
 	f.StringVar(&cfg.Name, "name", "", "The GKE cluster name.")
-	f.StringVar(&cfg.Region, "region", "us-central1", "The region to create the GKE cluster.")
+	f.StringVar(&cfg.Region, "region", "us-east1", "The region to create the GKE cluster.")
 	f.StringSliceVar(&cfg.BackupRegions, "backup-regions", []string{"us-west1", "us-east1"}, "The backup regions if the cluster creation runs into stockout issue in the primary region.")
```
chaodaiG, line 34: I think you also need to replace `us-east1` with `us-central1` here

dprotaso: good catch

PR #2641, opened 2021-01-25
Run conformance tests in mesh mode in net-istio

arturenault:
- Run conformance tests in mesh mode in net-istio; we've been bitten by bugs in Istio where injection breaks things before. These PR jobs would catch these sorts of bugs before they get merged.

Fixes knative-sandbox/net-istio#136

### DIFF
--- a/config/prod/prow/config_knative.yaml
+++ b/config/prod/prow/config_knative.yaml
@@ -448,6 +448,16 @@ presubmits:
     args:
     - --run-test
     - ./test/e2e-tests.sh --istio-version latest
+  - custom-test: latest-mesh
+    optional: true
+    args:
+    - --run-test
+    - ./test/e2e-tests.sh --istio-version latest --mesh
+  - custom-test: stable-mesh
+    optional: true
+    args:
+    - --run-test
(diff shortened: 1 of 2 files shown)


nak3:
net-istio's (networking's) conformance test does not use Knative serving but uses pure Kubernetes Pods.
So the conformance test cannot cover the serving's "mesh mode". But do we still want to enable the mesh test on the net-istio?

arturenault:
> net-istio's (networking's) conformance test does not use Knative serving but uses pure Kubernetes Pods.
> So the conformance test cannot cover the serving's "mesh mode". But do we still want to enable the mesh test on the net-istio?

As far as I can tell, all that mesh mode does is enable sidecar injection in system namespaces and test namespaces. This still would've caught the Istio 1.8.1 issue, and I think it's valuable in any case to ensure that net-istio passes conformance with injection enabled.

nak3:
/lgtm
/approve

I see. Thank you!

nak3:
@arturenault It seems this change does not work. Please see https://github.com/knative-sandbox/net-istio/pull/495 .

PR #2643, opened 2021-01-27
k8s 1.19 libs, go 1.15 remove replace

vaikas:

Remove the replace statements which are not necessary and cause problems.
Update to use k8s 1.19.7 libraries
use go 1.15 in go.mod

Fixes #


markusthoemmes (approved):
/lgtm
/approve

PR #2647, opened 2021-02-01
Move branch protector to another time

peterfeifanchen:
GKE keeps killing this job, so moving it to another time to see if it
works.

@chizhg

### DIFF
--- a/config/prod/prow/jobs/custom/branchprotector.yaml
+++ b/config/prod/prow/jobs/custom/branchprotector.yaml
@@ -47,8 +47,8 @@
 #          secretName: github-token-for-branchprotector
 
 periodics:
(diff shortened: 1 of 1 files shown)


config/prod/prow/jobs/custom/branchprotector.yaml
```diff
@@ -47,8 +47,8 @@
 #          secretName: github-token-for-branchprotector
 
 periodics:
-# Run at 10AM PST.
-- cron: "0 17 * * *"
+# Run at 8AM PST.
+- cron: "0 15 * * *"
```
chizhg, line 51: Hmm why do you think changing the time would help?

PR #2648, opened 2021-02-02
Allow for ignoring arbitrary files with shellcheck

arturenault:
- We currently hardcode that the `vendor/` directory should be ignored by shellcheck, but users leveraging our test-infra should be able to customize which files should be ignored.


arturenault:
Wrong repo.

Issue #2652, opened 2021-02-04
peribolos job fails silently, presubmit job does not appear to validate properly.


vaikas:
We tried to remove a group as per here:
https://github.com/knative/community/issues/91

And the PR is here:
https://github.com/knative/community/pull/425

Notice the presubmit job ran successfully here:
https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative_community/425/pull-knative-peribolos/1356294227448827904

But looking at the logs, it doesn't really seem to do anything useful since it skips everything?

```
{"component":"unset","file":"prow/flagutil/github.go:75","func":"k8s.io/test-infra/prow/flagutil.(*GitHubOptions).Validate","level":"warning","msg":"It doesn't look like you are using ghproxy to cache API calls to GitHub! This has become a required component of Prow and other components will soon be allowed to add features that may rapidly consume API ratelimit without caching. Starting May 1, 2020 use Prow components without ghproxy at your own risk! https://github.com/kubernetes/test-infra/tree/master/ghproxy#ghproxy","severity":"warning","time":"2021-02-01T17:32:02Z"}
{"client":"github","component":"unset","file":"prow/github/client.go:594","func":"k8s.io/test-infra/prow/github.(*client).log","level":"info","msg":"Throttle(300, 100)","severity":"info","time":"2021-02-01T17:32:02Z"}
{"component":"unset","file":"prow/cmd/peribolos/main.go:819","func":"main.configureOrg","level":"info","msg":"Skipping org metadata configuration","severity":"info","time":"2021-02-01T17:32:02Z"}
{"component":"unset","file":"prow/cmd/peribolos/main.go:831","func":"main.configureOrg","level":"info","msg":"Skipping org member configuration","severity":"info","time":"2021-02-01T17:32:02Z"}
{"component":"unset","file":"prow/cmd/peribolos/main.go:838","func":"main.configureOrg","level":"info","msg":"Skipping org repositories configuration","severity":"info","time":"2021-02-01T17:32:02Z"}
{"component":"unset","file":"prow/cmd/peribolos/main.go:844","func":"main.configureOrg","level":"info","msg":"Skipping team and team member configuration","severity":"info","time":"2021-02-01T17:32:02Z"}
{"component":"unset","file":"prow/cmd/peribolos/main.go:205","func":"main.main","level":"info","msg":"Finished syncing configuration.","severity":"info","time":"2021-02-01T17:32:02Z"}
``` 

[...]

PR #2653, opened 2021-02-04
run with same flags as in post except confirm stays as false

vaikas:
Add the following flags to pull-knative-peribolos and pull-knative-sandbox-peribolos runs. 

```
        - "--fix-org=true"
        - "--fix-org-members=true"
        - "--fix-teams=true"
        - "--fix-team-members=true"
        - "--fix-team-repos=true"
        - "--fix-repos=true"
```
Confirm still stays as false so no actions are performed. However, without these flags it does not seem like we do enough validation to prevent #2652 

Addresses #2652 not sure it completely fixes it, but at least it should process the file more thoroughly.

### DIFF
--- a/config/prod/prow/jobs/custom/peribolos.yaml
+++ b/config/prod/prow/jobs/custom/peribolos.yaml
@@ -35,6 +35,13 @@ presubmits:
         - "--config-path=peribolos/knative.yaml"
         - "--github-token-path=/etc/github/token"
         - "--min-admins=5"
+        - "--fix-org=true"
+        - "--fix-org-members=true"
+        - "--fix-teams=true"
+        - "--fix-team-members=true"
+        - "--fix-team-repos=true"
+        - "--fix-repos=true"
+        - "--tokens=1200"
         # Set --confirm=false to only validate the configuration file.
         - "--confirm=false"
         volumeMounts:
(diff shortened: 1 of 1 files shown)


chizhg:
The reason that we are not enabling these options in presubmit is it'll take a long time to finish, especially that GitHub enforces strict rate limit on the request sent from one single token. Also I don't think this will be able to catch all the errors, since running with `--confirm=false` will only do read but not write.

A better option IMO would be configuring the postsubmit and periodic jobs to send alerts to the `productivity` Slack channel if they fail, so we can get noticed as soon as there are errors which can make it much easier to root causing.

vaikas:
Yes, I'd love for these errors to surface on the slack channel, figuring out that things had failed and why was not the most obvious process.
I think when I looked at the running with those flags, it took about ~10 minutes. My thinking is that if it catches any more errors, even though it doesn't catch them all, it would be an improvement? Just from a quick look at the repo activity, does not seem like there's that much activity that it would cause a large slowdown in the ability to issue PRs, so if we add 10 minutes to each PR but it catches more of these errors, that seems better? Especially if the way you recover from a a failure is to, reissue another PR, wait for it to either fail or pass and then rinse and repeat?

n3wscott (commented):
/lgtm
/approve

n3wscott:
/hold

Just saw the other comments

n3wscott:
/unhold

If we find this causes a must larger than 10 minute delay in CI, we can revert.

PR #2659, opened 2021-02-06
Delete unused shellcheck filtering

bharattkukreja:
Shellcheck filtering which is supposed to make shellcheck run only on new files is not being used anywhere - deleting it.

### DIFF
--- a/scripts/library.sh
+++ b/scripts/library.sh
@@ -770,29 +770,6 @@ function get_latest_knative_yaml_source() {
   echo "https://storage.googleapis.com/knative-nightly/${repo_name}/latest/${yaml_name}.yaml"
 }
 
-function shellcheck_new_files() {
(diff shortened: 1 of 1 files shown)


mattmoor:
This broke verify-codegen 🙃

bharattkukreja:
Ugh I was concerned about that. Even though, the verify deps and codegen test failed, it automatically merged :(. Thanks for taking care of this @mattmoor

chizhg:
> Ugh I was concerned about that. Even though, the verify deps and codegen test failed, it automatically merged :(. Thanks for taking care of this @mattmoor

I didn't notice the GitHub actions failure when I approved that PR, sorry about that. @bharattkukreja if you still want this change, you can submit a PR in github.com/knative/hack instead, and the change will be automatically synced to this repo.

PR #2663, opened 2021-02-09
Add daily prow jobs running istio-head

arturenault:
**What this PR does, why we need it**
Starts running daily tests against Istio HEAD. The Istio team has requested this so we don't find issues only after they release.

Fixes knative-sandbox/net-istio/issues/496

None


howardjohn:
Nice!

PR #2664, opened 2021-02-10
Drop gloo, ambassador and kong from "/test all"

nak3:
When `/test all` is called, all tests are run. But gloo, ambassador
and kong are unstable now and the failure is not worth to investigate.

So this patch removes them from `/test all`.

/cc @tcnghia @markusthoemmes

### DIFF
--- a/config/prod/prow/jobs/config.yaml
+++ b/config/prod/prow/jobs/config.yaml
@@ -495,7 +495,7 @@ presubmits:
     optional: true
     run_if_changed: "^third_party/gloo-latest/*"
     rerun_command: "/test pull-knative-serving-gloo-0.17.1"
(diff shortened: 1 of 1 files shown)


tcnghia:
@nak3 I didn't see changed files?

nak3:
Sorry! I missed that all change was cleaned up by `./hack/update-codegen.sh`.

And it seems that [this change](https://github.com/knative/test-infra/pull/2664/commits/94437e9ceadfcaa4078c5f7959ec6f2280fee5c6) I wanted cannot be changed due to fixed template as
https://github.com/knative/test-infra/blob/master/tools/config-generator/templates/prow_presubmit_job.yaml#L9

I am closing this. Sorry for the noise.

PR #2675, opened 2021-02-22
Add Gerardo and Joshua as the owner of the Prow config files

chizhg:
/cc @albertomilan @joshua-bone @gerardo-lc

### DIFF
--- a/config/prod/prow/OWNERS
+++ b/config/prod/prow/OWNERS
@@ -1,8 +1,10 @@
 approvers:
-- chaodaiG
-- coryrc
-- chizhg
-- steuhs
-- peterfeifanchen
-- efiturri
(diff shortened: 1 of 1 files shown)


joshua-bone (approved):

config/prod/prow/OWNERS
```diff
@@ -1,8 +1,10 @@
 approvers:
```
n3wscott, line 1: I am not sure this file will do much in the near future. We are moving to CODEOWNERS.

chizhg: Yeah I can see that.. I'll close this PR then. Will try to bring it up again if needed after we move to CODEOWNERS.

PR #2676, opened 2021-02-24
chore: Add kn-plugin-admin releases

rhuss:
... and removed knative/client-contrib builds as this repo has been split
up and moved to the knative-sandbox


rhuss:
@dsimansk @maximilien actually found out that for kn-plugin-admin there was no release job set up.

@chizhg @chaodaiG we would need this landed to release kn-plugin-admin 0.21 (sorry we forgot to set that up before)

chaodaiG (approved):

PR #2683, opened 2021-03-03
Add all current 'main' based repos to the respective list

markusthoemmes:

As per title, this fills in all of the main repositories we currently have already. I've created those list with two quick curl commands:

```
curl -s 'https://api.github.com/orgs/knative/repos?per_page=200' | jq '.[] | select(.archived == false) | select(.default_branch == "main") | .full_name' | sort
curl -s 'https://api.github.com/orgs/knative-sandbox/repos?per_page=200' | jq '.[] | select(.archived == false) | select(.default_branch == "main") | .full_name' | sort
```

I suppose we could automate this too :thinking:. (Edit: tried automating it and rate limiting is a bit harsh so I'd go with this approach for now).

### DIFF
--- a/config/prod/prow/jobs/config.yaml
+++ b/config/prod/prow/jobs/config.yaml
@@ -7653,7 +7653,7 @@ periodics:
   - org: knative-sandbox
     repo: kn-plugin-diag
     path_alias: knative.dev/kn-plugin-diag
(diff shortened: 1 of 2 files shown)


n3wscott (approved):
/lgtm
/approve

PR #2684, opened 2021-03-03
Automate filling in the main exception list

### LINKED ISSUE #2632: All config/prod/prow/* config needs to be not hardcoded to master.
Move to not hardcode the HEAD branch of any repo. This is an effort to support inclusive naming and eliminate the usage of a `master` branch.

See https://github.com/knative/community/issues/293﻿

markusthoemmes:
Fixes #2632

A followup to https://github.com/knative/test-infra/pull/2683 where the list of main the default branches will be filled automatically for the `knative` and `knative-sandbox` org, so we "just" need to rerun the generator once flipping the respective switches.

Ideally, we'd send a request per repo and pick whatever default branch the respective repo has, but that makes us run into rate-limiting immediately and this seems to be a good 80% solution while we're in migration.

### DIFF
--- a/tools/config-generator/main.go
+++ b/tools/config-generator/main.go
@@ -21,6 +21,7 @@ package main
 
 import (
 	"bytes"
+	"context"
 	"flag"
 	"fmt"
 	"io"
@@ -35,6 +36,7 @@ import (
 	"text/template"
 	"time"
 
+	"github.com/google/go-github/v32/github"
 	"gopkg.in/yaml.v2"
 	"k8s.io/apimachinery/pkg/util/sets"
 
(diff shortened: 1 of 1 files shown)


tools/config-generator/main.go
```diff
@@ -760,6 +766,23 @@ func main() {
 		}
 	}
 
+	// Fill in the main exception list. We fetch this once for all of our orgs to avoid
+	// rate-limiting issues with Github.
+	client := github.NewClient(nil)
+	for org := range pathAliasOrgs {
+		repos, _, err := client.Repositories.List(context.Background(), org, &github.RepositoryListOptions{
```
chaodaiG, line 773: https://github.com/knative/test-infra/blob/e20e6a9df5009eb74eed35b453fcdb25cdb65f0a/pkg/ghutil/repo.go#L28
Can we use this instead?

markusthoemmes: Well, nope, cause it actually needs the entire response :thinking:

chaodaiG:
There is a merge conflict, will approve once it's fixed

chaodaiG (approved):
/lgtm

PR #2687, opened 2021-03-03
Update knative autobump job to new generic autobump image

mpherman2:
This PR replaces the (currently broken) autobump job with the new generic_autobump job. This is needed because we are hoping to deprecate the old autobump bash script. This should also fix the the autobump job. 

Fixes #


chaodaiG (approved):
/lgtm

Cool, it bumps itself as well 👍

PR #2693, opened 2021-03-04
Add ssh key for generic autobump job

mpherman2:
Generic autobump job is not working for 2 reasons: the robot used for the job config was wrong, and it was missing a ssh key secret. This fixes those two issues

Fixes #

### DIFF
--- a/config/prod/prow/jobs/custom/test-infra.yaml
+++ b/config/prod/prow/jobs/custom/test-infra.yaml
@@ -347,10 +347,16 @@ periodics:
       - name: prow-auto-bumper-github-token
         mountPath: /etc/prow-auto-bumper-github-token
         readOnly: true
+      - name: ssh
+        mountPath: /root/.ssh
     volumes:
     - name: prow-auto-bumper-github-token
       secret:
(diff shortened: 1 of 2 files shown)


config/prod/prow/jobs/custom/test-infra.yaml
```diff
@@ -347,10 +347,16 @@ periodics:
       - name: prow-auto-bumper-github-token
         mountPath: /etc/prow-auto-bumper-github-token
         readOnly: true
+      - name: ssh
```
chaodaiG, line 350: Should this be passed in as an arg?

chaodaiG:
/lgtm
/approve
/retest

/hold
in case you need to make change based on my comment

chaodaiG:
huh it's not prow job, retest won't work

config/prod/prow/jobs/custom/test-infra.yaml
mpherman2: No. It just needs to be in /root/.ssh when the autobumper is run

chaodaiG (approved):

PR #2694, opened 2021-03-04
Fix kubetest2 installation in prow-tests image by removing -u option from go get

chizhg:
Fix kubetest2 installation in prow-tests image by removing -u option from go get

### DIFF
--- a/images/prow-tests/Dockerfile
+++ b/images/prow-tests/Dockerfile
@@ -76,7 +76,7 @@ COPY images/prow-tests/in-gvm-env.sh /usr/local/bin
 ############################################################
 FROM stable-stuff AS external-go-gets
 
-ARG KUBETEST2_VERSION=43699a7ba223ddce766f62ecd7504c7233605710
+ARG KUBETEST2_VERSION=ea3c260668075a29dd2b7f3ba4ddf1fe78bdf898
 
 # Extra tools through go get
 # These run using the kubekins version of Go, not any defined by `gvm`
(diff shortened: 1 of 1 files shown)


coryrc:
Why doesn't `go get` work? It's supposed to. This seems like toil to maintain.

images/prow-tests/Dockerfile
```diff
@@ -88,10 +88,13 @@ RUN go get -u github.com/jstemmer/go-junit-report
 RUN GO111MODULE=on go get -u gotest.tools/gotestsum
 RUN GO111MODULE=on go get -u github.com/raviqqe/liche@v0.0.0-20200229003944-f57a5d1c5be4  # stable liche version for checking md links
 RUN GO111MODULE=on go get sigs.k8s.io/kind@v0.9.0
-RUN GO111MODULE=on go get -u sigs.k8s.io/kubetest2@${KUBETEST2_VERSION}
-RUN GO111MODULE=on go get -u sigs.k8s.io/kubetest2/kubetest2-gke@${KUBETEST2_VERSION}
-RUN GO111MODULE=on go get -u sigs.k8s.io/kubetest2/kubetest2-kind@${KUBETEST2_VERSION}
-RUN GO111MODULE=on go get -u sigs.k8s.io/kubetest2/kubetest2-tester-exec@${KUBETEST2_VERSION}
+RUN git clone https://github.com/kubernetes-sigs/kubetest2 --branch master --single-branch && \
+  cd kubetest2 && \
+  git checkout ${PROTOC_VERSION} && \
```
coryrc, line 93: also this is wrong

images/prow-tests/Dockerfile
```diff
@@ -88,10 +88,13 @@ RUN go get -u github.com/jstemmer/go-junit-report
 RUN GO111MODULE=on go get -u gotest.tools/gotestsum
 RUN GO111MODULE=on go get -u github.com/raviqqe/liche@v0.0.0-20200229003944-f57a5d1c5be4  # stable liche version for checking md links
 RUN GO111MODULE=on go get sigs.k8s.io/kind@v0.9.0
-RUN GO111MODULE=on go get -u sigs.k8s.io/kubetest2@${KUBETEST2_VERSION}
-RUN GO111MODULE=on go get -u sigs.k8s.io/kubetest2/kubetest2-gke@${KUBETEST2_VERSION}
-RUN GO111MODULE=on go get -u sigs.k8s.io/kubetest2/kubetest2-kind@${KUBETEST2_VERSION}
-RUN GO111MODULE=on go get -u sigs.k8s.io/kubetest2/kubetest2-tester-exec@${KUBETEST2_VERSION}
+RUN git clone https://github.com/kubernetes-sigs/kubetest2 --branch master --single-branch && \
```
coryrc, line 91: if you are going to do this (I prefer to just fix whatever is wrong with `go get`), then it should be:

* `git init`
* `git fetch --depth=1 https://github.com/kubernetes-sigs/kubetest2 $KUBETEST2_VERSION`
* `git checkout FETCH_HEAD`

chizhg:
> Why doesn't `go get` work? It's supposed to. This seems like toil to maintain.

Thanks Cory. I think previously it didn't work because `go get` does not support `replace` directives in go.mod, and we had `-u` which will update all the dependencies to the latest. But it seems `kubetest2` go.mod does not have `replace` anymore, so removing `-u` will simply make it work.

images/prow-tests/Dockerfile
chizhg: Fixed by removing `-u`.

images/prow-tests/Dockerfile
```diff
@@ -76,7 +76,7 @@ COPY images/prow-tests/in-gvm-env.sh /usr/local/bin
 ############################################################
 FROM stable-stuff AS external-go-gets
 
-ARG KUBETEST2_VERSION=43699a7ba223ddce766f62ecd7504c7233605710
+ARG KUBETEST2_VERSION=ea3c260668075a29dd2b7f3ba4ddf1fe78bdf898
```
chizhg, line 79: This line diff looks weird.. It's exactly the same as it's in the main branch.

coryrc:
I can't explain the diff, but change looks good

/lgtm
/approve

coryrc:
Yeah, I refreshed my tab and only observed the desired changes. Weird.

PR #2699, opened 2021-03-08
Switch all repos in knative to have main as the default branch

markusthoemmes:
As per title.

Reran automation + hand-adjusted peribolos etc.

/assign @chaodaiG


chaodaiG (approved):
/lgtm

PR #2700, opened 2021-03-08
Remove Github API calls and default generator to main

markusthoemmes:
This removes the defaulting logic that was necessary while having a mixed set of default branches. Much simpler now.

/assign @chaodaiG


chaodaiG (approved):
Awesome!
/lgtm

PR #2703, opened 2021-03-09
Drop unnecessary occurrences of master

markusthoemmes:
Went through and removed a few more unnecessary occurrences of `master`. I left buoy alone for this PR cc @n3wscott. Note the changes to the prow config, I suppose we forgot to adjust this piece before but it wasn't harmful?

/assign @chaodaiG


scripts/library.sh
```diff
@@ -787,18 +787,18 @@ function shellcheck_new_files() {
 function latest_version() {
```
chizhg, line 787: The scripts are vendored from github.com/knative/hack and copied to `scripts/` for backward compatibility, so you need to update the scripts in the hack repo first.

markusthoemmes:
Thanks @chizhg, should be good now!

chizhg (approved):
/lgtm

PR #2704, opened 2021-03-09
Remove knative- prefix from testgrid dashboard names

chaodaiG:
There is no need to prefix them with knative- on https://testgrid.k8s.io/r/knative-own-testgrid


coryrc:
tabs still have `knative-` in them

coryrc:
/lgtm
(also helps if you push new commits instead of force-pushing, so we can see changes)

PR #2717, opened 2021-03-17
Add eventing e2e tests on s390x architecture

barthy1:

This PR proposes the standard way to run Knative tests on non-Intel architectures, as first step with s390x eventing e2e tests. The approach can be applied to other architectures and components. 
The idea that main job is running on Intel architecture, as any other standard test job, but the actual tests will be executed on another k8s arch-specific cluster. From non-Intel architecture k8s cluster and image registry should be provided.
The details:
- the project and test images are built for another architecture (with ko parameter `--platform=linux/<arch>`), despite the actual build is done on x86_64
- images are pushed to special registry (`KO_DOCKER_REPO` parameter)
- the deployment after build is done on remote k8s arch-specific cluster (`KUBECONFIG` parameter points to non-Intel architecture k8s cluster) 
- the tests are running also on remote k8s arch-specific cluster (`KUBECONFIG` parameter parameter points to non-Intel architecture k8s cluster) 

The s390x eventing Prow job:
- is the daily job to run eventing e2e tests on s390x hardware architecture
- it expects from Prow to be able to use `s390x-cluster1` secret to get access to s390x k8s cluster(config file) and image registry(registry.crt) for temporary test images (secrets should be added by Prow maintainer?)

`s390x` hardware is ready and accessible via kubeconfig for k8s cluster. Registry is also up and ready.

The proposed job is tested using mkpj and phaino tools as local Prow job running on Intel hardware with actual tests running on s390x k8s cluster.

config-generator functionality is extended to specify volume/volume mount/env variable for the job (to be able to use secrets with k8s config and registry certificate).

### DIFF
--- a/config/prod/prow/config_knative.yaml
+++ b/config/prod/prow/config_knative.yaml
@@ -861,6 +861,26 @@ periodics:
         memory: 12Gi
       limits:
         memory: 16Gi
+  - custom-job: s390x-e2e-tests
+    cron: 0 8 * * *
+    command:
+    - bash
+    args:
+    - -c
+    - >-
+      mkdir -p /root/.kube &&
+      cp /opt/cluster/config /root/.kube/ &&
+      ./test/e2e-tests.sh --run-tests
+    env-vars:
+    - DISABLE_MD_LINTING="1"
+    - KO_FLAGS="--platform=linux/s390x"
+    - DEPLOY_KNATIVE_MONITORING="0"
+    - SYSTEM_NAMESPACE="knative-eventing"
+    - PLATFORM="linux/s390x"
+    - KUBECONFIG="/root/.kube/config"
+    - SSL_CERT_FILE="/opt/cluster/registry.crt"
+    external_cluster:
+      secret: s390x-cluster1
   knative/eventing-contrib:
   - continuous: true
     resources:
--- a/tools/config-generator/main.go
+++ b/tools/config-generator/main.go
@@ -342,6 +342,24 @@ func addExtraEnvVarsToJob(envVars []string, data *baseProwJobTemplateData) {
 	}
 }
 
+// addExtraClusterInfoToJob enables extra cluster information for the given job.
+// Information includes: k8s config(config file), registry certificate(registry.crt file)
+// and KO_DOCKER_REPO environment variable value (ko-docker-repo file).
+func addExtraClusterInfoToJob(cluster yaml.MapSlice, data *baseProwJobTemplateData) {
+	for _, secret := range cluster {
+		if secret.Key != "secret" {
+			return
+		}
+		secretName := getString(secret.Value)
+		// Volume mount for all data. Kubeconfig shoild be copied to readwrite directory in job command
+		addVolumeToJob(data, "/opt/cluster", secretName, true, nil)
+		// KO_DOCKER_REPO env varable mount from secret
+		env := []string{envNameToKey("KO_DOCKER_REPO"), "  valueFrom:", "    secretKeyRef: ", "      name: " + secretName, "      key: ko-docker-repo"}
+
+		data.Env = append(data.Env, env...)
+	}
+}
+
 // setupDockerInDockerForJob enables docker-in-docker for the given job.
 func setupDockerInDockerForJob(data *baseProwJobTemplateData) {
 	// These volumes are required for running docker command and creating kind clusters.
@@ -418,6 +436,8 @@ func parseBasicJobConfigOverrides(data *baseProwJobTemplateData, config yaml.Map
 			setResourcesReqForJob(getMapSlice(item.Value), data)
 		case "reporter_config":
 			setReporterConfigReqForJob(getMapSlice(item.Value), data)
+		case "external_cluster":
+			addExtraClusterInfoToJob(getMapSlice(item.Value), data)
 		case nil: // already processed
 			continue
 		default:
--- a/config/prod/prow/jobs/config.yaml
+++ b/config/prod/prow/jobs/config.yaml
@@ -10122,6 +10122,70 @@ periodics:
     - name: release-account
       secret:
         secretName: release-account
+- cron: "0 8 * * *"
+  name: ci-knative-eventing-s390x-e2e-tests
+  agent: kubernetes
+  decorate: true
+  decoration_config:
+    timeout: 120m
+  cluster: "build-knative"
+  extra_refs:
+  - org: knative
+    repo: eventing
+    path_alias: knative.dev/eventing
+    base_ref: main
+  annotations:
+    testgrid-dashboards: eventing
+    testgrid-tab-name: eventing-s390x-e2e-tests
+    testgrid-alert-stale-results-hours: "3"
+  spec:
+    containers:
+    - image: gcr.io/knative-tests/test-infra/prow-tests:stable
+      imagePullPolicy: Always
+      command:
+      - runner.sh
+      args:
+      - "bash"
+      - "-c"
+      - "mkdir -p /root/.kube && cp /opt/cluster/config /root/.kube/ && ./test/e2e-tests.sh --run-tests"
+      volumeMounts:
+      - name: s390x-cluster1
+        mountPath: /opt/cluster
+        readOnly: true
+      - name: test-account
+        mountPath: /etc/test-account
+        readOnly: true
+      env:
+      - name: DISABLE_MD_LINTING
+        value: "1"
+      - name: KO_FLAGS
+        value: "--platform=linux/s390x"
+      - name: DEPLOY_KNATIVE_MONITORING
+        value: "0"
+      - name: SYSTEM_NAMESPACE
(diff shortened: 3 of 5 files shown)


barthy1:
hmm, I guess I cannot just add directly to `config.yaml`, instead I need to use `config_knative.yaml. Will change my PR with necessary updates.

coryrc:
> it expects from Prow to be able to use s390x-registry and s390x-kubeconfig secrets to get access to s390x k8s cluster and image registry for temporary test images (secrets should be added by Prow maintainer?)

That's easy. How often will they need to be updated? I could see churn on the kubeconfig? I wonder if we could grant you the ability to modify the secret. I think currently we can't but it'd be nice.

> job expects that special service is running on the Prow cluster to hide k8s cluster ip address with hostname(setup should be installed by Prow maintainer?).

I'm pretty sure IPs get printed out in error messages so they wouldn't end up very hidden? But I also don't fully understand this part so I'm probably missing the point.

tools/config-generator/main.go
```diff
@@ -301,9 +301,9 @@ func addMonitoringPubsubLabelsToJob(data *baseProwJobTemplateData, runID string)
 }
 
 // addVolumeToJob adds the given mount path as volume for the job.
-func addVolumeToJob(data *baseProwJobTemplateData, mountPath, name string, isSecret bool, content []string) {
+func addVolumeToJob(data *baseProwJobTemplateData, mountPath, name string, isSecret bool, isReadOnly bool, content []string) {
```
coryrc, line 304: AFAICT you aren't using readwrite volumes, is this in anticipation of something else?

tools/config-generator/main.go
```diff
@@ -301,9 +301,9 @@ func addMonitoringPubsubLabelsToJob(data *baseProwJobTemplateData, runID string)
...
 	(*data).VolumeMounts = append((*data).VolumeMounts, []string{"- name: " + name, "  mountPath: " + mountPath}...)
-	if isSecret {
+	if isSecret && isReadOnly {
```
coryrc, line 306: Secrets are only readonly so should be an OR?

coryrc (changes requested):

tools/config-generator/main.go
barthy1: the kubeconfig is mounted to `/root/.kube`, and this dir should be readwrite, because extra files are written to it. With readonly setup, the erros occurs
```
error: open /root/.kube/config.lock: read-only file system
```

coryrc: Oh I see, the absence of readonly has made them mounted readwrite. Makes sense, could you throw a comment in somewhere with this info?

tools/config-generator/main.go
barthy1: well, secrets can be mounted as readwrite, as it is done in current proposal, so actually I guess here should be just check for readonly. Anyway, good point to update, thank you!

coryrc: Oh, I see now the intention was for this to only be for mount secrets as RW directories... it's confusing to understand in the other places. I think it'd be cleaner if instead of isSecret and isReadOnly variables, we had an enum:

```
Volume
SecretRO
SecretRW
```

or, I see little harm in all secrets being mounted as RW, which is even simpler! It's not like a job could harm another job via that method.

barthy1:
> That's easy. How often will they need to be updated? I could see churn on the kubeconfig? I wonder if we could grant you the ability to modify the secret. I think currently we can't but it'd be nice.

The cluster is stable, so far experience shows that cleanup after tests does its work and nothing needs to be updated/recreated. However of course some unexpected things can happen. So I'd say, yes, the kubeconfig might be updated sometimes, but not frequently. If you can find some easy way for me to update the secret, it would be just great!

> I'm pretty sure IPs get printed out in error messages so they wouldn't end up very hidden? But I also don't fully understand this part so I'm probably missing the point.

Yep, sorry, I had to be more specific. It's not a problem for me to use concrete ip address, however I think it'd be nice to hide it in some special config, just to be able to replace ip address, without touching job itself. However it's not something really mandatory, so up to reviewers to decide.

tools/config-generator/main.go
barthy1: Well, volumes can be mounted readwrite or readonly, and secret is just one possible source for volume. So for me the readonly and secret are just 2 different areas, I'd not mix them.

> I see little harm in all secrets being mounted as RW,

Best practices recommend not to do that, if possible :) However it will make the changes to the current setup smaller. So from my side, I'd still mount all current secrets readonly, as it is done now (except new ones). However of course it's up to maintainers to decide.

tools/config-generator/main.go
barthy1: sure, will add the comment with description

barthy1: done

coryrc:
> so I'd say, yes, the kubeconfig might be updated sometimes, but not frequently. If you can find some easy way for me to update the secret, it would be just great!

I'll ask around but we're probably stuck with this for now.

> Yep, sorry, I had to be more specific. It's not a problem for me to use concrete ip address, however I think it'd be nice to hide it in some special config, just to be able to replace ip address, without touching job itself.

Ah, because of the peculiarities of this setup, it'll be far easier to update if you have the IP address in a file in this repository and just have it as an ENV var. Then non-googlers can also update it as needed. Or really, the IP address is explicitly tied to the kubeconfig, is it not? so really both should be in the same secret == 50% fewer secrets to maintain!

barthy1:
@coryrc 

> Ah, because of the peculiarities of this setup, it'll be far easier to update if you have the IP address in a file in this repository and just have it as an ENV var. Then non-googlers can also update it as needed. Or really, the IP address is explicitly tied to the kubeconfig, is it not? so really both should be in the same secret == 50% fewer secrets to maintain!

I've rewrote the code to have all required info in 1 secret - k8s config file, registry.crt file and `KO_DOCKER_REPO` value. You are correct, all 3 pieces should be replaced in case if for instance IP address is changed. So better to keep them together + no changes in job config in case if something should be replaced in those values.

tools/config-generator/main.go
barthy1: I am deeply sorry, you were correct and I was wrong about mount of secrets and always readonly :) It's always readonly.
Only one (small) excuse for me that my test env bypasses this problem. Anyway, code is updated to resolve this challenge. Now I just copy k8s config to another directory, which is RW.

coryrc: Haha, no worries! I just know this config generator is already messy so I try to push back a bit on new features to it.

tools/config-generator/main.go
```diff
@@ -342,6 +342,24 @@ func addExtraEnvVarsToJob(envVars []string, data *baseProwJobTemplateData) {
 	}
 }
 
+// addExtraClusterInfoToJob enables extra cluster information for the given job.
+// Informatiion includes: k8s config(config file), registry certificate(registry.crt file)
```
coryrc, line 346: nit: s/Informatiion/Information

config/prod/prow/config_knative.yaml
```diff
@@ -861,6 +861,26 @@ periodics:
...
+      ./test/e2e-tests.sh --run-tests
+    env-vars:
+    - DISABLE_MD_LINTING="1"
+    - KO_FLAGS="--platform=linux/s390x"
+    - DEPLOY_KNATIVE_MONITORING="0"
+    - SYSTEM_NAMESPACE="knative-eventing"
+    - PLATFORM="linux/s390x"
+    - KUBECONFIG="/root/.kube/config"
+    - SSL_CERT_FILE="/opt/cluster/registry.crt"
+    extra_cluster:
```
coryrc, line 882: I think this would be clearer if it were called "external_cluster"? what do you think?
I had to reread the information/comments a few times to figure out why it looked like this, but `external_cluster` would be pretty clear to me.

coryrc (changes requested):
At some point someone from our team will contact you about setting up a GCP project for [this](https://docs.google.com/document/d/1Jo8dyNiJ1fnYh30M_6Gq2XDU9B3bGFmkD2QokyKYwEg/edit?ts=6055480a), so you will have the secret in your own GCP project and can change it at will. (They will also have better instructions soon so don't feel you need to start with it immediately).

But, to start with you can send me the secret and I can apply it to the cluster. My username at my company dot com (google.com) if you want to e-mail or slack it or... I'm open to whatever is easy :).

tools/config-generator/main.go
barthy1: fixed

config/prod/prow/config_knative.yaml
barthy1: yep, `external_cluster` sounds clearer. Thank you for suggestion!

coryrc:
/lgtm
/approve
/hold

(just holding until the secret is applied -- I think it could be an issue but also maybe not)

coryrc:
Good luck!

PR #2725, opened 2021-03-19
Remove nightly release job and disable go-coverage for net-ingressv2

nak3:
As net-ingressv2 adds test code only, nightly release is not
necessary. Also go-coverage should be disabled.

It is in the same manner as knative/networking repo.

/cc @ZhiminXiang @tcnghia @markusthoemmes


chizhg (approved):
/lgtm

PR #2746, opened 2021-04-12
Configure skip-review label to enable auto-merge a few trusted PRs

chizhg:
To address https://github.com/knative/test-infra/pull/2729#issuecomment-816994968

/cc

### DIFF
--- a/config/prod/prow/jobs/custom/test-infra.yaml
+++ b/config/prod/prow/jobs/custom/test-infra.yaml
@@ -358,7 +358,43 @@ periodics:
     - name: flaky-test-reporter-slack-token
       secret:
         secretName: flaky-test-reporter-slack-token
-- cron: "15 15-23 * * 1-5"  # Run at 7:15-15:15 PST (15:15 UTC) Mon-Fri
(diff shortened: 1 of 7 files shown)


config/prod/prow/jobs/custom/test-infra.yaml
```diff
@@ -358,7 +358,43 @@ periodics:
     - name: flaky-test-reporter-slack-token
       secret:
         secretName: flaky-test-reporter-slack-token
-- cron: "15 15-23 * * 1-5"  # Run at 7:15-15:15 PST (15:15 UTC) Mon-Fri
+- cron: "15 15-23 * * 1-5"  # Bump with label `skip-review`. Run at 7:15-15:15 PST (15:15 UTC) Mon-Fri
```
chaodaiG, line 361: Considering how stable prow is, most of the time there is no problem merging every day. But to be more cautious, I might suggest bumping this following prow update schedule, which is Monday every week

chaodaiG (approved):
/hold

config/prod/prow/jobs/custom/test-infra.yaml
chizhg: SGTM. Done.

chaodaiG (approved):

PR #2756, opened 2021-04-21
Added proper dot releases for eventing-kafka and eventing-kafka-broker

slinkydeveloper:
Signed-off-by: Francesco Guardiani <francescoguard@gmail.com>

### DIFF
--- a/config/prod/prow/config_knative.yaml
+++ b/config/prod/prow/config_knative.yaml
@@ -1197,6 +1197,13 @@ periodics:
   - nightly: true
     needs-dind: true
   - dot-release: true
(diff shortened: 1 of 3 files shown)


n3wscott (approved):
this is suppose to be automatic, but maybe the script only forwards on the dots and if they don't exist it does nothing. We are going to replace that in the near term I hope.

/lgtm
/approve

vaikas:
@n3wscott @slinkydeveloper, looks like it was bumped indeed for most other things and probably not for kafka because they didn't exist, they didn't get cut:
https://github.com/knative/test-infra/pull/2744

PR #2771, opened 2021-05-06
Bump Go patch versions

pierDipi:
**What this PR does, why we need it**:

Go bug fixes.

**Special notes to reviewers**:

- https://github.com/golang/go/releases/tag/go1.14.15
- https://github.com/golang/go/releases/tag/go1.15.12

PR #2776, opened 2021-05-11
Add go 1.16 and bump prior versions

benmoss:
Adds go 1.16.5 to the test image and bumps prior versions to the latest patch release. I just noticed that eventing has already bumped to 1.16 but the build still defaults to 1.15.5: https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative_eventing/5382/pull-knative-eventing-reconciler-tests/1392131979842949120


benmoss:
Just saw #2771, but still need the 1.16 addition too 😄

PR #2789, opened 2021-05-19
Add productivity-reviewers as the approver for images and config dir

chizhg:
1. In the future the `OWNERS_ALIASES` file will be generated and synced from the peribolos config file, update it to match with https://github.com/knative/community/pull/616
2. Add `productivity-reviewers` as the owner for `config` and `images` dir since these are the only files that need attention from oncall.

/cc @chaodaiG @efiturri


chaodaiG (commented):
/lgtm
/approve

PR #2804, opened 2021-06-08
Specify volume permission for s390x tests

BbolroC:

ATM, there is a permission problem on connecting to the external s390x test machine via ssh. The PR should resolve the issue.
 

Fixes #

### DIFF
--- a/config/prod/prow/jobs/config.yaml
+++ b/config/prod/prow/jobs/config.yaml
@@ -6958,6 +6958,7 @@ periodics:
     - name: s390x-cluster1
       secret:
         secretName: s390x-cluster1
+        defaultMode: 0600
     - name: test-account
       secret:
         secretName: test-account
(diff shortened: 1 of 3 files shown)


BbolroC:
@coryrc could you review this if you have time? thank you very much! 😉

coryrc:
I don't think this is helping though? Nothing changed here between when it was working and when it stopped.

coryrc:
`Unable to connect to the server: x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "kubernetes")`

Sounds like some certificate is expired? Or now self-signing is no longer good enough?
But it looks like it is reading the file (and looking at https://github.com/knative/eventing/commits/main I don't see any changes there that make me think it's any different)

coryrc:
Oops!

/lgtm
/approve

BbolroC:
Thank you very much and sorry for the unclear PR description. :wink:

PR #2812, opened 2021-06-16
Add serving-s390x-kourier-test

BbolroC:
The existing s390x-test for serving runs over the `contour` network layer. This PR is to add another test for `kourier`.

Fixes #


BbolroC:
@chizhg could you review this if you have time? :wink: Thank you very much!

chizhg:
It looks you have to update the `go-licenses` version on your local. The error is due to the change in https://github.com/google/go-licenses/pull/65

BbolroC:
> It looks you have to update the `go-licenses` version on your local. The error is due to the change in [google/go-licenses#65](https://github.com/google/go-licenses/pull/65)

I ran `./hack/update-codegen.sh` after updating the `go-licenses`. Am I correct to fix the problem? The number of the changed files is increased by the way.;;

chizhg:
> > It looks you have to update the `go-licenses` version on your local. The error is due to the change in [google/go-licenses#65](https://github.com/google/go-licenses/pull/65)
> 
> I ran `./hack/update-codegen.sh` after updating the `go-licenses`. Am I correct to fix the problem? The number of the changed files is increased by the way.;; Could you tell me the right way if I am wrong?

Thanks! The changes are expected.
/lgtm
/approve

PR #2817, opened 2021-06-22
Add new job trigger for mesh test with short option

nak3:
This patch adds new job trigger `pull-knative-serving-istio-stable-mesh-short` and `pull-knative-serving-istio-latest-mesh-short`.

About the jobs:
- Both jobs run mesh test with `--short` option.
- `/all` and `/pull-knative-serving-istio-{stable,latest}-mesh-short` in PR will trigger the job.
- `/pull-knative-serving-istio-stable-mesh-short` will be triggered when `third_party/istio-latest/net-istio.yaml` was changed.
- periodic job (test grid) does not run the jobs.

/cc @julz @dprotaso @ZhiminXiang @arturenault

### DIFF
--- a/config/prod/prow/config_knative.yaml
+++ b/config/prod/prow/config_knative.yaml
@@ -34,6 +34,18 @@ presubmits:
         memory: 12Gi
       limits:
         memory: 16Gi
+  - custom-test: istio-latest-mesh-short
+    needs-monitor: true
+    always-run: false
+    optional: true
+    args:
+    - --run-test
+    - ./test/e2e-tests.sh --istio-version latest --mesh --short
+    resources:
+      requests:
+        memory: 12Gi
+      limits:
+        memory: 16Gi
   - custom-test: istio-latest-mesh-tls
     needs-monitor: true
     always-run: false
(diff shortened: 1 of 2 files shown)


ZhiminXiang:
I assume this is for triggering mesh only test to verify internal traffic won't hit cluster local gateway right?

Why do we not want to run it periodically?

config/prod/prow/config_knative.yaml
```diff
@@ -83,6 +95,19 @@ presubmits:
         memory: 12Gi
       limits:
         memory: 16Gi
+  - custom-test: istio-stable-mesh-short
+    needs-monitor: true
+    always-run: false
+    optional: true
+    run-if-changed: ^third_party/istio-latest/net-istio.yaml
```
arturenault, line 102: shouldn't we run when istio-stable changes for this job?

nak3:
> I assume this is for triggering mesh only test to verify internal traffic won't hit cluster local gateway right?

It is different from the test to verify internal traffic. The `/pull-knative-serving-istio-{stable,latest}-mesh-short` just runs e2e test (scale test) with `-short` option. 

The background of this change is that current mesh test always red :red_circle:  so people who send PR are confused whether their change is broken or not. (We say that "mesh test **always** red :red_circle: . Please ignore the error." :facepalm: )
So, we add `/pull-knative-serving-istio-{stable,latest}-mesh-short` to expose whether their change will break mesh test or not.

config/prod/prow/config_knative.yaml
nak3: The path `^third_party/istio-latest/net-istio.yaml` is not net-istio repo's but for serving repo's. So only `istio-latest` is available https://github.com/knative/serving/tree/main/third_party/istio-latest :sweat_drops:

ZhiminXiang (approved):
/lgtm

nak3:
@chizhg  Could you please take a look?

PR #2821, opened 2021-06-28
Add istio labels per dr. max

evankanderson:

- Adds additional labels for categorizing Istio issues:
  `area/mesh`, `area/sidecar`, `area/perf`, `area/status`

(Request on Slack)

Adds additional labels to `knative-sandbox/net-istio`


evankanderson:
Apparently the dep checker was correctly complaining that we'd checked in a bunch of stuff with mode 100755 instead of 100644

PR #2840, opened 2021-07-15
Add kn-plugin-func

evankanderson:

Adds Prow support/commands for `kn-plugin-func` in the sandbox org.

Adds a new repo to Prow, per https://github.com/knative/test-infra/blob/main/guides/prow_knative_setup.md


evankanderson (commented):
/hold

Looking to add `knative/hack` scripts and `test/presubmit-tests.sh` to the repo.

config/prod/prow/config_knative.yaml
```diff
@@ -289,6 +289,10 @@ presubmits:
   - build-tests: true
   - unit-tests: true
   - integration-tests: true
+  knative-sandbox/kn-plugin-func:
+  - build-tests: true
+  - unit-tests: true
+  - integration-tests: true
```
rhuss, line 295: Could we start small, by disabling the Prow tests (func still has their tests running with GitHub Actions), and just enabling cla-bot and tide ? That way people already can work with the Prow flow while we are deciding how to continue with the CI setup.

Alternatively would could enable it and just add empty hook-scripts for the tests, which we will fill in later.

lance: We have been working pretty hard on our CI infrastructure for `func` and I would love it if we can continue using this while we work out the prow details. However that happens is fine with me, but if possible it would be great to not lose all of the work we've done on CI to date.

evankanderson:
If we're using GitHub Actions for all the CI, then we don't need to set up Prow; the tide/CLA stuff is applied at an org-wide level and doesn't need per-repo configuration.

/close

rhuss:
> If we're using GitHub Actions for all the CI, then we don't need to set up Prow; the tide/CLA stuff is applied at an org-wide level and doesn't need per-repo configuration

It looks like that the branch protection is not setup properly when not adding a repo specific prow configuration (at least I can't see anything for kn-plugin-func). @evankanderson would we need to setup branch protection manually ? (cc: @lance @dsimansk )

PR #2841, opened 2021-07-16
Add Calico CNI installation steps in Kind

shinigambit:

**Allows set up NetworkPolicy in a Kind cluster**:


chizhg:
I don't have much context on this change. Could you add some contexts please? Also it seems this is for net-contour and net-kourier, so I'm not sure if here is the right place to add this doc. But if it's indeed desirable, could you add the owner of these two projects to review it?

PR #2846, opened 2021-07-30
Enable cni plugin in prow clusters

shinigambit:
Allow network policy in prow tests


chizhg:
This is not how we update the Prow build cluster. I can manually run the commands for you if you still need them.

PR #2872, opened 2021-09-14
Add CI confing for client-pkg repository

dsimansk:

There's a new repository to share code between `kn` client and plugins. This PR enables CI jobs etc.

- Add CI confign for `client-pkg` repository

/cc @rhuss

### DIFF
--- a/config/prod/prow/config_knative.yaml
+++ b/config/prod/prow/config_knative.yaml
@@ -280,7 +280,6 @@ presubmits:
   knative/client-pkg:
   - build-tests: true
   - unit-tests: true
-  - integration-tests: true
   - go-coverage: false
   knative/client-contrib:
   - build-tests: true
@@ -901,7 +900,7 @@ periodics:
     - INGRESS_CLASS="contour.ingress.networking.knative.dev"
     external_cluster:
       secret: s390x-cluster1
-  knative/client:
+  knative/client-pkg:
   - continuous: true
   - auto-release: true
   - nightly: true
--- a/config/prod/prow/config_knative.yaml
+++ b/config/prod/prow/config_knative.yaml
@@ -277,6 +277,11 @@ presubmits:
     always-run: true
     command:
     - ./test/presubmit-integration-tests-latest-release.sh
+  knative/client-pkg:
+  - build-tests: true
+  - unit-tests: true
+  - integration-tests: true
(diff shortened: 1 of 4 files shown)


config/prod/prow/config_knative.yaml
```diff
@@ -277,6 +277,11 @@ presubmits:
     always-run: true
     command:
     - ./test/presubmit-integration-tests-latest-release.sh
+  knative/client-pkg:
+  - build-tests: true
+  - unit-tests: true
+  - integration-tests: true
```
dsimansk, line 283: @rhuss I'd say that we don't necessarily need integration tests for client-pkg. Wdyt? We should be fine with build and unit tests.

rhuss: yes, there is not really something we could test integration wise. Maybe the watch functionality, but I would add integration tests when we would need it the first time. let's start small.

config/prod/prow/config_knative.yaml
```diff
@@ -277,6 +277,11 @@ presubmits:
...
+  - go-coverage: false
```
rhuss, line 284: no coverage ?

config/prod/prow/config_knative.yaml
dsimansk: Or maybe we'll keep it blank and come up with something meaningful? :)

config/prod/prow/config_knative.yaml
```diff
@@ -896,6 +901,10 @@ periodics:
     - INGRESS_CLASS="contour.ingress.networking.knative.dev"
     external_cluster:
       secret: s390x-cluster1
+  knative/client:
```
rhuss, line 904: was this missing before ?

dsimansk: :facepalm: that's a copy paste error.

config/prod/prow/config_knative.yaml
dsimansk: We'll have still have Codecov and only skip the go-coverage from Prow.

E.g.: https://github.com/knative/test-infra/blob/main/config/prod/prow/config_knative.yaml#L24

rhuss (commented):
/approve
/lgtm

PR #2874, opened 2021-09-15
Add kn-plugin-source-kamelet for autorelease

rhuss:


rhuss:
@chizhg can we get this merged please so that we have some nightly runs before the release next week ?

PR #2879, opened 2021-09-20
Add prow config for container-freezer

psschwei:

Add prow configs for https://github.com/knative-sandbox/container-freezer

https://github.com/knative/community/issues/742

cc @julz @markusthoemmes


config/prod/prow/config_knative.yaml
```diff
@@ -620,6 +620,11 @@ presubmits:
   - build-tests: true
   - unit-tests: true
   - integration-tests: false
+  knative-sandbox/container-freezer:
+  - build-tests: true
+  - unit-tests: true
+  - integration-tests: true
+  - go-coverage: false
```
julz, line 627: not that Im personally a huge fan of coverage checkers, but any reason we're explicitly disabling the coverage checker here?

psschwei: Mental lapse? I thought I had set all four of these to `true`, but I guess not...

psschwei: @julz this has been fixed

PR #2889, opened 2021-09-27
Workaround Prow CI failures with iptables change

chizhg:
`run.sh` is the entrypoint for all the Knative Prow jobs. Adding the command here will automatically apply to all Knative Prow jobs so we don't have to make the change to each Knative repo.


pierDipi (approved):
/lgtm

PR #2901, opened 2021-10-14
use go1.17.2 by default

dprotaso:
we no longer read the go.mod file as the toolchain should support compiling
modules that target a lower min version

Follow up to: https://github.com/knative/test-infra/pull/2898

### DIFF
--- a/images/prow-tests/runner.sh
+++ b/images/prow-tests/runner.sh
@@ -18,7 +18,7 @@ ORIGINAL_GOPATH="${GOPATH}"
 
 source "${HOME}/.gvm/scripts/gvm"
 
-# By default do not switch the Go version and use the default.
+# By default use the latest go version we have installed
 version="go1.17.2"
 
 # If GO_VERSION is defined, use it as the version.
--- a/images/prow-tests/runner.sh
+++ b/images/prow-tests/runner.sh
@@ -19,12 +19,8 @@ ORIGINAL_GOPATH="${GOPATH}"
 source "${HOME}/.gvm/scripts/gvm"
 
 # By default do not switch the Go version and use the default.
-version=""
-# Extract the go version if the project is using go mod.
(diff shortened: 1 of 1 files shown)


images/prow-tests/runner.sh
```diff
@@ -18,13 +18,9 @@ ORIGINAL_GOPATH="${GOPATH}"
...
 
-# By default do not switch the Go version and use the default.
-version=""
-# Extract the go version if the project is using go mod.
-if [[ -f "go.mod" ]]; then
-  version="go$(sed -n 's/^go //p' go.mod | tr -d '[:space:]')"
-  echo "Found go.mod file, using Go version '${version}' specified by it."
-fi
+# By default use the latest go version we have installed
+version="go1.17.2"
```
chizhg, line 22: Can we run the `gvm` command to set the default Go version when building the Docker image? Maintaining the version at two different places is not a good practice and can easily gets out of sync.

Below is the `gvm` help:
```
chizhg@chizhg-macbookpro:~$ gvm help
Usage: gvm [command]

Description:
  GVM is the Go Version Manager

Commands:
  version    - print the gvm version number
  get        - gets the latest code (for debugging)
  use        - select a go version to use (--default to set permanently)
  diff       - view changes to Go root
  help       - display this usage text
  implode    - completely remove gvm
  install    - install go versions
  uninstall  - uninstall go versions
  cross      - install go cross compilers
  linkthis   - link this directory into GOPATH
  list       - list installed go versions
  listall    - list available versions
  alias      - manage go version aliases
  pkgset     - manage go packages sets
  pkgenv     - edit the environment for a package set
```

coryrc:
The problem with this change is we use this to compile old releases. So if you have to patch older releases, you've making bigger changes (patch + new compiler) instead of just the intended patch change.

PR #2913, opened 2021-10-19
Fix code coverage

### LINKED ISSUE #2910: go-coverage is broken on all repo
All repos in knative and knative-sandbox has a broken go-coverage test now. The logs says the following:

```
2021/10/19 02:01:24 concerned Files=map[go.mod:true go.sum:true vendor/knative.dev/hack/presubmit-tests.sh:false vendor/modules.txt:false]
2021/10/19 02:01:24 repo path (../../../../../../home/prow/go/src/knative.dev/net-istio) is not a prefix of filepath (knative.dev/net-istio/pkg/reconciler/accessor/errors.go):
+ EXIT_VALUE=1
+ set +o xtrace
```

Example logs:
https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative_test-infra/2907/pull-knative-test-infra-go-coverage/1450087179987456000
https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative-sandbox_net-istio/789/pull-knative-sandbox-net-istio-go-coverage/1450280363568926720

benmoss:
GVM will set it to some crazy /root/.gvm directory, and now that we're using `gvm use` in our Dockerfile we can't rely on the "original gopath" to be "$HOME/go" anymore. 

#2905 introduced this problem by switching to `gvm use --default` in the Dockerfile

Fixes #2910

Is there any way we can test these changes before merging them in?

### DIFF
--- a/images/prow-tests/runner.sh
+++ b/images/prow-tests/runner.sh
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-ORIGINAL_GOPATH="${GOPATH}"
+ORIGINAL_GOPATH="${HOME}/go"
 
 source "${HOME}/.gvm/scripts/gvm"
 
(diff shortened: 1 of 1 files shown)


chizhg:
Thanks for the fix! Unfortunately there is not a good way to test it now. Let's merge it to test since the coverage test is broken anyway.

/lgtm
/approve

PR #2915, opened 2021-10-19
Remove transfigure jobs

chaodaiG:
Transfigure was a prior art that creates pullrequests in k8s/test-infra, for serving knative prow jobs on testgrid.k8s.io. Knative testgrid testgrid.knative.dev is now served by configurator only, removing transfigure which is no longer used


chaodaiG:
cc @chases2

PR #2916, opened 2021-10-19
Bugfix for #2914: Proper quoting for release jobs

### LINKED ISSUE #2914: Auto-release prow jobs calls release script with invalid parameters quoting
The prow job that runs auto-release job calls it like that:

```
./hack/release.sh --auto-release '--release-gcs knative-releases/kn-plugin-event' '--release-gcr gcr.io/knative-releases' '--github-token /etc/hub-token/token'
```

but in fact it should call it like that (so the flag parsing works properly):

```
./hack/release.sh --auto-release --release-gcs knative-releases/kn-plugin-event --release-gcr gcr.io/knative-releases --github-token /etc/hub-token/token
```

The first call fails on kn-plugin-event (see [build 1450497752243900416](https://prow.knative.dev/view/gs/knative-prow/logs/ci-knative-sandbox-kn-plugin-event-auto-release/1450497752243900416)), as I quote parameters in release script, properly (see [hack/release.sh#L45](https://github.com/knative-sandbox/kn-plugin-event/blob/595440b339fcad669facf51767a3cc5d82c27c63/hack/release.sh#L45)):

```bash
main "$@"
```

Current call works on other repos because they use shellcheck invalid [[SC2068]](https://github.com/koalaman/shellcheck/wiki/SC2068):

```bash
main $@
```

Prow should call release script properly, so I could continue using lint free code, and proper quoting should also work on other repos with lint error on that invocation.

cardil:
Fixes #2914

### DIFF
--- a/config/prod/prow/jobs/config.yaml
+++ b/config/prod/prow/jobs/config.yaml
@@ -7741,7 +7741,8 @@ periodics:
       - "gcr.io/knative-releases"
       - "--github-token"
       - "/etc/hub-token/token"
-      - "--branch release-0.23"
+      - "--branch"
+      - "release-0.23"
       volumeMounts:
       - name: hub-token
         mountPath: /etc/hub-token
@@ -7799,7 +7800,8 @@ periodics:
       - "gcr.io/knative-releases"
       - "--github-token"
       - "/etc/hub-token/token"
-      - "--branch release-0.24"
+      - "--branch"
+      - "release-0.24"
       volumeMounts:
       - name: hub-token
         mountPath: /etc/hub-token
@@ -7857,7 +7859,8 @@ periodics:
       - "gcr.io/knative-releases"
       - "--github-token"
       - "/etc/hub-token/token"
-      - "--branch release-0.25"
+      - "--branch"
+      - "release-0.25"
       volumeMounts:
       - name: hub-token
         mountPath: /etc/hub-token
@@ -7915,7 +7918,8 @@ periodics:
       - "gcr.io/knative-releases"
       - "--github-token"
       - "/etc/hub-token/token"
-      - "--branch release-0.26"
+      - "--branch"
+      - "release-0.26"
       volumeMounts:
       - name: hub-token
         mountPath: /etc/hub-token
@@ -8670,7 +8674,8 @@ periodics:
       - "gcr.io/knative-releases"
       - "--github-token"
       - "/etc/hub-token/token"
-      - "--branch release-0.23"
+      - "--branch"
+      - "release-0.23"
       volumeMounts:
       - name: hub-token
         mountPath: /etc/hub-token
@@ -8723,7 +8728,8 @@ periodics:
       - "gcr.io/knative-releases"
       - "--github-token"
       - "/etc/hub-token/token"
-      - "--branch release-0.24"
+      - "--branch"
+      - "release-0.24"
       volumeMounts:
       - name: hub-token
         mountPath: /etc/hub-token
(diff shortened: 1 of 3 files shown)


config/prod/prow/config_knative.yaml
```diff
@@ -1280,74 +1280,80 @@ periodics:
       limits:
         memory: 16Gi
   - nightly: true
-    args:
-    - --publish
-    - --tag-release
-    - --release-gcs knative-gcp-nightly
-    - --release-gcr gcr.io/knative-gcp-nightly
+    args: [
```
mgencur, line 1283: IMO, it would be better to keep the original style with dashes and not brackets because in the other file below (on other places) you're keeping the "dash" style.
```
args:
-
-
```

.editorconfig
```diff
@@ -0,0 +1,15 @@
+# EditorConfig is awesome: https://EditorConfig.org
```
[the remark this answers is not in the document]
cardil: Right. I thought to add this to help to format, especially the YAMLs. But, yeah, maybe better to add a linter to knobots.

.editorconfig
```diff
@@ -0,0 +1,15 @@
...
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+indent_style = space
+indent_size = 2
```
mgencur, line 12: Will this clash with `gofmt` ? indent size 2 and indent style space doesn't look like what is used everywhere.

config/prod/prow/jobs/config.yaml
```diff
@@ -21771,9 +22125,12 @@ periodics:
...
-      - "--release-gcs knative-releases/eventing-rabbitmq"
-      - "--release-gcr gcr.io/knative-releases"
-      - "--github-token /etc/hub-token/token"
+      - "--release-gcs"
+      - "knative-releases/eventing-rabbitmq"
+      - "--release-gcr"
+      - "gcr.io/knative-releases"
+      - "--github-token"
+      - "/etc/hub-token/token"
       - "--branch release-0.25"
```
mgencur, line 22134: This line `- "--branch release-0.25"` seems like not converted yet? Same on many other places

mgencur: Also there's a line such as this one: https://github.com/knative/test-infra/blob/357314fc91670759d6a7e64de3372129e8ebf23a/config/prod/prow/jobs/config.yaml#L214 (and others, just searched `args:` and found a few)
Should they all be converted?

config/prod/prow/config_knative.yaml
cardil: The other file `config/prod/prow/jobs/config.yaml` is generated, so we have no control how it's written. As it's generated the readability of it makes less of a concern to me.

I did such change here to keep parameters that relates to each other on the same line. I think it's easier to understand in this way:

```yaml
args: [
  '--tag-release',
  '--release-gcs', 'knative-gcp-nightly'
]
```
instead of:

```yaml
args:
- --tag-release
- --release-gcs
- knative-gcp-nightly
```
In the second example, I think it's harder to understand that `knative-gcp-nightly` is a value of `--release-gcs` flag.

config/prod/prow/jobs/config.yaml
cardil: The following is an entirely different case (best to talk on separate thread). And in fact it looks proper. As a whole line is passed as a single argument to the other script.

https://github.com/knative/test-infra/blob/357314fc91670759d6a7e64de3372129e8ebf23a/config/prod/prow/jobs/config.yaml#L214

cardil:
Maybe I should have said in the PR that quoting those args properly should work for other repos as well. I said that on the issue, here: https://github.com/knative/test-infra/issues/2914

> _proper quoting should also work on other repos with lint error on that invocation._

PR #2951, opened 2021-11-27
Drop istio stable from test

nak3:
https://github.com/knative-sandbox/net-istio/issues/782 described several reasons to remove `istio-stable`.
In short, we should maintain `head` and `latest` rather than adding `stable` which makes complication.


nak3:
Asking on Slack https://knative.slack.com/archives/CCSNR4FCH/p1638150980127000 if there is any objections.

julz:
I'll give others time to weigh in in case they want to, but sgtm

shinigambit:
lgtm

nak3:
Thank you. I will wait for one more day and unhold if nobody has objections.

nak3:
/hold cancel

I think this is ready to go now.

PR #2953, opened 2021-11-30
Add net-gateway-api job to serving CI

nak3:
This patch adds a job:
- to run gateway-api test by `/test pull-knative-serving-gateway-api-latest`
- to run automatically when `third_party/gateway-api-latest/` was changed.

`tls` is not added as net-gateway-api does not support it yet.

/cc @chizhg @markusthoemmes


markusthoemmes (approved):
/lgtm
/approve

PR #2995, opened 2022-01-07
:gift: Go lists tags

cardil:
This PR adds a command line tool to list Go build tools. It supports both old and new style of tags. See: https://go.googlesource.com/proposal/+/master/design/draft-gobuild.md

It will simplify the listing of tags, making it more bug-free (see: https://github.com/knative/hack/issues/134) and robust.

When having Go 1.17 using it simply by (for older versions we can use wrapper script to locally install tool):

```shell
tags="$(go run knative.dev/test-infra/tools/go-ls-tags@latest)"
```

Related to https://github.com/knative/hack/issues/134


tools/go-ls-tags/files/wd.go
```diff
@@ -0,0 +1,11 @@
+package files
+
+import "os"
+
+func WorkingDirectoryOrDie() string {
```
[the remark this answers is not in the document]
cardil: Fixed

coryrc:
A few minor requests:

Could `tools/go-ls-tags/tags/tagged.go` be put in a `testdata` folder? like https://github.com/kubernetes/gengo/tree/de9496dff47bef4ff8f45a56096bf50798cba2a5/testdata sort-of

Could you add a README.md with install instructions and purpose of the program and example?

cardil:
Re @coryrc:

> _Could tools/go-ls-tags/tags/tagged.go be put in a testdata folder? like https://github.com/kubernetes/gengo/tree/de9496dff47bef4ff8f45a56096bf50798cba2a5/testdata sort-of_
> 
> _Could you add a README.md with install instructions and purpose of the program and example?_

I like both your ideas! On it.

coryrc:
Ah, I knew there should be a knative example of testdata: https://github.com/knative/serving/tree/ad94366251dda975d2bc2bef1dedc5b83b6ad4d4/pkg/reconciler/configuration/config/testdata

coryrc:
Personally, I like using cobra for every single CLI program just so I am used to it, but I'll lgtm regardless.

chizhg:
@cardil do we still need this PR? I saw https://github.com/knative/hack/pull/135 already simplified the list_tag function a lot.

dprotaso:
Following up here - an alternative could be to make additional tags opt-in

This is currently done with a config file for our linter - https://github.com/knative/serving/blob/3446ca3f8d523d41774c555fafef7b05c319a4ac/.golangci.yaml#L4-L6

We could just read that file to get tags

kvmware:
Closing due to inactivity. Feel free to reopen when you get time.

cardil:
Re @chizhg:

> _do we still need this PR? I saw https://github.com/knative/hack/pull/135 already simplified the list_tag function a lot._

Yes. Couple of reasons:

- We use different complex Bash scripts to find those tags. That's cumbersome to maintain.
- The old tag syntax `+build` will eventually go away, and new syntax isn't so easy to parse with bash.

cardil:
/test pull-knative-test-infra-flaky-test-reporter-image-build

Infra error

cardil:
Fixes https://github.com/knative/hack/issues/219

kvmware:
I like the general idea! so LGTM, but someone with more time/depth should review closer

tools/go-ls-tags/cli/logger.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	"knative.dev/test-infra/pkg/logging"
+)
+
+func contextWithLogger() context.Context {
+	ctx := context.Background()
+	fallback := logging.FromContext(ctx)
+	return logging.WithLogger(context.Background(), newLogger(fallback))
```
mgencur, line 31: Nit: You can either save some line by 
```
fallback := logging.FromContext(context.Background())
```
or better use the `ctx` one line below as well:
```
return logging.WithLogger(ctx, newLogger(fallback))
```

tools/go-ls-tags/files/reader.go
```diff
@@ -0,0 +1,54 @@
...
+	"knative.dev/test-infra/tools/go-ls-tags/timeout"
+)
+
+// ReadLines will read lines from a file and pass each line to a func handler.
+func ReadLines(ctx context.Context, path string, fn func(line string) error) error {
+	f, err := os.Open(path)
+	if err != nil {
+		return err
+	}
+	defer func(f *os.File) {
```
mgencur, line 33: Interesting, why isn't it just `defer f.Close()` ?

tools/go-ls-tags/tags/files.go
```diff
@@ -0,0 +1,67 @@
...
+		return nil, errwrap(err)
+	}
+	return files, nil
+}
+
+func (l Lister) isExcluded(ctx context.Context, fullpath string) bool {
+	log := logging.FromContext(ctx)
+	relative, err := filepath.Rel(l.Directory, fullpath)
+	if err != nil {
+		log.Fatal(err)
```
mgencur, line 54: I guess this should either return `true, err` at this point or just end quickly and return `true` ? Not sure if continuing this function after error makes sense.

tools/go-ls-tags/tags/ignored.go
```diff
@@ -0,0 +1,65 @@
...
+		ignored = append(ignored, line)
+		return nil
+	})
+	if err != nil {
+		return nil, errwrap(err)
+	}
+
+	log.Infof("Ignoring tags: %q", ignored)
+	result := make([]string, 0, len(tags))
+OUTER:
```
mgencur, line 55: Hmm, what is it supposed to do? I suppose in this case, when the tag is ignored you will continue again from the very beginning and this will be an infinite loop? If the tag is supposed to be added only if it's ignored I'd change it to this (without using continue XYZ):
```
        for _, tag := range tags {
		tagIgnored := false
		for _, ignore := range ignored {
			if tag == ignore {
				tagIgnored = true 
			}
		}
		if !tagIgnored {
			result = append(result, tag)
		}
	}
```

tools/go-ls-tags/timeout/deadline.go
```diff
@@ -0,0 +1,21 @@
...
+	"context"
+	"errors"
+	"fmt"
+	"time"
+)
+
+var ErrTimeoutExceeded = errors.New("timeout exceeded")
+
+func WithDeadline(ctx context.Context, fn func() error) error {
+	deadline, ok := ctx.Deadline()
```
mgencur, line 13: Perhaps the idiomatic way to work with deadlines would be something like this (you could place it directly in reader.go and get rid of this deadline.go file:
```
func scanFile(ctx context.Context, f *os.File, fn func(line string) error) error {
	scanner := bufio.NewScanner(f)
	for scanner.Scan() {
		select {
		case <-ctx.Done():
			return ctx.Err()
		default:
			return fn(scanner.Text())
		}
	}

	if err := scanner.Err(); err != nil {
		return err
	}
	return nil
}
```
If the deadline is exceeded the `ctx.Err()` will return `DeadlineExceeded`

mgencur (commented):
Hey @cardil , a few nits here and there. Otherwise looks OK.

go.mod
```diff
@@ -20,13 +19,66 @@ require (
...
 	google.golang.org/api v0.70.0
-	google.golang.org/genproto v0.0.0-20220222213610-43724f9ea8cf // indirect
 	k8s.io/apimachinery v0.20.6
-	knative.dev/hack v0.0.0-20220823140917-8d1e4ccf9dc3
+	knative.dev/hack v0.0.0-20220823140717-f472a96ba780
 	sigs.k8s.io/boskos v0.0.0-20200729174948-794df80db9c9
 	sigs.k8s.io/yaml v1.2.0
 )
+
+require github.com/hashicorp/errwrap v1.0.0
```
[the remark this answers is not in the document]
cardil: I wonder why do Go formatted this like that :shrug:

tools/coverage/test_output/tmp_artifacts/TestProfiling_195043851/stdout.txt
```diff
@@ -0,0 +1,2 @@
+ok  	knative.dev/test-infra/tools/coverage/testTarget/subPkg1	0.001s	coverage: 66.7% of statements
```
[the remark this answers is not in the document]
cardil: Whoa! This shouldn't be committed! :bow:

tools/go-ls-tags/cli/logger.go
cardil: Yes, this is weird. The `context.Background()` should be executed only once...

tools/go-ls-tags/files/reader.go
cardil: To silence the linter in IDE.

With `defer f.Close()` the IDE complains that return value from `Close` isn't utilized. This way, it should be obvious, that omitting the return value is intentional.

The `f.Close() could return error in some cases. For example, you are writing to a disk that is rapidly filling. The contents of the write are small, so it gets buffered, and `f.Flush()` is called within `.Close()`. Then the error would be returned.

This file is used just for reading, so I believe there is no way an error could be returned.

tools/go-ls-tags/tags/files.go
cardil: From [`log.Fatal`](https://pkg.go.dev/go.uber.org/zap#SugaredLogger.Fatal) doc:

> _Fatal uses fmt.Sprint to construct and log a message, then calls os.Exit._

So, the function does not continue afterwards. Program end with error.

go.mod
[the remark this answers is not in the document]
mgencur: Well, I have no idea. The section below is for "indirect" dependencies. The section above for direct ones.

tools/go-ls-tags/files/reader.go
mgencur: Good to know. But it was not clear to me at first that this would be there to silence the linter. I know that Close() can return an error and calling defer with it directly makes it ignore the error. Personally, I'd use this to make it clear:
 ```
defer f.Close() //nolint // Ignore returned error
```

mgencur: Or maybe just a comment would be helpful. Thanks

tools/go-ls-tags/tags/files.go
mgencur: Cool. Thanks

tools/go-ls-tags/tags/ignored.go
mgencur: Nevermind, I think your code will work well.

cardil: Both versions will work.

I rarely use `LABELS:` in code, but I think this is justified. The `continue OUTER` statement will escape both loops, making the function a bit faster.

tools/go-ls-tags/cli/logger.go
cardil: Fixed in c71881c5f700a78b006f2a5835b6008222bb6a87

tools/coverage/test_output/tmp_artifacts/TestProfiling_195043851/stdout.txt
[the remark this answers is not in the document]
cardil: Fixed in fa27e5a903021f6aacb21ae97d57693523fb0f86

go.mod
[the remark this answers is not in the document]
cardil: Fixed in 0e9b1a3d4ec91c8dcd32de8789b6d05d8bebcd8f

Go formats the `require` block, but respect it you create additional one.

tools/go-ls-tags/timeout/deadline.go
cardil: Fixed in 7d95365e

mgencur:
I suppose it needs another rebase. Otherwise LGTM. Thanks !

cardil:
Rebased.

PR #3011, opened 2022-01-13
Configure Docker-in-Docker for prow-tests image build presubmit job

chizhg:
The job is using `docker` command to build the image, so we need to configure Docker-in-Docker for it.


upodroid (approved):
LGTM

Issue #2957, opened 2021-12-06
Build the prow-tests image from scratch instead of from the kubekins-e2e image


chizhg:
The current prow-tests image we use for running all Prow jobs is based on the kubekins-e2e image - see https://github.com/knative/test-infra/blob/main/images/prow-tests/Dockerfile#L15, which is large and contains a lot of tools that are not needed by Knative. The latest one of gcr.io/knative-tests/test-infra/prow-tests is 2.9 GB.

To reduce its size and make it more clear on the version control, it'll be more ideal to build the image from scratch and only install the tools that are needed. We can learn from how Istio built their images - https://github.com/istio/tools/blob/master/docker/build-tools/Dockerfile

PR #3012, opened 2022-01-14
Use cloudbuild to validate the image build in presubmit

chizhg:
I forgot that we tried building prow-tests image in presubmit before, but it had some issues with memory limit so we dropped it (see https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative_test-infra/2960/pull-knative-test-infra-prow-tests-image-build/1481787082845196288). 

This PR is a workaround to also verify the image build with Google cloud build. We may be able to switch back to `docker build` if https://github.com/knative/test-infra/issues/2957 is done.


upodroid (approved):
/lgtm

PR #3026, opened 2022-01-19
Add the label for cncf-infra migration, so we can use it across all Knative repos

chizhg:
/cc @upodroid

### DIFF
--- a/config/label_sync/labels.yaml
+++ b/config/label_sync/labels.yaml
@@ -89,6 +89,14 @@ default:
       target: both
       prowPlugin: label
       addedBy: humans
+    - color: 2596be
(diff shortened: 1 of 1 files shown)


upodroid:
The error doesn't make sense. s label doesn't exist

chizhg:
> 

Yeah this error looks weird.. 

> The error doesn't make sense. s label doesn't exist

There are two `size/S` labels in https://github.com/knative-sandbox/kn-plugin-func/labels?q=size.. Not sure why it happened.

chizhg:
@upodroid it's passing now. May I get a stamp please?

upodroid (approved):
LGTM

PR #3032, opened 2022-01-24
Run kubetest2 --down directly to save some costs

chizhg:
Update kubetest2 to the latest version, and run it with `--down` flag to directly release GCP projects after the test flows are finished, instead of the relying on the heartbeat timeout to automatically release projects. This'll help save some compute time thus reducing some costs.

/cc @upodroid


upodroid (approved):
LGTM

PR #3062, opened 2022-02-02
Additional jobs to run KafkaChannel tests with auth cases

aliok:
- With https://github.com/knative-sandbox/eventing-kafka-broker/pull/1816, we introduce capability to run e2e and reconciler tests of KafkaChannel with auth cases.
- We would like to run the entire test suites for KafkaChannel with different auth scenarios (TLS, SASL_SSL, SASL_PLAIN)
- As I wrote in the comment https://github.com/knative-sandbox/eventing-kafka-broker/pull/1816#issuecomment-1027993010, this PR won't cause any job failures with the HEAD at https://github.com/knative-sandbox/eventing-kafka-broker because it just adds new jobs with new command line arguments. Those arguments are currently ignored. So, the only "bad" thing will be that same thing (no auth) is tested multiple times.
- I want to merge https://github.com/knative-sandbox/eventing-kafka-broker/pull/1816 after this one is merged to see if my scripting changes there are actually good.

Fixes #

### DIFF
--- a/config/prow/config_knative.yaml
+++ b/config/prow/config_knative.yaml
@@ -504,6 +504,36 @@ presubmits:
     - --run-test
     - ./test/reconciler-tests.sh
   - go-coverage: true
+  - custom-test: channel-integration-tests-tls
+    needs-dind: true
+    args:
+    - --run-test
+    - ./test/e2e-tests.sh --channel-tls
+  - custom-test: channel-integration-tests-sasl-ssl
+    needs-dind: true
+    args:
+    - --run-test
+    - ./test/e2e-tests.sh --channel-sasl-ssl
+  - custom-test: channel-integration-tests-sasl-plain
+    needs-dind: true
+    args:
+    - --run-test
+    - ./test/e2e-tests.sh --channel-sasl-plain
+  - custom-test: channel-reconciler-tests-tls
+    needs-dind: true
+    args:
+    - --run-test
+    - ./test/reconciler-tests.sh channel-tls
+  - custom-test: channel-reconciler-tests-sasl-ssl
+    needs-dind: true
+    args:
+    - --run-test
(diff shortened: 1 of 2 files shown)


aliok:
cc @cardil @chizhg

config/prow/config_knative.yaml
```diff
@@ -504,6 +504,36 @@ presubmits:
...
+  - custom-test: channel-reconciler-tests-sasl-ssl
+    needs-dind: true
+    args:
+    - --run-test
+    - ./test/reconciler-tests.sh --channel-sasl-ssl
+  - custom-test: channel-reconciler-tests-sasl-plain
+    needs-dind: true
+    args:
+    - --run-test
+    - ./test/reconciler-tests.sh --channel-sasl-plain
```
cardil, line 536: But, the knative-sandbox/eventing-kafka-broker#1816 isn't merged. When we merge https://github.com/knative/test-infra/pull/3062 first, it's likely it will block other PRs!

Better, would be to spit knative-sandbox/eventing-kafka-broker#1816 in two. 

1. First introduce, new flags, and do NOP to make sure the test will not fail. 
2. Then merge https://github.com/knative/test-infra/pull/3062. 
3. Lastly, merge the knative-sandbox/eventing-kafka-broker#1816, so you will have a clear signal.

cardil (changes requested):

config/prow/config_knative.yaml
aliok: @cardil I explained pretty much the same thing in this PR's description at the top.

New flags don't have an effect. So, everything is already like you explained in your comment above.

cardil: Hope you're right about it.

cardil (approved):
/lgtm
/approve

PR #3066, opened 2022-02-03
Fix KafkaChannel jobs for eventing-kafka-broker

pierDipi:
Instead of using a flag, I'm using an environment variable
`EVENTING_KAFKA_BROKER_CHANNEL_AUTH_SCENARIO`
to distinguish between different auth contexts.


cardil (approved):
This should be safe

/lgtm
/approve

Issue #3060, opened 2022-02-01
create-build-cluster.sh tries to curl a non existent file


kvmware:
https://github.com/knative/test-infra/blob/main/config/prow/build-cluster/create-build-cluster.sh#L36 results in a 404.

psschwei:
It shows up for me

<img width="546" alt="image" src="https://user-images.githubusercontent.com/44698588/151997934-8c0cb9e3-85b3-4589-b0a2-663ac99cc613.png">

kvmware:
The link on that line points to a non existent file.

![image](https://user-images.githubusercontent.com/84356882/152005974-8cb19131-2c30-4e5e-91bf-82a85f8ea997.png)

psschwei:
Sorry, misread the issue title.

cardil:
Let's remove this. Invoking a shell script, directly downloaded from the network without checking for consistency, isn't safe.

kvmware:
I agree with the downloading a file and running it is not the best, but sorta seems like an idiom used for now (I am guessing to avoid duplicating and manually syncing files) https://github.com/GoogleCloudPlatform/oss-test-infra/blob/master/prow/knative/create-build-cluster.sh#L36

cardil:
At least, we should check the download sha, before executing it blindly. It's not an excuse that other people does that. 

Yes, people do that, then get hacked and stop doing that.

Issue #2982, opened 2022-01-02
Move Knative Prow config files from GoogleCloudPlatform/oss-test-infra repo to the knative/test-infra repo


chizhg:
As the outcome of [Proposal: Reducing Productivity Oncall Duties](https://docs.google.com/document/d/1tktwX9qHR9DKiA6yZCEMRYouNtScyRVW4ztVea-MJ1M/edit?resourcekey=0-G8IkKBaNP2s7xc2j92nvAQ#), Knative Prow config files were moved from GoogleCloudPlatform/oss-test-infra repo to the knative/test-infra repo, which are managed and maintained by the Prow team at Google.

Now that Knative has been donated to CNCF, we'd want to move the Knative Prow to be community-managed. This issue is for tracking the effort on moving the Prow config files back to the knative/test-infra repo and configuring autobump.

The steps will be:
1. Move everything from https://github.com/GoogleCloudPlatform/oss-test-infra/tree/master/prow/knative, which includes Prow deployment config files and some other script files.
2. Move the autobump Prow job from https://github.com/GoogleCloudPlatform/oss-test-infra/blob/master/prow/prowjobs/GoogleCloudPlatform/oss-test-infra/gcp-oss-test-infra-config.yaml#L167-L239, which is for automatically creating PRs to bump the image versions for Prow.
> Note there is also an autobump Prow job in the knative/test-infra repo (https://github.com/knative/test-infra/blob/main/config/prod/prow/knative-autobump-config.yaml), so we may consider combining them.
3. Move the postsubmit Prow job from https://github.com/GoogleCloudPlatform/oss-test-infra/blob/master/prow/prowjobs/GoogleCloudPlatform/oss-test-infra/gcp-oss-test-infra-config.yaml#L26-L58 to auto-deploy Prow.

More details can be found in the [Knative test infrastructure 1.0](https://docs.google.com/document/d/1U6wybYjCslEhsKOiennrwhoKZmSVnd1NGbFI8gnnnOg/edit#) doc

chizhg:
On a side note for this issue, @upodroid is working on https://github.com/knative/community/issues/443 to automate management of GCP resources, and likely we'll want all these config files to be in the same repo.

@upodroid is your plan to have the k8s tools in the test-infra repo?

upodroid:
1. https://github.com/kubernetes/k8s.io/tree/master/groups will live at https://github.com/knative/community/tree/main/groups
[...]

PR #3069, opened 2022-02-03
Port over oss-test-infra prow config and jobs.

kvmware:
Fixes https://github.com/knative/test-infra/issues/2982

Please read through the commits. I tried to make them very easy and bite size to review.
Note: I moved all prow related things from `config/prow` to `./prow`. The testgrid and gubernator (that needs to be removed) I left in `config/prow` for now. This might be confusing for someone who comes from kubernetes/test-infra as usually the root `prow` directory is used for prow code not yaml, but for people joining our team and a little documentation this new placement will make much more sense.

EDIT: should fix https://github.com/knative/test-infra/issues/3060 as well.


prow/create-build-cluster.sh
```diff
@@ -1,23 +1,46 @@
 #!/usr/bin/env bash
+
+# Copyright 2020 The Knative Authors
```
chizhg, line 3: Year should be 2022. Ditto for elsewhere that are newly licensed.

prow/config.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# Copyright 2021 The Knative Authors
+# Copyright 2022 The Knative Authors
```
chizhg, line 3: I think the rule is if we are only modifying the file, no need to change the copyright year.

chizhg (commented):
Could you fix the lint test?

prow/config.sh
kvmware: Ok so only when we are changing from `Google` to `Knative Authors`?

chizhg: Yep

PR #3074, opened 2022-02-03
Don't set SOURCE_DATE_EPOCH

dprotaso:
Since we're bumping community files and go.mod deps across our project it's
creating new nightlies with no actual code changes.

In serving this produces at least 5 PRs a day which bump the nightlies which take forever to
get through the merge pool.

This is unnecessary so this drops the use of `SOURCE_DATE_EPOCH` and
ko will now just zero out the timestamps.

### DIFF
--- a/images/prow-tests/runner.sh
+++ b/images/prow-tests/runner.sh
@@ -95,10 +95,6 @@ EOF
     gcloud auth activate-service-account --key-file="${GOOGLE_APPLICATION_CREDENTIALS}" || true
   fi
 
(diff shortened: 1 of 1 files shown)


dprotaso:
note: the other option is setting this to a specific date `SOURCE_DATE_EPOCH=1` - supposedly `0` can break stuff when I googled it

chizhg:
/lgtm
/hold
In case others have concerns.

kvmware:
No worries from me LGTM :+1:

PR #3082, opened 2022-02-11
Some fixes to the Prow config files after the move

chizhg:
1. Fix some broken Prow config files due to the move
2. Clean up some unused jobs
3. Fix some issues with `config.sh`
4. Move the autobump-prow jobs to a separate config file

/cc @kvmware


chizhg:
/hold

Just found some more issues. Working on the fix.

prow/jobs/custom/prow-autobump.yaml
```diff
@@ -1,88 +0,0 @@
-# Copyright 2022 The Knative Authors
```
kvmware, line 1: I sorta liked having the job separated out. It would be nice to actually split up test-infra.yaml into more bite size chunks in general. However, I might not understand the implication of having a bunch of custom jobs vs a monolithic yaml.

chizhg:
It looks Krsna is offline today. @dprotaso could you help stamp on this? It's mainly for fixing some Prow config issues.

/cc @dprotaso

kvmware (commented):
Is the performance test not being used?

PR #3083, opened 2022-02-11
More fixes to the Prow config files

chizhg:
1. Clean up some Makefiles
2. Fix regex for `pull-test-infra-validate-prow-yaml`
3. Remove `post-knative-prow-cluster-config-updater` as it's not needed anymore after we have `make -C prow/ deploy`
4. Fixes some config paths

/cc @kvmware

### DIFF
--- a/prow/knative-autobump-config.yaml
+++ b/prow/knative-autobump-config.yaml
@@ -17,7 +17,7 @@ extraFiles:
 prefixes:
   - name: "Prow"
     prefix: "gcr.io/k8s-prow/"
-    refConfigFile: "prow/cluster/deck_deployment.yaml"
+    refConfigFile: "config/prow/cluster/deck_deployment.yaml"
     repo: "https://github.com/kubernetes/test-infra"
     summarise: false
     consistentImages: true
--- a/config/Makefile
+++ b/config/Makefile
@@ -42,8 +42,6 @@ PROW_JOB_CONFIG                 ?= prow/jobs
 PROW_GCS                        ?= knative-prow
 PROW_CONFIG_GCS                 ?= gs://$(PROW_GCS)/configs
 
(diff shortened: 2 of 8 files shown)


prow/knative-autobump-config.yaml
```diff
@@ -17,7 +17,7 @@ extraFiles:
 prefixes:
   - name: "Prow"
     prefix: "gcr.io/k8s-prow/"
-    refConfigFile: "prow/cluster/deck_deployment.yaml"
+    refConfigFile: "config/prow/cluster/deck_deployment.yaml"
```
kvmware, line 20: There shouldn't be any config/prow folder?

chizhg: This is referring to the config file in the kubernetes/test-infra repo, for which the path is https://github.com/kubernetes/test-infra/blob/master/config/prow/cluster/deck_deployment.yaml

kvmware: Did I miss you put testgrid there I guess? However,  there is no cluster folder there. As intended

chizhg: Sorry not sure I understood your question...

The Prow team is using k8s Prow instance as a "staging" environment, where the config files are in https://github.com/kubernetes/test-infra/blob/master/config/prow/cluster. Downstream Prows such as Knative Prow are can use the `autobump` tool to bump to the same version as k8s Prow, and it parses the up-to-date image version from https://github.com/kubernetes/test-infra/blob/master/config/prow/cluster/deck_deployment.yaml, which is currently gcr.io/k8s-prow/deck:`v20220211-9762cdc433`

kvmware: I missed that it was referring to k8s/test-infra my bad sorry

PR #3087, opened 2022-02-13
Move kntest and rundk to the tools folder

chizhg:
We should also move `buoy` to the tools folder, but moving it will be more involved since it's now used by multiple GitHub Actions. We can do it in follow-up PRs.


kvmware:
/lgtm
Should this go under `testing/` instead?
/hold
Unhold it if tools you think is the right place :)

chizhg:
> /lgtm Should this go under `testing/` instead? /hold Unhold it if tools you think is the right place :)

TBH I don't like having too many root folders and that's why I'm moving them. I think `tools` is a good folder to hold all "binary tools", testing might cause confusion since there is a `test` folder and it has different meaning in other repos such as https://github.com/knative/pkg/tree/main/testing

PR #3088, opened 2022-02-13
Clean up some Prow jobs for knative-gcp

chizhg:
knative-gcp is under very low maintenance now and in the foreseeable future, dropping most of its periodic Prow jobs to save some resource.


kvmware:
/lgtm
My favorite kind of PR :)

PR #3111, opened 2022-02-22
Use kubetest2 to retry on GKE cluster creations

### LINKED ISSUE #3096: Use the retry logic in kubetest2 gke deployer to retry on cluster creations
We have logic in https://github.com/knative/test-infra/blob/main/pkg/clustermanager/kubetest2/gke.go#L134-L171 to retry on GKE cluster creation when there is GCE stockout. Now that kubetest2 gke deployer also supports smart retry, we should switch to use that instead.

This can also potentially save some compute resources since kubetest2 gke deployer will clean up the failed cluster before retrying.

chizhg:
Fixes #3096

### DIFF
--- a/pkg/clustermanager/kubetest2/gke.go
+++ b/pkg/clustermanager/kubetest2/gke.go
@@ -21,8 +21,8 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"os"
 	"os/exec"
-	"regexp"
 	"strconv"
 	"strings"
 
@@ -45,10 +45,9 @@ var (
 	// If one of the error patterns below is matched, it would be recommended to
 	// retry creating the cluster in a different region.
 	// - stockout (https://github.com/knative/test-infra/issues/592)
-	retryableCreationErrors = []*regexp.Regexp{
-		regexp.MustCompile(`.*does not have enough resources available to fulfill.*`),
-		regexp.MustCompile(`.*only \d+ nodes out of \d+ have registered; this is likely due to Nodes failing to start correctly.*`),
-	}
+	retryableErrorPatterns = `.*does not have enough resources available to fulfill.*` +
+		`,.*only \d+ nodes out of \d+ have registered; this is likely due to Nodes failing to start correctly.*` +
+		`,.*All cluster resources were brought up.+ but: component .+ from endpoint .+ is unhealthy.*`
 )
 
 // GKEClusterConfig are the supported configurations for creating a GKE cluster.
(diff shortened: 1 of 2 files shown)


pkg/clustermanager/kubetest2/gke.go
```diff
@@ -21,8 +21,8 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"os"
 	"os/exec"
```
kvmware, line 25: Since we are importing os couldn't we remove the `os/exec` import below and use `os.exec.*`?

kvmware: Not a blocker or anything just something to think about. It might be annoying to have `os.exec` everywhere.

pkg/clustermanager/kubetest2/gke.go
```diff
@@ -128,69 +126,45 @@ func Run(opts *Options, cc *GKEClusterConfig) error {
 		kubetest2Flags = append(kubetest2Flags, "--project="+cc.GCPProjectID)
 	}
 
-	return createGKEClusterWithRetries(kubetest2Flags, opts, cc)
-}
+	if cc.PrivateClusterAccessLevel != "" {
+		kubetest2Flags = append(kubetest2Flags, "--private-cluster-access-level="+cc.PrivateClusterAccessLevel)
+		kubetest2Flags = append(kubetest2Flags, "--private-cluster-master-ip-range="+strings.Join(cc.PrivateClusterMasterIPSubnetRange, ","))
```
kvmware, line 131: Are the network masks not needed? I see in the code removed the `--private-cluster-master-ip-range` seemes to expect the mask?

pkg/clustermanager/kubetest2/gke.go
```diff
@@ -128,69 +126,45 @@ func Run(opts *Options, cc *GKEClusterConfig) error {
...
-	for _, regx := range retryableCreationErrors {
-		if regx.MatchString(errMsg) {
-			return true
-		}
+	// Only save the metadata if it's in CI environment and meta data is asked to be saved.
+	if prow.IsCI() && opts.SaveMetaData {
+		return saveMetaData(cc, strings.Join(regions, ","))
 	}
-	return false
+	return nil
```
kvmware, line 156: My golang is not great, but isn't not having any return statement the same as returning nil?

pkg/clustermanager/kubetest2/gke.go
chizhg: Golang does not support the `.*` syntax so we'll have to write them like this, see https://golangdocs.com/import-in-golang

pkg/clustermanager/kubetest2/gke.go
chizhg: It requires returning explicit `nil`. Named return values can make this simpler, but it's not very commonly used - https://go.dev/tour/basics/7

pkg/clustermanager/kubetest2/gke.go
chizhg: They were needed before to prevent IP conflicts when retrying on the private cluster creations, but now we are switching to the retry feature for kubetest2 gke deployer, which already supports this, so this field can be safely dropped.

PR #3118, opened 2022-02-24
Run kubetest2 with --down arg

### LINKED ISSUE #3117: Run kubetest2 with --down arg to release the Boskos project directly
Currently the E2E test flow relies on the timeout setting in Boskos reaper to release the project resource, which will add ~ 15 minutes of delay for cleaning up the projects. We can add --down flag when running `kubetest2` command to directly release the projects in the end of E2E test flow, which'll help save some resource.

chizhg:
Fixes #3117 

Update kubetest2 to the latest version, and run it with --down arg to directly release GCP projects after the test flows are finished, instead of the relying on the heartbeat timeout to automatically release projects. This'll help save some compute time thus reducing some costs.

Previously attempt failed due to one bug in kubetest2, now the bug has been fixed by https://github.com/kubernetes-sigs/kubetest2/pull/180 so trying this again.

/cc @upodroid @kvmware

### DIFF
--- a/images/prow-tests/Dockerfile
+++ b/images/prow-tests/Dockerfile
@@ -175,7 +175,7 @@ COPY images/prow-tests/in-gvm-env.sh /usr/local/bin
 ############################################################
 FROM golang:1.17 AS external-go-gets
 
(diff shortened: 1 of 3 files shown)


upodroid (approved):
LGTM

PR #3148, opened 2022-03-09
Remove beta-prow-tests flows to cut some infra cost

chizhg:
I'm still working on switching to Istio config generator, but this is a cleanup we'll need anyways.

/cc @kvmware @upodroid


kvmware:
/lgtm
:heart:

PR #3150, opened 2022-03-10
Increase resources for eventing-kafka auto-release

devguyio:
It seems that the [auto-release job](https://prow.knative.dev/job-history/gs/knative-prow/logs/ci-knative-sandbox-eventing-kafka-auto-release) for eventing-kafka gets killed by the end of the job randomly, which might be due to resources  starvation. This PR sets the resources similar to serving and other repos resources.

### DIFF
--- a/prow/config_knative.yaml
+++ b/prow/config_knative.yaml
@@ -1636,6 +1636,11 @@ periodics:
     release: "1.3"
     needs-dind: true
   - auto-release: true
+    resources:
+      requests:
+        memory: 12Gi
+      limits:
+        memory: 16Gi
     needs-dind: true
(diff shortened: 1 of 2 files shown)


kvmware:
Could we trigger the autorelease job using this commit to ensure that this will indeed fix the current issue? I would hate for us to eat more resources if it is not needed

devguyio:
> 

I'm totally fine with that, how can we do that :)

devguyio:
/hold
Going to wait for knative-sandbox/eventing-kafka#1131 , maybe it does the trick and we avoid increasing the resources.

devguyio:
OOMed again. We need this PR
/unhold

PR #3154, opened 2022-03-12
downgrade ko to v0.9.3

dprotaso:
We're seeing tons of OOM errors

Related issue: https://github.com/google/ko/issues/642

Failures running nightly sample-controller

https://testgrid.k8s.io/r/knative-own-testgrid/sample-controller#nightly-release


dprotaso:
cc @upodroid 

/assign @chizhg

PR #3162, opened 2022-03-15
Apply ci-script before e2e-tests for s390x

BbolroC:
This is an extended work in line with https://github.com/GoogleCloudPlatform/oss-test-infra/pull/1523.

The PR includes the following changes:
- Apply the `s390x` specific ci-script fetched from the external secret before the e2e test is triggered
- Adjust a job schedule as a preparation for the upcoming new jobs

### DIFF
--- a/prow/config_knative.yaml
+++ b/prow/config_knative.yaml
@@ -656,7 +656,7 @@ periodics:
     - bash
     args:
     - -c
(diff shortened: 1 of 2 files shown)


prow/config_knative.yaml
```diff
@@ -670,12 +670,12 @@ periodics:
       secret: s390x-cluster1
   - custom-job: s390x-kourier-tests
     release: "1.0"
-    cron: 0 15 * * *
+    cron: 20 3 * * *
```
chizhg, line 673: Just out of curiosity, how did you determine these crons?

chizhg:
/ok-to-test
/lgtm
/approve
/hold

Feel free to unhold when you are ready to merge it.

prow/config_knative.yaml
BbolroC: Thanks for the question. The CI is not a big scale atm. I set the schedule just on my own (just simple math based on the job running time and an amount of jobs). :wink:

BbolroC: At this time, another 45 sets of custom jobs will be put in place. So I need to pack the current schedules as much as possible to accommodate them.

PR #3165, opened 2022-03-16
Add 1.2 dot release job for knaitve/eventing

pierDipi:

we aren't getting dot releases for eventing release-1.2 branch
since there is no job for 1.2.

See individual commits for details.

### DIFF
--- a/prow/config_knative.yaml
+++ b/prow/config_knative.yaml
@@ -1064,6 +1064,8 @@ periodics:
   - branch-ci: true
     release: "1.1"
   - branch-ci: true
+    release: "1.2"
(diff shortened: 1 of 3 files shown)


prow/config_knative.yaml
```diff
@@ -1063,6 +1063,8 @@ periodics:
     release: "1.0"
   - branch-ci: true
     release: "1.1"
+  - branch-ci: true
```
chizhg, line 1066: Please also add the dot-release job for release 1.2 - above L103.

pierDipi: Done

PR #3168, opened 2022-03-16
Fix a bug on missing ingressendpoint for s390x custom job

BbolroC:
This PR is to fix a bug from https://github.com/knative/test-infra/pull/3162 where an e2e test option `ingressendpoint` is missed.


BbolroC:
/assign @chizhg 

Thanks!

PR #3174, opened 2022-03-21
Move knativeteam-groups Prow jobs to a separate file and delete unused backup Prow job

chizhg:
/cc @upodroid @kvmware

### DIFF
--- a/prow/jobs/custom/test-infra.yaml
+++ b/prow/jobs/custom/test-infra.yaml
@@ -104,25 +104,6 @@ presubmits:
       - name: test-account
         secret:
           secretName: test-account
-  knative/community:
-  - name: pull-knativeteam-groups-test
-    decorate: true
-    run_if_changed: "^groups/"
-    branches:
-    - ^main$
-    spec:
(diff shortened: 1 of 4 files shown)


prow/jobs/custom/test-infra.yaml
```diff
@@ -154,32 +135,6 @@ periodics:
         requests:
           cpu: 100m
           memory: 1Gi
-- cron: "15 9 * * *"
-  name: ci-knative-backup-artifacts
```
kvmware, line 158: Was it intentional to remove this job?

kvmware: ah, seems like unit tests caught this was missing. If it is an intentional removal then looks like unit tests will need to be updated as well.

chizhg: Yes, it has been failing for a long while - https://prow.knative.dev/?job=ci-knative-backup-artifacts

upodroid (commented):
👍

Issue #2311, opened 2020-07-20
[oncall toil] Use inrepoconfig for prow jobs in all repos


chaodaiG:
Inrepoconfig(https://github.com/kubernetes/test-infra/blob/master/prow/inrepoconfig.md) allows each repo setting up prow jobs in its own repo instead of everything in test-infra, the benefits:

- Promotes self-service
- Presubmit flows get validated in PRs before applying

To make sure there is no regression doing this, it might be required to refactor prow job config so that it still serves each repo as it is today

/kind enhancement

PR #3175, opened 2022-03-22
Use istio prowgen to generate Prow jobs

### LINKED ISSUE #2986: Use the Istio Prow config generator tool, and deprecate the bespoke Knative tool
There is a plan to make https://github.com/istio/test-infra/tree/master/prow/config as the default config templating solution for Prow. After that happens, it'll be ideal to use it to generate the Prow config files for Knative, and deprecate the bespoke config generator tool in https://github.com/knative/test-infra/tree/main/tools/config-generator

More details can be found in the [Knative test infrastructure 1.0](https://docs.google.com/document/d/1U6wybYjCslEhsKOiennrwhoKZmSVnd1NGbFI8gnnnOg/edit#) doc

chizhg:
Fixes #2986

This PR is a huge overhaul of the all the Prow job and TestGrid config files, as well as the config generation process. The most important change is replacing the gigantic bespoke config-generator tool with a more lightweight tool, and the README is at https://github.com/chizhg/test-infra/blob/e96e34f0816b9d4c58d0f832656233e9f0ad6f2f/tools/configgen/README.md

This PR is probably too large to be unfeasible for reviewing, but I don't have a good approach to break it down. Since I'm confident the major flow is not broken, and I'm pretty sure it's much much better than the previous tool, I would suggest us merging it for now and iterating on it with follow-up PRs.

/cc @kvmware @upodroid


kvmware:
`+965,521 −62,832` is a little worrisome. I notice that there is lots more go vendored packages being pulled in. Why so much more things? 900 thousand! more things?

chizhg:
> `+965,521 −62,832` is a little worrisome. I notice that there is lots more go vendored packages being pulled in. Why so much more things? 900 thousand! more things?
> 
> EDIT: Does not seem like a more lightweight tool.

Most of the files were go vendored packages since I added `iistio.io/test-infra/tools/prowgen` as a dependency which introduced a number of indirect dependencies, also since we run `go mod vendor` for this repo so the vendored files had to be added.

I just moved it to a separate Go module and removed the vendored packages, now this PR should be much smaller.

kvmware:
lgtm
@evankanderson brought up this point with google groups work on an earlier PR and I think it applies here to. Is there a way for us just to use the tool (released version) instead of building it?

chizhg:
> lgtm @evankanderson brought up this point with google groups work on an earlier PR and I think it applies here to. Is there a way for us just to use the tool (released version) instead of building it?

We can build it into a Docker image and publish it to GCR with a postsubmit Prow job, then this tool can be used elsewhere, but I don't think it simplifies things unless we enable in-repo configs for every Knative repo, see https://github.com/knative/test-infra/issues/2311, since the only place we need this tool is knative/test-infra (or knative/infra after we move the config files).

upodroid:
Why don't you consume the code at https://github.com/istio/test-infra/tree/master/tools/prowgen directly?

Something like this:
```
go install istio.io/test-infra/tools/prowgen@v0.0.0-20220322144828-745ae0daa7d5
prowgen --foo=bar something
```

chizhg:
> Why don't you consume the code at https://github.com/istio/test-infra/tree/master/tools/prowgen directly?
> 
> Something like this:
> 
> ```
> go install istio.io/test-infra/tools/prowgen@v0.0.0-20220322144828-745ae0daa7d5
> prowgen --foo=bar something
> ```

The major reason is for Knative Prow config generation, there is some specific logic that is not generalized and is only for Knative, e.g. automatically calculating and generating cron schedules for periodic Prow jobs, to inject this logic, we need custom preprocessor to the meta config files which needs to be implemented by importing the structs defined in https://github.com/istio/test-infra/blob/master/tools/prowgen/pkg/spec/spec.go.

Since for the above reason, we already imports the istio.io/test-infra repo, it's better to call the `prowgen` tool as a library to keep version consistency.

Issue #2985, opened 2022-01-02
Remove the build and unit tests run by Prow, and use the GitHub actions to block on presubmit


chizhg:
Currently there are some overlaps between the Prow test jobs and GitHub actions, the most obvious ones are the build and unit tests.

Since the tests run with GitHub actions are faster and easier to configure, it's more ideal to keep them and remove the Prow jobs.

The work will include:
1. Add a GitHub action in https://github.com/knative-sandbox/knobots to update the branchprotector rule config in https://github.com/GoogleCloudPlatform/oss-test-infra/blob/a76b9d08758ad62b15d0b94ff107f053218f642e/prow/knative/config.yaml#L187-L223 to mark these GitHub actions as required status checks
2. Auto-merge the PRs created by step 1
3. Remove the Prow build and unit test jobs

More details and discussions can be checked from the [Knative test infrastructure 1.0](https://docs.google.com/document/d/1U6wybYjCslEhsKOiennrwhoKZmSVnd1NGbFI8gnnnOg/edit#) doc

Note: one catch for this work is Tide merge queue does not support GitHub Actions, so there'll be a small chance to break them if two conflicting PRs are merged together. But since the chance is pretty low, and most of the time it'll be easy to fix the broken GitHub Actions, this should be fine before the auto-merge queue support.

upodroid:
@dprotaso @kvmware We have an interesting problem. It seems github actions doesn't report contexts in their status.

Sample PR: https://github.com/knative/serving/pull/12775

Use the gh cli to explore the API responses.

```
 REDACTED  MCW0CDP3YY  ~  Desktop  Git  mgcp-infrastructure   main  6+  1⚑  $   gh api repos/knative/serving/pulls/12775 | jq | grep status
  "statuses_url": "https://api.github.com/repos/knative/serving/statuses/efc837471c066ff8f0424156e9f6295b55a5ab7f",
      "statuses_url": "https://api.github.com/repos/knative-automation/serving/statuses/{sha}",
      "statuses_url": "https://api.github.com/repos/knative/serving/statuses/{sha}",
    "statuses": {
      "href": "https://api.github.com/repos/knative/serving/statuses/efc837471c066ff8f0424156e9f6295b55a5ab7f"
[...]

PR #3187, opened 2022-03-24
Mark gh-actions as required for eventing/serving repos initially.

upodroid:
Part of https://github.com/knative/test-infra/issues/2985

/cc @chizhg @dprotaso 

/hold Need to merge https://github.com/knative/actions/pull/11 and find out the new context for the unit test for it.

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -219,6 +219,11 @@ branch-protection:
         contexts:
         - cla/google
         - tide
+        - "Build (1.17.x, ubuntu-latest)"
+        - "Auto-format and Check (goimports)"
+        - "Auto-format and Check (gofmt)"
+        - "Unit Tests (1.17.x, ubuntu-latest)"
+        - "Lint"
       include:
       - "^main$"
       - "^master$"
@@ -229,10 +234,6 @@ branch-protection:
       repos:
         .github:
           protect: false
-        serving:
-          required_status_checks:
-            contexts:
-            - "Build (1.17.x, ubuntu-latest)"  
 
     google:
       repos:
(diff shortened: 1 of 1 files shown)


chizhg:
Please fix the lint error.

chizhg:
In the longer term we would want this to be automated, but it's okay for now.

/lgtm

chizhg:
@googlebot

dprotaso:
note: this will change as the re-usable actions gets rolled out 

Secondly it's very fragile to name changes etc.  I wonder if it's ok just let admins of the repo add actions manually?

prow/config.yaml
```diff
@@ -219,6 +219,11 @@ branch-protection:
         contexts:
         - cla/google
         - tide
+        - "Build (1.17.x, ubuntu-latest)"
```
chizhg, line 222: Oh I just realized this will change will apply to all repos in knative org, which is not correct since some of the repos do not enable these GitHub Actions.

chizhg:
> note: this will change as the re-usable actions gets rolled out
> 
> Secondly it's very fragile to name changes etc. I wonder if it's ok just let admins of the repo add actions manually?

The problem is the branch-protector job will delete the rules that are not declared in this file. 

I agree the name changes could be fragile, but not sure how the re-usable actions rollout impacts this since we only need to sync the action names, could you elaborate?

dprotaso:
>  but not sure how the re-usable actions rollout impacts this since we only need to sync the action names, could you elaborate?

I think switching to reusuable actions will change the name

upodroid:
I'll wait for the knobots and the reusable workflows to settle and take another look.

kvmware:
Closing due to inactivity. Feel free to reopen when you have time.

Issue #2962, opened 2021-12-09
Convert Knative Prow into a GitHub app


chizhg:
As the number of repos in Knative grows, there'll be issues in authenticating to the GitHub APIs with the GitHub Personal Access Token, since the rate limit is always 5000 per hour. To make it more scalable and secure, it'll be nice to convert Knative Prow into a GitHub app.

After it's converted into a GitHub app, the Prow jobs for peribolos, label_sync and branchprotector should also be switched to authenticate as GitHub app, to bypass the token rate limit.

More contexts in https://docs.google.com/document/d/1q9NagH0xB2vcuK-TC_X6V_LHc0syBrNrLqiuAQ8zGWY/edit#heading=h.66y4kqbj468a

upodroid:
kubernetes's prow instance has been running as Github App for sometime now.

References: 
- https://github.com/kubernetes/test-infra/issues/10423
- https://github.com/kubernetes/test-infra/pull/23630
- https://github.com/kubernetes/test-infra/blob/master/prow/getting_started_deploy.md

chizhg:
Followed the documentations to manually apply the changes and manually did the test.

Below are the PRs needed for the migration:
- https://github.com/GoogleCloudPlatform/oss-test-infra/pull/1553
- https://github.com/GoogleCloudPlatform/oss-test-infra/pull/1554
- https://github.com/knative/test-infra/pull/3198

chizhg:
https://github.com/knative/test-infra/pull/3200 for fixing the GitHub ops Prow jobs.

chizhg:
The migration has been done. Marking this issue as fixed.

PR #3188, opened 2022-03-25
Update timeout for branchprotector to 5h

chizhg:
It cannot be finished within 3h - https://prow.knative.dev/?job=ci-knative-test-infra-branchprotector

/cc @kvmware @upodroid


chizhg:
We probably need to prioritize https://github.com/knative/test-infra/issues/2962, since the reason for this taking so long is the GitHub API queries kept getting throttled.

upodroid (approved):
LGTM

PR #3189, opened 2022-03-25
Fix one config error for flaky-test-reporter

chizhg:
/cc @kvmware


chizhg:
@kvmware can I get another `/lgtm`? It looks the previous one didn't take into effect.

PR #3191, opened 2022-03-25
Add back the release-job-syncer tool to sync Prow jobs when there is new release

chizhg:
This is one follow-up for https://github.com/knative/test-infra/issues/2986 since the `release-job-syncer` tool was broken by switching to `configgen`.

This is for automatically creating PRs like https://github.com/knative/test-infra/pull/3160

/cc @kvmware 

FYI @BbolroC
This is the PR I mentioned to you. It contains some specific logic to sync s390x Prow jobs when there is a new release.


BbolroC:
@chizhg Then I could make a PR to set the previous time schedules after this, right?

chizhg:
> @chizhg Then I could make a PR to set the previous time schedules after this, right?

Yes, that is correct.

PR #3198, opened 2022-03-28
Move the label-sync, branchprotector, peribolos Prow jobs to authenticate as GitHub app

chizhg:
/cc @kvmware @upodroid


upodroid (approved):
LGTM

PR #3219, opened 2022-03-31
Delete config.sh and replace with actual Prow jobs

chizhg:
/cc @kvmware


chizhg:
This will need a force merge since it inevitably breaks the presubmit test.

kvmware:
Who should we ping for a force push? I handed my superuser permissions over to the next release lead rotation

Issue #3176, opened 2022-03-22
Get rid of go mod vendor across all Knative repos


chizhg:
Currently all the Knative projects are using `go mod vendor`, resulting in a huge amount of files in the `vendor` directory - e.g. https://github.com/knative/test-infra/tree/main/vendor, which can add a lot of noise when reviewing PRs that introduce extra dependencies. 

AFAICT the only reason we need them is we need to vendor a bunch of non-go files - e.g. https://github.com/knative/test-infra/tree/main/vendor/knative.dev/hack, so if we can use another automated way to update them, for example, by using a knobot Action to directly copy these files, these vendored packages can be deleted.

/cc @kvmware @upodroid @cardil

dprotaso:
> which can add a lot of noise when reviewing PRs that introduce extra dependencies.

If this is the sole reason I would split this into a separate commit and review the commits independently 

Just as an FYI dropping this practice will be quite a loot of work since the majority reason we vendor deps is forpulling in non-go code artifacts 

Some examples:
1. code-gen scripts from https://github.com/kubernetes/code-generator. 
2.  reconciler gen scripts in `knative.dev/pkg` that are used for generating injection clients.
3. `knative.dev/hack` scripts being propagated everywhere

Another extreme case is carrying patches for dependencies while we wait for them to be merged upstream. We've done that before for some K8s issues.

I actually _do like_ seeing the vendored dependencies in PRs since it gives reviewers the ability to sanity check the changes being pulled in. This makes it easier to debug at times. If we drop vendoring I would hope we put an effort to improve our tooling around visualizing the changes - ie. see this dependabot PR - https://github.com/google/ko/pull/663

chizhg:
> If this is the sole reason I would split this into a separate commit and review the commits independently
That's true, but we don't have a way to enforce it and usually it's tedious (at least for me) to keep that in mind.

If we keep vendor/, I found the pull request file tree does help reduce the noise though:
![image](https://user-images.githubusercontent.com/19399934/161338864-e3fdf02a-920e-4b3a-8f57-1fee6be18a51.png)

cardil:
I will list here a couple of reasons why we should get rid of the vendor, and address some of the concerns people often raise.

## Reasons for removal

### Maintenance burden for productivity

[...]

PR #3222, opened 2022-04-01
Fix cleanup Prow job and setup alert

chizhg:
1. `github.com/google/go-containerregistry` needs to be updated to the latest version. It's now using a very old one.
2. `gcloud auth configure-docker` command is needed for running the `cleanup` tool
3. Configure the `cleanup` Prow job to send alerts to Slack if it fails

/cc @upodroid @kvmware


upodroid:
This is not needed. We already authenticate to gcr.io

https://github.com/knative/test-infra/blob/23df24633ad4cf3bc45f96d069a08593bb24b277/images/prow-tests/Dockerfile#L242

Unless we are logging in to other registries.

chizhg:
> This is not needed. We already authenticate to gcr.io
> 
> https://github.com/knative/test-infra/blob/23df24633ad4cf3bc45f96d069a08593bb24b277/images/prow-tests/Dockerfile#L242
> 
> Unless we are logging in to other registries.

I did some more tests, and have confirmed we need two changes together to make it working:
1. github.com/google/go-containerregistry dep version needs to be updated. It's now using a very old version.
2. `gcloud auth configure-docker` needs to be run after `gcloud auth activate-service-account`.

I'm not entirely sure about the reason, but I have tested it at https://prow.knative.dev/view/gs/knative-prow/logs/ci-knative-cleanup-chizhg-test/1509776600835559424 and it's working.

chizhg:
@upodroid can I get an LGTM if this looks good to you? Thanks!

upodroid (approved):
LGTM, large diff though

chizhg:
Yeah, mostly vendor/ updates though. There are some discussions in https://github.com/knative/test-infra/issues/3176 about it.

PR #3227, opened 2022-04-01
boskos_resources.yaml config file is part of Boskos deployment so moving it back

chizhg:
/cc @upodroid


upodroid (approved):
LGTM, didn't think this file was used by prow

PR #3237, opened 2022-04-05
Add the Prow job to run config-bootstrapper periodically

### LINKED ISSUE #3220: Configure config-bootstrapper Prow job to do full update for Prow configs
Currently Prow is using the [config-updater](https://github.com/kubernetes/test-infra/tree/master/prow/plugins/updateconfig) plugin to do the Prow configs update. But since it relies on GitHub webhook, and only does incremental updates, the configmaps on the Prow cluster will be out of sync if a GitHub webhook event gets lost and it'll be very tedious to recover when it happens.

To mitigate this, one recommendation is to run [config-bootstrapper](https://github.com/kubernetes/test-infra/tree/master/prow/cmd/config-bootstrapper) as a periodic Prow job to do the full update. Ideally it can be scheduled to run in off-hours to prevent interruptions.

See https://github.com/kubernetes/test-infra/issues/24485 for more contexts.

chizhg:
Fixes #3220

/cc @kvmware @upodroid

### DIFF
--- a/prow/jobs/custom/autodeploy-prow.yaml
+++ b/prow/jobs/custom/autodeploy-prow.yaml
@@ -43,3 +43,51 @@ postsubmits:
         - deploy
         - deploy-build
         - deploy-monitoring
+
+periodics:
+# Run config-bootstrapper Prow job every day to make sure the configmaps on the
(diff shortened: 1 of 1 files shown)


upodroid (approved):
LGTM

Issue #2965, opened 2021-12-10
Auto merge PRs created by the knative-automation bot


chizhg:
For PRs created by [knobot](https://github.com/knative-sandbox/knobots) GitHub actions, such as:
- https://github.com/knative/client-pkg/pull/24
-  https://github.com/knative/client-pkg/pull/23
- https://github.com/knative/client-pkg/pull/22

At this point there shouldn't be issues on skipping the code review and auto-merging them.

This can be achieved by adding a label such as `skip-review` to the PRs and configuring Tide. One example we previously did for knative/test-infra repo is in https://github.com/knative/test-infra/blob/66d6a1f645ff585bfd1bce0eee0cb3446c7405b9/prow/config.yaml#L164-L171

xtreme-sameer-vohra:
I'd want to be cautious for https://github.com/knative/client-pkg/pull/24 type PRs. If they get merged without all the checks passing and leave repos in a broken state that should be considered undesirable.

chizhg:
> If they get merged without all the checks passing and leave repos in a broken state that should be considered undesirable.

We should set these GitHub Actions checks as required, ideally by automation with branch-protector https://github.com/knative/test-infra/blob/c2f5bece9217fe1a5377dbdf4777fe9d5bc81fd3/prow/config.yaml#L194-L243. 

@shinigambit 's work for https://github.com/knative/test-infra/issues/2985 is for build and unit tests, but we can extend it to work for any custom GitHub Actions checks.

chizhg:
The other two are more involved and more dangerous to enable auto-merge, and some prerequisite work will be needed, at least https://github.com/knative/test-infra/issues/2985. At this point I don't have bandwidth to take on this so unassgin myself.

Please feel free to continue on this issue if anyone is interested.

/unassign

PR #3245, opened 2022-04-07
Enable auto-merge for PRs created by knative-automation robot

chizhg:
With this change, if a PR is created by `knative-automation` robot **and** it has `skip-review` label **and** all the required tests pass, it will be automatically merged without requiring manual approval.

Part of https://github.com/knative/test-infra/issues/2965

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -141,7 +141,7 @@ tide:
   - orgs:
     - "knative"
     - "knative-sandbox"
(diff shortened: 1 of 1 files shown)


dprotaso:
Actually - I wonder about repos that don't have required checks configured (ie. if they're using only actions and not prow)- this could merge in bad PRs no?

Issue #3253, opened 2022-04-11
Update Knative Prow to use the new monitoring stack


chizhg:
As per the Prow owners, Knative Prow is currently using a legacy monitoring stack (in https://github.com/knative/test-infra/tree/5094d5004d5b9188e508024789d0e0dee2b1a866/prow/cluster/monitoring), which has been deprecated in favor of https://github.com/GoogleCloudPlatform/oss-test-infra/tree/master/prow/oss/terraform.

Since the legcy monitoring stack is not actively being maintained, and the new one is easier to setup and measure more valid metrics, Prow team's recommendation is to make the switch to it.

Please reach out to Prow team via the k8s/Prow slack channel if there are any questions.

/cc @kvmware @upodroid

PR #3254, opened 2022-04-11
Update Prow alerts recipient to kn-infra-gcp-org-admins@knative.team

chizhg:
I have confirmed the alerts are working well, so updating the recipient to the group that has access to the infra, which is kn-infra-gcp-org-admins@knative.team

In the future Knative Prow can make the switch to use the new monitoring stack, see https://github.com/knative/test-infra/issues/3253, with which alerts can be configured to send to the Productivity Slack channel.

/cc @kvmware @upodroid


upodroid (approved):
LGTM, I'll go and work on that issue soon

PR #3267, opened 2022-04-18
Fix label-sync presumbit Prow job, when config on community repo

kvmware:
Since we moved `labels.yaml` to community the label_sync prow job is failing which seemed to happen to peribolos before as well.

### DIFF
--- a/prow/jobs/custom/label-sync.yaml
+++ b/prow/jobs/custom/label-sync.yaml
@@ -36,25 +36,19 @@ presubmits:
         # Set --confirm=false to only validate the configuration file.
         - --confirm=false
         - --orgs=knative-sandbox,knative
-        - --github-app-id=$(GITHUB_APP_ID)
-        - --github-app-private-key-path=/etc/github/cert
-        - "--github-hourly-tokens=2000"
-        - "--github-allowed-burst=1000"
+        # We have to still use the GitHub oauth token for authentication, since
+        # https://github.com/knative/test-infra/pull/3198 change and having
+        # --confirm=false. This happens with peribolos as well.
+        - --token=/etc/github/oauth
         - --debug
         volumeMounts:
-        - name: github-token
+        - name: oauth
           mountPath: /etc/github
           readOnly: true
-        env:
-        - name: GITHUB_APP_ID
-          valueFrom:
-            secretKeyRef:
-              name: github-token
-              key: appid
       volumes:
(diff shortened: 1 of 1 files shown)


prow/jobs/custom/label-sync.yaml
```diff
@@ -36,25 +36,34 @@ presubmits:
         # Set --confirm=false to only validate the configuration file.
         - --confirm=false
         - --orgs=knative-sandbox,knative
-        - --github-app-id=$(GITHUB_APP_ID)
-        - --github-app-private-key-path=/etc/github/cert
-        - "--github-hourly-tokens=2000"
-        - "--github-allowed-burst=1000"
+        # Below had to be comment out when we moved labels.yaml to community.
```
chizhg, line 39: This issue isn't related to the repo move. It is due to the change I made in https://github.com/knative/test-infra/pull/3198, and it only happens to the presubmit Prow job when `--confirm=false`.

prow/jobs/custom/label-sync.yaml
```diff
@@ -36,25 +36,34 @@ presubmits:
...
+            # secretKeyRef:
+              # name: github-token
+              # key: appid
       volumes:
-      - name: github-token
+      #- name: github-token
+      - name: oauth
         secret:
-          secretName: github-token
+          secretName: github-token-for-label-sync
```
chizhg, line 66: Per https://github.com/knative/test-infra/pull/3198/files#diff-7bcbed15b9fafd436e2545bb373c5d97fc1640d4b8cc7d333caec232e880fb22L126, it was `github-token` before the change.

prow/jobs/custom/label-sync.yaml
```diff
@@ -36,25 +36,34 @@ presubmits:
...
+        - name: oauth
           mountPath: /etc/github
           readOnly: true
-        env:
-        - name: GITHUB_APP_ID
-          valueFrom:
-            secretKeyRef:
-              name: github-token
-              key: appid
+        # env:
```
chizhg, line 56: I would suggest removing these instead of commenting them in yaml config for cosmetic reasons, unless you are planning to uncomment them soon. Mostly my personal preference, though.

prow/jobs/custom/label-sync.yaml
kvmware: done also included the `--token`` option

PR #3283, opened 2022-04-20
Bump kind version

kvmware:
Fixes # https://github.com/knative/hack/pull/169#issuecomment-1097409577
although I think that maybe we need that k8s min version as it does not look like a kind issue but just an older version of k8s version we are using.


kvmware:
cc @chizhg

PR #3287, opened 2022-04-25
Move generated Prow Jobs over to new `prow-build` cluster

upodroid:
~~Busy~~ PR but:
- ~~changed default gcs-credential-secret. The new secret doesn't exist on prow-trusted so I overrode it manually on each manual prowjob running on prow-trusted.~~
- SOme Generated jobs are moved to prow-build cluster and new secrets are being loaded.


prow/jobs/custom/autodeploy-prow.yaml
```diff
@@ -18,8 +18,10 @@ postsubmits:
     cluster: prow-trusted
     run_if_changed: '^prow/cluster/'
     decorate: true
+    decoration_config: # will remove this temp override
```
kvmware, line 21: Assuming it is intentional that this is opposite from prow/config.yaml ?

prow/jobs/custom/autodeploy-prow.yaml
```diff
@@ -18,8 +18,10 @@ postsubmits:
...
+      gcs_credentials_secret: "gcs-upload"
     branches:
-    - main
+    - test-account
```
chizhg, line 24: This does not look right..

prow/jobs/custom/autodeploy-prow.yaml
```diff
@@ -55,7 +57,7 @@ periodics:
   extra_refs:
   - org: knative
     repo: test-infra
-    base_ref: main
+    base_ref: test-account
```
chizhg, line 60: Ditto.

prow/jobs/custom/autodeploy-prow.yaml
chizhg: If that's the plan, shouldn't `gcs-upload` be replaced as `test-account` here?

prow/config.yaml
```diff
@@ -32,7 +32,7 @@ plank:
       gcs_configuration:
         bucket: "knative-prow"
         path_strategy: "explicit"
-      gcs_credentials_secret: "test-account"
+      gcs_credentials_secret: "gcs-upload"
```
chizhg, line 35: IIRC Prow supports configuring `default_decoration_configs` per build cluster, AKA replacing `*` to be build cluster names, so you won't have to override this per job.

You can check the Prow docs or config spec to check whether Prow supports this and/or how to configure it.

prow/jobs/generated/knative-sandbox/async-component-main.gen.yaml
```diff
@@ -43,11 +43,14 @@ periodics:
     volumes:
     - name: test-account
       secret:
-        secretName: test-account
+        items:
+        - key: service-account-key.json
+          path: service-account.json
```
chizhg, line 48: Are you planning to move these jobs to use workload identity? If yes you may need some changes to Istio `prowgen` since the preset does not support configuring serviceAccoutName now.

prow/config.yaml
upodroid: I didn't know that. That would my life so much easier

prow/jobs/generated/knative-sandbox/async-component-main.gen.yaml
upodroid: I noticed that as well.

prow/config.yaml
chizhg: Prow does support this - see an example at https://github.com/GoogleCloudPlatform/oss-test-infra/blob/master/prow/oss/config.yaml#L93-L175

prow/jobs/custom/autodeploy-prow.yaml
upodroid: Typo Fixed.

prow/config.yaml
upodroid: Done. Change is much leaner now.

PR #3292, opened 2022-04-27
Tweaks and fixes to the infra config

chizhg:
See context in https://knative.slack.com/archives/CCSNR4FCH/p1651023361697399?thread_ts=1650986265.675059&cid=CCSNR4FCH

/cc @upodroid @kvmware


infra/gcp/tests/gke.tf
```diff
@@ -141,6 +145,9 @@ module "prow_build" {
       auto_repair        = true
       auto_upgrade       = true
       service_account    = google_service_account.gke_nodes.email
+      oauth_scopes       = [
+        "https://www.googleapis.com/auth/cloud-platform"
```
upodroid, line 149: This scope is already present. 

```
 REDACTED  MCW0CDP3YY  ~  Desktop  Git  kn-hack   rename-master-function  $   gcloud container clusters describe prow-build --region us-central1 --project knative-tests --format json | jq .nodePools[1]
{
  "autoscaling": {
    "enabled": true,
    "maxNodeCount": 4
  },
  "config": {
    "diskSizeGb": 100,
    "diskType": "pd-ssd",
    "imageType": "COS_CONTAINERD",
    "labels": {
      "cluster_name": "prow-build",
      "environment": "production",
      "node_pool": "testing-pool-v1",
      "type": "testing"
    },
    "localSsdCount": 1,
    "machineType": "n2d-standard-16",
    "metadata": {
      "cluster_name": "prow-build",
      "disable-legacy-endpoints": "true",
      "node_pool": "testing-pool-v1"
    },
    "oauthScopes": [
      "https://www.googleapis.com/auth/cloud-platform"
    ],
    "serviceAccount": "gke-nodes@knative-tests.iam.gserviceaccount.com",
    "shieldedInstanceConfig": {
      "enableIntegrityMonitoring": true,
      "enableSecureBoot": true
    },
    "tags": [
      "gke-prow-build",
      "gke-prow-build-testing-pool-v1"
    ],
    "workloadMetadataConfig": {
      "mode": "GKE_METADATA"
    }
  },
  "initialNodeCount": 1,
  "instanceGroupUrls": [
    "https://www.googleapis.com/compute/v1/projects/knative-tests/zones/us-central1-c/instanceGroupManagers/gke-prow-build-testing-pool-v1-69a0049f-grp",
    "https://www.googleapis.com/compute/v1/projects/knative-tests/zones/us-central1-f/instanceGroupManagers/gke-prow-build-testing-pool-v1-2ca0560d-grp"
  ],
  "locations": [
    "us-central1-c",
    "us-central1-f"
  ],
  "management": {
    "autoRepair": true,
    "autoUpgrade": true
  },
  "maxPodsConstraint": {
    "maxPodsPerNode": "110"
  },
  "name": "testing-pool-v1",
  "networkConfig": {
    "podIpv4CidrBlock": "10.250.64.0/20",
    "podRange": "prow-build-pods"
  },
  "podIpv4CidrSize": 24,
  "selfLink": "https://container.googleapis.com/v1/projects/knative-tests/locations/us-central1/clusters/prow-build/nodePools/testing-pool-v1",
  "status": "RUNNING",
  "upgradeSettings": {
    "maxSurge": 1
  },
  "version": "1.22.8-gke.200"
}
```

Also, e2 machine types don't support local ssds.

upodroid (changes requested):

infra/gcp/tests/gke.tf
chizhg: I see. Then it's very weird. I logged into a boskos janitor pod and ran `gcloud auth activate-service-account --key-file=/etc/test-account/service-account.json`, it just got stuck for a while and then said `Killed`: 

```
root@boskos-janitor-5b7788867d-hl2hl:/# gcloud auth activate-service-account --key-file=/etc/test-account/service-account.json
Killed
```

chizhg: Reverted those changes. The other changes in this PR might still be desirable.

infra/gcp/tests/gke.tf
```diff
@@ -105,6 +105,10 @@ module "prow_build" {
   filestore_csi_driver       = false
   create_service_account     = false
   remove_default_node_pool   = true
+  cluster_autoscaling = {
```
upodroid, line 108: Can you remove this?
https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1beta1/projects.locations.clusters#clusterautoscaling

This is already enabled in a different way. 

https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-autoscaler#how_cluster_autoscaler_works

Autoscaling on GKE can done by either:
- Enabling autoscaling on a nodepool - This adds an extra instance to the existing nodepool. Faster as the GCE MIG is already there and as well the nodepool. We have sane limits for now. The cap is 8 instances but we can raise that instead.
- Enabling autoscaling on a cluster - This spins up new nodepools when the cluster needs more resources. Slow as GKE creates a nodepool and the backing GCE infrastructure. An extra config to write

kvmware: Probably want to capture this (if it is not already) somewhere in the `gcp/` folder @upodroid .

chizhg: @upodroid the main thing I want is the `autoscaling_profile = "OPTIMIZE_UTILIZATION"` here which enables scaling down the cluster more aggressively - https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-autoscaler#autoscaling_profiles. I don't think the cluster-level autoscaling config is **only** for spinning up new nodepools.

chizhg: @upodroid another further thoughts on this =) ? Is it okay to merge now?

kvmware:
/lgtm
/hold
@upodroid please remove hold when you are happy with the change

infra/gcp/tests/gke.tf
upodroid: This is on hold for a bit longer.

Current diff is failing with the following error.

```
 REDACTED  MCW0CDP3YY  ~  Desktop  …  kn-test-infra  infra  gcp   move-last-custom-jobs  1✎  5⚑  $    terraform apply -target module.tests.module.prow_build
╷
│ Error: Invalid value for module argument
│ 
│   on tests/gke.tf line 108, in module "prow_build":
│  108:   cluster_autoscaling = {
│  109:     enabled = true
│  110:     autoscaling_profile = "OPTIMIZE_UTILIZATION"
│  111:   }
│ 
│ The given value is not suitable for child module variable "cluster_autoscaling" defined at
│ .terraform/modules/tests.prow_build/variables.tf:179,1-31: attributes "gpu_resources", "max_cpu_cores", "max_memory_gb", "min_cpu_cores",
│ and "min_memory_gb" are required.
╵

```

We need to switch to the beta submodule

So apply the following diff:

```
 REDACTED  MCW0CDP3YY  ~  Desktop  …  kn-test-infra  infra  gcp   move-last-custom-jobs  2✎  5⚑  ERROR  $    git diff
diff --git a/infra/gcp/tests/gke.tf b/infra/gcp/tests/gke.tf
index 8f6af499..b53dde17 100644
--- a/infra/gcp/tests/gke.tf
+++ b/infra/gcp/tests/gke.tf
@@ -90,7 +90,7 @@ module "prow" {
 }
 
 module "prow_build" {
-  source                     = "terraform-google-modules/kubernetes-engine/google"
+  source                     = "terraform-google-modules/kubernetes-engine/google//modules/beta-public-cluster"
   project_id                 = module.project.project_id
   name                       = "prow-build"
   region                     = "us-central1"
@@ -105,6 +105,17 @@ module "prow_build" {
   filestore_csi_driver       = false
   create_service_account     = false
   remove_default_node_pool   = true
+  gce_pd_csi_driver          = true
+  cluster_autoscaling = {
+    enabled             = false
+    autoscaling_profile = "OPTIMIZE_UTILIZATION"
+    gpu_resources       = []
+    max_cpu_cores       = null
+    max_memory_gb       = null
+    min_cpu_cores       = null
+    min_memory_gb       = null
+
+  }
 
```
`

upodroid (changes requested):

infra/gcp/tests/gke.tf
chizhg: Thanks for checking on this.
Do we have to add the empty fields? Also shouldn't `cluster_autoscaling.enabled` be `true`?

upodroid: - Empty variables: yes it is a terraform limitation/feature 
- setting enabled to false: This is correct, that block creates nodepool auto provisioning which we don't want but we want to set the Autoscaling profile.

chizhg: Thanks for the clarification. Done.

PR #3285, opened 2022-04-22
Move some Prow Jobs over

upodroid:
Moved some handwritten Prow Jobs over to new cluster.

Lets see how they perform. All the required secrets are stored in Google Secret Manager and synced correctly.

I have manually updated kubeconfig secret in prow(us-central1-f) with a service account from the new prow-build cluster.

/cc @chizhg @kvmware

PR #3295, opened 2022-04-27
Move all generated jobs to the new `prow-build` cluster

upodroid:
Move all generated jobs to the new prow-build cluster. I feel more comfortable with this change now given all the bugs we fixed when we moved autobump(#3285) and client(#3287).

If we see catastrophic failures, we can revert this PR and investigate.

/cc @dprotaso @chizhg @kvmware


prow/test.yaml
```diff
@@ -0,0 +1,74 @@
+apiVersion: v1
```
kvmware, line 1: What is this file?

upodroid: random test file, shouldn't be here. Deleting

kvmware:
/lgtm
/hold
unhold when you ready to monitor and ensure all is good

PR #3297, opened 2022-04-27
Give Boskos More Resources

upodroid:
Reduced replica count but made the resource limits much higher.

### DIFF
--- a/prow/cluster/build/400-boskos-janitor.yaml
+++ b/prow/cluster/build/400-boskos-janitor.yaml
@@ -21,7 +21,7 @@ metadata:
     app: boskos-janitor
   namespace: test-pods
 spec:
-  replicas: 6  # Number of distributed janitor instances; k8s has ~1/16 as many projects
+  replicas: 4  # Number of distributed janitor instances; k8s has ~1/16 as many projects
   selector:
     matchLabels:
       app: boskos-janitor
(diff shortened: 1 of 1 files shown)


prow/cluster/build/400-boskos-janitor.yaml
```diff
@@ -21,7 +21,7 @@ metadata:
     app: boskos-janitor
   namespace: test-pods
 spec:
-  replicas: 6  # Number of distributed janitor instances; k8s has ~1/16 as many projects
+  replicas: 4  # Number of distributed janitor instances; k8s has ~1/16 as many projects
```
chizhg, line 24: Why reducing this number?

prow/cluster/build/400-boskos-janitor.yaml
```diff
@@ -47,11 +47,11 @@ spec:
           readOnly: true
         resources:
           requests:
-            cpu: 200m
-            memory: 20Mi
+            cpu: 2000m
```
chizhg, line 50: This is likely more than enough. For now let's keep the requests and only bump up the limits, and see how it goes?

```
requests:
    cpu: 200m
    memory: 20Mi
limits:
    cpu: 2000m
    memory: 2560Mi
```

prow/cluster/build/400-boskos-janitor.yaml
upodroid: I'm thinking fewer replicas but larger ones. I can revert it if it won't work.

prow/cluster/build/400-boskos-janitor.yaml
upodroid: Fixed

PR #3300, opened 2022-04-28
add optional job to run perf tests using mako stub

nader-ziada:

 - add optional job to run perf tests using mako stub in serving


chizhg:
Are you still going to use Mako? Do we have a plan documented somewhere?

nader-ziada:
> Are you still going to use Mako? Do we have a plan documented somewhere?

not using mako, just leveraging the existing tests with the mako stub that already exists here: https://github.com/knative/pkg/tree/main/test/mako/stub-sidecar and used by eventing: https://github.com/knative/eventing/tree/main/test/performance#run-without-mako

PR #3307, opened 2022-05-03
Move remaining custom jobs to new cluster

upodroid:
Moved the remaining jobs to prow-build cluster. We can delete the old cluster after this PR is merged and jobs look good.


kvmware:
/lgtm
/hold
Maybe Chi wants to take a look. It a bit early for him.

PR #3308, opened 2022-05-03
Remove the config fields for the previous build cluster

chizhg:
Now that all jobs have been migrated to the new build cluster, removing all the config fields added during the migration process.

/cc @upodroid @kvmware


upodroid (changes requested):
Isn't that needed for default and prow-trusted clusters?

chizhg:
> Isn't that needed for default and prow-trusted clusters?

You are right. Added that back.

upodroid (approved):
LGTM

PR #3315, opened 2022-05-10
Update max number of nodes for the build cluster to 20

chizhg:
As per discussions in https://knative.slack.com/archives/CCSNR4FCH/p1652212376542809

/cc @upodroid @kvmware


kvmware:
Looks good to me. @upodroid please push it through if looks good to you

kvmware:
/lgtm
we can change this later if need be

PR #3334, opened 2022-05-24
run kperf performance tests nightly

nader-ziada:
- add periodic job for running performance tests suing kperf


prow/jobs/generated/knative/serving-main.gen.yaml
```diff
@@ -443,6 +443,53 @@ periodics:
         - key: service-account-key.json
           path: service-account.json
         secretName: prow-google-credentials
+- annotations:
+    testgrid-dashboards: serving
+    testgrid-tab-name: performance-tests-kperf
+  cluster: prow-build
+  cron: 39 */9 * * *
```
psschwei, line 450: The three Istio periodic tests all run very close to this one. Do we want to space this out from those tests a bit? I think they're all a bit resource-intensive, which might cause a bottleneck (though honestly, worst case is that we just adjust the time if we start to see a slowdown...).

Also, do you think we could get by running this once a day? Not a sticking point for me, I'd be ok either way.

psschwei (commented):
Minor question about the scheduling, but otherwise lgtm...

prow/jobs/generated/knative/serving-main.gen.yaml
nader-ziada: this part is auto generated, I just put the entry in the serving file and run update-codegen.

we can specify a cron if needed, I just went with default behaviour, what do you think it should be, once at night?

psschwei: Yeah, I think once at night should be good to start. We can always bump it up down the line. 

That said, there's a good chance I'm over thinking things here and the default is fine too... I'm ok with whatever you think is best.

nader-ziada: how about we stay with default and adjust if it becomes an issue

psschwei: works for me

PR #3367, opened 2022-06-07
bump go version to 1.18

dprotaso:
go1.18 has been out for a few months so I'm assuming all the major bugs have been shaken out.

Confirmed with some K8s folks that 1.18.3 is 👍


psschwei (commented):
I've been building serving with go 1.18 for a while now without issue
/lgtm

PR #3387, opened 2022-06-17
Add DNS Record to Verify GitHub Organization

upodroid:
Fixes: https://github.com/knative/community/issues/1100

### DIFF
--- a/infra/gcp/dns/dns.tf
+++ b/infra/gcp/dns/dns.tf
@@ -90,6 +90,14 @@ module "knative_dev" {
         "ghs.googlehosted.com.",
       ]
     },
+    {
+      name = "_github-challenge-knative-org"
+      type = "TXT"
+      ttl  = 300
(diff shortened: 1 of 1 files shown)


infra/gcp/dns/dns.tf
```diff
@@ -126,7 +134,7 @@ module "knative_team" {
     {
       name = "google._domainkey"
       type = "TXT"
-      ttl = 300
+      ttl  = 300
```
evankanderson, line 137: This looks like an accident, but it also doesn't seem super-harmful, so I'm not going to block the PR.

evankanderson (commented):
/lgtm
/approve

infra/gcp/dns/dns.tf
upodroid: it is not an accident. ran `terraform fmt`

PR #3393, opened 2022-06-24
Set presubmit peribolos jobs to optional

kvmware:
Set presubmit peribolos config check jobs to optional until we figure out https://github.com/kubernetes/test-infra/issues/26542


kvmware:
cc @upodroid @dprotaso @lance @chizhg

upodroid (approved):
LGTM

PR #3396, opened 2022-06-27
Remove vendor folder and copy `knative/hack` contents to `hack/upstream` folder

upodroid:
Part of #3176 

I have removed vendor folder and copied knative/hack in to hack/upstream.

Next steps:
- verify behaviour
- add a new action in knobots to sync knative/hack in to all repos
- I will make the changes in serving and eventing. Other WGs will sort out their own repos.

/cc @dprotaso @evankanderson @kvmware


upodroid:
This a bit more complex than I though?

- Need to upstream changes I made in hack/upstream around vendor/* folder.
- Do we need third-party/* folder?

evankanderson:
> This a bit more complex than I though.
> 
> - Need to upstream changes I made in hack/upstream around vendor/* folder.
> - Do we need third-party/* folder?
> - Fix lint errors in knative/hack schema folder.

Generally, `vendor` gets managed via `go mod`. We've been using third_party to track / keep items which can't be imported by `go mod`, but for which we're not the source of truth.

PR #3398, opened 2022-06-28
s390x: disabled TLS check for remote registry with self-signed certificate

haubenr:
Signed-off-by: Dirk Haubenreisser <haubenr@de.ibm.com>

This PR is to update the existing custom jobs for s390x to enable communication with a remote image registry that uses a self-signed certificate for HTTPS support by introducing a new flag for `ko`.

### DIFF
--- a/prow/jobs/generated/knative-sandbox/eventing-kafka-main.gen.yaml
+++ b/prow/jobs/generated/knative-sandbox/eventing-kafka-main.gen.yaml
@@ -246,7 +246,7 @@ periodics:
       - name: GOOGLE_APPLICATION_CREDENTIALS
         value: /etc/test-account/service-account.json
       - name: KO_FLAGS
-        value: --platform=linux/s390x
+        value: --platform=linux/s390x --insecure-registry
       - name: PLATFORM
         value: linux/s390x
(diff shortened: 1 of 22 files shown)


upodroid (approved):
LGTM

Can you tell us more about this registry?

/hold

You can cancel the hold when you are ready

haubenr:
> Can you tell us more about this registry?

Sure, it's a registry used internally for the s390x e2e tests only. The registry will be created on a dedicated server in IBM Cloud prior to the prow tests being kicked off. The prow test pod gets configured with the registries connection data (IP, authentication credentials) at startup and uses the registry to push the test artifacts (aka images) to that registry. As those images are of concern to the tests only the registry will be destroyed automatically after the test run. As the registry is self-hosted I need to utilize a self-signed certificate for HTTPS support which can only be used in conjunction with ko if the flag `--insecure-registry` is set. It's important to note that the registry itself actually is *not* unsecured - it uses HTTPS as well as basic authentication to guard against unauthorized access. The ko flag just turns off the TLS certificate verification step in ko.

upodroid:
Thanks

/hold cancel

PR #3414, opened 2022-07-12
Add auto-release config for kn-plugin-operator

dsimansk:
We'd like to onboard operator plugin to release train. Therefore this PR adds release job config for `knative-sandbox/kn-plugin-operator` project.

/cc @houshengbo @maximilien @knative/knative-release-leads

### DIFF
--- a/prow/jobs/generated/knative-sandbox/kn-plugin-operator-main.gen.yaml
+++ b/prow/jobs/generated/knative-sandbox/kn-plugin-operator-main.gen.yaml
@@ -47,6 +47,104 @@ periodics:
         - key: service-account-key.json
           path: service-account.json
         secretName: prow-google-credentials
+- annotations:
+    testgrid-dashboards: kn-plugin-operator
+    testgrid-tab-name: nightly
+  cluster: prow-build
+  cron: 25 9 * * *
+  decorate: true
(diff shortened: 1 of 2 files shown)


psschwei:
Since today's the day of the release, would it make sense to cut branch `release-1.6` manually and then get the automation in place for the next release?

dsimansk:
> Since today's the day of the release, would it make sense to cut branch `release-1.6` manually and then get the automation in place for the next release?

Afaik, auto-release job needs to be in place to produce output from e.g. `release-1.6`.

psschwei:
Yeah, you're right... don't know what I was thinking there

/lgtm
/approve

psschwei:
@dsimansk should the release leads cut this one or will you handle it today?

Also, if you could update https://github.com/knative/release by adding the operator plugin to the lists [here](https://github.com/knative/release/blob/main/TIMELINE.md#this-group-can-be-started-after-all-the-previous-repos-have-been-successfully-published) and [here](https://github.com/knative/release/blob/main/.github/ISSUE_TEMPLATE/release-checklist.yaml) and [here](https://github.com/knative/release/blob/main/PROCEDURES.md#homebrew-kn-plugins)

PR #3443, opened 2022-07-25
Updated s390x prow job schedule

haubenr:
This PR updates all prow job schedules for the s390x platform. This schedule update is required due to a rework of the internal IBM Cloud zSystems VSI worker preparation logic / schedule.

As a result of this rework all s390x-related prow jobs are expected to fail for 72h max.


upodroid (approved):
LGTM

/approve
/ok-to-test

Issue #3345, opened 2022-05-27
Release Improvements Project


upodroid:
This is an epic style issue tracking Release Improvements Project:

- Start generating SBOMs: https://github.com/knative/test-infra/pull/3342 ko 0.11+ generates and attaches SBOMs to docker images.

upodroid:
Most images have been signed for 1.8 except serving, eventing and 3 eventing plugins.

Image references were not being created properly for the unsigned projects. 

Details https://github.com/knative/serving/pull/13408

I intend to complete the remaining images before 1.9 and verify them in the nightly releases.

PR #3448, opened 2022-07-25
Workload Identity Enablement - Part Two

upodroid:
Part of #3345 

This PR does the following:
- Enables `gcsweb` on deck after #3444 is merged
- Enables Workload Identity on trusted and default cluster.
- Adds more resources under IaC management.
- Enables GitOps for prow-trusted cluster.

/cc @kvmware


upodroid:
/hold cancel

gcsweb is up

PR #3453, opened 2022-07-27
Make all contexts required unless set to optional.

kvmware:
This should make all GHA required unless set to be optional contexts here.
Commented out is an example way of doing that.

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -217,6 +217,7 @@ tide:
   squash_label: tide/merge-method-squash
   rebase_label: tide/merge-method-rebase
   merge_label: tide/merge-method-merge
+  # https://github.com/kubernetes/test-infra/blob/master/prow/cmd/tide/config.md#context-policy-options
   context_options:
     # Treat unknown contexts as optional
     skip-unknown-contexts: false
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -219,7 +219,22 @@ tide:
   merge_label: tide/merge-method-merge
   context_options:
     # Treat unknown contexts as optional
-    skip-unknown-contexts: true
+    skip-unknown-contexts: false
+#     orgs:
+#       org:
+#         required-contexts:
+#         - "check-required-for-all-repos"
+#         repos:
+#           repo:
+#             required-contexts:
+#              - "check-required-for-all-branches"
+#             branches:
+#               branch:
+#                 from-branch-protection: false
+#                 required-contexts:
+#                 - "required_test"
+#                 optional-contexts:
+#                 - "optional_test"
 
 # Configure GitHub branch protection rules according to the specified policy.
 branch-protection:


kvmware:
cc @dprotaso @evankanderson

kvmware:
cc @pierDipi @lionelvillard @rhuss @navidshaikh @csantanapr @lance @nak3 @houshengbo 

Sorry for the large tag, this would be a breaking change in some people's workflow and want to make sure everyone gets a heads up, in case there is any checks people don't want to be required.

prow/config.yaml
```diff
@@ -219,7 +219,22 @@ tide:
   merge_label: tide/merge-method-merge
   context_options:
     # Treat unknown contexts as optional
-    skip-unknown-contexts: true
+    skip-unknown-contexts: false
+#     orgs:
```
evankanderson, line 223: I assume that this would be `context_options.orgs` when un-commented?

prow/config.yaml
```diff
@@ -219,7 +219,22 @@ tide:
   merge_label: tide/merge-method-merge
   context_options:
     # Treat unknown contexts as optional
```
evankanderson, line 221: I'm not sure I understand what "contexts" are here.

Is it worth adding a link to documentation?

Do we know what this will end up blocking?

evankanderson (commented):
My comments aren't blocking, just making sure I understand.

prow/config.yaml
kvmware: From the docs `If we want to skip unknown checks (ie checks that are not defined in Prow Config), we can set skip-unknown-contexts to true. This option can be set globally or per org, repo and branch.` In this case it is global.

prow/config.yaml
kvmware: This should not block anyone. I did ping all leads just in case, but this makes it so that all github actions are set to required pass rather than optional as they are now. Essentially instead of whitelisting which should be required this is making it so all are required to pass unless stated otherwise ie blacklist.

dprotaso (commented):
Just to confirm - if some action is completely broken are we able to use `/override` and what name do I supply?

ie. from serving `/override Code Style / style / Boilerplate Check (sh) (pull_request)` ?

Also how do you mark a GitHub action optional?

kvmware:
> Just to confirm - if some action is completely broken are we able to use `/override` and what name do I supply?
> 
> ie. from serving `/override Code Style / style / Boilerplate Check (sh) (pull_request)` ?
> 
Override I don't know about. We can see if that will work and revert if it too much of a pain.
> Also how do you mark a GitHub action optional?
One would follow the instructions in this commit (the commented out part, and docs) to add the test has optional https://github.com/knative/test-infra/pull/3453/files#diff-43597815f705840f0d74fd38408b1dfab023521863d730099df2dad36f7183a5R237

dprotaso:
last question - is there any way to configure the optional checks in each repo vs the centralized config in test-infra?

pierDipi:
This is breaking for some repositories that I'm aware of, can we make it opt-it instead of the reverse?

I also have another question, what exactly triggered this change? Any link to a discussion?

kvmware:
> last question - is there any way to configure the optional checks in each repo vs the centralized config in test-infra?
Tide is what dictates what is required and or not. However, each repo's github actions can be added or removed or made to be optional in their own repo. So, I guess the answer is yes, just depends on how you write your GHAs.

kvmware:
> This is breaking for some repositories that I'm aware of, can we make it opt-it instead of the reverse?
> 
Could you please share which ones? I would like to have a chat with them.
> I also have another question, what exactly triggered this change? Any link to a discussion?
>
GHAs are currently not required which is an issue. I know docs team for instance has important GHAs that they want to pass before a merge. If anyone has GHA that they don't care if they pass or fail should remove them, make them optional in the GHA itself, or explicitly set the action to optional as described above.

The rationale is that it is easier to blacklist GHA that we want optional than the inverse which is what it is currently. We want to empower repos to be more self service and be able to make GHAs and hopefully their prow jobs in their control.

kvmware:
@dprotaso @pierDipi https://stackoverflow.com/questions/62441607/mark-github-action-as-optional

upodroid (approved):
LGTM

Marking this as required every should be fine. We do have a few levers:
- https://stackoverflow.com/questions/62441607/mark-github-action-as-optional good idea to turn it on for flaky flows
- /override shoud work
- particularly problematic contexts(status checks) can be marked as optional in the prow config.
- Our other option is to manually enumerate all github actions for everyrepo and add it to prow config which is not feasible.

pierDipi:
I have explicitly expressed concerns that this will break many people's workflow, you are requiring changes without notice or giving people time to migrate to this new way of working without even considering other's people opinions.

I'd suggest reverting and giving other WG a way of saying something.

psschwei:
Will `/retest` (or some equivalent) work for actions? (I seem to recall not being able to rerun them in repos I wasn't an approver for, but still currently have release lead superpowers so can't verify that)

pierDipi:
> > This is breaking for some repositories that I'm aware of, can we make it opt-it instead of the reverse?
> 
> Could you please share which ones? I would like to have a chat with them.

All repositories with downstream tests (pkg, eventing, rekt, etc), repositories with flaky tests (retest doesn't work for everybody with actions, afaik)

> 
> > I also have another question, what exactly triggered this change? Any link to a discussion?
> 
> GHAs are currently not required which is an issue. I know docs team for instance has important GHAs that they want to pass before a merge. If anyone has GHA that they don't care if they pass or fail should remove them, make them optional in the GHA itself, or explicitly set the action to optional as described above.
> 
> The rationale is that it is easier to blacklist GHA that we want optional than the inverse which is what it is currently. We want to empower repos to be more self service and be able to make GHAs and hopefully their prow jobs in their control.
> 
> EDIT: Realized maybe I didn't answer your question fully. This was a decision from Productivity, and if anyone's workflow will be interrupted they can let us know and we can help them to exclude themselves and I would for sure like to have a chat with them on why they have that type of workflow and see if I can help them get to something more optimal

kvmware:
@pierDipi Sorry, please let me know anyone's workflow that will be interrupted and I will be sure to exclude them in a jiffy so they are not blocked.

kvmware:
> Will `/retest` (or some equivalent) work for actions? (I seem to recall not being able to rerun them in repos I wasn't an approver for, but still currently have release lead superpowers so can't verify that)

`/retest` only works for prow jobs currently we have issue open to extend that for GHAs however it is low priority as one can rerun actions via github UI

kvmware:
> All repositories with downstream tests (pkg, eventing, rekt, etc), repositories with flaky tests (retest doesn't work for everybody with actions, afaik)
> 
You are right, I will try to go through and identify which tests those are or you can provide a list and we can add them to the exclude list

psschwei:
>  as one can rerun actions via github UI

I think only approvers and/or admins have that ability. I just checked on tekton/pipelines, where I'm a member but not approver, and I can't rerun actions there. (can't check knative for the reason mentioned above)

kvmware:
> > as one can rerun actions via github UI
> 
> I think only approvers and/or admins have that ability. I just checked on tekton/pipelines, where I'm a member but not approver, and I can't rerun actions there. (can't check knative for the reason mentioned above)

Thanks for looking into that, I will bring that up today in our meeting and we will try to address that ASAP

upodroid:
> I think only approvers and/or admins have that ability. I just checked on tekton/pipelines, where I'm a member but not approver, and I can't rerun actions there. (can't check knative for the reason mentioned above)

https://docs.github.com/en/actions/managing-workflow-runs/re-running-workflows-and-jobs

Anyone with write permission on a repo can rerun workflows.

https://github.com/knative/community/blob/main/peribolos/knative.yaml Most *Writers teams have write permissions on the relevant repositories.

@psschwei You should be able to rerun actions on knative/serving and related repos

psschwei:
> > I think only approvers and/or admins have that ability. I just checked on tekton/pipelines, where I'm a member but not approver, and I can't rerun actions there. (can't check knative for the reason mentioned above)
> 
> https://docs.github.com/en/actions/managing-workflow-runs/re-running-workflows-and-jobs
> 
> Anyone with write permission on a repo can rerun workflows.
> 
> https://github.com/knative/community/blob/main/peribolos/knative.yaml Most *Writers teams have write permissions on the relevant repositories.
> 
> @psschwei You should be able to rerun actions on knative/serving and related repos

For non-approvers though, they'll need an approver to trigger retests for the actions, which may add a bit of toil.... though as I'm thinking about it now, maybe not, as instead of running the actions they could just push an empty commit (though that would trigger them all...)

Not opposed to the change, btw, just thinking a bit out loud...

kvmware:
@psschwei All feedback is welcomed and appreciate your thinking!

PR #3454, opened 2022-07-27
Remove dry-run fixing teams

kvmware:
Peribolos, in dry-run mode, naively checks for team members even when the team is being created which causes the job to fail. Instead remove checking for teams in the dry-run and let https://github.com/knative/community/pull/1136 take care of validation.

This also removes the optional flag I put in to get us around this bug as this and https://github.com/knative/community/pull/1136 should fix that now.


kvmware:
@upodroid 
@lance this should allow you to finally add you func team to knative(-sandbox).

upodroid (approved):
LGTM

Issue #3440, opened 2022-07-23
SLSA 1 Compliance


upodroid:
https://slsa.dev/spec/v0.1/requirements

For SLSA 1 we need to do the following:

- Build:
  - ~~Scripted Build: All build steps were fully defined in some sort of “build script”. The only manual command, if any, was to invoke the build script.~~ ✅ 
- Provenance:
  - Available: The provenance is available to the consumer in a format that the consumer accepts. The format SHOULD be in-toto [SLSA Provenance](https://slsa.dev/provenance), but another format MAY be used if both producer and consumer agree and it meets all the other requirements. I'm thinking of just attesting the SBOMs generated by `ko` by running `cosign attest --predicate sbom.json --key gcpkms://projects/knative-nightly/locations/global/keyRings/cosign/cryptoKeys/signing-key gcr.io/knative-nightly/knative.dev/container-freezer/cmd/daemon` as an example.
  - Identifies the artifact: The provenance MUST identify the output artifact via at least one cryptographic hash. The provenance MAY provide multiple identifying cryptographic hashes using different algorithms. When only one hash is provided, the RECOMMENDED algorithm is SHA-256 for cross-system compatibility. If another algorithm is used, it SHOULD be resistant to collisions and second preimages.
  - Identifies the builder: The provenance identifies the entity that performed the build and generated the provenance. This represents the entity that the consumer must trust. Examples: “GitHub Actions with a GitHub-hosted worker”, “[jdoe@example.com](mailto:jdoe@example.com)’s machine”.
  - Identifies the build instructions: The provenance identifies the top-level instructions used to execute the build. The identified instructions SHOULD be at the highest level available to the build (e.g. if the build is told to run build.sh it should list build.sh and NOT the individual instructions in build.sh).
   
/kind security
/priority important-soon

upodroid:
@mattmoor 

I'm having a hardtime working out the Provenance piece.

Using net-certmanager as an example.

To build images, we call https://github.com/knative-sandbox/net-certmanager/blob/main/vendor/knative.dev/hack/release.sh and https://github.com/knative-sandbox/net-certmanager/blob/main/hack/release.sh.

A typical run looks like this. https://storage.googleapis.com/knative-prow/logs/nightly_net-certmanager_main_periodic/1550767763880939520/build-log.txt

[...]

PR #3456, opened 2022-07-28
Add SLSA Provenance Generator

upodroid:
Part of: #3440

I wrote a provenance generator that generates the following statement:

```json
{
    "_type": "https://in-toto.io/Statement/v0.1",
    "predicateType": "https://slsa.dev/provenance/v0.2",
    "subject": [
      {
        "name": "gcr.io/knative-releases/knative.dev/serving/cmd/controller",
        "digest": {
          "sha256": "bac158dfb0c73d13ed42266ba287f1a86192c0ba581e23fbe012d30a1c34837c"
        }
      },
      {
        "name": "gcr.io/knative-releases/knative.dev/serving/cmd/queue",
        "digest": {
          "sha256": "83f6888ea9561495f67334d044ffa8ad067d251ad953358dda7ea5183390cc69"
        }
      }
    ],
    "predicate": {
      "builder": {
        "id": "https://prow.knative.dev"
      },
      "buildType": "https://prow.knative.dev/ProwJob@v1",
      "invocation": {
        "configSource": {
          "entryPoint": "https://github.com/knative/test-infra/tree/main/prow/jobs/generated/knative"
        }
      },
      "buildConfig": {
        "command": [
          "runner.sh",
          "./hack/release.sh",
          "--publish",
          "--tag-release"
        ],
        "entrypoint": {
          "timeout": 10800000000000,
          "grace_period": 15000000000,
          "artifact_dir": "/logs/artifacts",
          "args": [
            "runner.sh",
            "./hack/release.sh",
            "--publish",
            "--tag-release"
          ],
          "container_name": "test",
          "process_log": "/logs/process-log.txt",
          "marker_file": "/logs/marker-file.txt",
          "metadata_file": "/logs/artifacts/metadata.json"
        },
        "prowjob": {
          "metadata": {
            "name": "b71fa037-0d8c-11ed-ab62-2ace146f4dd8",
            "namespace": "default",
            "uid": "dc818b0b-df93-44a8-86fe-fee9c02182a0",
            "resourceVersion": "71593618",
            "generation": 7,
            "creationTimestamp": "2022-07-27T09:15:53Z",
            "labels": {
              "created-by-prow": "true",
              "prow.k8s.io/build-id": "1552221225705541632",
              "prow.k8s.io/context": "",
              "prow.k8s.io/id": "b71fa037-0d8c-11ed-ab62-2ace146f4dd8",
              "prow.k8s.io/job": "nightly_serving_main_periodic",
              "prow.k8s.io/refs.base_ref": "main",
              "prow.k8s.io/refs.org": "knative",
              "prow.k8s.io/refs.repo": "serving",
              "prow.k8s.io/type": "periodic"
            },
            "annotations": {
              "prow.k8s.io/context": "",
              "prow.k8s.io/job": "nightly_serving_main_periodic",
              "testgrid-dashboards": "serving",
              "testgrid-tab-name": "nightly"
            }
          },
          "spec": {
            "type": "periodic",
            "agent": "kubernetes",
            "cluster": "prow-build",
            "namespace": "test-pods",
            "job": "nightly_serving_main_periodic",
            "extra_refs": [
              {
                "org": "knative",
                "repo": "serving",
                "base_ref": "main",
                "path_alias": "knative.dev/serving"
              }
            ],
            "report": true,
            "pod_spec": {
              "volumes": [
                {
                  "name": "docker-graph",
                  "emptyDir": {}
                },
                {
                  "name": "modules",
                  "hostPath": {
                    "path": "/lib/modules",
                    "type": "Directory"
                  }
                },
                {
                  "name": "cgroup",
                  "hostPath": {
                    "path": "/sys/fs/cgroup",
                    "type": "Directory"
                  }
                },
                {
                  "name": "nightly-account",
                  "secret": {
                    "secretName": "prow-google-credentials",
                    "items": [
                      {
                        "key": "nightly.json",
                        "path": "service-account.json"
                      }
                    ]
                  }
                }
              ],
              "containers": [
                {
                  "name": "",
                  "image": "gcr.io/knative-tests/test-infra/prow-tests:v20220725-a4aaff33",
                  "command": [
                    "runner.sh",
                    "./hack/release.sh",
                    "--publish",
                    "--tag-release"
                  ],
                  "env": [
                    {
                      "name": "DOCKER_IN_DOCKER_ENABLED",
                      "value": "true"
                    },
                    {
                      "name": "GOOGLE_APPLICATION_CREDENTIALS",
                      "value": "/etc/nightly-account/service-account.json"
                    }
                  ],
                  "resources": {
                    "limits": {
                      "memory": "16Gi"
                    },
                    "requests": {
                      "memory": "12Gi"
                    }
                  },
                  "volumeMounts": [
                    {
                      "name": "docker-graph",
                      "mountPath": "/docker-graph"
                    },
                    {
                      "name": "modules",
                      "mountPath": "/lib/modules"
                    },
                    {
                      "name": "cgroup",
                      "mountPath": "/sys/fs/cgroup"
                    },
                    {
                      "name": "nightly-account",
                      "readOnly": true,
                      "mountPath": "/etc/nightly-account"
                    }
                  ],
                  "securityContext": {
                    "privileged": true
                  }
                }
              ],
              "nodeSelector": {
                "type": "testing"
              }
            },
            "decoration_config": {
              "timeout": "3h0m0s",
              "grace_period": "15s",
              "utility_images": {
                "clonerefs": "gcr.io/k8s-prow/clonerefs:v20220721-cf42b99a12",
                "initupload": "gcr.io/k8s-prow/initupload:v20220721-cf42b99a12",
                "entrypoint": "gcr.io/k8s-prow/entrypoint:v20220721-cf42b99a12",
                "sidecar": "gcr.io/k8s-prow/sidecar:v20220721-cf42b99a12"
              },
              "resources": {
                "sidecar": {
                  "requests": {
                    "cpu": "100m",
                    "memory": "20Mi"
                  }
                }
              },
              "gcs_configuration": {
                "bucket": "knative-prow",
                "path_strategy": "explicit"
              },
              "gcs_credentials_secret": "gcs-upload"
            },
            "reporter_config": {
              "slack": {
                "channel": "serving",
                "job_states_to_report": [
                  "failure"
                ],
                "report_template": "\"The nightly release job fails, check the log: <{{.Status.URL}}|View logs>\"\n",
                "report": true
              }
            },
            "prowjob_defaults": {
              "tenant_id": "GlobalDefaultID"
            }
          },
          "status": {
            "startTime": "2022-07-27T09:15:53Z",
            "pendingTime": "2022-07-27T09:15:53Z",
            "completionTime": "2022-07-27T11:15:41Z",
            "state": "success",
            "description": "Job succeeded.",
            "url": "https://prow.knative.dev/view/gs/knative-prow/logs/nightly_serving_main_periodic/1552221225705541632",
            "pod_name": "b71fa037-0d8c-11ed-ab62-2ace146f4dd8",
            "build_id": "1552221225705541632",
            "prev_report_states": {
              "gcsk8sreporter": "success",
              "gcsreporter": "success"
            }
          }
        }
      },
      "metadata": {
        "buildInvocationID": "1552221225705541632",
        "buildStartedOn": "2022-07-27T09:15:53Z",
        "buildFinishedOn": "2022-07-27T11:15:41Z",
        "completeness": {
          "parameters": true,
          "environment": true,
          "materials": true
        },
        "reproducible": false
      },
      "materials": [
        {
          "uri": "git+https://github.com/knative/serving",
          "digest": {
            "sha1": "c82be271867f137d0923be34acd18b6aca452446"
          }
        }
      ]
    }
 }

```

/cc @evankanderson @kvmware 
@puerco 

The code is a bit rough but it works. Need to add some tests.


kvmware:
@cardil

go.mod
```diff
@@ -1,30 +1,90 @@
 module knative.dev/test-infra
 
-go 1.15
+go 1.17
```
evankanderson, line 3: Bump to go 1.18? It turns out 1.15 has been out of support for almost a year.

tools/provenance-generator/README.md
```diff
@@ -0,0 +1,72 @@
+# Knative Provenance Generator
+
+As part of achieving SLSA Compliance, we need to generate [provenance](https://slsa.dev/provenance/v0.2).
+
+Kubernetes project already has a library that can generate provenances but we will wrap it with some go code that can pull our build info from Prow.
+
+We will pull the following from Prow jobs:
+
+- https://prow.knative.dev/view/gs/knative-prow/logs/nightly_kn-plugin-quickstart_main_periodic/1552222987300638720
```
evankanderson, line 9: I don't understand what this first thing is.

tools/provenance-generator/README.md
```diff
@@ -0,0 +1,72 @@
...
+PROW_JOB_ID=46282424-0d8d-11ed-ab62-2ace146f4dd8
+ENTRYPOINT_OPTIONS={"timeout":7200000000000,"grace_period":15000000000,"artifact_dir":"/logs/artifacts","args":["runner.sh","./hack/release.sh","--auto-release","--release-gcs","knative-releases/kn-plugin-func","--release-gcr","gcr.io/knative-releases","--github-token","/etc/hub-token/token"],"container_name":"test","process_log":"/logs/process-log.txt","marker_file":"/logs/marker-file.txt","metadata_file":"/logs/artifacts/metadata.json"}
+```
+
+We need to read this and build an in-toto statement that looks like this:
+
+```json
+  "predicate": {
+    "builder": {
+      "id": "pkg:github/puerco/release@provenance"
```
evankanderson, line 37: Is this really the builder id?

tools/provenance-generator/README.md
```diff
@@ -0,0 +1,72 @@
...
+    },
+    "recipe": {
+      "type": "https://cloudbuild.googleapis.com/CloudBuildYaml@v1",
```
evankanderson, line 40: ??

tools/provenance-generator/attestation.json
```diff
@@ -0,0 +1 @@
...https://in-toto.io/Statement/v0.1","predicateType":"https://slsa.dev/provenance/v0.2","subject":[{"name":"gcr.io/knative-releases/knative.dev/serving/cmd/controller","digest":{"sha256":"bac158dfb0c73d13ed42266ba287f1a86192c0ba581e23fbe012d30a1c34837c"}},{"name":"gcr.io/knative-releases/knative.dev/serving/cmd/queue","digest":{"sha256":"83f6888ea9561495f67334d044ffa8ad067d251ad953358dda7ea5183390cc69"}}],"predicate":{"builder":{"id":"https://prow.knative.dev"},"buildType":"https://cloudbuild.googleapis.com/CloudBuildYaml@v1","invocation":{"configSource":{"entryPoint":"https://github.com/knative/serving/blob/main/hack/release.sh"},"parameters":["runner.sh","./hack/release.sh","--auto-release","--release-gcs=knative-releases/async-component","--release-gcr=gcr.io/knative-releases","--github-token=/etc/hub-token/token"]},"metadata":{"buildInvocationID":"1552230788685107200","buildStartedOn":"2022-07-28T10:21:41.996820966Z","buildFinishedOn":"2022-07-28T10:28:08.373467Z","completeness":{"parameters":true,"environment":false,"materials":true},"reproducible":false},"materials":[{"uri":"git+https://github.com/knative/serving","digest":{"sha1":"c82be271867f137d0923be34acd18b6aca452446"}}]}}
```
evankanderson, line 1: Can you pretty-print this JSON?

Otherwise, no one's ever going to check a diff in it, and it feels like maybe we want people to notice the changes here.

tools/provenance-generator/kn-attestation.json
```diff
@@ -0,0 +1,56 @@
...
+        "digest": {
+          "sha256": "83f6888ea9561495f67334d044ffa8ad067d251ad953358dda7ea5183390cc69"
+        }
+      }
+    ],
+    "predicate": {
+      "builder": {
+        "id": "https://prow.knative.dev"
+      },
+      "buildType": "https://cloudbuild.googleapis.com/CloudBuildYaml@v1",
```
evankanderson, line 22: Is this right?

tools/provenance-generator/kn-attestation.json
```diff
@@ -0,0 +1,56 @@
...
+      "metadata": {
+        "buildInvocationID": "1552230788685107200",
+        "buildStartedOn": "2022-07-28T10:21:41.996820966Z",
+        "buildFinishedOn": "2022-07-28T10:25:12.818292Z",
+        "completeness": {
+          "parameters": true,
+          "environment": false,
+          "materials": true
+        },
+        "reproducible": false
```
evankanderson, line 45: I think our builds might be reproducible... have we checked?

tools/provenance-generator/kn-attestation.json
```diff
@@ -0,0 +1,56 @@
...
+          "--github-token=/etc/hub-token/token"
+        ]
+      },
+      "metadata": {
+        "buildInvocationID": "1552230788685107200",
+        "buildStartedOn": "2022-07-28T10:21:41.996820966Z",
+        "buildFinishedOn": "2022-07-28T10:25:12.818292Z",
+        "completeness": {
+          "parameters": true,
+          "environment": false,
```
evankanderson, line 42: Why are `parameters` and `environment` not complete?

tools/provenance-generator/main.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"flag"
+	"log"
+
+	"knative.dev/test-infra/tools/provenance-generator/pkg"
+)
+
+func main() {
+	cloneLogPath := flag.String("clone-json", "", "Path to clone.json")
```
evankanderson, line 30: What is `clone.json`? Is this a template?

tools/provenance-generator/pkg/parse.go
```diff
@@ -0,0 +1,169 @@
...
+	// Parse Entrypoint of the prowjob to pull args, path, etc
+	entryPointOpsRaw := os.Getenv(config.EntryPointOptsVariable)
+	entryPointOpts := &entrypoint.Options{}
+	err := json.Unmarshal([]byte(entryPointOpsRaw), entryPointOpts)
+	if err != nil {
+		log.Fatalf("failed to unmarshal ENTRYPOINTS_OPTIONS env variable %v", err)
+	}
+	config.EntryPointOpts = entryPointOpts
+
+	// Sanity checks
```
evankanderson, line 98: Given that there's just one check, not a section, maybe omit this header?

tools/provenance-generator/pkg/parse.go
```diff
@@ -0,0 +1,169 @@
...
+			config.Arguments[params[0]] = ""
+			keys = append(keys, params[0])
+		} else if len(params) == 2 {
+			config.Arguments[params[0]] = params[1]
+			keys = append(keys, params[0])
+		}
+	}
+	config.ArgumentsInsertOrder = keys
+
+	//Appoximate start time of job
```
evankanderson, line 119: Do we not get the exact time from Prow?

tools/provenance-generator/pkg/parse.go
```diff
@@ -0,0 +1,169 @@
...
+	if config.EntryPointOpts.Args[0] != "runner.sh" {
+		log.Fatal("this prowjob is misconfigured, expecting runner.sh to be called first")
+	}
+
+	config.Arguments = map[string]string{}
+	keys := []string{}
+
+	// Parse entrypoint flags
+	for _, elem := range config.EntryPointOpts.Args {
+		params := strings.Split(elem, "=") // expected values --something=foo
```
evankanderson, line 108: Prefer go 1.18:

```golang
flag, val, _ = strings.Cut(elem, "=")
config.Arguments[flag] = val
keys = append(keys, flag)
```

tools/provenance-generator/pkg/parse.go
```diff
@@ -0,0 +1,169 @@
...
+	config.ArgumentsInsertOrder = keys
+
+	//Appoximate start time of job
+	fileInfo, err := os.Stat(config.CloneLogPath)
+	if err != nil {
+		log.Fatalf("failed to open file %v", err)
+	}
+	config.StartTime = fileInfo.ModTime()
+
+	byteValue, _ := os.ReadFile(config.CloneLogPath)
```
evankanderson, line 126: intended to drop the `error` here?

tools/provenance-generator/pkg/parse.go
```diff
@@ -0,0 +1,169 @@
...
+	cloneLogs := []Record{}
+
+	err = json.Unmarshal(byteValue, &cloneLogs)
+	if err != nil {
```
evankanderson, line 130: You can write this as:

```golang
if err := json.Unmarshal(byteValue, &cloneLogs); err != nil {
  log.Fatalf(...)
}
```

tools/provenance-generator/pkg/parse.go
```diff
@@ -0,0 +1,169 @@
...
+	config.CloneRecords = cloneLogs
+
+	// Lets strip out the empty refs that is inserted by prow
+	for i := len(config.CloneRecords) - 1; i >= 0; i-- {
+		cloneRecord := config.CloneRecords[i]
+		if cloneRecord.FinalSHA == "" {
+			config.CloneRecords = append(config.CloneRecords[:i],
+				config.CloneRecords[i+1:]...)
+		}
+	}
```
evankanderson, line 142: It's probably slightly more efficient to keep two counters and trim the slice at the end, but it may not be important:

```golang
for i, j := 0; i < len(config.CloneRecords) -1; i++ {
  if config.CloneRecords[i].FinalSHA == "" {
    continue
  }
  if i != j {
    config.CloneRecords[j] = config.CloneRecords[i]
  }
  j++
}
config.CloneRecords = config.CloneRecords[:j-1]
```

tools/provenance-generator/pkg/parse.go
```diff
@@ -0,0 +1,169 @@
...
+	if err != nil {
+		log.Fatalf("failed to open file %v", err)
+	}
+	lines := strings.Split(string(imageRefFile), "\n")
+	subjects := []in_toto.Subject{}
+	for _, elem := range lines {
+		name, err := reference.ParseNamed(elem)
+		if err != nil {
+			break
+		}
```
evankanderson, line 155: This suggests a failure in `ParseNamed` will be skipped silently. Intentional?

tools/provenance-generator/pkg/parse.go
```diff
@@ -0,0 +1,169 @@
...
+		digest := strings.TrimPrefix(elem, name.Name()+"@")
+		digestComponents := strings.Split(digest, ":")
```
evankanderson, line 157: How will we detect if this code gets confused. Should we make the digest part of the object returned from `reference.ParseNamed`, such that it could return an error in that case?

evankanderson: Also, `strings.Cut` here!

evankanderson: Ah, this is coming from `docker/distribution`...

Consider using [`google/go-containerregistry`](https://pkg.go.dev/github.com/google/go-containerregistry@v0.11.0/pkg/name#pkg-types), particularly the `Digest` object.

tools/provenance-generator/main.go
```diff
@@ -0,0 +1,50 @@
...
+	flag.Parse()
+
+	config := pkg.Config{
+		CloneLogPath:           *cloneLogPath,
+		EntryPointOptsVariable: "ENTRYPOINT_OPTIONS",
+		ImageReferencePath:     *imageRefPath,
+	}
+
+	config = pkg.LoadParameters(config)
+	pkg.GenerateAttestation(config)
```
evankanderson, line 48: Minor style point -- I'd have `GenerateAttestation` return a `provenance.Statement`, and then do the output writing here. It will make it easier to test in the future, I expect.

tools/provenance-generator/pkg/generate.go
```diff
@@ -0,0 +1,53 @@
...
+	"time"
+
+	slsa "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v0.2"
+	"sigs.k8s.io/bom/pkg/provenance"
+)
+
+func GenerateAttestation(config Config) {
+	// Build the arguments RawMessage:
+
+	cloneRecord := config.CloneRecords[0] // Safe as earlier step deletes all other entries
```
evankanderson, line 15: Do you want to assert this in some way?

tools/provenance-generator/pkg/generate.go
```diff
@@ -0,0 +1,53 @@
...
+
+	p.AddMaterial(("git+" + repo), slsa.DigestSet{"sha1": commitSHA})
+
+	// Create the new attestation and attach the predicate
+	attestation := &provenance.Statement{}
+	attestation = provenance.NewSLSAStatement()
+	attestation.Predicate = p
+	attestation.Subject = config.Subject
+
+	attestation.Write("attestation.json")
```
evankanderson, line 51: Capture any `error` here, or move the call to `Write` to main and capture there?

evankanderson (commented):
I don't really know what this should be like, so I'm asking a lot of what may be annoying questions.

tools/provenance-generator/README.md
upodroid: It is the .env file in testdata. When this tool runs in the CI it will pull info from the environment, the typical values are posted at that link. 

Need to write better docs.

tools/provenance-generator/README.md
upodroid: Sample provenance, will swap out for the test provenance created by this package.

tools/provenance-generator/attestation.json
upodroid: No, I'll need to fix it upstream at k-sigs/bom. Also, this file should be gitignored.

tools/provenance-generator/kn-attestation.json
upodroid: No, I still don't know the right values. Will need to talk to SLSA and puerco about that.

tools/provenance-generator/kn-attestation.json
upodroid: No, it is only required for SLSA 4 which we aren't targeting at all now.

tools/provenance-generator/kn-attestation.json
upodroid: Typo, both are complete.

tools/provenance-generator/main.go
upodroid: This file created by the clonerefs init container.

https://storage.googleapis.com/knative-prow/logs/nightly_net-istio_main_periodic/1548602272080990208/podinfo.json

tools/provenance-generator/pkg/parse.go
upodroid: It is not exposed as an env variable, so reading the timestamp of when the clone log was written is the best approximation. After the clone is complete, the test container starts.

tools/provenance-generator/pkg/parse.go
upodroid: Yes, I did os.Stat on it a few lines earlier and that error is catched.

tools/provenance-generator/pkg/parse.go
upodroid: Yes, to avoid empty lines and malformed strings from being added as subject.

tools/provenance-generator/pkg/parse.go
upodroid: I spent alot of time looking for a package that can parse docker image reference but the best one I found was the docker/distribution package. Still annoyed that it doesn't have a function that parses a image ref and returns hostname, digest, digest-algo, name, etc

upodroid:
Also, the in-toto attestation will be signed and pushed to the registry, right next to the images. Most of this is bleeding edge work with minimal docs and reference code.

vaikas:
I assume we can't use something like this because of prow? And furthermore I assume no reusable generator is for prow somewhere in k8s tools?
https://github.com/slsa-framework/slsa-github-generator#generation-of-provenance

Also, I wonder if this is something that we might want to upstream or have it be reusable so other projects could benefit from this also? Just thoughts up-front.

upodroid:
> I assume we can't use something like this because of prow? And furthermore I assume no reusable generator is for prow somewhere in k8s tools? https://github.com/slsa-framework/slsa-github-generator#generation-of-provenance
> 
> Also, I wonder if this is something that we might want to upstream or have it be reusable so other projects could benefit from this also? Just thoughts up-front.

I saw the generators in that repo but they are for github actions and not relevant to us. My generator will work for any jobs that run on Prow so Istio can benefit from this.

The Kubernetes tooling that Adolfo wrote at https://github.com/kubernetes/release/blob/master/pkg/anago/stage.go isn't useful for us as their release process is very different and they use GCB and multistage releases.

vaikas:
Super thanks @upodroid just wanted to check my assumptions!

tools/provenance-generator/kn-attestation.json
```diff
@@ -0,0 +1,56 @@
...
+          "entryPoint": "https://github.com/knative/serving/blob/main/hack/release.sh"
+        },
+        "parameters": [
+          "runner.sh",
+          "./hack/release.sh",
+          "--auto-release",
+          "--release-gcs=knative-releases/async-component",
+          "--release-gcr=gcr.io/knative-releases",
+          "--github-token=/etc/hub-token/token"
+        ]
```
puerco, line 34: How is the build triggered? This section should capture the details of whatever triggered the build.

I think the parameters you need to create the resulting `buildConfig` are captured in the [container `ENTRYPOINT_OPTIONS`](https://storage.googleapis.com/knative-prow/logs/nightly_net-istio_main_periodic/1548602272080990208/podinfo.json#:~:text=%22name%22%3A%20%22ENTRYPOINT_OPTIONS%22%2C), which is great because you don't have to figure out how to capture them at the start of the job:

```json
{
  "name": "ENTRYPOINT_OPTIONS",
  "value": "{\"timeout\":7200000000000,\"grace_period\":15000000000,\"artifact_dir\":\"/logs/artifacts\",\"args\":[\"runner.sh\",\"./hack/release.sh\",\"--publish\",\"--tag-release\"],\"container_name\":\"test\",\"process_log\":\"/logs/process-log.txt\",\"marker_file\":\"/logs/marker-file.txt\",\"metadata_file\":\"/logs/artifacts/metadata.json\"}"
}
```

The goal of these sections of the provenance attestation is to express something like:

>  "prow (builder id) did (invocation) this (buildConfig) to the source (materials) and as a result we now have these artifacts (the subjects)"

tools/provenance-generator/kn-attestation.json
puerco: In this case we need a URL to let others what kind of build we are running. In general, I think this is most useful when you are running a build-as-code scenario. This would point reviewers to  the build definition spec.

My opinion here is that since prow is running a shell script, it is probably self-explanatory and we can point the URL to the shell script.

tools/provenance-generator/attestation.json
puerco: Sure we can fix it upstream! Lets chat soon to align these projects

tools/provenance-generator/pkg/generate.go
```diff
@@ -0,0 +1,53 @@
...
+	p.Metadata.Completeness.Parameters = true  // The parameters are complete as we know the from prow
+	p.Metadata.Completeness.Materials = true   // The materials are complete as we only use the github repo
+	p.Metadata.Completeness.Environment = true // We don't use environment values to build images/binaries
+	startTime := config.StartTime.UTC()
+	endTime := time.Now().UTC()
+	p.Metadata.BuildStartedOn = &startTime
+	p.Metadata.BuildFinishedOn = &endTime
+	p.Invocation.ConfigSource.EntryPoint = repo + "/blob/" + cloneRecord.Refs.BaseRef + strings.Trim(config.EntryPointOpts.Args[1], ".")
+	p.BuildType = "https://cloudbuild.googleapis.com/CloudBuildYaml@v1"
+	p.Invocation.Parameters = config.EntryPointOpts.Args
```
puerco, line 40: Move these parameters to the `buildConfig` section. It should be a struct describing what prow ran. You can list anything there so think of a way to express it which can give a process or human a way to rerun the step.

In the invocation we need to document the trigger for the prow job.

puerco (commented):
This is awesome work @upodroid! The parameters need a little bit of fine-tuning but I think you got the most important part right: Where to get the data for the run. Querying the build system will give you the data you need from a source out of reach from the build process.

I would like to suggest we sync and share some code, we are working on some newer features which will be able to help you a bit more and the prow query you did here is one feature I was aiming to build later in the year, but have it already figured out!

I left some notes below, let me know if I can help with anything!

upodroid:
We really need to get rid vendor soon(sorry about the large diff), anyway:
- provenance-generator is using the root go.mod, has the side-effect if vendoring more files
- root go.mod deps didn't play nice with the deps specified by tools/configgen package. So I bumped those deps. As a result, imagePullPolicy field had to be removed from jobs_config as it was never generated properly.
- Lining up the deps was painful.
- Applied all the suggestions from Adolfo and Evan.
- Added tests
- go.work* is breaking some knative/hack scripts

dprotaso:
Approving to unblock - further changes can be done in follow up PRs

PR #3470, opened 2022-08-05
add secret for perf tests

nader-ziada:
- add secret to the serving perf tests job so it can access influxdb running on the community cluster, secret already added in gcp secret manager


upodroid (approved):
LGTM

/approve

PR #3472, opened 2022-08-08
update influx secret name in config

nader-ziada:

update secret name in external secrets so it matches with jobs config


upodroid (approved):
LGTM

/approve

PR #3478, opened 2022-08-11
update influx external secrets

nader-ziada:
Trying to get the influxdb external secrets working with prow


nader-ziada:
@upodroid giving it another try, followed the example in the docs here https://external-secrets.io/v0.5.9/provider-google-secrets-manager/

upodroid (approved):
LGTM

btw, the secret shouldn't be stored as base64 but as a string. The operator will convert it to b64. The configuration of other secrets is a more complex way of doing it and requires templating.

I changed the influx-url-secret to secret, the other one looked a bit strange so i'll let you do that one.

PR #3484, opened 2022-08-15
Upgrade Boskos

upodroid:
- Boskos is now autobumped
- Switched authentication to workload identity
- Gave it more CPU

/cc @dprotaso


prow/cluster/build/boskos-janitor.yaml
```diff
@@ -30,32 +30,21 @@ spec:
       labels:
         app: boskos-janitor
     spec:
-      serviceAccountName: "boskos"
+      serviceAccountName: boskos
       terminationGracePeriodSeconds: 300
       containers:
       - name: boskos-janitor
-        image: gcr.io/k8s-staging-boskos/janitor:v20200819-984516e
```
dprotaso, line 37: wow that's old

PR #3498, opened 2022-08-24
Fix `run_job.sh` scripts

upodroid:
We moved some files around in the repo sometime ago + moved to regional cluster and the run_job.sh script hasn't been updated.

/cc @dprotaso @kvmware

### DIFF
--- a/guides/manual_release.md
+++ b/guides/manual_release.md
@@ -169,7 +169,8 @@ Example:
    `[RELEASE_BRANCH_VERSION-]` with the release branch number (e.g., `1.4`)
 
    ```
-   cd config/prow
+   # run gcloud auth login first and pull the latest commit in main branch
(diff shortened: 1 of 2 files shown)


guides/manual_release.md
```diff
@@ -169,7 +169,8 @@ Example:
    `[RELEASE_BRANCH_VERSION-]` with the release branch number (e.g., `1.4`)
 
    ```
-   cd config/prow
+   # run gcloud auth login first and pull the latest commit in main branch
```
kvmware, line 172: can this be scriptified? ie. shouldn't we just call `gcloud auth login` before here?

upodroid: No, because gcloud auth login is a browser based flow. You are already expected to run that for access to the clusters and other things.

PR #3500, opened 2022-08-25
Deploy Grafana

upodroid:
Busy PR but:
- Grafana has been deployed and is reading metrics from Managed Prometheus https://grafana.knative.dev
- I need to create the prom rules and deploy them. (WIP)
- Created backing infra in GCP
- Gateway API is in beta but it is much better than ingress.

/cc @kvmware


infra/gcp/boskos/main.tf
```diff
@@ -1,12 +1,12 @@
...
-  required_providers {
-    google = {
-      version = "4.15"
-    }
-    google-beta = {
-      version = "4.15"
-    }
-  }
-}
+# terraform {
```
kvmware, line 3: Lots of commenting out through out this PR, is that necessary? Can't we just remove and if we need it back we can revert?

kvmware:
Other than the comment i left LGTM

infra/gcp/boskos/main.tf
upodroid: Fixed

upodroid: Oops forgot to push, afk

upodroid: This is ready now.

PR #3503, opened 2022-08-26
Remove some projects from boskos pool for testing

upodroid:
I need to test boskos in a separate deployment so I'm sharding off 10 projects for testing.

/cc @kvmware


kvmware:
Why we deleting the README? if it is out of date we should update it.

upodroid:
I'm deleting it because the config is in a separate file. Also, that readme is no longer needed. It should be clear where the configmap is at now.

PR #3506, opened 2022-08-29
Workload Identity Enablement - Part Four

upodroid:
Part of #3345 

This PR does the following:
- Fixes a bug from #3503, a prowjob was expecting something there so I pointed it to the new file.
- Removed service account keys from all test-infra jobs
- One job didn't even need access to Google Cloud at all. 

/cc @kvmware


kvmware:
Whats up with all the vendor files being removed?

upodroid:
I need to bump some deps and it is a bit trickier than expected.

upodroid:
This should be good now.

1. Diffs in prow/jobs/* is caused by https://github.com/istio/test-infra/pull/4241 a nice change by istio
2. This PR should reduce the vendor changes for #3456

kvmware:
We might have to discuss this one over video chat. +1mil changes and -14k is something that does not make me feel to easy. I like more removal than adds :P

kvmware:
CC @cardil for go expertise

upodroid:
ok, lets talk about it at the productivity meeting. It is mostly vendor/* changes

kvmware:
Whatever is pulling in +1mil changes in vendor files we don't need imo

upodroid:
PR is ready to merged

upodroid:
you need to redo your last comment and put thanks on a separate line

PR #3508, opened 2022-09-01
Expand boskos pool and enable more logging

upodroid:
This PR:
- Changes how we create projects for the boskos pool. It is done via Terraform and I got rid of knative-gcp specific stuff as we don't do those anymore.
- Changed log severity to debug to see extra logs
- Increased replica to 6 and reduces concurrency of janitor. Should allow more frequent attempts to cleanup


kvmware:
We should at some point add some tests/validation for TF files as we have more of those now.

PR #3509, opened 2022-09-01
Bump Deps Manually

upodroid:
Makes reviewing much easier #3456

/cc @dprotaso @kvmware

PR #3510, opened 2022-09-02
Updating deps with submodules

cardil:
Related to #3509


upodroid:
/hold

We are landing a very large go.mod changes in #3509 next week which is required for the SLSA Generator that I'm working on.

go.mod
```diff
@@ -1,30 +1,92 @@
...
+	golang.org/x/oauth2 v0.0.0-20220608161450-d0670ef3b1eb
+	google.golang.org/api v0.83.0
+	k8s.io/apimachinery v0.24.2
+	knative.dev/hack v0.0.0-20220902140220-97fd663bf52b
+	sigs.k8s.io/boskos v0.0.0-20210730172138-093b54882439
+	sigs.k8s.io/yaml v1.3.0
+)
+
+// TODO: remove this once we have a fixed knative-hack (https://github.com/knative/hack/issues/216)
+replace knative.dev/hack => github.com/cardil/knative-hack v0.0.0-20220902210110-5d40dd4a1f82
```
mgencur, line 31: I think this one was already merged? Let's remove the TODO

mgencur (commented):
Sorry if this is a dumb question but this current PR should replace https://github.com/knative/test-infra/pull/3509 if I'm not mistaken. 
The current PR adds 2,822,424 lines and https://github.com/knative/test-infra/pull/3509 adds only 1,355,412 . Do we really need those files. Do we have a good reason to have this change?

go.mod
cardil: Yes, this should be removed.

cardil:
Re @mgencur:

> _Sorry if this is a dumb question but this current PR should replace https://github.com/knative/test-infra/pull/3509 if I'm not mistaken.
> The current PR adds 2,822,424 lines and https://github.com/knative/test-infra/pull/3509 adds only 1,355,412 . Do we really need those files. Do we have a good reason to have this change?_

That depends on what we want. This PR adds the "hidden" modules `tools/configgen`, `tools/release-jobs-syncer`, `tools/rundk` to Go workspace. This makes them work properly. Together with https://github.com/knative/hack/issues/216 it ends up vendoring all thei dependencies in appopriate `vendor` directories.

The #3509 is doing the oposite - it is removing the `go.mod` for those tools. In effect all dependencies are being vendored on root level `vendor` dir.

Both PRs are updating the deps, and fixing the invalid go modules structure.

The key difference between those is that mine is a bit safer, becouse the dependencies for those submodules are kept at the same version. #3509 is aliging all those version to one defined in root `go.mod`.

I'm fine with both approaches.

upodroid:
I'm leaning towards #3509 as well (bit biased). In any case, Chris is working on removing vendor anyway so the line count doesn't really matter.

Just to note, no one should be using `knative/test-infra` as a library so a messy root `go.mod` is fine. The root `go.mod` is a mess because of dependency on `kubernetes/test-infra` and `istio-test-infra` (istio relies on k8s/test-infra). 

If life wasn't hard enough, our k8s version is determined by `knative/pkg` which always lags 

https://github.com/kubernetes/test-infra/blob/master/go.mod

cardil:
Lets go with https://github.com/knative/test-infra/pull/3509 approach. It's simpler.

PR #3513, opened 2022-09-05
:gift: Modscope tool

cardil:
Related knative/hack#220

**Why is this needed?**
To properly fix knative/hack#220 we need a better tooling for resolving the list of modules listed in `go.work`, and the current module as well. This tool adds both, by porting the same code from Golang https://github.com/golang/go/blob/4c1ca42aa295d68b6172b7f49e4fb4fd14d86475/src/cmd/go/internal/modload/init.go#L1274-L1319 and using development version of [golang.org/x/mod/modfile](https://cs.opensource.google/go/x/mod/+/master:modfile/) package.


.editorconfig
```diff
@@ -0,0 +1,15 @@
...
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+indent_style = space
+indent_size = 2
+
+[{*.go,Makefile}]
+indent_style = tab
```
mgencur, line 15: Should this editorconfig be sent in a separate PR? how does it belong to this one?

mgencur (commented):
@cardil This PR is quite big. Can you give an overview which stuff you wrote and which stuff is imported from somewhere else? There's a lot of stuff under tools/modscope, where does it come from? Just some notes about things that could make the review easier.

.editorconfig
cardil: I could remove it, if you feel like it. It's not crucial. I just piggyback it here for simplicity.

cardil:
> _@cardil This PR is quite big. Can you give an overview which stuff you wrote and which stuff is imported from somewhere else? (and from where). There's a lot of stuff under tools/modscope, where does it come from? Just some notes about things that could make the review easier._

Well. I think the PR description (together with linked issue) answers all of those questions you've asked. All the stuff in this PR is my original work. Some functions are only inspired by Golang implementation, as stated in PR description.

upodroid (commented):
The code looks good.

Are you planning on running this on a prow job? If so, you need to add the install the binary on the prow-tests image.

kvmware:
LGTM

upodroid (approved):
LGTM

thank you

PR #3517, opened 2022-09-07
Stop running docs presubmit tests

psschwei:
Signed-off-by: Paul S. Schweigert <paul@paulschweigert.com>

The unit / build / continuous tests for the docs repo aren't doing meaningful work (they test the code samples, not the main docs site), and with the recent hack update to support multi-module projects (https://github.com/knative/hack/pull/205), the build tests are actually failing. 

Given that these tests are basically noise, we should stop running them (see discussion in slack thread [here](https://knative.slack.com/archives/C9CV04DNJ/p1662554973989079)). 

For DUX signoff:
/assign @abrennan89 @snneji 

For Prow update (I dropped the docs config file and regenerated the config, not sure if I should've kept the file and just deleted the jobs):
/assign @kvmware @upodroid


upodroid:
+1 from me

Leaving LGTM to DUX leads

PR #3519, opened 2022-09-08
Sign Images and Binaries for net-certmanager and client nightlies

upodroid:
I need to bump the deps in knative/client and knative-sandbox/net-certmanager.

/hold

/cc @kvmware


upodroid:
Can I get lgtm?

PR #3523, opened 2022-09-13
create separate module for the provenance-generator & release-job-syncer

dprotaso:
this allows us to drop the replace directives and allow kntest and other tools to be installed using go install


upodroid (approved):
LGTM

PR #3533, opened 2022-09-19
Sign all nightlies and release binaries and images

upodroid:
We have been successfully signing netcertmanager nightlies for a while successfully #3519

This PR signs all nightlies and release images/binaries

/cc @cardil


cardil (approved):
/lgtm

PR #3535, opened 2022-09-21
Enabled short e2e test mode for periodic s390x tests due to environme…

### LINKED ISSUE #3534: s390x: run perodic e2e tests in 'short' mode
The e2e tests contain dedicated scale tests (e.g. AutoScaling) that can only be run on a dedicated cluster with appropriate hardware resources. At the moment, all tests on s390x are run on a KinD cluster with the biggest underlying hardware configuration being 16 IFLs and 128 GB RAM which seems not to be sufficient.

Due to this and the following code: https://github.com/knative/serving/blob/58cce5440428e7e1671106ba0951538e378dc890/test/e2e/autoscale_test.go#L102
the e2e tests will be run in 'short' mode only, effectively disabling the scale tests.

haubenr:
…nt limitations

Signed-off-by: Dirk Haubenreisser <haubenr@de.ibm.com>

Fixes #3534

### DIFF
--- a/prow/jobs/generated/knative/serving-main.gen.yaml
+++ b/prow/jobs/generated/knative/serving-main.gen.yaml
@@ -683,7 +683,7 @@ periodics:
       - name: SYSTEM_NAMESPACE
         value: knative-serving
       - name: TEST_OPTIONS
-        value: --enable-alpha --enable-beta --resolvabledomain=false
+        value: --enable-alpha --enable-beta --resolvabledomain=false --short
       - name: E2E_CLUSTER_REGION
         value: us-central1
       - name: DOCKER_IN_DOCKER_ENABLED
@@ -772,7 +772,7 @@ periodics:
       - name: SYSTEM_NAMESPACE
         value: knative-serving
       - name: TEST_OPTIONS
-        value: --enable-alpha --enable-beta --resolvabledomain=false
+        value: --enable-alpha --enable-beta --resolvabledomain=false --short
       - name: E2E_CLUSTER_REGION
         value: us-central1
       - name: DOCKER_IN_DOCKER_ENABLED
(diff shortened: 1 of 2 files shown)


prow/jobs/generated/knative/serving-main.gen.yaml
```diff
@@ -683,7 +683,7 @@ periodics:
       - name: SYSTEM_NAMESPACE
         value: knative-serving
       - name: TEST_OPTIONS
-        value: --enable-alpha --enable-beta --resolvabledomain=false
+        value: --enable-alpha --enable-beta --resolvabledomain=false --short
```
cardil, line 686: The `-short` doc: https://github.com/golang/go/blob/d11c58eedbee29b4912dd50508b36ad15ebb739e/src/cmd/go/internal/test/test.go#L311-L315

I think the `-short` flag should be used for unit test.

It is also used in default unit test runner in Knative: https://github.com/knative/hack/blob/d1317b08c0c3a7180e1dd9e9faec7532fdfe615f/presubmit-tests.sh#L205

cardil (changes requested):

prow/jobs/generated/knative/serving-main.gen.yaml
cardil: Better would be to introduce a clear, flake related flag/envvar like: `-skip-flakes`

haubenr: Well, it would seem that knative is using the '--short' flag differently to what golang proposes - at least in context of the e2e tests. If you look at the current serving test cases you'll find two occurrences of `testing.Short()` which are used to disable specific scale tests, one of them directly noting that KinD clusters are insufficient for running those tests. My goal with this PR is to piggyback on existing integrations, not coming up with new ways of how to do things.

cardil: I clearly get your intention of piggybacking. But, I think that's wrong. We shouldn't put flaky tests to slow bucket. That's why I'm proposing to introduce flake-specific flag instead.

haubenr: Thanks Chris for your input. I'll think of something else. For now I will close this PR.

PR #3568, opened 2022-10-12
Fix a bug in the Provenance Generator

upodroid:
Fixing a few bugs in the provenance-generator

- endTime was being fetched from Prow which doesn't exist when the generator is being run.
- `config.ProwJob.Status` has been emptied now as it is not complete when the attestation is ran

/cc @dprotaso @kvmware

### DIFF
--- a/tools/provenance-generator/pkg/generate.go
+++ b/tools/provenance-generator/pkg/generate.go
@@ -2,6 +2,7 @@ package pkg
 
 import (
 	"os"
+	"time"
 
 	slsa "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v0.2"
 	prowapi "k8s.io/test-infra/prow/apis/prowjobs/v1"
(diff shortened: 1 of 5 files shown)


tools/provenance-generator/pkg/generate.go
```diff
@@ -34,16 +35,18 @@ func GenerateAttestation(config Config) *provenance.Statement {
...
 	startTime := config.ProwJob.CreationTimestamp.Time.UTC()
-	endTime := config.ProwJob.Status.CompletionTime.Time.UTC()
+	endTime := time.Now().UTC() // The prowjob is still running when this timestamp is needed. Attestation happens after the images/binaries are built
 	p.Metadata.BuildStartedOn = &startTime
 	p.Metadata.BuildFinishedOn = &endTime
 	p.Invocation.ConfigSource.EntryPoint = "https://github.com/knative/test-infra/tree/main/prow/jobs/generated/" + cloneRecord.Refs.Org
 	p.BuildType = "https://prow.knative.dev/ProwJob@v1"
 	var buildConfig interface{}
+	// set config.ProwJob.Status to empty as it is not complete when the generator runs
+	config.ProwJob.Status = prowapi.ProwJobStatus{}
```
upodroid, line 49: This doesn't empty the struct properly.

```
        -                               "status":   map[string]any{"startTime": nil},
        +                               "status":   map[string]any{"startTime": string("")},
```

tools/provenance-generator/pkg/testdata/prowjob.yaml
```diff
@@ -111,15 +111,3 @@ spec:
       report: true
   prowjob_defaults:
     tenant_id: GlobalDefaultID
-status:
```
kvmware, line 114: why is this being removed both in json and yaml?

upodroid: The yaml is loaded as a fakeclient to test fetching ProwJobs. I have decided not to include the status field as it is not complete.

PR #3587, opened 2022-10-24
Fix release process for eventing-kafka-broker

aliok:
Fixes #

- Fix release process for eventing-kafka-broker
- Add missing env var for the release process
- That env var is not really needed for the release itself, but it is used in the tests that are ran just before the release


aliok:
I am not sure how prow/jobs_config/knative-sandbox/eventing-kafka-broker-release-1.8.yaml is created. If it is a copy paste of 1.7 and manually changing things, then we should be good for 1.9.
If that file is generated, I should modify the source.

@pierDipi maybe you know

upodroid (approved):
LGTM

1.9 release will copy the release job as it is configured when the release is cut

PR #3593, opened 2022-10-31
Allow jobs to be cancelled in the UI + run only one prowjob at a time for nightly/release jobs

upodroid:
This change allows:
- Prowjobs can now be cancelled from the web browser without the need to login to the cluster. https://kubernetes.slack.com/archives/CDECRSC5U/p1666217793138119
- Various teams listed here https://github.com/knative/test-infra/blob/main/prow/config.yaml#L140 can cancel and rerun jobs.
-  I have set max_concurrency for all release/nightly jobs to 1 to avoid multiple concurrent runs.

/cc @kvmware

### DIFF
--- a/prow/cluster/control-plane/301-rbac.yaml
+++ b/prow/cluster/control-plane/301-rbac.yaml
@@ -104,6 +104,7 @@ rules:
       - watch
       # Required when deck runs with `--rerun-creates-job=true`
       - create
+      - patch
 ---
 kind: Role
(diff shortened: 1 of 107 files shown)


kvmware:
/lgtm
i think the gofmt errors are from files that were not changed

upodroid:
/override "style / Auto-format and Check (gofmt)" "style / Auto-format and Check (goimports)"

PR #3609, opened 2022-11-08
Use the GKE auth plugin in order to avoid spammy test output

dprotaso:
Details here: https://cloud.google.com/blog/products/containers-kubernetes/kubectl-auth-changes-in-gke

/assign @upodroid


upodroid (approved):
LGTM

Interesting

PR #3622, opened 2022-11-18
Remove notarisation flags from func 1.7 jobs

upodroid:
When the repo got renamed, the 1.7 job was deleted so a new one was made with the latest flag and it uses features not backported to 1.7.

/cc @lance @evankanderson


lance:
/lgtm
Thanks!

PR #3639, opened 2022-12-07
bump go versions

dprotaso:
CVE annoucements
https://groups.google.com/g/golang-announce/c/L_3rmdT0BMU


dprotaso:
/assign @kvmware 
cc @evankanderson

Issue #3638, opened 2022-12-07
Drop infrastructure support for 1.5 jobs


dprotaso:
It's EOL https://github.com/knative/community/pull/1222

/assign @kvmware

kvmware:
https://github.com/knative/test-infra/pull/3651

PR #3640, opened 2022-12-07
Remove unsupported 1.5 relase

kvmware:
Fixes # https://github.com/knative/test-infra/issues/3638


dprotaso:
I see there was a job that did this - https://github.com/knative/test-infra/pull/3585

I wonder why it's not pruning 1.5 for EOL - unless it's hard coded to support the last 4 releases - which we're in a transition to support the last two

dprotaso:
Would merging this PR trigger the automation to create a new PR adding the manifests back?

PR #3649, opened 2022-12-16
add new job to run internal encryption tests for Contour

KauzClay:
* run test added in https://github.com/knative/serving/pull/13559 on each PR
* test does nothing now, will be implemented in a later PR (https://github.com/knative/serving/pull/13536)
* Part of a series of PRs to add an e2e test for Internal Encryption feature
* In support of implementing internal encryption for net-contour here: https://github.com/knative-sandbox/net-contour/pull/819.


kvmware:
@KauzClay Looks like update-codegen script needs to be ran. The prow job_config yaml is just a template the codegen tool will generate the actual job yaml

PR #3672, opened 2023-01-17
Fix config-bootstrapper and test-runner RBAC

upodroid:
/cc @dprotaso @kvmware 

1st commit fixes the e2e serving errors @dprotaso mentioned
2nd commit fixes the config-bootstrapper errors. Basically, it was pointing to the old prow control plane cluster which I deleted this month. 

Contents of the kubeconfig:

```
 REDACTED  C02CW0CDP3YY  ~  $  kg secret -n test-pods prow-kubeconfig -o yaml | yq '.data."kubeconfig.yaml"' | base64 -d | yq .contexts
- context:
    cluster: prow-build
    user: prow-build
  name: prow-build
- context:
    cluster: prow-trusted
    user: prow-trusted
  name: prow-trusted
- context:
    cluster: prow-control-plane
    user: prow-control-plane
  name: prow-control-plane
```

I'm working with sig-testing at Kubernetes on getting rid of the bearer tokens and using the gke auth plugins.

https://github.com/kubernetes/test-infra/issues/28142


kvmware:
You have some no newline issues else lgtm

upodroid:
/hold

I had to revert my change earlier. Prow unfortunately doesn't understand different contexts properly. It assumes all users have cluster-admin which isn't good.

First order of business is extending config-bootstrapper to use a context.

upodroid:
A quick fix would be to load a kubeconfig with just only the config-bootstrapper context.

upodroid:
Also, it is 7 days till the release day so I don't really want to make some changes this week and cause testgrid to show extra flakes from infra changes/snafu

upodroid:
/hold cancel

I created a dedicated secret for it with just a single context. I will experiment with the contexts after the release is done.

PR #3677, opened 2023-01-24
Generate Provenances for Release jobs under knative-sandbox

upodroid:
I forgot that the release jobs for knative-sandbox are overriden.

Follow up from #3673 

/cc @psschwei @evankanderson


psschwei (commented):
/lgtm

PR #3678, opened 2023-01-25
Make eventing-kafka upgrade job optional

creydr:
Due to some deprecated/removed API versions we have some chicken-egg problems on the upgrade test jobs (see https://github.com/knative-sandbox/eventing-kafka/pull/1287 for example).
This PR marks them as optional in the meantime


upodroid (commented):
/approve

pierDipi (approved):
/lgtm

PR #3694, opened 2023-01-30
Revert "add new job to run internal encryption tests for Contour (#36…

KauzClay:
This reverts commit 71d847859a56204be4c81e380f8b2b5f6effd4dc.

Based on feedback here: https://github.com/knative/serving/pull/13536

There is a different approach that will get much better test coverage than this test job.


psschwei (commented):
/lgtm

PR #3696, opened 2023-01-31
:bug: Buoy will not float current project

cardil:
#### Which issue(s) this PR fixes:

Fixes https://github.com/knative/hack/issues/264

/lint

### DIFF
--- a/buoy/commands/needs.go
+++ b/buoy/commands/needs.go
@@ -34,7 +34,11 @@ func addNeedsCmd(root *cobra.Command) {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			gomods := args
 
-			_, packages, err := gomod.Modules(gomods, domain)
+			selector, err := gomod.DefaultSelector(domain)
+			if err != nil {
+				return err
+			}
+			_, packages, err := gomod.Modules(gomods, selector)
 			if err != nil {
 				return err
 			}
--- a/pkg/gomod/selector.go
+++ b/pkg/gomod/selector.go
@@ -1,0 +1,104 @@
+package gomod
+
+import (
+	"log"
+	"strings"
+	"sync"
+
+	"github.com/pkg/errors"
+	"knative.dev/test-infra/pkg/gowork"
+)
+
+var (
+	// ErrNoDomain is returned when no domain is provided.
+	ErrNoDomain = errors.New("no domain provided")
+)
+
+// Matcher matches a module name.
+type Matcher interface {
+	// Match returns true if the module name matches.
+	Match(modname string) bool
+}
+
+// MatcherFunc is a function that implements the Matcher interface.
+func MatcherFunc(fn matcherFunc) Matcher {
+	return fn
+}
+
+type matcherFunc func(modname string) bool
+
+func (fn matcherFunc) Match(modname string) bool {
+	return fn(modname)
+}
+
+// Selector is a set of matchers that can be used to select a subset of modules.
+type Selector struct {
+	Includes []Matcher
+	Excludes []Matcher
+}
+
+// Select returns true if the module should be selected.
+func (s Selector) Select(modname string) bool {
+	selected := false
+	for _, m := range s.Includes {
+		if m.Match(modname) {
+			selected = true
+			break
+		}
+	}
+	if !selected {
+		return false
+	}
+	for _, m := range s.Excludes {
+		if m.Match(modname) {
+			return false
+		}
+	}
+	return true
+}
+
+// CurrentModulesMatcher matches the current project modules.
+type CurrentModulesMatcher struct {
+	once sync.Once
+	mods []gowork.Module
+	err  error
+}
+
+func (c *CurrentModulesMatcher) Match(modname string) bool {
+	c.once.Do(func() {
+		os := gowork.RealSystem{}
+		c.mods, c.err = gowork.List(os, os)
+		if c.err != nil && errors.Is(c.err, gowork.ErrInvalidGowork) {
+			var m *gowork.Module
+			m, c.err = gowork.Current(os, os)
+			if m != nil {
+				c.mods = []gowork.Module{*m}
+			}
+		}
+	})
+	if c.err != nil {
+		log.Fatal(c.err)
+	}
+	for _, m := range c.mods {
+		if m.Name == modname {
+			return true
+		}
+	}
+	return false
+}
+
+// DefaultSelector returns a selector that includes modules with given domain,
+// but excludes the current project modules.
+func DefaultSelector(domain string) (Selector, error) {
+	domain = strings.TrimSpace(domain)
+	if len(domain) == 0 {
+		return Selector{}, ErrNoDomain
+	}
+
+	return Selector{
+		Includes: []Matcher{MatcherFunc(func(modname string) bool {
+			return strings.HasPrefix(modname, domain)
+		})},
+		Excludes: []Matcher{&CurrentModulesMatcher{}},
+	}, nil
+}
--- a/pkg/gowork/system.go
+++ b/pkg/gowork/system.go
@@ -1,0 +1,54 @@
+package gowork
+
+import (
+	"fmt"
+	"io/fs"
+	"os"
+	"path"
+	"path/filepath"
+)
+
+const rootDir = "/"
+
+var (
+	systemFS                 = os.DirFS(rootDir) //nolint:gochecknoglobals
+	_        operatingsystem = RealSystem{}
+)
+
+// RealSystem is the real, live OS implementation.
+type RealSystem struct{}
+
+func (s RealSystem) Get(name string) string {
+	return os.Getenv(name)
+}
+
+func (s RealSystem) Open(name string) (fs.File, error) {
+	f, err := systemFS.Open(name)
+	if err != nil {
+		return nil, fmt.Errorf("%w: %v", ErrBug, err)
+	}
+	return f, nil
+}
+
+func (s RealSystem) Cwd() (string, error) {
+	dir, err := os.Getwd()
+	if err != nil {
+		return "", fmt.Errorf("%w: %v", ErrBug, err)
+	}
+	p, err := filepath.Rel(rootDir, dir)
+	if err != nil {
+		return "", fmt.Errorf("%w: %v", ErrBug, err)
+	}
+	return p, nil
+}
+
+func (s RealSystem) Abs(filepath string) string {
+	return path.Clean(path.Join(rootDir, filepath))
+}
+
+// OS represents a virtual operating system.
+type operatingsystem interface {
+	FileSystem
+	Environment
+	Abs(filepath string) string
+}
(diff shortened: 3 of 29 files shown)


buoy/commands/needs.go
```diff
@@ -34,7 +34,11 @@ func addNeedsCmd(root *cobra.Command) {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			gomods := args
 
-			_, packages, err := gomod.Modules(gomods, domain)
+			selector, err := gomod.DefaultSelector(domain)
```
kvmware, line 37: This bit of logic is used 3 different places, probably should pull this out into a funtion that all 3 files can call

kvmware: EDIT: 8 different places. Should for sure be pulled out to a common function

cardil: Hi @kvmware. I think, I do not follow. What code in specific you have in mind?

```golang
selector, err := gomod.DefaultSelector(domain)
if err != nil {
  return err
}
```

The above code couldn't really be shortened in the Golang. The syntax can't let us do it. At least that's what I'm thinking.

kvmware:
LGTM, but someone else might wanna look over it before.

cardil:
/hold

I'd like to confirm on kn-plugin-event...

cardil:
## Confirmed to work as intended on kn-plugin-event.

### Current `latest` buoy

```shell
$ cd ~/git/kn-plugin-event/build
$ go run knative.dev/test-infra/buoy@latest float \
  --domain knative.dev \
  --release 1.9 \
  --module-release 0.36 go.mod
knative.dev/kn-plugin-event@main
$ cd ..
$ go run knative.dev/test-infra/buoy@latest float \
  --domain knative.dev \
  --release 1.9 \
  --module-release 0.36 go.mod
knative.dev/client-pkg@release-1.9
knative.dev/eventing@v0.36.1
knative.dev/hack@release-1.9
knative.dev/pkg@release-1.9
knative.dev/reconciler-test@release-1.9
knative.dev/serving@v0.36.0
```

_**NOTE:** Notice the buoy running in `build/` dir prints the `knative.dev/kn-plugin-event@main`, which is wrong._

### Buoy compiled from this PR

```shell
$ go build -o ~/bin/buoy ./buoy
$ cd ~/git/kn-plugin-event/build
$ ~/bin/buoy float \
  --domain knative.dev \
  --release 1.9 \
  --module-release 0.36 go.mod
$ cd ..
$ ~/bin/buoy float \
  --domain knative.dev \
  --release 1.9 \
  --module-release 0.36 go.mod                               
knative.dev/client-pkg@release-1.9
knative.dev/eventing@v0.36.1
knative.dev/hack@release-1.9
knative.dev/pkg@release-1.9
knative.dev/reconciler-test@release-1.9
knative.dev/serving@v0.36.0
```

/hold cancel

pkg/gomod/selector.go
```diff
@@ -0,0 +1,104 @@
...
+// Matcher matches a module name.
+type Matcher interface {
+	// Match returns true if the module name matches.
+	Match(modname string) bool
+}
+
+// MatcherFunc is a function that implements the Matcher interface.
+func MatcherFunc(fn matcherFunc) Matcher {
+	return fn
+}
```
evankanderson, line 26: Do we need an interface / adapter here, or would a simple type alias like `matcherFunc` do? i.e. what if we changed Selector to:

```golang

// Filter is a function which accepts or rejects a module name
type Filter func(modname tring) bool


// Selector filters modules based on a set of include and exclude functions.
type Selector struct {
  Includes []Filter
  Excludes []Filter
}
```

pkg/gomod/selector.go
```diff
@@ -0,0 +1,104 @@
...
+
+// Selector is a set of matchers that can be used to select a subset of modules.
+type Selector struct {
+	Includes []Matcher
+	Excludes []Matcher
+}
+
+// Select returns true if the module should be selected.
+func (s Selector) Select(modname string) bool {
```
evankanderson, line 41: If you rename `Select` to `Match`, `Selector` becomes a `Matcher` implementation.

(Yes, this might push against the removal of the `Matcher` interface separate from the type alias.)

pkg/gomod/selector.go
```diff
@@ -0,0 +1,104 @@
...
+	return fn
+}
+
+type matcherFunc func(modname string) bool
+
+func (fn matcherFunc) Match(modname string) bool {
+	return fn(modname)
+}
+
+// Selector is a set of matchers that can be used to select a subset of modules.
```
evankanderson, line 34: With a list of Includes and Excludes, it feels worth documenting whether "Include wins" or "Exclude wins", which also controls whether the default behavior is to include or exclude.

pkg/gomod/selector.go
```diff
@@ -0,0 +1,104 @@
...
+			var m *gowork.Module
+			m, c.err = gowork.Current(os, os)
+			if m != nil {
+				c.mods = []gowork.Module{*m}
+			}
+		}
+	})
+	if c.err != nil {
+		log.Fatal(c.err)
+	}
```
evankanderson, line 81: Rather than calling `log.Fatal` here to panic, what about making a constructor for a `currentModulesMatcher` which can return `Matcher, error`?

pkg/gomod/selector.go
```diff
@@ -0,0 +1,104 @@
...
+	for _, m := range c.mods {
+		if m.Name == modname {
```
evankanderson, line 83: If we only use `m.Name`, does it make sense to extract them into something like a `sets.Set[string]`?

I think [`sets.String[string].Has`](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#Set.Has) would satisfy the `Filter` function interface I described above.

pkg/gomod/selector.go
```diff
@@ -0,0 +1,104 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+// DefaultSelector returns a selector that includes modules with given domain,
+// but excludes the current project modules.
+func DefaultSelector(domain string) (Selector, error) {
+	domain = strings.TrimSpace(domain)
```
evankanderson, line 93: Do we expect the `domain` to have surrounding space?

pkg/gomod/selector.go
```diff
@@ -0,0 +1,104 @@
...
+	if len(domain) == 0 {
+		return Selector{}, ErrNoDomain
+	}
+
+	return Selector{
+		Includes: []Matcher{MatcherFunc(func(modname string) bool {
+			return strings.HasPrefix(modname, domain)
+		})},
+		Excludes: []Matcher{&CurrentModulesMatcher{}},
```
evankanderson, line 102: Do we need `Selector` as a general interface outside this module?  Can it be non-exported (`selector`)?

pkg/gowork/system.go
```diff
@@ -0,0 +1,54 @@
+package gowork
```
evankanderson, line 1: Do we no longer require copyright headers on files?

upodroid (approved):
LGTM

pkg/gowork/system.go
```diff
@@ -0,0 +1,54 @@
...
+
+func (s RealSystem) Abs(filepath string) string {
+	return path.Clean(path.Join(rootDir, filepath))
+}
+
+// OS represents a virtual operating system.
+type operatingsystem interface {
+	FileSystem
+	Environment
+	Abs(filepath string) string
```
evankanderson, line 53: Do we need a new interface here, or something that can take a `FileSystem` and `path string` and return the absolute path?

evankanderson: In particular, I'm worried about hoisting what was an implementation in the `modscope` CLI into a general-purpose module without good reason for expanding the interface.  Can we keep the `operatingsystem` abstraction in `modscope/cli`?

evankanderson (commented):
I had a couple code style comments, but it looks like this has merged.

Including them for future consideration.

pkg/gomod/selector.go
cardil: I think the interface should stay, as there are 2 implementations: `matcherFunc` and `CurrentModulesMatcher` struct. The `CurrentModulesMatcher` struct holds `sync.Once` to collect project information just once.

pkg/gomod/selector.go
cardil: Agree. A bit more docs, could be beneficial.

pkg/gomod/selector.go
cardil: Maybe a good idea to replace `Select -> Match`. That would allow users to build composite `Matcher`'s using `Selector` as a Matcher itself. :grin:

pkg/gomod/selector.go
cardil: Yeah, maybe. Although, I wanted to have the lazy `CurrentModulesMatcher`. If none of the includes matches, there's no point going through the excludes. That's the reason for `sync.Once`.

Constructor will allow for better error handling, but will make the `CurrentModulesMatcher` always calculate the stuff...

pkg/gowork/system.go
cardil: Well, this `operatingsystem` interface is private. It isn't used anywhere. The `modscope/cli` have different `OS` interface (compatible OFC).

I use it just to simplify the ensure assignment on line 15:

```golang
var _ operatingsystem = RealSystem{}
```

Without it, I would need to type:

```golang
var (
  _ FileSystem  = RealSystem{}
  _ Environment = RealSystem{}
)
```

and that doesn't ensure the `Abs(filepath string) string` func.

OFC I could live without those assures...

pkg/gomod/selector.go
cardil: The modscope tool uses `m.Path`. Both attributes are used.

pkg/gomod/selector.go
cardil: The original code had this (by @n3wscott). Probably not required, as Bash scripts should pass parameters properly, but at the same time - it doesn't hurt to call it.

pkg/gomod/selector.go
cardil: Before this PR the functions `Module` and `Modules` both were public with `domain string` as parameter. 

I'm replacing that `domain string`, with the `selector Selector`. That's why it is public. It allows users of `Module` and `Modules` to pass whatever `Selector` they like (even not necessarily connected to the domain).

pkg/gowork/system.go
cardil: I haven't added them for 2y. There's a LICENSE on root repo level.

From: https://softwareengineering.stackexchange.com/a/125841

> _Your software is automatically covered under copyright, as agreed in international law._

I think adding the license header was a Google policy. I think it is pointless and outdated. The copyright law is active, no matter if you add a file header or not. Otherwise, it would be pretty silly law. The GPLv3 strongly encourage people to attach the header, but I doubt it's grounded in any legal stuff.

evankanderson (commented):
I've sent #3704 for these comments

PR #3704, opened 2023-02-03
Refactor selector with feedback from 3696

evankanderson:
I decided to follow up on my comments on #3696 for filtering Buoy modules.

I figured writing the refactor was the easiest way to communicate my ideas; this removes the need for `MatcherFunc` and simplifies the calling inferface for `Module` and `Modules` in a few places.  I'm considering replacing `Selector` with 'In()` and `Except`, but I'm not sure it is a substantial improvement.  (`Excludes` is only used in one place, and `In(string...)` and `Except(Matcher, Matcher)` would cover the same cases.)

/assign @cardil

### DIFF
--- a/pkg/gomod/selector.go
+++ b/pkg/gomod/selector.go
@@ -16,33 +16,6 @@ var (
 // Matcher matches a module name.
 type Matcher func(modname string) bool
 
-// Selector is a set of matchers that can be used to select a subset of modules.
-type Selector struct {
-	Includes []Matcher
-	Excludes []Matcher
-}
-
-// Select returns true if the module should be selected.
-func (s Selector) Match(modname string) bool {
-	for _, match := range s.Excludes {
-		if match(modname) {
-			return false
-		}
-	}
-
-	for _, match := range s.Includes {
-		if match(modname) {
-			return true
-		}
-	}
-
-	return false
-}
-
-// Assert that Selector.Match implements Matcher
-var selector_type_assert = Selector{}
-var _ Matcher = selector_type_assert.Match
-
 // CurrentModulesMatcher matches the current project modules.
 func CurrentModulesMatcher() (Matcher, error) {
 	os := gowork.RealSystem{}
@@ -81,10 +54,9 @@ func DefaultSelector(domain string) (Matcher, error) {
 		return nil, err
 	}
 
-	return Selector{
-		Includes: []Matcher{func(modname string) bool {
-			return strings.HasPrefix(modname, domain)
-		}},
-		Excludes: []Matcher{currentModules},
-	}.Match, nil
+	inDomainButNotCurrentWorkspace := func(modname string) bool {
+		return strings.HasPrefix(modname, domain) && !currentModules(modname)
+	}
+
+	return inDomainButNotCurrentWorkspace, nil
 }
--- a/pkg/gomod/selector.go
+++ b/pkg/gomod/selector.go
@@ -1,11 +1,10 @@
 package gomod
 
 import (
-	"log"
 	"strings"
-	"sync"
 
 	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"knative.dev/test-infra/pkg/gowork"
 )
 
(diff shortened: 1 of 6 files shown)


kvmware:
@cardil looks good to you?

evankanderson:
I think Chris is only on mobile this week, according to another message on Slack.

This is anti-urgent, so I think it can wait.

I simplified the code one more time, removing `Selector` as an exported type in favor of returning an anonymous function which captures the "module under prefix but not in my current workspace" requirement.

pkg/gomod/selector.go
```diff
@@ -15,90 +14,49 @@ var (
...
-	selected := false
-	for _, m := range s.Includes {
-		if m.Match(modname) {
-			selected = true
-			break
+// CurrentModulesMatcher matches the current project modules.
+func CurrentModulesMatcher() (Matcher, error) {
+	os := gowork.RealSystem{}
+	knownMods := sets.NewString()
+	mods, err := gowork.List(os, os)
```
cardil, line 23: As I said in #3696 such refactor will remove the laziness, which was being done using `sync.Once`.

Now all work is done in factory function `CurrentModulesMatcher() (Matcher, error)`, which is undesired, and unexpected behavior. Factory function shouldn't do a real costly work, especially if it isn't required.

If you really like to pursuit this refactor, I would advise changing the `Matcher` func to be:

```golang
type Matcher func(modname string) (bool, error)
```

That would allow moving the "work" to the interior of the matcher func, keeping the current lazy behavior.

pkg/gomod/selector.go
```diff
@@ -15,90 +14,49 @@ var (
...
 // but excludes the current project modules.
-func DefaultSelector(domain string) (Selector, error) {
+func DefaultSelector(domain string) (Matcher, error) {
 	domain = strings.TrimSpace(domain)
 	if len(domain) == 0 {
-		return Selector{}, ErrNoDomain
+		return func(string) bool { return true }, ErrNoDomain
+	}
+
+	currentModules, err := CurrentModulesMatcher()
```
cardil, line 52: Calling `DefaultSelector(domain string) (Matcher, error)` will also do "the work", of looking for project modules, regardless if the given module matches the given domain name.

cardil (changes requested):

pkg/gomod/selector.go
evankanderson: Are we taking advantage of that laziness anywhere?

The advantage to doing the work up-front is that `Matcher` can have a simpler signature that doesn't need to report errors, as we do the possibly-error-creating work up front.  As far as I can tell, `CurrentModulesMatcher` is actually only ever used in this file, so we could even avoid exporting it at all.

(Yes, we could make this lazy, but I'm concerned about deferring the error to a place where it's more difficult to handle.)

pkg/gomod/selector.go
evankanderson: Is it really important to defer the error handling into late in the game? It looked to me that we were only using `DefaultSelector` in the `RunE` commands for `check`, `float`, and `needs`, each of which invokes the `DefaultSelector` once.

pkg/gomod/selector.go
kvmware: Take this with a grain of salt because of my ignorance of golang best practices and all, but for me readability > almost anything.

cardil: Okay. I think I'm convinced by you both. This lazy behavior isn't needed. It could be easily, re-introduced if it became a problem.

Also, removal of this `log.Fatal` is a great idea.

pkg/gomod/selector.go
cardil: Same as https://github.com/knative/test-infra/pull/3704/files#r1107629477

pkg/gomod/selector.go
```diff
@@ -15,90 +14,49 @@ var (
 )
 
 // Matcher matches a module name.
-type Matcher interface {
```
cardil, line 18: I would really like to keep at least this interface.

It allows people passing not only functions, but compatible structs as well.

It can be easily implemented by simple func:

```go
type MatcherFn func(modname string) bool
```

and to return it:

```go
return MatcherFn(func(modname string) bool {
  return strings.HasPrefix(modname, domain) && !currentModules(modname)
})
```

cardil: But, if you don't feel it is worth it, I can give LGTM right away.

evankanderson: Did you see how I used `StringSet.Has` as a `Matcher` below on line 41?

```golang
func CurrentModulesMatcher() (Matcher, error) {
  knownMods := sets.NewString()
  // ...
  return knownMods.Has, nil
}
```

I'm not convinced we need to declare an `interface` to define an interface (`Matcher` as a type alias for `func(string) bool`).

evankanderson: So in your example above, you could have:

```golang
  return MatcherFn(myMatcher).Match
```

cardil: /lgtm
/approve

cardil (approved):

Issue #3479, opened 2022-08-11
[EPIC] Are we Github Actions yet?


kvmware:
This is an epic to evaluate what is possible for Prow vs Github Actions(GHA). For some context please see https://docs.google.com/document/d/1Jf2lZ9V63oxItIv2__JoEA8gQ52s9_NXCYR-VzkEkPg/edit?resourcekey=0-doRA3EyTBo9N8EN9yRiRAw#heading=h.4guahjqktlld. I will add in potential solutions and how much of feature parity they have in separate issues and link them here.

<This is a placeholder for now>

PR #3722, opened 2023-02-16
Revert "Process stale issues/PRs via Prow (#3532)"

kvmware:
This reverts commit 3fe80a667adf66613eb8f95ccf627537f7c7844d.

There already exists a knative/actions stale action. As already discussed with the community a few times now, we are trying to minimize prow usage - https://github.com/knative/test-infra/issues/3479

### DIFF
--- a/prow/jobs/custom/knativeteam-groups.yaml
+++ b/prow/jobs/custom/knativeteam-groups.yaml
@@ -8,7 +8,7 @@ presubmits:
     cluster: prow-trusted
     spec:
       containers:
-      - image: gcr.io/knative-tests/test-infra/prow-tests:v20230217-021b0c69
+      - image: gcr.io/knative-tests/test-infra/prow-tests:v20230220-8e42b241
         command:
         - runner.sh
         args:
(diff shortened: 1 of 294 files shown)


cardil:
Having both robots is definitely not a good idea. I can confirm we discussed we should move away from using Prow where we can, so I'm also interested of reasons for #3532

upodroid:
I deployed #3532 for a few reasons:
- better label selections, it is very easy to adjust the query that the commenter runs on
- Prow still processes the webhook events and that will not be replaced by github actions anytime soon.
- repo exclusions
- not-as-planned closure (this is a new feature that the stale action can't do)

If Prow still processes the github events(issue_comment) then assuming we removed the prow dependancy by swapping the job that writes comments for an action that only writes comments isn't that helpful.

I can fix @ReToCode issue by adjusting the query label to exclude triage-accepted issues.

ReToCode:
> I can fix @ReToCode issue by adjusting the query label to exclude triage-accepted issues.

Not sure about that. It would solve it when we add triage labels, but not that both bots interfere, right? One removes the `stale` label to add `rotten` while the gh-actions bot only relies on `stale`, right?

upodroid:
In addition to disabling the GitHub stale action as well when this PR is closed.

PR #3737, opened 2023-02-24
drop release-1.7

dprotaso:
Fixes


kvmware:
This should be done automatically with https://github.com/knative/test-infra/tree/main/tools/release-jobs-syncer, if it is not that is a issue with the tool.

dprotaso:
good point

Issue #3671, opened 2023-01-17
Automate buming golang patch versions


kvmware:
We should be bumping patch versions of golang (https://github.com/knative/test-infra/pull/3669#event-8238996488) automatically. Dependabot might be a good candidate for this.

PR #3746, opened 2023-03-02
Cleanup golang install for prow-tests image

kvmware:
Step 1 to fix https://github.com/knative/test-infra/issues/3671

What this does is uses whatever the FROM golang:X version as the latest golang version installed with gvm as well as a manual specified version that should be bumped manually when project is ready to use the new latest stable version. The latest stable version will be bumped with dependabot.

### DIFF
--- a/images/prow-tests/Dockerfile
+++ b/images/prow-tests/Dockerfile
@@ -151,43 +151,17 @@ RUN java -version && \
 # END: JAVA SETUP
 #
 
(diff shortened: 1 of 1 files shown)


evankanderson (commented):
That's quite a Dockerfile.

/lgtm

Though I didn't read all the other parts that weren't in the diff to double check no other side effects.

PR #3748, opened 2023-03-03
Configure Prow URL via flags

upodroid:
/cc @kfaseela @puerco

You should be able to use this library in istio release tooling


kfaseela:
thanks @upodroid

PR #3777, opened 2023-03-30
Update slack channels

dprotaso:
See mappings here: https://docs.google.com/spreadsheets/d/1jYrkQ_nHqdQr8K4aomVnL0WSA8GTY2W24NxZBPbp-iY/edit#gid=0


upodroid:
The creds you got from CNCF won't work. We need an access token.

https://github.com/clarketm/kubernetes_test-infra/blob/b75a14a3781cc6b5b196b6d6328293a23938c2c1/prow/cmd/crier/README.md#slack-reporter

PR #3786, opened 2023-04-05
we only care about patch versions

dprotaso:
Dependabot bumped our 1.19 image to 1.20 which is something we don't want.

I'm hoping this change will mean we'll only bump respective patches

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,4 +3,9 @@ updates:
   - package-ecosystem: "docker"
     directory: "/images/prow-tests/"
     schedule:
-      interval: "weekly"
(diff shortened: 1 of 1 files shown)


kvmware:
initial behavior was to have a latest stable go version and manually bump older versions https://github.com/knative/test-infra/pull/3746/files#diff-3a809bc2ddb42dd8cd4627d729d61452cc1ed660a8d82ae02476cf740e7740faR242

https://github.com/knative/test-infra/commit/1305d234c0e41dad2db9539d7a8ef24791a95161 this broke that assumption.

I see the logic you want to implement though.

PR #3787, opened 2023-04-10
added support for s390x for kn-plugin-event

dilipgb:
This PR updates the prow related jobs for s390x for kn-plugin-event.


upodroid (approved):
LGTM

PR #3804, opened 2023-04-25
Change the billing accounts used by GCP projects

upodroid:
/cc @kvmware 

Thank you @jeefy and @puerco for sorting out the billing issues.

All the projects are using the invoiced billing account that jeefy sorted out.


puerco (approved):
Thank you! And thanks for wiring it here in the TF code. 

I don't have visibility to check if the account ID is right so leaving the LGTM here, feel free to lift the hold when ready.

/lgtm
/hold

infra/gcp/boskos/main.tf
```diff
@@ -20,7 +20,7 @@ module "project" {
   project_id      = each.key
   folder_id       = google_folder.boskos.id
   org_id          = "22054930418"
-  billing_account = "01DDAC-7D2015-6E152D"
+  billing_account = "018CEF-0F96A6-4D1A14"
```
kvmware, line 23: shouldn't this also be `var.billing_account`?

upodroid: No, it is a separate Terraform state and doesn't inherit variables

kvmware: cool. well lgtm just unhold

upodroid: /hold cancel

PR #3832, opened 2023-05-23
[s390x]: Adjusted cron schedule for s390x e2e-eventing-reconciler tests

dilipgb:
This PR fixes the cron schedule time for e2e-eventing-reconciler tests for 1.10.


dilipgb:
@upodroid can you please take a look at this PR. Thanks.

PR #3848, opened 2023-07-06
buoy: cut extra URL path for module

nak3:
Some module use sub-directory for the module like https://github.com/blang/semver/tree/master/v4

In that case, buoy fails to fetch metadata with the full path:

```
$ curl https://github.com/blang/semver/v4?go-get=1
```

so needs to use the top directory and path.

```
$ curl https://github.com/blang/semver?go-get=1
```

Hence, this patch cuts extra URL path for module.


kvmware:
as discussed in slack please make changes to respected repo https://github.com/knative/toolbox/tree/main/buoy

PR #3857, opened 2023-11-20
Make SECURITY.md consistent

evankanderson:
We're missing these across a lot of Knative repos, this is copied/improved from the 3 that existed.


davidhadas:
@upodroid hey, 
we are adding this file to all repos (as part of the security audit
if there is also away to make sure it is added to all future repos, we should do that as well. 
/lgtm

dprotaso:
this repo shoudl be archived

